PR #48, opened 2016-07-21
Avoid problems with SSL and versions prior to 2.7.9

ethervoid:
https://github.com/PokemonGoF/PokemonGo-Bot/blob/master/pokecli.py#L38 this part is only for python 2.7.9 or above


PokemonGoF:
Thanks a lot for the pull request, would you mind to create a pull request with you in contributor?

PR #78, opened 2016-07-21
Print the pokemons in inventory and profile on each step

diordache:


PokemonGoF:
Should we wait for the refactoring ? @douglascamata

douglascamata:
@diordache you mind rebasing your work with the master? We got some big refactoring done. Also, please consider a reworking the profile print to something with less lines, otherwise it's gonna spam the terminal a lot and can hide more important things (like Pokemons caught, Pokestops, etc).

Thanks!

PR #91, opened 2016-07-21
print to log, fort worker now know item names

berend:
remade my previous PR after refactoring

fort_worker had the item_list missing, after splitting of files

### DIFF
--- a/bot.py
+++ b/bot.py
@@ -10,12 +10,14 @@ from stepper import Stepper
 from geopy.geocoders import GoogleV3
 from math import radians, sqrt, sin, cos, atan2
 
+
 class PokemonGoBot(object):
 
     def __init__(self, config):
         self.config = config
         self.pokemon_list=json.load(open('pokemon.json'))
         self.item_list=json.load(open('items.json'))
+        self.log = logging.getLogger(__name__)
 
     def start(self):
         self._setup_logging()
@@ -27,7 +29,7 @@ class PokemonGoBot(object):
 
     def work_on_cell(self, cell, position):
         if 'catchable_pokemons' in cell:
-            print 'Something rustles nearby!'
+            self.log.debug('Something rustles nearby!')
             for pokemon in cell['catchable_pokemons']:
                 worker = PokemonCatchWorker(pokemon, self)
                 worker.work()
@@ -42,7 +44,7 @@ class PokemonGoBot(object):
                         worker = SeenFortWorker(fort, self)
                         hack_chain = worker.work()
                         if hack_chain > 10:
-                            print('need a rest')
+                            self.log.info('need a rest')
                             break
 
     def _setup_logging(self):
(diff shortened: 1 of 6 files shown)


douglascamata:
Weird, github says you still have conflicts to merge with master.

bot.py
```diff
@@ -10,12 +10,14 @@
...
 from math import radians, sqrt, sin, cos, atan2
 
+
 class PokemonGoBot(object):
 
     def __init__(self, config):
         self.config = config
         self.pokemon_list=json.load(open('pokemon.json'))
         self.item_list=json.load(open('items.json'))
+        self.log = logging.getLogger(__name__)
```
douglascamata: There's no need for this line here. `self.log` is assigned later on the `_setup_logging` function.

cell_workers/pokemon_catch_worker.py
```diff
@@ -8,6 +9,7 @@ def __init__(self, pokemon, bot):
         self.position = bot.position
         self.config = bot.config
         self.pokemon_list = bot.pokemon_list
+        self.log = logging.getLogger(__name__)
```
douglascamata: You can use the logger available at `bot.log`, there's no need to create the logger again here.

cell_workers/pokemon_catch_worker.py
```diff
@@ -1,3 +1,4 @@
+import logging
```
douglascamata: This is not necessary, explanation is in a comment below.

douglascamata:
btw, IMHO, it's better if we use just one logger for the whole app. What you think?

douglascamata:
@berend please ignore my messages. I like more loggers better, but I think it's better if we use more meaningful strings to name them, like the one for the `Stepper` class.

berend:
Yes to all. Way past midnight,  will change that tomorrow

pokecli.py
```diff
@@ -78,14 +81,19 @@ def init_config():
...
     logging.getLogger("requests").setLevel(logging.WARNING)
     # log level for main pgoapi class
     logging.getLogger("pgoapi").setLevel(logging.INFO)
     # log level for internal pgoapi class
     logging.getLogger("rpc_api").setLevel(logging.INFO)
 
+    logging.getLogger("bot").setLevel(logging.INFO)
+    logging.getLogger("stepper").setLevel(logging.INFO)
+    logging.getLogger("seen_fort_worker").setLevel(logging.INFO)
+
```
douglascamata: @berend I think these `setLevel` calls above should be put inside the `Bot` class to keep things isolated. Also you forgot the `PokemonCatchWorker` logger.

stepper.py
```diff
@@ -22,20 +23,20 @@ def __init__(self, bot):
         self.steplimit2 = self.steplimit**2
         self.origin_lat = self.bot.position[0]
         self.origin_lon = self.bot.position[1]
+        self.log = logging.getLogger('stepper')
```
douglascamata: @berend can you update all `getLogger` calls related to the bot and workers to have simple and meaningful names like this?

bot.py
```diff
@@ -74,26 +76,30 @@ def _setup_api(self):
...
+
+        self.log.info('#' * 30)
+        self.log.info('Profile:')
+        self.log.info('Username: {username}'.format(**player_profile))
+        self.log.info('Bag Size: {item_storage}'.format(**player_profile))
+        self.log.info('Pokemon Storage: {poke_storage}'.format(**player_profile))
+        self.log.info('Account Creation: {creation_time}'.format(**player_profile))
+        self.log.info('Pokecoin: {POKECOIN}'.format(**player_profile))
+        self.log.info('Stardust: {STARDUST}'.format(**player_profile))
+        self.log.info('#' * 30)
```
douglascamata: @berend great refactoring with `str.format`, awesome! I didn't know this was possible.

stepper.py
berend: yes, will do

jtdroste:
@eggins why did you close this?

berend:
I just finished cleaning up all log handling. All is now centralized in the bot at one place. But I guess, master has again moved into some silly direction.

@douglascamata, please have a look.

To be able to do stuff here, we need a proper output, and not just `print`s all over.

PR #102, opened 2016-07-22
if cp is lower, transfer only caught pokemon

techedlaksh:
This changes will not transfer pokemon from the bag as some people have mentioned that they need this feature. In future, pokemon from bag will transfer too unless its only 1 left of its own type.

### DIFF
--- a/cell_workers/pokemon_catch_worker.py
+++ b/cell_workers/pokemon_catch_worker.py
@@ -1,4 +1,5 @@
 import time
+from sets import Set
 
 class PokemonCatchWorker(object):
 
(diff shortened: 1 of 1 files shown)


PokemonGoF:
Hmm, do we really need this ? @techedlaksh

techedlaksh:
Right now it only deletes the pokemon we just caught, if its cp is lower but i was gonna change this to removing all duplicates (leaving one) lower than the set cp.

tonz134:
I'm actually using it since i don't want the bot to delete my pokemon in the bag.

PokemonGoF:
OK, then we need another place to drop many pokemon in bag.

PokemonGoF:
Thanks for your vote  and pull request :)

Issue #95, opened 2016-07-21
 Why the bot not chooses the nearest pokestop ?


vincentbeutin:
I start in a location with a pokestop at 50 meters, why it chooses a pokestop far from here ?

`python pokecli.py -a google -u *** -p *** --location "Pont Boieldieu, 76000 Rouen" -w 2.5 --spinstop`

-> Found fort 69d333ccf5574064a350752d4ef74706.16 at distance 605.598593412m

An other thing, egg hatching seems not working... Meters aren't counted.

douglascamata:
@vincentbeutin because right now there's not a lot of work on how we choose PokeStops. If you look at bot.py:40 you will see that we just get a list for the server and than literally walk from stop to stop in this list. Feel welcome to improve it to walk to the nearest pokestop, it's a bit late for me and I gotta get some sleep. Maybe I'll work on it tomorrow.

vincentbeutin:
@douglascamata That's what I thought. I'll try to do something.

Do you have any idea for the second problem ? (egg hatching)

douglascamata:
No idea for egg hatching, I'm sorry. I just started playing this afternoon.

ironbishop:
the stop issue is a duplicate of #141 , while the egg hatching is not implemented yet at current time.

TheGoldenXY:
Duplicate of https://github.com/PokemonGoF/PokemonGo-Bot/issues/141

/closed

PR #104, opened 2016-07-22
Add basic fort search logic

CapCap:
#95

### DIFF
--- a/bot.py
+++ b/bot.py
@@ -41,7 +41,6 @@ class PokemonGoBot(object):
                 forts = [fort for fort in cell['forts'] if 'latitude' in fort]
                 # Sort all by distance from current pos- eventually this should build graph & A* it
                 forts.sort(key=lambda x: SeenFortWorker.geocalc(self.position[0], self.position[1], fort['latitude'], fort['longitude'])) 
-                print forts
                 for fort in cell['forts']:
                     worker = SeenFortWorker(fort, self)
                     hack_chain = worker.work()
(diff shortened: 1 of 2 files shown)


CapCap:
Eventually should refactor globally useful things like geo dist calc to it's own utility class, etc

cell_workers/seen_fort_worker.py
```diff
@@ -90,9 +90,10 @@ def work(self):
                 print('may search too often, lets have a rest')
                 return 11
         time.sleep(8)
-        return 0
+        return
```
PokemonGoF: Is this a typo?

bot.py
```diff
@@ -37,13 +37,17 @@ def work_on_cell(self, cell, position):
                 worker.work()
         if self.config.spinstop:
             if 'forts' in cell:
+                # Only include those with a lat/long
+                forts = [fort for fort in cell['forts'] if 'latitude' in fort]
+                # Sort all by distance from current pos- eventually this should build graph & A* it
+                forts.sort(key=lambda x: SeenFortWorker.geocalc(self.position[0], self.position[1], fort['latitude'], fort['longitude'])) 
+                print forts
                 for fort in cell['forts']:
-                    if 'type' in fort:
```
PokemonGoF: you need the 'type', or it will be a gym.

vincentbeutin:
@CapCap this should be done also here ? : 

bot.py l.60
`for cell in map_cells:
     self.bot.work_on_cell(cell,position)`

For my initial example i think the nearest pokestop is not in the first cell.

What do you think ?

CapCap:
@vincentbeutin I think you're absolutely right

CapCap:
I also think we shouldn't split everything up like this in terms of distance calculation- it should go to the nearest pokemon/fort, not nearest pokemon then nearest forts

PR #107, opened 2016-07-22
find pokemons while moving to pokestop

alawing:
let possible to find pokemons while moving to pokestop


cell_workers/seen_fort_worker.py
```diff
@@ -21,9 +21,15 @@ def work(self):
         print('Found fort {} at distance {}m'.format(fortID, distance))
         if distance > 10:
             print('Need to move closer to Pokestop')
-            position = (lat, lng, 0.0)
+            #position = (lat, lng, 0.0)
+            position = (10 * (lat - self.position[0]) / distance, 10 * (lng - self.position[1]) / distance, 0.0)
             if self.config.walk > 0:
                 self.api.walk(self.config.walk, *position)
+                self.api.player_update(latitude=lat,longitude=lng)
```
PokemonGoF: I think this won't catch pokemons while moving to pokestop, am I correct?

alawing: I thought you are not going to nearest pokestop from your previous version. So if moving to a long distance pokestop, it may have chance to get pokemons near player. My action is to split this long moving into little steps.

CapCap: should this use a generic linear interpolation function?

alawing:
This time works

PR #117, opened 2016-07-22
Fix Issue #115 + Fix a KeyError bug that occurs on new accounts

### LINKED ISSUE #115: "Exchange" not functioning
Confirming that low CP pokemon are not being exchanged at all, using the -c 200 command. 

It is doing the same thing without the -c command as well. 

Even pokemon with CP < 100 are not being exchanged. The Debug log says they are being exchanged, but they are not.

haykuro:
Fixes #115 : Seems like the actual release_pokemon API call wasn't being made. Just added that in..

Also, fixed a bug at the bottom of bot.py the playerData will not return certain keys like poke_stop_visits if it was never used before..

image

### DIFF
--- a/bot.py
+++ b/bot.py
@@ -206,8 +206,15 @@ class PokemonGoBot(object):
 
                                     nextlvlxp = (int(playerdata['next_level_xp']) - int(playerdata['experience']))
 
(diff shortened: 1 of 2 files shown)


PokemonGoF:
Great! We were confusing about the RELEASE.

Issue #52, opened 2016-07-21
Walking speed bugged?


harsh1412:
So i set the speed -w 4.40 and got this in console
This is PokeStop
distant is 43.8958391159m
need setup the postion to farming fort
((xxx, xxx, 0.0), xxxxxxxxxxxxxxxxxxxxxxxxxxx')
Walking from (xxx, xxx) to (xxx, xxxx) for approx. 95.0sec

If i am correct speed of walking should be low right?

douglascamata:
Can be faster than 4.40 m/s. I'm using 20 m/s, which is equivalent to a can running 100 km/h and it works.

holywend:
Hahaha but 20m/s wont hatch egg isnt it?

crack00r:
yep, if you want to hatch eggs, you can max go 15km/h wich is 4.16

harsh1412:
@crack00r exactly..! then why it was showing 95 seconds? :3

LilDice:
I cant set the walkk speed at all I am using -w 2.2 and it gives error

millejosh:
@LilDice what is the error?

holywend:
im confused I set the walking speed as 4 m/s but this happen

This is PokeStop
distant is 790.359542282m
need setup the postion to farming fort
((52.630008, -1.142752, 0.0), u'1c4bf55a09c642a295e588895401fe5f.16')
Walking from (52.632314, -1.145138) to (52.630008, -1.142752) for approx. 76.0sec

distant is 790.359 while the approx is 76 sec it doesnt make sense coz
790.359 / 76 = 10.4 m/s

different from what I set (4 m/s)

LilDice:
unrecognized arguments: -w 4.16

holywend:
@LilDice perhaps you can try to redownload the zip.

LilDice:
That worked thankyou!

tstumm:
@holywend Your calculations are correct but the distance estimation by the bot itself is wrong so the displayed distance is wrong. The walk routine uses a different distance estimator which is more precise.

PR #124, opened 2016-07-22
Added utf8 compatibility and fixes wrong distance estimator

### LINKED ISSUE #96: not spinning forts with non ascii char in names
or smthing like that. 

Raising UnicodeEncodeError here
File "D:\pokemonbot\cell_workers\seen_fort_worker.py", line 38, in work
  print('Now at Pokestop: ' + fort_details['name'] + ' - Spinning...')

works fine when `fort_details['name']` removed

tstumm:
Fixes #96

### DIFF
--- a/README.md
+++ b/README.md
@@ -60,7 +60,7 @@ This project uses Google Maps. There's one map coupled with the project, but as 
       -u USERNAME, --username USERNAME              Username
       -p PASSWORD, --password PASSWORD              Password
       -l LOCATION, --location LOCATION              Location (Address or 'xx.yyyy,zz.ttttt')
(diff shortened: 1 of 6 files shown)


tstumm:
Also fixes #52 since ae4d842

eggins:
Appreciate this commit!

tstumm:
Please do not forget to close the corresponding issues :)

eggins:
Just did @tstumm  :)

douglascamata:
@tstumm @PokemonGoF `pgoapi` is an external package and its folder is going to be removed in favor of only importing as a dependency from setup.py. Fixes to `pgoapi` should be done in https://github.com/tejado/pgoapi and we should not add code to `pgoapi` like this.

tstumm:
It's actually not a fix in pgoapi since I added the code for walking by myself. I just moved my code to utilities.

PR #126, opened 2016-07-22
Hotfix: Missing imports

tstumm:

PR #137, opened 2016-07-22
Math importr on utilities were missing

ineedblood:
There were missing cos, asin, sqrt imports missing at utilities.py


tstumm:
Already fixed by #126

PokemonGoF:
Thanks anyway :)

PR #152, opened 2016-07-22
index.html for viewing location

Reaver01:
Still need to update position whenever it walks.. Can't test until servers are up.


Reaver01:
forgot to remove the git stuff from stepper file

Issue #142, opened 2016-07-22
Refactoring Proposal: Behavior Trees, Tasks, and Plugins


TheSavior:
Problems:
It is hard to configure the bot the way each person wants.
It is hard for engineers to add / extend the functionality of the bot
As this bot grows, there isn't great separation of behavior, causing additional complexity and problems with merge conflicts.

Solution:
Previous bots for games like World of Warcraft have grown to huge heights by being easily configurable, extensible, and easy to understand. The classic solution to this is to implement the bot core as a configurable behavior tree.

The basics are actually pretty straight forward in this case.

Imagine there are tasks like this:

ClearBag
TransferPokemon
CatchPokemon
SpinPokestops
WalkToLocations
WalkToNearestPokestop

These tasks would be implemented as classes that have a run method.

We could then read in a configuration file that people can write which defines how the bot runs. This would make it significantly easier to configure than adding more cli flags.

I imagine people could then write configs like this:

```
{
	"tasks": [
		{
			"type": "TransferPokemon",
			"arguments": {
				"minimumRemainingSlots": 10,
				"minimumCP": 100
			}
		}, {
			"type": "CatchPokemon",
			"arguments": {
				"greatBallCP": 300,
				"ultraBallCP": 500,
			}
		}, {
			"type": "ClearBag",
			"arguments": {
				"minimumRemainingSlots": 5
			}
		}, {
			type: "SpinPokestops"
		}, {
			type: "WalkToLocations",
			arguments: {
				"locations": [
					"40.781362, -73.969046", 
					"40.768075, -73.976339"
				]
			}
		}
	]
}
```

The tool would take the configuration, instantiate an instance of the task type for each element in the tasks array passing the configuration arguments to the constructor. It then uses a game loop, essentially a while loop that on each tick iterates through each task calling run. 

By putting tasks higher in the task list, it essentially says it is a higher priority. 

In the example above it is saying that it needs to make sure there is space in the pokebox before catching pokemon, space in the bag before spinning pokestops, and it should catch the pokemon around before starting to move in the direction of another location. 

By doing this, we can enable other people to create repos that implement new tasks that can be more complex and swapped in for other things, and those classes can be extended as well. 

I'm curious what people's thoughts are on this. Is this a direction people would want to go?

tstumm:
A more modular approach would be nice, but configuration should be more easy. It is "just" Pokemon Go, it is not that complex that we need full "if this than that" configuration.

[...]

PR #155, opened 2016-07-22
(WIP) Behavior Tree Proof of Concept

TheSavior:
As talked about in #142, I want to explore the possibility of breaking the bot into extensible, possibly shared, configurable tasks. This is the first step in that direction and I would love any feedback people have on the implementation


bot.py
```diff
@@ -179,6 +181,16 @@ def _setup_api(self):
              print('Exception during print player profile')
         self.update_inventory();
 
+    def _setup_tree(self):
+        tasks = []
+
+        for taskConfig in self.config.tasks:
+            taskRef = globals()[taskConfig['type']]
```
douglascamata: `globals()` is really bad behaviour. Please encapsulate properly. 

Also `taskConfig` is CamelCase and we use snake_case for variables and functions. CamelCase only for class definitions.

douglascamata:
@TheSavior you tasks implementation looks a lot like our workers. Please keep them in mind when you do your work, because they are used already to process cells (which can be pokemons) and then action are taken in each of those cells by these workers.

TheSavior:
@douglascamata I agree that the tasks and workers seem very similar. I think the biggest difference is that I want to get these tasks more standalone than the workers are. Currently a lot of the logic is not actually in the worker. For example, the stepper essentially tells the worker which pokemon to catch. Since it iterates over each visible pokemon, if you fill up your bag while catching, the worker will be stuck.

So in the case of the catch pokemon task, the goal would be for it to figure out what is visible, pick one, catch it, and return.

bot.py
TheSavior: I'm not very familiar with python, the thinking here is that we want to be able to instantiate these tasks by name. It's probably the right choice to just put a switch statement in for now. If we were trying to support a plugin system at this point, we would need to figure out some way to dynamically pull those in and instantiate by string name. Overengineering at this point though. :)

douglascamata:
@TheSavior that's expected, because the `Stepper` class will be the one to decide where to go. So we can add logic in it to avoid walking to pokemons when have no balls, for example.

douglascamata:
And then we can have many classes to decide where to go based on the current state.

TheSavior:
@douglascamata I messaged you on gitter. I'd love to chat this out, and make sure I'm heading in the right direction. 

One of the side effects of the problem I'm trying to solve in https://github.com/PokemonGoF/PokemonGo-Bot/issues/142 would actually pull out that decision making logic from the stepper. Having it in the stepper makes it hard to extend, configure, and swap out. Take a look at the linked issue where I walk through it a bit more and see if it makes sense where I'm coming from.

jtdroste:
@TheSavior we moved to slack.  To get an invite, [go here](http://pokemongo-bot.herokuapp.com/)

TheSavior:
Closing the PR for now. After talking to @douglascamata it sounds like there are a bunch of bigger things that need to be worked on now that will cause problems in the short term for this refactor. 

We'll revisit this soon, so let's continue the conversation in #142 and try to get a good grasp on the approach and implication. It sounds like @douglascamata had something a tad different in mind, so it would be good to figure this stuff out in the mean time.

PR #163, opened 2016-07-22
Indentation fix for config.json.example

Mordana:
Small update since notepad++ messed up some of the indentation (replaced by tabs were the original was with spaces)


tstumm:
Can you also change the values to the best fitting datatypes?
Like `"True"` should be `true`, `"3.141"` should be `3.141` and so on.

jtdroste:
Looks good, merging.

PR #167, opened 2016-07-22
Add a little jitter to sleep

brantje:
Should make it a bit harder for them to detect us.

### DIFF
--- a/cell_workers/pokemon_catch_worker.py
+++ b/cell_workers/pokemon_catch_worker.py
@@ -1,6 +1,6 @@
 import time
 from sets import Set
-
+import humanBehaviour
 class PokemonCatchWorker(object):
 
     def __init__(self, pokemon, bot):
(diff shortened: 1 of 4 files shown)


douglascamata:
@brantje thanks!

tstumm:
I'm not sure if your calculations are correct.

tstumm:
Lets imagine you want so sleep 5 secs.

jitter = 5 * 1000 / 10 = 5 * 100 = 500

so sleep_time is something in [5-500, 5+500] = [-495, 505] ....

brantje:
And that is jitter

tstumm:
Yep but we did not invent time machines yet, so negative seconds are not possible.

tstumm:
Jitter should be like 10% of seconds, so `jitter = 0.1 * seconds`

PR #169, opened 2016-07-22
else if does not exists in python (at least python 2)

jdeniau:
replaced by `elif`


jdeniau:
declined in favor of #164

PR #170, opened 2016-07-22
Master

debosset:


douglascamata:
It needs to be rebased with dev again, lol

jtdroste:
As discussed on the slack channel, we will be forcing a push from dev to master when we are ready for a release.  The two branches (dev and master) have diverged too much to easily merge them.

PR #202, opened 2016-07-22
Do not encode received string

giomasce:
Otherwise the following print instruction will fail with a
UnicodeDecodeError when there are non-ASCII characters in the
Pokestop name.


douglascamata:
@giomasce thanks for the fix, but your branch has conflicts with the upstream. Can you run `git pull origin dev` and fix it?

giomasce:
Done.

jtdroste:
Unicode issues should be resolved, as the print_* functions can now output unicode.  I'm going to close this, but thanks!

PR #203, opened 2016-07-22
Use entrypoint instead of ENV variables

lucasilverentand:
The use of entrypoint instead of ENV variables allows the container to be run like an executable. This way everything remains uniform, in my opinion.

I would also suggest this repository to be auto build on the docker hub: https://docs.docker.com/docker-hub/builds/

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,10 +1,3 @@
 FROM python:2.7-onbuild
-ENV AUTH_SERVICE ""
-ENV USERNAME ""
-ENV PASSWORD ""
-ENV LOCATION ""
-ENV WALK ""
-ENV CP ""
 
-CMD python pokecli.py -a $AUTH_SERVICE -u $USERNAME -p $PASSWORD -l $LOCATION -w $WALK -s -c $CP
-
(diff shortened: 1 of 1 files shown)


mercuriete:
+1

jtdroste:
I don't have a bunch of state on docker, otherwise I'd be able to evaluate this better.  Could you show me how you would run it with our current dockerfile vs yours (i.e. the command you would enter in the terminal)

lucasilverentand:
Sure thing!

Previous:
`docker run <DockerName> -e AUTH_SERVICE=google -e USERAME=someuser@gmail.com -e PASSWORD=somepassword -e LOCATION=33.8118,-117.9189 -e CP=50 -e WALK=2`

Current:
`docker run <DockerName> -a google -u someuser@gmail.com -p somepassword -l 33.8118,-117.9189 -c 50 -w 2`

The new approach I'm proposing is that basically the arguments just get passed into the docker, we wouldn't need to add a ENV variable for every argument we come up with.

jtdroste:
Yup - makes total sense.  lgtm - merging...

PR #213, opened 2016-07-22
Fixes UnicodeEncodeError on initial location

vieira:
Short Description: 
When initial location has special characters the current version crashes with a UnicodeEncodeError: 'ascii' codec can't encode character u'\xe7' in position 14: ordinal not in range(128).

Fixes:
- UnicodeEncodeError when initial location has special characters, e.g. "Paço"

### DIFF
--- a/pokemongo_bot/__init__.py
+++ b/pokemongo_bot/__init__.py
@@ -203,7 +203,7 @@ class PokemonGoBot(object):
         self.position = self._get_pos_by_name(self.config.location)
         self.api.set_position(*self.position)
 
-        print('[x] Address found: {}'.format(self.config.location.decode('utf-8')))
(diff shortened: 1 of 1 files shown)


douglascamata:
@vieira how well you know Python? I think better solution is to add the utf-8 shebang to all your source files. If you don't know about it, I can merge this and do the proper fix to avoid any more utf-8 problems later.

vieira:
@douglascamata I might be wrong but I think that will only affect the encoding of the source code itself and any static strings defined there but not what comes in the wire as binary.

I tried putting
```
#!/usr/bin/env python
# -*- coding: utf-8 -*-
```
but it did not solve the problem. Is this what you mean?

jtdroste:
lgtm.

Please rebase your changes so we can merge it.  Thanks!

PR #223, opened 2016-07-22
Fix encoding error when printing messages

vieira:
Short Description: 
Some messages that will be printed will contain utf-8 chars, e.g. Pokestops in European locations.

Fixes:
- Printing messages that contain special chars, e.g. Pokestops in Europe.


douglascamata:
thanks!

PR #242, opened 2016-07-22
Cleanup code

haykuro:
Short Description: Cleanup headers and imports.

Make code more "pythonic"

Fixes:
- Caught KeyboardInterrupt error to show user aborted message instead of "crashing"
- Fixed a crash where if "-a" had an invalid value (not ptc/google) "log.error" was not a function, changed to "logging.error" which uses the logging import to output the error message propery.
- Other maintenance / cleanup.


douglascamata:
Thanks!

PR #212, opened 2016-07-22
Adding initial REST Framework; Adding method to get user stats

bstascavage:
This is the same as #204 , which was reverted with no explanation.  Basically looking for feedback as to the revert and how this code can comply better with the main branch.  Thank you!

PR #253, opened 2016-07-22
Adding initial REST Framework; Adding method to get user stats

bstascavage:
Short Description: 
Right now there isn't a way for a developer to get information on the bot while the bot is running. Sure, you can check information by logging into the app or debugging, but there is no way to query the bot for various statistics. I added a REST Framework to the bot so that a user can retrieve this information by accessing various endpoints.

Some use cases:

- Developers trying to gather information/test methods
- Developers wanting to write other applications that interact with the bot, such as web views or chat commands
- A way to access the bot without logging in; very useful for people with multiple accounts (or a bot test account)

Once the framework is approved, more accessor REST methods can be implemented.  These can be used by the new `web` section for displaying stats on the map or for any other utilities. 

This PR fixes the issues raised by @jtdroste in #212 .

Fixes:
- Flask and Flask-Classy are added to `requirements.txt`
- `flask.ext.classy` is updated to `flask_classy`
- Server now binds to `127.0.0.1` instead of `0.0.0.0`
- Flask's request logging now defaults to WARNING; HTTP requests will only be logged when `config.debug`
- Port defaults to 5000, can be set via config/commandline

### DIFF
--- a/pokecli.py
+++ b/pokecli.py
@@ -41,6 +41,7 @@ if sys.version_info >= (2, 7, 9):
 
 from pokemongo_bot import PokemonGoBot
 from pokemongo_bot.cell_workers.utils import print_green, print_yellow, print_red
+from server import PokemonGoServer
 
 def init_config():
     parser = argparse.ArgumentParser()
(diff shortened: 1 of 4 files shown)


douglascamata:
@bstascavage, please, communicate with us first. We decided to implement SocketIO server on top of Flask to providing information to applications that interact with the bot. You think you can update this to user socketio-python?

bstascavage:
@douglascamata I would love to communicate with you folks, but with the Slack invite down there is no way I can ATM, right?  If there is an alternative to being invited, or another avenue of communication, I would be very appreciated.

As for using socketio, I don't have an issue with that, depending on the implementation.

douglascamata:
@bstascavage We will only have more invites tomorrow because of rate limit :/

We want to use it because it enables better communication to send events like pokemon capture, for example. So if you create a web socket server with flask and a small api that we can use inside the bot to send events to every listener in this server, we would greatly appreciate.

douglascamata:
@bstascavage after the web socket server works and we can send events there from the bot code, we can create a map app that is updated dynamically and notifies about everything!

bstascavage:
Righto.  I am working on a socketio-flask solution now

jtdroste:
@bstascavage you have also been invited to the slack.  Check your GitHub email address :)

solderzzc:

PR #255, opened 2016-07-22
Fix player info response parsing

okushchenko:
Short Description: 
I don't know if pogo API changed, but it works for me this way.

Fixes:
- player info response parsing


douglascamata:
@okushchenko you're environment is messed up, please consider using `virtualenv` for a cleaner one and reset everything.

Issue #258, opened 2016-07-22
the config.json is not overiden by the parameter that has value and it's not NONE


felixzzz:
### Other Information
OS: 
Git Commit: (run 'git log -n 1 --pretty=format:"%H"' and paste it here)

PR #259, opened 2016-07-22
hotfix config.json overide the argument with default value and not None

felixzzz:
for issue #258 
for example I want the minimum cp to be 200 or transfer
this way will handle that, and also I figure the json should be in the correct data type.
please CMIIW :)


solderzzc:
you forgot to modify contributor field. ...

PR #268, opened 2016-07-22
Protect from exception when get inventory list return []

solderzzc:
Short Description: 
To fix the exception if get inventory []
Fixes:
- Lots of report...
- 
-


douglascamata:
No no, I have proper fix. I'm committing

solderzzc:
ok

PR #269, opened 2016-07-22
update for webview

Reaver01:
Short Description: 

Fixes:
- added framework for pokedex
- added json dump for pokedex
-


douglascamata:
the dump is never updated... but will also take quite some time to dump it every time, I guess, because it requires an api call

Reaver01:
Yes I know... If there is a different API call to only get what is in inventory_req['responses']['GET_INVENTORY']['inventory_delta']['inventory_items'].  It doesn't take too long to do.  At least it's not noticeable on my end.

PR #277, opened 2016-07-22
smart catch while farming

mohitmun:
Short Description: 
While Farming, Keep single copy of pokemon with highest CP,  and exchange other for candy

### DIFF
--- a/cell_workers/pokemon_catch_worker.py
+++ b/cell_workers/pokemon_catch_worker.py
@@ -1,6 +1,7 @@
 import time
 from sets import Set
 from pgoapi.utilities import f2i, h2f, distance
+import json
 
 class PokemonCatchWorker(object):
 
@@ -105,6 +106,19 @@ class PokemonCatchWorker(object):
                                 if status is 3:
                                     print('[x] Oh no! ' + str(pokemon_name) + ' vanished! :(')
                                 if status is 1:
+                                    self.api.get_inventory()
+                                    response_dict = self.api.call()
+                                    if self.config.smart_catch:
+                                        id_cp_tuples = self.get_id_cp_tuples_for_pokemonid(pokemon['pokemon_data']['pokemon_id'],response_dict)
+                                        prev_id, prev_cp = (0,0)
+                                        for id_cp in id_cp_tuples:
+                                            next_id,next_cp = id_cp
+                                            if next_cp < prev_cp:
+                                                self.transfer_pokemon(next_id)
+                                            elif prev_id != 0:
+                                                self.transfer_pokemon(prev_id)
+                                            prev_id,prev_cp = id_cp
+
                                     if cp < self.config.cp:
                                         print('[x] Captured ' + str(pokemon_name) + '! [CP' + str(cp) + '] - exchanging for candy')
                                         id_list2 = self.count_pokemon_inventory()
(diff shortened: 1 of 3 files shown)


cell_workers/pokemon_catch_worker.py
```diff
@@ -105,6 +106,19 @@ def work(self):
                                 if status is 3:
                                     print('[x] Oh no! ' + str(pokemon_name) + ' vanished! :(')
                                 if status is 1:
+                                    self.api.get_inventory()
```
solderzzc: I remember there's another function to get_inventory(), right?

douglascamata:
you sure this works?

mohitmun:
yeah It works for me

cell_workers/pokemon_catch_worker.py
mohitmun: there are multiple calls to 'count_pokemon_inventory' which calls get_inventory. so I moved this api call up and use response to do other operations

douglascamata: smart!

douglascamata:
@mohitmun please add information about this feature in the README in the command-line description.

cell_workers/pokemon_catch_worker.py
mohitmun: exachange is super fast. get_inventory moves lots of data on higher level

mohitmun: we can remove 'count_pokemon_inventory' and use directly response

mohitmun:
Yeah sure!

cell_workers/pokemon_catch_worker.py
douglascamata: @mohitmun leave this for another pull request, because that thing is big and dangerous.

cell_workers/pokemon_catch_worker.py
```diff
@@ -105,6 +106,19 @@ def work(self):
...
+                                        prev_id, prev_cp = (0,0)
+                                        for id_cp in id_cp_tuples:
+                                            next_id,next_cp = id_cp
+                                            if next_cp < prev_cp:
+                                                self.transfer_pokemon(next_id)
+                                            elif prev_id != 0:
+                                                self.transfer_pokemon(prev_id)
+                                            prev_id,prev_cp = id_cp
+
                                     if cp < self.config.cp:
```
douglascamata: @mohitmun I still see a possibility here of code getting inside `smart_catch` if and inside this one. So I have a new idea! Instead of adding a new option for smart catching, why not use `-c` for this? You can make `-c` accept a string and then you check if the string is `smart` to use smart catch. If the string isn't `smart`, then you try to convert it to int.

mohitmun: make sense

douglascamata:
@mohitmun I was eager for your update. Gonna review!

cell_workers/pokemon_catch_worker.py
```diff
@@ -105,7 +106,24 @@ def work(self):
...
+                                        id_cp_tuples = self.get_id_cp_tuples_for_pokemonid(pokemon['pokemon_data']['pokemon_id'],response_dict)
+                                        prev_id, prev_cp = (0,0)
+                                        for id_cp in id_cp_tuples:
+                                            next_id,next_cp = id_cp
+                                            if next_cp < prev_cp:
+                                                self.transfer_pokemon(next_id)
+                                            elif prev_id != 0:
+                                                self.transfer_pokemon(prev_id)
+                                            prev_id,prev_cp = id_cp
+                                    try:
```
douglascamata: @mohitmun please add an `else` clause to the `if self.config.cp == "smart"` if and move the `try` block inside it.

solderzzc:
I little bit confusing..... merged....

mohitmun:
New to python :P

mohitmun:
@solderzzc can you please try it. It works for me, not confident

PR #289, opened 2016-07-22
Add a option to continue on saved location

mzupan:
Short Description: 

This adds a last known location flag to pickup.

it will override the location flag if set and there is a location.json file already created with valid data in it


douglascamata:
@mzupan the option name is a bit misleading. You mind changing to something like `--use-location-cache` and short version being `-lc`?

mzupan:
@douglascamata good idea.. fixed the names

douglascamata:
Thanks!

PR #319, opened 2016-07-22
Use Great and Ultra Balls when no Poke Balls

lucasilverentand:
Short Description: 

I propose to use Ultra- and Great Balls when there are no Poke Balls, since I have many Great Balls being collected by the bot it might as well use them too, right :) . But I also realise we don't want to deplete them, so I set a limit of at least 20 in the stock.


solderzzc:
I'm not sure if there's something similar with @Gary-BOT 's working.
@douglascamata should we merge ?

douglascamata:
@solderzzc not yet, IMHO. @Gary-BOT will surprise us, as he always does.

lucasilverentand:
Sounds great, let's see the wizard at work then 👍

Issue #318, opened 2016-07-22
IRC channel instead of Slack


MaxLeiter:
Why use closed-source software when IRC runs perfectly well? I'm willing to run a version of [TheLounge](https://www.github.com/thelounge/lounge) purely for a PokemonGo channel, if people are unwilling to setup their own clients.

douglascamata:
Slack as awesome as soon as we stop hitting the invite rate limit. Much more feature rich.

MaxLeiter:
What are some features that Slack has over IRC/TheLounge?

solderzzc:
IRC maybe same as Gitter, we can get Gitter back ?

MaxLeiter:
Or just use IRC? Gitter would be a regression from Slack, if anything.

solderzzc:
but we will lose the github integration

MaxLeiter:
What Github integration? You can setup webhooks on the repo to announce commits, PRs, etc, plus I wrote a helpful bot for referencing issues and PR's which is available [here](https://www.github.com/lounge/lounge-bot). I'm willing to run lounge-bot (under a different name) and a github [announce bot](https://github.com/xpaw/wendysharp) on the channel if necessary.

solderzzc:
Nice :)

MaxLeiter:
I registered the channel and started a Lounge instance here: http://avatar.playat.ch:5000, and am currently setting the bots up

douglascamata:
I don't see any advantages of IRC over Slack. Why should we change?

MaxLeiter:
Not proprietary, open-source, multiple clients, more general use, no invite limit, etc.

https://drewdevault.com/2015/11/01/Please-stop-using-slack.html

douglascamata:
@MaxLeiter and then we lose: code highlight with github flavoured markdown, text files hosting, image hosting, very powerful notifications control, powerful built-in search... I can name more if u want.

MaxLeiter:
It all depends on the client you setup....people can use imgur, pastebin, gist, etc. for file hosting. Notifications are all handled by the client (IRC has many, Slack has one).

douglascamata:
@MaxLeiter this way you're trusting the users to have the best experience by themselves. We should give them the best experience by default.

TheSavior:
It sounds like you guys are arguing over something like tabs or spaces. It doesn't really matter as long as we pick one. The project has currently picked slack.

[...]

PR #322, opened 2016-07-23
Update README to have IRC channel instead of Slack 

MaxLeiter:


douglascamata:
??????

MaxLeiter:
We should use IRC instead of slack. See #318

douglascamata:
@MaxLeiter not everyone agreed to IRC channel yet. So this MR should wait.

MaxLeiter:
Doesn't need to be merged immediately. Not everyone will agree, but plenty of people can't access slack at the moment due to invite limitations.

PR #326, opened 2016-07-23
improve ball type decision maker

earthchie:
Now if you ran out of pokeballs it will use Great balls instead. If ran out both - use Ultra balls.

### DIFF
--- a/README.md
+++ b/README.md
@@ -139,9 +139,9 @@ ethervoid
 Bashin  
 tstumm  
 TheGoldenXY  
-Reaver01
-rarshonsky
-earthchie
+Reaver01  
+rarshonsky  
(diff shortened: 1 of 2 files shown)


solderzzc:
I want to this commit be merged :)

douglascamata:
shouldn't we wait for Gary's version?

solderzzc:
the merge of Gary should be very simple, there're two commits on this line mean users need it heavily.
since https://github.com/PokemonGoF/PokemonGo-Bot/pull/319
is easily and with a keep 20 in stock, we should merge #319

solderzzc:
Thanks for the PR, let's close for now.

PR #329, opened 2016-07-23
Renames --firsttrans to --initial-transfer and allow it to be combined with -c

douglascamata:
`--firsttrans` was renamed to `--initial-transfer`.

Now it can be combined with `-c CP_LIMIT` so it will NEVER release pokemon above the CP_LIMIT.


pokemongo_bot/__init__.py
```diff
@@ -121,55 +121,63 @@ def _setup_api(self):
...
+    def _initial_transfer_get_groups(self):
         pokemon_groups = {}
         self.api.get_player().get_inventory()
         inventory_req = self.api.call()
-        inventory_dict = inventory_req['responses']['GET_INVENTORY']['inventory_delta']['inventory_items']  
-        
+        inventory_dict = inventory_req['responses']['GET_INVENTORY']['inventory_delta']['inventory_items']
+
         for pokemon in inventory_dict:
             try:
```
solderzzc: What will happen when there's exception?

douglascamata: shit, this look messed up, w8

douglascamata: Whole block is:
```
try:
     reduce(dict.__getitem__, ["inventory_item_data", "pokemon_data", "pokemon_id"], pokemon)
except KeyError:
      continue
```

Which is basically a test to see if key at `pokemon['inventory_item_data']['pokemon_data']['pokemon_id']` can be pulled correctly. If it cannot, it means this is not a pokemon (probably an egg), so we skip it.

solderzzc: are you doing this part, I think there's no such code existing in tree.

PR #330, opened 2016-07-23
Transfer Pomemon if bag is full instead of flooding "Something rustles nearby!"

haykuro:
Short Description: Adds a "[x] Pokemon Bag is full!" message if you encounter a pokemon but can't catch it because your pokemon bag is full. (a.k.a. status 7 is returned when encounter is sent)

Fixes:
```
[#] Walking from (34.0092419, -118.4976037) to (34.013523, -118.492144) for approx. 278.0 seconds
[#] Something rustles nearby!
[#] Something rustles nearby!
[#] Something rustles nearby!
[#] Something rustles nearby!
[#] Something rustles nearby!
[#] Something rustles nearby!
...
```

Will instead do:
```
[#] Walking from (34.0092419, -118.4976037) to (34.013523, -118.492144) for approx. 278.0 seconds
[#] Something rustles nearby!
[x] Pokemon Bag is full!
[x] Initial Transfer.
[x] Preparing to transfer all Pokemon duplicates, keeping the highest CP of each one type.
[x] Releasing Pokemon #129 and CP 28
[x] Releasing Pokemon #129 and CP 23
[x] Releasing Pokemon #129 and CP 10
[x] Releasing Pokemon #140 and CP 48
[x] Releasing Pokemon #21 and CP 10
...
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -129,26 +129,27 @@ Quick Tip: When using this script, use a Lucky egg to double the XP for 30 mins.
  * s2sphere
  * googlemaps
  * pgoapi
+ * haykuro
 To install the pgoapi use `pip install -e git://github.com/tejado/pgoapi.git#egg=pgoapi`
 
 
 ## Contributors (Don't forget add yours here when you create PR:)
-eggins -- The first pull request :)  
-crack00r  
-ethervoid  
-Bashin  
-tstumm  
-TheGoldenXY  
+eggins -- The first pull request :)
+crack00r
+ethervoid
+Bashin
+tstumm
+TheGoldenXY
 Reaver01
 rarshonsky
 earthchie
 
 ## Credits
 ### The works are based on the Pokemon Go API
-[tejado](https://github.com/tejado) many thanks for the API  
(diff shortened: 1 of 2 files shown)


douglascamata:
@haykuro please raise a RuntimeError with the message instead of just printing, because the bot will be stuck forever, so it's better to stop it.

haykuro:
Fixed to run the transfer if the bag is full.

haykuro:
Just realized this is similar to #277 ..

However, I'm utilizing the bot.initial_transfer function which already does smart transfers..

Perhaps we can use his switch to enable/disable this feature?

solderzzc:
Looks good let me merge.

TSM-EVO:
Very similar to this: 

If you run out of bag space for balls, you'll continue to "rustle" nearby pokemon, and then say that you have no balls, but you won't continue walking to the next pokestop to get it.

Maybe there needs to be some sort of flag to say 'I'm out of pokeballs' and then only run the 'search' function if you have useable balls available. Or perhaps keep a running count of pokeballs , and include a count check before trying to catch a pokemon.

PR #331, opened 2016-07-23
Add Transfer list

tienkhoi:
Short Description: 

Fixes:
- Change pokemon.json, items.json to relative so we can call pokecli from anywhere in cli
- Add feature: Transfer list (Transfer all of these pokemons regardless their CP)

### DIFF
--- a/pokecli.py
+++ b/pokecli.py
@@ -65,7 +65,7 @@ def init_config():
     parser.add_argument("--maxsteps",help="Set the steps around your initial location(DEFAULT 5 mean 25 cells around your location)",type=int,default=5)
     parser.add_argument("-d", "--debug", help="Debug Mode", action='store_true')
     parser.add_argument("-t", "--test", help="Only parse the specified location", action='store_true')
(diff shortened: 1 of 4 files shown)


pokecli.py
```diff
@@ -65,8 +65,10 @@ def init_config():
     parser.add_argument("--maxsteps",help="Set the steps around your initial location(DEFAULT 5 mean 25 cells around your location)",type=int,default=5)
     parser.add_argument("-d", "--debug", help="Debug Mode", action='store_true')
     parser.add_argument("-t", "--test", help="Only parse the specified location", action='store_true')
+    parser.add_argument("-tl", "--transfer_list", help="Transfer these pokemons regardless cp(pidgey,drowzee,rattata)")
```
solderzzc: you need a default value, or is there an exception without -tl?

tienkhoi: Added

solderzzc:
Thanks for the PR.

PR #332, opened 2016-07-23
Customize how to display distance in the output with --distance_unit

Grace:
**Short Description:** 
Added the argument option --distance_unit to customize how to display distance in the output. E.g., km instead of meters, miles instead of meters.

**How to Use:**
- Use kilometers (km) for distance output
`python pokecli.py -a 'google' -u [user_email] -p [password] -l [location] -w 5 --distance_unit km`
- Use miles (mi) for distance output
`python pokecli.py -a 'google' -u [user_email] -p [password] -l [location] -w 5 --distance_unit mi`
- Use feet (ft) for distance output
`python pokecli.py -a 'google' -u [user_email] -p [password] -l [location] -w 5 --distance_unit ft`

**Fixes:**
- Implemented the argument option --distance_unit to customize how to display distance in the output. 
- The first line of requirements.txt to install pgoapi was changed to `-e git+https://github.com/tejado/pgoapi.git#egg=pgoapi` to correct a "branch not found" error when running `pip install -r requirements.txt`.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,4 +1,4 @@
--e git+https://github.com/tejado/pgoapi.git#egg=pgoapi
+-e git://github.com/tejado/pgoapi.git@1f25e907f3e5f1b603330e71041e1ad7bee7580f#egg=pgoapi
 geopy==1.11.0
 protobuf==3.0.0b4
 requests==2.10.0
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,4 +1,4 @@
--e git://github.com/tejado/pgoapi.git@1f25e907f3e5f1b603330e71041e1ad7bee7580f#egg=pgoapi
+-e git+https://github.com/tejado/pgoapi.git#egg=pgoapi
 geopy==1.11.0
 protobuf==3.0.0b4
 requests==2.10.0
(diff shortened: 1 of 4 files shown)


requirements.txt
```diff
@@ -1,9 +1,9 @@
--e git://github.com/tejado/pgoapi.git@1f25e907f3e5f1b603330e71041e1ad7bee7580f#egg=pgoapi
+-e git+https://github.com/tejado/pgoapi.git#egg=pgoapi
```
VirtualSatai: Is this change suppsed to be here?

douglascamata: I don't think so. @Grace we need to get `pgoapi` at that specific commit. Please don't remove the commit reference, otherwise ppl that install after/before others may get different versions of `pgoapi`.

douglascamata: With this fixed it's good to merge, IMHO.

solderzzc: We need fixed on specific version of pgoapi, before upgrade, test need to be done.

douglascamata:
Looks good. Merging. Thanks!

requirements.txt
Grace: Okay, I fixed it! Thank you all for letting me know.

PR #333, opened 2016-07-23
Added auto switching between capture and farm mode.

solderzzc:
Short Description: 
Added auto switching between capture and farm mode.
Fixes:
-  Stupid running without AI at all.
- 
-


VirtualSatai:
Only issue with this is that it will wait until the inventory is full before attempting to catch any pokemon again, and it will try to catch each pokemon in range before finally stopping. e.g.: 
```
[#] Something rustles nearby!
[#] A Wild Pidgey appeared! [CP 104]
[x] Out of pokeballs, switching to farming mode...
[#] A Wild Poliwag appeared! [CP 176]
[x] Out of pokeballs, switching to farming mode...
[#] Finished walking
```

solderzzc:
@VirtualSatai  You are correct. Need do more fine tune.

PR #340, opened 2016-07-23
[Fix] Multiple instances shouldn't fight over dominance of the outfiles

haykuro:
Short Description: Allows multiple instances of the bot to run, with different users, without having to fight over dominance of the `outfiles`.

### DIFF
--- a/pokemongo_bot/__init__.py
+++ b/pokemongo_bot/__init__.py
@@ -34,11 +34,11 @@ class PokemonGoBot(object):
         if (self.config.mode == "all" or self.config.mode == "poke") and 'catchable_pokemons' in cell:
             print '[#] Something rustles nearby!'
             for pokemon in cell['catchable_pokemons']:
-                with open('web/catchable.json', 'w') as outfile:
+                with open('web/catchable-%s.json' % (self.config.username), 'w') as outfile:
                     json.dump(pokemon, outfile)
                 worker = PokemonCatchWorker(pokemon, self)
                 worker.work()
-                with open('web/catchable.json', 'w') as outfile:
+                with open('web/catchable-%s.json' % (self.config.username), 'w') as outfile:
                     json.dump({}, outfile)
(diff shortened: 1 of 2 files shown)


solderzzc:
could you please explain how does work? I'm not a Python guy, I'm nodejs/cloud/mobile app/... guy.

haykuro:
If I run multiple instances of the script, and they all want to write to 1 file, e.g. `web/location.json` or `web/catchable.json` they will fight to write to the same file.

The more instances you run, the slower it will get because they want to constantly try to open/lock/write/release the file. It becomes a power struggle and will eat up CPU / IO read/writes.

This PR is sorta incomplete (i just realized what those files do), I will have to update the [`web/index.html`](https://github.com/PokemonGoF/PokemonGo-Bot/blob/dev/web/index.html#L53-L75) javascript to allow the user to perhaps define the usernames you're running?

solderzzc:
I suggest you use different folder or docker for multiple account, that's a good way to do cluster.

haykuro:
This would be in extension to Docker, not really a replacement..

It would be overkill to spin up multiple installations of the bot if 1 install is already capable of running as many as it wants (or as many as the allocated resources allow)..

The only thing preventing one from running multiple bots effectively is the use of the outfiles, and I see this as an enhancement for someone who wants to run 2 docker instances with 5 accounts on each one (as an example)

solderzzc:
The cost of docker is very small, we should keep logic simple for maintain.

haykuro:
No worries, this may be a personal preference.

We can close this for now.

solderzzc:
Thanks, let's discuss later.

MaxLeiter:
I see no downside to this; not everyone wants to use docker or have multiple instances (disk space and all that). :+1: from me

TheSavior:
The approach could be as simple as naming the files based on the username currently signed in.

MaxLeiter:
Did you check the modified files? That's basically what this PR is doing; it appends the username to the end of the file

PR #342, opened 2016-07-23
allowed no --location if location cache argument present

Devagamster:
Fixes:
- Requiring location argument even when location cache is specified


solderzzc:
great !

PR #351, opened 2016-07-23
Go to nearest pokemon/fort first

CapCap:
Short Description: 

Fixes:
- Problems with not visiting closest fort first

Adds: 
- Visiting nearest pokemon first


pokemongo_bot/stepper.py
```diff
@@ -5,6 +5,7 @@
 
 from math import ceil
 from s2sphere import CellId, LatLng
+import s2sphere
```
solderzzc: Is this still needed?

PR #353, opened 2016-07-23
Displays minutes as well as seconds in the output

Grace:
Short Description: 
Displays minutes as well as seconds in the output. If seconds are > 60.0, time is displayed in minutes.


solderzzc:
This looks good for me.

PR #354, opened 2016-07-23
ignore certain pokemon from being collected

mzupan:
Short Description: 

This allows the user to create a `./data/catch-ignore.yml` file that looks like

```
ignore:
  - Pidgey
  - Rattata
  - Pidgeotto
  - Spearow
  - Ekans
  - Zubat
```

It will remove the ignored pokemon from the cell data so it won't even collect.

The idea is we probably don't want to run up candy numbers on common pokemon

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -96,3 +96,4 @@ info.json
 inventory.json
 pokedex.json
 web/catchable.json
+data/catch-ignore.yml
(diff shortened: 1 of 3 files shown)


solderzzc:
Many thanks, this is wanted PR.

haykuro:
you need to add the import for `yaml` into `requirements.txt`

```
  File "pokecli.py", line 43, in <module>
    from pokemongo_bot import PokemonGoBot
  File "/Users/a206450328/Documents/Repositories/pgo/PokemonGo-Bot/pokemongo_bot/__init__.py", line 9, in <module>
    import yaml
ImportError: No module named yaml
```

PR #357, opened 2016-07-23
[ Feature ] multi user support

haykuro:
Allows you to run multiple instances of the bot and see the output correctly in the web interface.

Must configure the array found in [`index.html`](https://github.com/haykuro/PokemonGo-Bot/blob/feature/multi-user-support/web/index.html#L23-L28).

such as:
```
var users = [
  'user1',
  'user2',
  'user3'
];
```

# Preview:

out

# Fixes:

Not only does this let you monitor all your accounts on one map, it's also a fix for a big file i/o bug where if you run multiple bots on 1 machine you'll experience a race condition when they fight for the output files.

### DIFF
--- a/pokemongo_bot/__init__.py
+++ b/pokemongo_bot/__init__.py
@@ -73,11 +73,11 @@ class PokemonGoBot(object):
             # Sort all by distance from current pos- eventually this should build graph & A* it
             cell['catchable_pokemons'].sort(key=lambda x: distance(self.position[0], self.position[1], x['latitude'], x['longitude']))
             for pokemon in cell['catchable_pokemons']:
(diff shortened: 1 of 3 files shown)


TheSavior:
The need for listing an array of the users is so that the frontend knows the filenames to request. Ideally we can get rid of that need. What if we write the username to the file on startup, and listen for script exit to remove ourselves from that list?

solderzzc:
What will happen when this two account don't in one map cell, I just wondering lmao

solderzzc:
I'm ok with this commit. Many thanks :)

haykuro:
@solderzzc looks kinda like this:

image

PR #388, opened 2016-07-23
Corrected Pokeball usage

millejosh:
Short Description: 

Fixes:
- 
- 
- 

Corrected Pokeball issue preventing great/ultra ball usage


solderzzc:
without .value the value will not be int.

millejosh:
@solderzzc Yeah it messes with the readout when you first launch but it actually uses them correctly now

Still trying to figure out how to do both.

PR #398, opened 2016-07-23
Avoid using uninitilized variable

VirtualSatai:
Short Description: 
In some cases, if dist > 10, then the local variable position, would not be initilized, but still used. Since the global position (self.position) is updated in the if statement, we can simply use the self.position later on.

Fixes:
- Havn't made an issue for the problem.


solderzzc:
the description is very good. please help to review code:)

VirtualSatai:
@solderzzc Will do :)

PR #419, opened 2016-07-23
ensure cell has pokestops before trying to iterate over them

julesferreira:
`Uncaught TypeError: Cannot read property 'length' of undefined.`


jtdroste:
lgtm - merging.

Thanks!

brantje:
After this merge i don't see any pokestops at all

PR #426, opened 2016-07-23
Logger

brantje:
Short Description: 
Add a small logging system.
In the future we can extend this to use HD44780 lcd display's


jtdroste:
I'm not sure about this - I believe in the future we'll want to use / tie in the logging module.  Any other devs want to tie in?

brantje:
We still can use the Logging module.

instead of print we can use `Logging.log` and still have a possibility to drive LCD's with ease

brantje:
@jtdroste Ok if i add the lcd support?
Need to find away to let users configure the i2c address.

brantje:
Some examples

img_20160723_152854
img_20160723_152917

brantje:
If wanted, i can write a short how to setup the display.

pokemongo_bot/Logger.py
```diff
@@ -0,0 +1,23 @@
...
+
+def log(string, color = 'white'):
+    colorHex = ''
+    if(color == 'green'):
+        colorHex = '92m'
+    if(color == 'yellow'):
+        colorHex = '93m'
+    if(colorHex == 'red'):
+        colorHex = '91m'
+    if colorHex == '':
```
douglascamata: Please, consider replacing this big `if` logic with a dict.

douglascamata: And why not use the logging module? I want to use it properly because it's a default choice for logging in Python. We can create a wrapper around the logging module to log to LCD too, that's no problem.

brantje: Would be nice, can we discus that on slack?

douglascamata:
Please do not merge this yet, I'm going to work on @brantje fork to make it use the `logging` module without losing support for the LCD.

douglascamata:
he's going to rebase on dev and do some cleanup

PR #435, opened 2016-07-23
[README] Setting up a simple webserver

s1x:
Short Description:  Added description of setting up a simple webserver

### DIFF
--- a/README.md
+++ b/README.md
@@ -176,6 +176,27 @@ This project uses Google Maps. There's one map coupled with the project, but as 
         ```  
     5. You can now debug on the log to see if get what you need  
 
+## How to set up a simple webserver with nginx
+### Nginx on Ubuntu 14.x, 16.x
+#### 1. Install nginx on your Ubuntu machine (e.g. on locally or AWS)
+```
+sudo apt-get update
+sudo apt-get install nginx
+```
+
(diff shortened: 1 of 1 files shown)


jtdroste:
q_q

that removed my name from the contributors section.  Anyway, I think this will be better off in the wiki - maybe called "Advanced Installation" or something

ferbeb:
Just what I was looking for when I randomly clicked on this pull request. Certainly would not have found it in the wiki.

aditardy:
i start my webserver end change web directory but not see anyting just  option OAN and zoom
run multiple id on bot ... this posible to show on map .. ? or bug ..?

s1x:
@aditardy Thats releated to the map feature of this bot, not to the webserver itself. Check your .json for the correct username.

aditardy:
@s1x  thanks !   <3

haykuro:
FYI, nginx is one of many ways to setup a webserver..

Could even just do `python -m SimpleHTTPServer 8000` and it will run a server on port 8000.

s1x:
@haykuro Totally true!

solderzzc:
I'm confusing why this commit has no conflict, it was 12hrs ago !@!

PR #439, opened 2016-07-23
Pokeballs count fix.

necrozis:
Short Description: 
Fixing # get player balls stock
Fixes:
Fixing # get player balls stock


Shoh:
These same changes were made in #388 and I believe it got reverted a few minutes after it got merged.  Not sure about all the issues it causes but the comments seem to suggest the logging will be messed up since it's not an int

PR #456, opened 2016-07-23
change walking_hook with flush

xtrsyz:
Short Description: 
change walking_hook process with flush
Fixes:
- 
- 
-


solderzzc:
currently we don't accept pull request on master, all work are based on dev branch.
let me close this.
Walk_hook is not possible in dev branch since we don't track 3rd pgoapi for now, just download when use it.

PR #448, opened 2016-07-23
Advance Releasing Configuration

earthchie:
Sometimes filter desired pokemon with ``-iv`` is not enough. Some Pokemons have decent stats but their potential value is low because how the value distribted between stamina, attack, and defense. E.g. Charmeleon with IVs 15/15/0, its potential value is 0.67 which quite low - the truth is this dude is badass, you don't really need to care much about defense for this guy. He is fragile anyway xD
   
To config the whitelist, copy file ``pkmn_whitelist.json.example`` and rename it to  ``pkmn_whitelist.json``
Edit this file however you like, but keep in mind:

1. Pokemon name is always capitalize and case-sensitive
2. These settings will override ``-c`` and ``-iv``. Especially field ``any``, so use with caution.

image

PR #460, opened 2016-07-23
Changed IV defaults and added a VICP parameter.

05-032:
Short Description:  
The VICP parameter prevents the transfer of low IV Pokemon with high CP if the user so chooses. 
Changed the IV default to 0 so people are not surprised.

Fixes:
- Removed an extraneous print statement.


douglascamata:
We're freezing release logic feature in favour of https://github.com/PokemonGoF/PokemonGo-Bot/pull/448, which implements per pokemon release parameters. I'm working on it too.

05-032:
Oh, I see. That's certainly a lot better than what I have right now.

PR #473, opened 2016-07-23
[Untested] Increases fort movement efficiency

Cameronlund4:
at the beginning of the fort movement. This should make walking times between pokestops much
less in certain cases. (For example, on the old loop, if fort a was
2.5mi north, fort b was 2.7 mi north, and fort c was 2.6mi south, the
program would go a->c->b, which is insanely inefficient)

Note, unsure of the change on line 116. Without that change the current
code will get in an infinite loop once it goes through all forts it
knows, but might undo whatever that line was trying to do. Instead of
breaking, it just moves on to the next fort.


Cameronlund4:
Forgot to change tabs to spaces, whoops....

PR #474, opened 2016-07-23
[Untested] Increases fort movement efficiency

Cameronlund4:
at the beginning. This should make walking times between pokestops much
less in certain cases. (For example, on the old loop, if fort a was
2.5mi north, fort b was 2.7 mi north, and fort c was 2.6mi south, the
program would go a->c->b, which is insanely inefficient)

Note, unsure of the change on line 116. Without that change the current
code will get in an infinite loop once it goes through all forts it
knows, but might undo whatever that line was trying to do. Instead of
breaking, it just moves on to the next fort.


pokemongo_bot/__init__.py
```diff
@@ -100,17 +100,20 @@ def work_on_cell(self, cell, position, include_fort_on_path):
...
-                forts.sort(key=lambda x: distance(self.position[
-                           0], self.position[1], x['latitude'], x['longitude']))
-                for fort in cell['forts']:
+           
+                while len(forts) > 0:
+                    # Sort all by distance from current pos- eventually this should
+                    # build graph & A* it
+                    forts.sort(key=lambda x: distance(self.position[
+                               0], self.position[1], x['latitude'], x['longitude']))
+                    fort = forts[0]
```
solderzzc: Did it exclude the current one?

Cameronlund4:
@solderzzc If you mean does it replace the old loop, then yes. Otherwise not sure what you mean.

solderzzc:
I mean we need delete the current position from list, is that correct?

pokemongo_bot/__init__.py
```diff
@@ -110,17 +110,20 @@ def work_on_cell(self, cell, position, include_fort_on_path):
...
                     worker = SeenFortWorker(fort, self)
                     hack_chain = worker.work()
```
ChronoAS: Shouldn't you set hack_chain to a value after you remove a fort? Like moving a like 121 to line 124

solderzzc: NO, hack_chain is the record on server side , mean server know how frequently you search fort.

pokemongo_bot/__init__.py
```diff
@@ -110,17 +110,20 @@ def work_on_cell(self, cell, position, include_fort_on_path):
...
-                    if hack_chain > 10:
+                    if fort in forts:
+                        fort.remove(fort)
```
ChronoAS: forts.remove(fort)

then

solderzzc:
You should test it first them merge.

solderzzc:

PR #475, opened 2016-07-23
Add date and time to logger.

ChrisStory:
Short Description: 
Adds the date and time to the logger function.

I am not a Python programmer. The location of 'import time' may be incorrect, but Python isn't a language I typically use.
Fixes:
- 
- 
-

### DIFF
--- a/pokemongo_bot/logger.py
+++ b/pokemongo_bot/logger.py
@@ -1,3 +1,4 @@
+import time
 try:
     import lcd
     lcd = lcd.lcd()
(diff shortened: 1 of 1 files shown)


pokemongo_bot/logger.py
```diff
@@ -6,14 +7,18 @@
...
-
+def log(string, color = 'white'):
+    colorHex = {
+        'green': '92m',
+        'yellow': '93m',
+        'red': '91m'
+    }
     if color not in colorHex:
+        print(time.strftime("%Y-%m-%d %H:%M:%S"))
         print(string)
```
brantje: This will results in eg:
```
2016-07-23 20:59:33
string
```
Can you make like like 
```
[2016-07-23 20:59:33] string
```
Will be much cleaner then
Corrected code:
```
print(time.strftime("%Y-%m-%d %H:%M:%S") + ' '+ string)
```

ChrisStory:
Updated per brantje comment.

PR #477, opened 2016-07-23
Fixed logic problem when both CP and IV are specified

jmeyer2k:
Short Description: 
If both CP and IV are specified, it will throw out Pokemons even when one of the conditions doesn't match.

Fixes:
- If both CP and IV are specified, it will throw out Pokemons even when one of the conditions doesn't match.


fnutrica:
I think we should find a better way to do this one. If a really high CP pokemon appears I would want it to still be kept, even if IV is not high. So maybe a third parameter for really high CP to be kept and a CP-IV parameter for the behavior above.

fnutrica:
Also default IV should be 0.0 or 0.1

jmeyer2k:
Yeah, we'll be replacing this with the config options: "min_cp", "min_iv", and "cp_iv_logic".

See: https://pokemongo-bot.slack.com/archives/dev/p1469300379003015

PR #479, opened 2016-07-23
Added ignore init transfer argument

JohanJu:
Short Description: 

Added ignore init transfer argument


JohanJu:
run example it supports both name and number: 
-ign Pidgey,13,Rattata

PR #499, opened 2016-07-23
file should be ignored and used as example

surfaace:
release_config.json will be edited by all users


solderzzc:
@douglascamata  How do you think about?

tehp:
Agreed. Looks good to me.

solderzzc:
But without this, will the bot works normallY?

tehp:
Yeah, all the gitignore does is prevent git from tracking changes on the file.

solderzzc:
but without default value, will this work?
release_config.json → release_config.json.example

PR #502, opened 2016-07-23
Recycling unwanted new items when spinning Pokestops

dmateusp:
I added a "filter_list" parameter that allows users to pass unwanted items ID that should be recycled by the bot when received spinning PokeStops

Parameter / config structure:
      -if LIST, --item_filter LIST                  Pass a list of unwanted items to recycle when collected at a Pokestop (e.g, [\"101\",\"102\",\"103\",\"104\"] to recycle potions when collected)" 

this example filters out all potions:

"item_filter": ["101","102","103","104"]

I also updated the api of drop_item with an example of a good request for people that would like to recycle items or implement an inventory manager (they should check my example !)


solderzzc:
This is good for me.

kazyka:
looks good

skydrome:
#517

PR #523, opened 2016-07-23
README.md update for usage on --initial_transfer

MEGAMERICAN:
Short Description: 

--initial_transfer no longer used -c as an upper limit.

Fixes:
- 
- 
-


solderzzc:
we should also modify the cli usage.

PR #542, opened 2016-07-24
condition logic semantics fix

reddivision:
Short Description: 
I know this was just fixed only minutes ago, but the current fix doesn't sit well with me as it doesn't "fix" the conditionals themselves, just short-circuits them. There's the (not-so-realistic) danger that if someone wants special processing later on "within" the conditionals when those values are 0, that logic will never be hit. 

At any rate, if no one else cares or folks see this as pedantic, feel free to close! :)

Fixes:
- conditional logic for `release_config` key checks


solderzzc:

solderzzc:

PR #545, opened 2016-07-24
Renamed min_cp to release_under_cp and min_iv to release_under_iv

Grace:
Short Description: 
Renamed min_cp to release_under_cp and min_iv to release_under_iv in release_config.json to make their purpose clearer.


Calcyfer:
Nice, makes a lot more sense this way.

pokemongo_bot/cell_workers/utils.py
```diff
@@ -82,7 +82,9 @@ def format_dist(distance, unit):
 
 def format_time(seconds):
     # Return a string displaying the time given as seconds or minutes
-    if seconds <= 1.0:
+    if seconds <= 0.0:
+        return '{:.2f} seconds'.format(seconds)
+    elif seconds <= 1.0:
```
reddivision: this code doesn't seem to be tied to the PR purpose, but as a point of feedback, it seems the intent here is that if `seconds` is exactly equal to 1 it should be "second", but when less than 1 it should be seconds. i.e. `1.0 second` and `0.99 seconds`

PR #553, opened 2016-07-24
Fixed broken images on Web View when it's in a subdirectory

wblondel:
Fixed broken images on Web View when it's in a subdirectory


gcsoftware:
Confirmed to fix the broken images.

solderzzc:
thanks for the confirmation @gcsoftware

PR #556, opened 2016-07-24
Refactoring SeenFortWorker move logic into MoveToFortWorker

TheSavior:
Extracting some of the logic from SeenFortWorker related to moving towards the fort into MoveToFortWorker


solderzzc:
The restructure is good for me.

PR #559, opened 2016-07-24
Update main.js

gcsoftware:
Short Description: 

Fixes:
- Fixes broken path for item images when using web server.


gcsoftware:
Realized there's already an existing PR waiting to be approved that does same thing.

PR #564, opened 2016-07-24
Allow people to select another config file for multiple bot support

ColeGreenlee:


solderzzc:
we don't merge code into master this way, when we have a stable code base, I'll delete master and create a new one form that point.

PR #568, opened 2016-07-24
Merging release_config.json with config.json

TheSavior:
Short Description: Instead of having two required configuration files, we are merging them together.


ionTea:
Should perhaps consider adding in the catch-ignore.yml file as well to the config merge, keeping all the config files in one place. Especially since this might be considered a sort of "release configuration".

config.json.example
```diff
@@ -12,5 +12,31 @@
...
     "distance_unit": "km",
-    "item_filter": ["101","102","103","104"]
+    "item_filter": ["101","102","103","104"],
+    "release_config": {
+        "any": {
+            "release_under_cp": 400,
+            "release_under_iv": 0.8,
+            "cp_iv_logic": "or"
+        },
+        "My Pokemon": {
```
thellimist: This is not clear enough. Always put the default values in. Adding all the pokemon names would be more clear

thellimist:
We should write the config file just like [sublime text config files](https://github.com/bgreenlee/sublime-github/blob/master/GitHub.sublime-settings). Lots of comments with default value initially set. In my opinion having all the settings in a single config file would be to complex. A good structure could be

/PokemonGo/AdvanceSettings/catch-ignore.json
/PokemonGo/AdvanceSettings/release.json
...
/PokemonGo/config.js  (maybe less settings, only settings which are must for the user should be here)
```
{
    "auth_service": "google",
    "username": "EMAIL@gmail.com",
    "password": "PASS",
    "location": "LOC EXAMPLE",
    "gmapkey": "KEY",
    "debug": false,
    "test": false,
    "location_cache": false,
    "distance_unit": "km"
}
```

In the future AdvanceSettings can be renamed as /PokemonGo/Packages and people can write custom packages with custom package settings as `json` file. We should only use one type of setting file.

TheSavior:
@thellimist We need to have some approach that supports the direction of https://github.com/PokemonGoF/PokemonGo-Bot/issues/142

SteffanLong:
#766 wants to allow multiple config files. It would be simpler to have everything in one file for that PR.

ColeGreenlee:
Fixed that PR mentioned above (#766) here: https://github.com/PokemonGoF/PokemonGo-Bot/pull/771

paledrewnem:
Please do not! Keep release_config, but additional parameter in config.json which release_config it should use!

surfaace:
function may already be coming by @SteffanLong. if not, you can collaborate

SteffanLong:
If we're going to move config files, and merge them, we may as well merge three different ways of catching/releasing as well. I have a draft we're working on. Join the Slack so we can perfect it. :)

SteffanLong:
This is fully implemented in the dev branch now.

Issue #82, opened 2016-07-21
Evolving Pokemon


tellomichmich:
working.py :`

    def _transfer_low_cp_pokemon(api,value,pokemon):
        if 'cp' in pokemon and pokemon['cp'] < value:
                #print('need release this pokemon({}): {}'.format(value,pokemon))
                api.evolve_pokemon(pokemon_id=pokemon['id'])
                response_dict = api.call()
                #print('Response dictionary: \n\r{}'.format(json.dumps(response_dict, indent=2)))
                status = response_dict['responses']['EVOLVE_POKEMON']['result']
                if status == 1:
                        print('Evolved succesfuly !')
                else:
                        print('Evolve failed... Exchange !')
                        api.release_pokemon(pokemon_id=pokemon['id'])
                        response_dict = api.call()
                        #print('Response dictionary: \n\r{}'.format(json.dumps(response_dict, indent=2)))
                        print('Exchanged successfuly!')
`

pgoapi/protos/RpcSub.proto :`
    
    message EvolvePokemonRequest {
      optional fixed64 pokemon_id = 1;
    }

    message EvolvePokemonResponse {
      optional Result result = 1;
    
      enum Result {
        UNSET = 0;
        SUCCESS = 1;
        POKEMON_DEPLOYED = 2;
        FAILED = 3;
        ERROR_POKEMON_IS_EGG = 4;
      }
    }
`

Need to compile protobuf:
>protoc --proto_path=pgoapi/protos/ --python_out=pgoapi/protos/ pgoapi/protos/RpcSub.proto

Working :)

LOG :

> Walking from (XXX) to (XXX) for approx. 10.0sec
> Finished walking
> (XXX)
> Something rustles nearby!
> A Wild Pidgey appeared! [CP92]
> Captured Pidgey! [CP92] - exchanging for candy
> Evolved succesfuly !
> looping: step 7 of 100
> steplimit: 100 x: -1 y: -1 pos: 1 dx: 0 dy -1
> Walking from (XXX) to (XXX) for approx. 10.0sec
> Finished walking
> (XXX)
> Something rustles nearby!
> A Wild Rattata appeared! [CP81]
> Captured Rattata! [CP81] - exchanging for candy
> Evolve failed... Exchange !
> Exchanged successfuly!
> A Wild Staryu appeared! [CP134]
> Captured Staryu! [CP134]
> A Wild Spearow appeared! [CP66]
[...]

PR #570, opened 2016-07-24
Add Evolve (#82)

DavidK1m:
Short Description: 
Tries to evolve everything that it catches. If it fails (if it doesn't have enough candy), it checks if it should transfer the Pokémon.
Should add some settings to let user customize what to evolve.
Would be nice if there's a way to find out remaining candies.

Fixes:
- Inability to evolve a Pokémon

### DIFF
--- a/pokemongo_bot/cell_workers/pokemon_catch_worker.py
+++ b/pokemongo_bot/cell_workers/pokemon_catch_worker.py
@@ -134,21 +134,28 @@ class PokemonCatchWorker(object):
                                             pokemon_potential
                                         ), 'green'
                                     )
-                                    if self.should_release_pokemon(pokemon_name, cp, pokemon_potential, response_dict):
-                                        id_list2 = self.count_pokemon_inventory()
-                                        # Transfering Pokemon
-                                        pokemon_to_transfer = list(
-                                            Set(id_list2) - Set(id_list1))
-                                        if len(pokemon_to_transfer) == 0:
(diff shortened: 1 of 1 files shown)


solderzzc:
Great, this is finally coming, is my README.me GOOD? or you figure it out all by yourself?
Have you already join our slack, would you mind drop me a PM.

pokemongo_bot/cell_workers/pokemon_catch_worker.py
```diff
@@ -134,21 +134,28 @@ def work(self):
...
-                                            Set(id_list2) - Set(id_list1))
-                                        if len(pokemon_to_transfer) == 0:
-                                            raise RuntimeError(
-                                                'Trying to transfer 0 pokemons!')
-                                        self.transfer_pokemon(
-                                            pokemon_to_transfer[0])
-                                        logger.log(
-                                            '[#] {} has been exchanged for candy!'.format(pokemon_name), 'green')
+                                    pokemon_to_transfer = list(Set(id_list2) - Set(id_list1))
+                                    self.api.evolve_pokemon(pokemon_id=pokemon_to_transfer[0])
```
TheSavior: Does it make sense to pull this out into a new worker called EvolvePokemonWorker? It would make this significantly easier to refactor later.

solderzzc: Yes, I think you are correct @TheSavior

PR #574, opened 2016-07-24
Added evolve all to script

guyz:
Short Description: 

Can now run evolve all pokemons on bot startup. This is set to false by default, but can be changed using the -ev/--evolve_all argument. Some notes:
- Takes awhile (but significantly faster than doing it manually).
- Caches pokemons that cannot be evolved and then skips any subsequent versions of them.
- Really great after popping a lucky egg!


solderzzc:
You forgot to change the Contributor...
Great PR.

guyz:
Thanks - sent another PR in with the README change!

Issue #573, opened 2016-07-24
Proposal: Enable multiple configuration files and overriding from CLI


TheGoldenXY:
The command line and the parameters is very essential for linux users.

https://github.com/PokemonGoF/PokemonGo-Bot/commit/647bac8322dcebcccc116c81343d8378445236ff
I'd rather prefer using the command line. A lot of people are confused of the "new way". Please add the old way back and make it default, also add an option inside `config.json` to enable the `release_config.json` for those who rather want to use that.

TheSavior:
As stated in Slack:

It seems like as this bot grows, it is going to be more and more configurable. At a certain point, it doesn't make sense to configure everything through cli args and will have to move to something like json configuration. 

If the problem is that for multiple bots, the only real way is via command line args, we can probably find a way to fix that.

I could imagine a cli flag that passed a path to a config file. Would that solve the problem at hand?

Perhaps an approach would be to pass either a cli arg that tells it what config file to use, or there is a command line arg that take a string of the json it should use as config

TheGoldenXY:
@TheSavior Sure, but if people wants to use the parameters, make the parameters (i.e. `-c -iv`) overwrite the config file.

curtis2447:
I think having both would just add more functionality, if someone wants to run 3-4 accounts at once using .bat files and parameters it is easier than having 4 separate bots

TheSavior:
I think that is reasonable, but how would you handle overwriting deeply nested configuration? For example, if we merged release_config into config and you wanted to overwrite a field like this: https://github.com/TheSavior/PokemonGo-Bot/blob/0d44885fcd8d710e54b718d3468094965801d56d/config.json.example#L18

Perhaps 

`--release_config.any.release_under_cp 300`

TheGoldenXY:
@TheSavior Great statement. How about `--release_config.any.cp 200` and `--release_config.any.iv 300`?

TheSavior:
One of the downsides of that approach is that it may not make sense to support the shorthand cli args anymore so that the configuration can be defined in only one place `config.json` and it becomes clear in code how they are being overridden. What are your thoughts on that?

The result would be that it would be very clear what the command line flags are, because they match the config.json perfectly, but it would mean that the length of the command to run would be longer.

TheGoldenXY:
@TheSavior What do you recommend then?

TheSavior:
Here is what I propose:

[...]

PR #580, opened 2016-07-24
Multiple Config Files Functionality

ColeGreenlee:


TheSavior:
We should hold off on this until #573 is figured out.

ColeGreenlee:
The people arguing about #573 did agree that having an optional parameter to use a config file would be a good idea. This doesn't require a config file, and will fall back on params if one isn't present.

solderzzc:
looks good for me. Nothing broken.

PR #585, opened 2016-07-24
Evolve captured only if user wants

DavidK1m:
Short Description:
adds -ec to let user choose if they want to evolve everything they capture.

Fixes:

Bot evolving no matter what


solderzzc:
Heavily wanted, before we have better AI.

PR #596, opened 2016-07-24
Fix for .split(',') error in pokecli.py

petercunha:
The bot is refusing to start up due to a syntax error in pokecli.py that involves attempting to split a list. This commit solves the problem.


reddivision:
this might need discussion... isn't the new syntax for item_filter just a string of comma-separated IDs? if so, the split is necessary.

realwork.txt
```diff
@@ -0,0 +1 @@
+Sun Jul 17 00:00:00 2016  -0400454721
```
TheSavior: This file isn't needed.

petercunha: Yeah accidentally ran greenhat.py in the wrong directory. Tons of junk in here now.

PR #597, opened 2016-07-24
Refactor: Pull out initial transfer into a worker

TheSavior:


pokemongo_bot/cell_workers/pokemon_catch_worker.py
```diff
@@ -33,7 +33,9 @@ def work(self):
                 if 'status' in response_dict['responses']['ENCOUNTER']:
                     if response_dict['responses']['ENCOUNTER']['status'] is 7:
                         logger.log('[x] Pokemon Bag is full!', 'red')
-                        self.bot.initial_transfer()
+                        worker = InitialTransferWorker(self.bot)
```
TheSavior: This is the one part that bothers me. Ideally, workers shouldn't be calling workers. If we were doing behavior trees, the pokemon_catch_worker wouldn't call this when it gets full, instead it would just fail out and on the next tick, a TransferWhenFull task would be run before this one, making sure this never gets hit when the bag is full. 

I think it is fine to leave this here for now but be aware that this will need to change. Happy to change this though if someone has a great idea.

reddivision: what about implementing a return code for this case that you can handle in the if-block at around line 69 in `pokemongo_bot/__init__.py`?

TheSavior: great call

solderzzc:
lol, @s1x is still working :)

solderzzc:
I'd like to merge.

TheSavior:
Followed up with the recommended refactor in #603

PR #598, opened 2016-07-24
fix error when user first login

Xuefeng-Zhu:
Short Description: 
User will get an error when login in first time, since there is no experience in profile

Fixes:
- 
- 
-


solderzzc:
Yes, you are python expert.

Issue #437, opened 2016-07-23
Bot spinning gyms?


go-on-git:
Noticed that the bot will 'spin' gyms and get items. Is this possible with the app? If not it should be removed to prevent ban detection.

(Using master)

millejosh:
@go-on-git Can you paste the log where it is spinning gyms?

dddbliss:
+1

Don't have a log available, but noticed that the bot would spin a gym near my house. Since it is the only nearby fort there was no confusion.

go-on-git:
I don't have exact log but grabbed the coords:
42.267592, -83.027075 fort ID (33f17840024f4ba9bc1e25e9ece4cc2f.16) is a gym
42.266746, -83.027786 fort ID (c0e6653e57d54f64b705d3f7fa2f4435.16) is a stop

Not sure if fort ID is useful or if unique to each process but included it just in case.

boivie:
Got some logs:

```
[#] Found fort 5daf0b3c8c7d45be86bf2c7c3fa2054f.11 at distance 0.03km
[#] Need to move closer to Pokestop
[#] Walking from (55.603532, 13.01478) to (55.601481, 13.016047) for approx. 1.27 minutes
[#] Finished walking
[#] Arrived at Pokestop
[#] Now at Pokestop: Ängel och fåglar - Spinning...
[+] Loot:
[+] 50 xp
[+] 3x Pokeball
[#] PokeStop on cooldown. Time left: 4.98 minutes
```

Teiby1:
Same here.
Gym:
```
[#] Found fort 0ec80690643349ab99f4bfc473ca2dda.16 at distance 0.07km
[#] Need to move closer to Pokestop
[#] Walking from (48.047734140154596, 8.21031510128687) to (48.047535, 8.209441) for approx. 20.00 seconds
[#] Finished walking
[#] Arrived at Pokestop
[#] Now at Pokestop: Furtwangen Speicher - Spinning...
[+] Loot: 
[+] 50 xp
[+] 4x Pokeball
[#] PokeStop on cooldown. Time left: 4.98 minutes
```

Pokestop:
```
[#] Found fort d16d818b0c2a44e0be0261775bae683a.16 at distance 0.06km
[#] Need to move closer to Pokestop
[#] Walking from (48.047535, 8.209441) to (48.048055, 8.210982) for approx. 37.00 seconds
[#] Finished walking
[#] Arrived at Pokestop
[...]

PR #605, opened 2016-07-24
stop spinning gyms

reddivision:
Short Description: 
Gyms are currently being treated as pokestops (and somehow we're getting items from "spinning" them). Putting an end to this nonsense.

Fixes:
- Spinning gyms for items

ref #437


solderzzc:
oh, yes, you are right. This is stupid.

PR #637, opened 2016-07-24
zero means out of stock ;)

jdeniau:
Short Description: 
if pokeball sock is 0, it mean you need to use other balls


eggins:
haha - just fixed this up man

jdeniau:
You are too quick :o !

PR #638, opened 2016-07-24
Fixing ITEM_FILTER and Attribute Error: no position Issues

dmateusp:
Fixes:
- lots of users were using the syntax "101,102,103" instead of the correct syntax ["101","102","103"]
- a commit changed the correct syntax to the string syntax without changing the logic which made item_filter throw every item away
- I changed the logic to accept both syntaxes in a "correct" way now
- fixed character escaping in the README
- fixed a new issue that was preventing cached_location from being used when an initial location was given

Please note that I implemented this feature so my commit is more relevant than other fixes that do not understand the code


dmateusp:
please don't merge, need to change something

solderzzc:
we won't, we need fix CI error first.

dmateusp:
Now working @solderzzc just notifying, I know you need to fix CI errors first

README.md
```diff
@@ -163,7 +163,7 @@ This project uses Google Maps. There's one map coupled with the project, but as
       -it, --initial_transfer                       Transfer all duplicate pokemon with same ID on bot start, except pokemon with highest CP. Accepts a number to prevent transferring pokemon with a CP above the provided value.  Default is 0 (aka transfer none).
       -ms, --max_steps MAX_STEP                     Set the steps around your initial location (DEFAULT 5 mean 25 cells around your location)
       -iv IV, --pokemon_potential                   Set the ratio for the IV values to transfer (DEFAULT 0.4 eg. 0.4 will transfer a pokemon with IV 0.3)
-      -if LIST, --item_filter LIST                  Pass a list of unwanted items to recycle when collected at a Pokestop (e.g, \"101,102,103,104\" to recycle potions when collected)"
```
solderzzc: We can not keep changing the config.json format time by time... It's huge work to solve users question in slack help channel.

dmateusp:
@solderzzc please see update

README.md
```diff
@@ -163,7 +163,7 @@ This project uses Google Maps. There's one map coupled with the project, but as
...
+      -if LIST, --item_filter LIST                  Pass a list of unwanted items to recycle when collected at a Pokestop (e.g, SYNTAX FOR CONFIG.JSON : ["101","102","103","104"] to recycle potions when collected, SYNTAX FOR CONSOLE ARGUMENT : "101,102,103,104" 
       -d, --debug                                   Debug Mode
       -t, --test                                    Only parse the specified location
```
solderzzc: The console config is working? Now I only use pycli -fc config.json, is other params acceptable ?

dmateusp: actually let me redo the whole pull request, the CI not working earlier made me do changes I should not have done

README.md
dmateusp: it's not changing it, people were still using it both ways, but one way was causing unexpected behavior

README.md
solderzzc: sure.

dmateusp:
@solderzzc Here you go :) down to 2 commits, I tested it earlier but can't test it now as dev branch is unstable. The changes don't change the way arguments work, it just corrects unwanted behavior for users that misunderstood how to use the feature

README.md
dmateusp: Can't test the argument because of the state of the dev branch but config works both ways

solderzzc:
Will it break current users' configuration ? If so, need add it to the breaking thing in README.

dmateusp:
@solderzzc it won't, see the initial configuration was ["101","102","103","104"] as an example to filter out potions, but people used it as "101,102,103,104" but the app would still run and filter out everything. So I just changed the logic so both syntaxes are now correct and have the expected behavior

pokemongo_bot/__init__.py
```diff
@@ -324,14 +329,12 @@ def _set_starting_position(self):
...
-                if not self.config.location:
-                    sys.exit(
-                        "No cached Location. Please specify initial location.")
-                else:
-                    pass
+                sys.exit(
+                    "No cached Location. Please specify initial location.")
 
     def _get_pos_by_name(self, location_name):
         # Check if the given location is already a coordinate.
```
solderzzc: release_config.json.example → release_config.json
This should not be tracked.

Invitation sent, please join our team.

dmateusp: Thank you :)

solderzzc: You are more than welcome.

PR #641, opened 2016-07-24
Lcd support

brantje:
Short Description: 
Better integration with lcd's.
When users specify the i2c address of the lcd in config, it will show stats and print output to the display.

My LCD is on i2c bus 0x23 which is 35 as an int
```
"lcd_address": "35"
```

### DIFF
--- a/pokecli.py
+++ b/pokecli.py
@@ -42,7 +42,7 @@ if sys.version_info >= (2, 7, 9):
 from getpass import getpass
 from pokemongo_bot import PokemonGoBot
 from pokemongo_bot.cell_workers.utils import print_green, print_yellow, print_red
-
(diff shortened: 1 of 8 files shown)


TJSoler:
is it possible that this generates conflicts for all the systems that don't have smbus installed? 

```
python2 pokecli.py                                        *[dev]
  File "pokecli.py", line 46, in <module>
    from pokemongo_bot import lcd
  File "/Users/username/Projects/PokemonGo-Bot/pokemongo_bot/lcd.py", line 14, in <module>
    import smbus
ImportError: No module named smbus
```

I'm using an osx that can't compile smbus.

brantje:
Pushing fix

PR #642, opened 2016-07-24
Only check for syntax errors

TheSavior:


TheSavior:
Declining. Some people are working on fixing the linter errors. That is a much better fix.

PR #644, opened 2016-07-24
Fix ImportError

brantje:
Mentioned in https://github.com/PokemonGoF/PokemonGo-Bot/pull/641#issuecomment-234771163


pokemongo_bot/cell_workers/pokemon_catch_worker.py
```diff
@@ -163,7 +163,7 @@ def work(self):
                                     )
                                     if hasattr(self.config, 'lcd'):
                                         self.config.lcd.message(
-                                            'Captured {}! [CP {}] [IV {}] - Checking Release Config'.format(
```
TheSavior: Please don't change lines like this when you are trying to fix a broken build.

PR #657, opened 2016-07-24
save ultra ball for super hight cp pokemon

tranhuanltv:
We don't want to miss a super strong pokemon, so temp solution for this first

### DIFF
--- a/pokemongo_bot/cell_workers/pokemon_catch_worker.py
+++ b/pokemongo_bot/cell_workers/pokemon_catch_worker.py
@@ -86,22 +86,29 @@ class PokemonCatchWorker(object):
                         while(True):
                         
                             pokeball = 1 # default:poke ball
+                            usable_balls_stock = balls_stock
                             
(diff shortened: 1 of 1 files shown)


pokemongo_bot/cell_workers/pokemon_catch_worker.py
```diff
@@ -86,22 +86,29 @@ def work(self):
                         while(True):
                         
                             pokeball = 1 # default:poke ball
+                            usable_balls_stock = balls_stock
                             
-                            if balls_stock[1] <= 0: # if poke ball are out of stock
-                                if balls_stock[2] > 0: # and player has great balls in stock...
+                            usable_balls_stock[3] -= 10 # save 10 ultra ball for super hight cp pokemon
```
thellimist: This should be added to config file.

thellimist:
In config it `"reserve_ultra_ball_for_high_cp": true` would be  a good check for people to decide to use this feature or not

ProjectBarks:
This need to be a config option also you are missing logic to say what level to save them for

PR #659, opened 2016-07-24
Print walking step coordinates

sinistance:
Add new configuration to print walking step coordinates. This will be helpful if you don't use any gui.


sinistance:
need cleanup

PR #679, opened 2016-07-24
Update main.js

jadbalout:
Short Description:  When you have seen the pokemon but not captured it, Times Caught: undefined occurs.

**# Fixes:**
Times Caught: undefined fixed, now Times Caught: 0


Reaver01:
https://github.com/OpenPoGo/OpenPoGoWeb
moved to submodule

PR #683, opened 2016-07-24
Updated README and example files

surfaace:
* Fixed readme so it reflects application switches
* Added example files if someone doesnt reads README


ProjectBarks:
We have made so many changes that this has pretty much been already outdated

PR #687, opened 2016-07-24
new dockerfile to use config.json and release_config.json

mercuriete:
Short Description: 
Now we use config.json to start application.
With this dockerfile we can provide this 2 files using ONBUILD COPY.

Fixes:
- Use of config.json and release.json
- Alpine image to get the image more slim (from 200 MB to 93 MB)
      you can check it there -> https://hub.docker.com/r/mercuriete/pokemongo-bot/tags/
- -u parameter to fix python issue https://github.com/docker/docker/issues/12447

Suggestion: please dont use git dependencies to pgoapi because this leads in a dependency to git and to GCC libraries.


mercuriete:
Please check travis.
Travis is givin this as invalid but if you see the logs is not my fault.

solderzzc:
this is not good for build a public version of Docker Image.
we can't publish image with our own config.json.
need another way to do it.
Use volume to manage config file will be ok.

Feel free to reopen when ready for review again.

PR #701, opened 2016-07-24
1 hour after we cant do anyting: Temp fix

unequaled86:


MikeDX:
Interesting option, but a bit of a hack...

When we get the key from login, I'm sure it has an expiry time. Wouldn't it be better to check that on each loop instead?

PR #704, opened 2016-07-24
Update inventory after spinning pokestop or catch pokemon

piersimak:
Auto update inventory-user.json  after spinning pokestop or catch pokemon


ProjectBarks:
As far as I can tell this has already been resolved. I am marking as closed

PR #705, opened 2016-07-24
After 1 hour relogin

unequaled86:
Fixes:
- because after 1 hour bot gonna stuck loop

### DIFF
--- a/pokemongo_bot/__init__.py
+++ b/pokemongo_bot/__init__.py
@@ -10,6 +10,7 @@ import sys
 import yaml
 import logger
 import re
+import time
 from pgoapi import PGoApi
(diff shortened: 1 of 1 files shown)


MikeDX:
Would be much better to check the expiry time of the login token and relogin when that has failed rather than some arbitrary time which may be wrong is the auth token expires sooner

MikeDX:
Just tested this, and it doesn't work since you are setting the statedtime to the current time each loop, so it will never reach startedtime+60*60.

unequaled86:
i tested man its worked. startedtime setted on bot login

MikeDX:
It might work but it is a bodge. I've submitted a correct patch that checks the session time, which then refreshes once the token is about to expire.

ProjectBarks:
#735 fixes this but thanks

PR #706, opened 2016-07-24
add signal/event systems

namlehong:
Short Description: 
Implement signal system.

### DIFF
--- a/test_receiver.py
+++ b/test_receiver.py
@@ -1,0 +1,35 @@
+#!/usr/bin/env python
+# -*- coding: utf-8 -*-
+
+from dispatcher import Signal, receiver
+
+sign_in_signal = Signal(providing_args=['player', 'instance'])
+
+
+class MyBot(object):
+    def __init__(self, player=None):
+        if not player:
+            player = {}
+        self.player = player
+
+    def login(self):
+        sign_in_signal.send(sender=self.__class__, instance=self, player=self.player)
+
+
+def sign_in_handler(sender, instance, player, **kwargs):
+    print 'sign_in_handler'
+    print 'sender', sender
+    print 'instance', instance
+    print 'actor', player
+
+
+@receiver([sign_in_signal], sender=MyBot)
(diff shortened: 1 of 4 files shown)


test_receiver.py
```diff
@@ -0,0 +1,35 @@
+#!/usr/bin/env python
```
TheSavior: If this is meant to be a test, can you write this in a way that it will actually pass / fail on travis? Aka, make it a unit test instead of printing things and requiring manual work to verify that it is printing the right stuff.

TheSavior:
I'm concerned that a dispatcher + event system will be very incompatible with the longer term approach of using behavior trees. 

It might be compatible, but I'm not sure how we intend to use this / what we intend to dispatch. 

I'll talk to @namlehong offline about this and see if I can understand better.

TheSavior:
From @namlehong:

```
For example, a sign_in_signal

people can add ton of receiver like
- check inventory
- do transfer old pokemons

then 
pre_catch_pokemon:
- check pokeball

post_catch_pokemon:
- do transfer or not
- update pokeball count
```

```
One of the typical problems with hooks is that you end up having to have a bunch of logic to handle people who want to order hooks, or hooks that need to dispatch other things and thus call more hooks which call more hooks.
```

I'm worried that this event system will actually cause more problems than it solves.

TheSavior:
It has been mentioned in slack that the intention for the event system is for things like the logger. Which would provide plugin level support for something like the lcd display and the web ui.

I think that is a reasonable decision, but it becomes very important that we do not trigger bot *behavior* with events. For example, we should *not* have events like in the above snippet where we send an event on `pre_catch` that checks the pokemon inventory and transfers old pokemon. This type of logic should be handled by the behavior tree.

test_receiver.py
namlehong: @TheSavior :( sorry about my laziness.
@ProjectBarks may you be my reinforcement?

ProjectBarks:
this isn't completed but will be added in the future

PR #709, opened 2016-07-24
Add try/catch to file read operations

Shoh:
Short Description: 

Fixes:
- Add try/catch where necessary so it doesn't crash on startup
- Various cleanup


douglascamata:
Catching all exception is a code crime. You should just check if the file exists using the `os` module.

Shoh:
I was using os.path.isfile before but I changed it to just try/catch because that is how it has been used in other places.  For things that aren't absolutely necessary (like this map feature), I think silently continuing is fine.

Also, when you overwrote all of my changes yesterday - that was a crime. :)

solderzzc:
lol :) so what's the decision now >>>>>>>>>>

Shoh:
If it needs to be changed to use os.path, I'm fine with it since that is what I had it as before.  Like I said, I only changed it to try/catch because that is how I saw it being done in other files that don't import os module.

solderzzc:
Oh, let's follow @douglascamata 's instruction. If you don't mind.

Shoh:
Updated to use os.path.isfile

solderzzc:
nice, thanks, just viewed all code.

Shoh:
Thanks @solderzzc.  I'm probably going to work on adding back my previous logic for saving Great Balls/Ultra Balls for more important pokemon but I'm going to make it configurable.  I don't like the current implementation because it wastes all of my ultra balls :(

lennartc4:
There is a problem with this pull. The ` if os.path.isfile(user_web_location):` causes that there is no location file created for new trainers. You have to create it by hand now otherwise the location ist not written and can't be seen in the webinterface.

PR #712, opened 2016-07-24
for travis-ci to check pylint on .py files

surfaace:
checks for files recursivly, should not cause error code 30 on exit


TheSavior:
I think this caused problems. All travis builds are passing even though the linter is failing.

PR #723, opened 2016-07-24
Fix missing comma in config.json.example

krook1024:
Short Description: There was a missing comma in the config.json.example file.

Fixes:
- missing comma in config.json.example


solderzzc:
great catch.

PR #726, opened 2016-07-24
Lucky Egg - Evolve All

codybaldwin:
Short Description: 
Provides the ability to use a lucky egg before evolve_all.  This will maximize xp gain.
Fixes:
- Provides option via config.json
- If user has "use_lucky_egg" set to true and they have no lucky eggs, it will skip evolve_all


solderzzc:
I voted +1 for the feature and a quick code review.

codybaldwin:
Attempting to merge commits

PR #727, opened 2016-07-24
change the access to the response hash

sebyx07:
The api from pokemon go changed, but the data from the GET_INVENTORY is gone

### DIFF
--- a/pokemongo_bot/__init__.py
+++ b/pokemongo_bot/__init__.py
@@ -147,28 +147,28 @@ class PokemonGoBot(object):
         currency_1 = "0"
         currency_2 = "0"
 
-        player = response_dict['responses']['GET_PLAYER']['player_data']
+        player_profile = response_dict['responses']['GET_PLAYER']['profile']
 
         # @@@ TODO: Convert this to d/m/Y H:M:S
         creation_date = datetime.datetime.fromtimestamp(
-            player['creation_timestamp_ms'] / 1e3)
+            player_profile['creation_time'] / 1e3)
 
         pokecoins = '0'
         stardust = '0'
         balls_stock = self.pokeball_inventory()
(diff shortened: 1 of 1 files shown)


solderzzc:
I'm not quite sure what this for..

solderzzc:
👍

TheSavior:
I don't understand. If this stuff was changed, wouldn't we all see the broken behavior? Mine prints my account creation time on startup implying this is a non issue.

solderzzc:
From my side, the print has error.
Should we try to protect both working?
I'm using PTC account.

solderzzc:
`
[2016-07-24 18:16:29] 
  File "pokecli.py", line 222, in <module>
    main()
  File "pokecli.py", line 208, in main
    bot.start()
  File "/Users/simba/Projects/pokemon/PokemonGo-Bot/pokemongo_bot/__init__.py", line 32, in start
    self._setup_api()
  File "/Users/simba/Projects/pokemon/PokemonGo-Bot/pokemongo_bot/__init__.py", line 153, in _setup_api
    player = response_dict['responses']['GET_PLAYER']['player_data']
KeyError: 'player_data'
`

TheSavior:
I'm using PTC too. We should get more input from people on slack. I'd be worried that we fix this for a few people and break this for many.

solderzzc:
@sebyx07  Could you changed to support both format, I guess the servers will not change at the same time.

solderzzc:
It caused by existing  pgoapi
`
rm -rf pgoapi/
`

solderzzc:
It caused by switching from main branch without rm pgoapi.

 rm -rf pgoapi/ 
Solved my issue, please reopen if still an issue.
Let me close here.

PR #728, opened 2016-07-24
Removed web/index.html GMAP updating

zynjec:
Removes regex search to replace GMAP key, destroys page and that projects handles keys differently now.


SteffanLong:
Also fixed in #730

PR #748, opened 2016-07-25
Evolving based on IV first, then CP

guyz:
Short Description: 

Instead of evolving pokemons by their CP, evolving is now done by IV first. To prevent situations where ridiculously low CP pokemons with high IVs are evolved, there is a new optional config value `cp_min` (default: 300). This is used as a threshold - first, all larger (than the threshold) CP pokemons are evolved based on their IV. Then, if possible, lower CP pokemons are evolved based on their CP alone, since presumably IV doesn't matter that much then.


solderzzc:
Please have example config as your suggestion for the users.

solderzzc:
Please have the conflicts resolved then open :)

PR #757, opened 2016-07-25
Run Configs From a 'config' Directory

ColeGreenlee:
People are able to throw all of their test configs into this folder and
they are ignored whilst doing a commit, minimizing the risk that someone
uploads a password-containing file to GitHub. Also moved and refactored where the program looks fro the release config file.


solderzzc:
CLI/Config part, we'd better keep it as now. thanks for the PR.

PR #761, opened 2016-07-25
only_track_filter: whitelist of pokemon ids to track and catch

julesferreira:
defaults to normal tracking/catching behavior. when list of ids is supplied, only those corresponding pokemon will be tracked.

uses:
- drop bot in abra nest with `"only_track_filter": "63,64,65"` and everything other than abra, kadabra, and alakazam will be ignored
- when completing pokedex, insert ids of missing pokemon

### DIFF
--- a/pokecli.py
+++ b/pokecli.py
@@ -126,6 +126,14 @@ def init_config():
         default="km")
 
     parser.add_argument(
+        "-otf",
+        "--only_track_filter",
+        help=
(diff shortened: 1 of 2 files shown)


solderzzc:
Please modify README.me as well for the usage. and the contributor.
An example or suggested config is also welcomed.

solderzzc:
Please help to update the README.me, to have this feature documented well. This is a killer feature.

SteffanLong:
This is fully implemented in the new config.json setup that is on dev.

PR #771, opened 2016-07-25
Multiple Config File Support

ColeGreenlee:
Adds ability to drop config files into the 'configs' dir whilst developing and have them ignored by git


paledrewnem:
Does it support run bot with multiple configs as once?

ColeGreenlee:
The bot can currently use any .json file to run, all this PR will do is move all the configs to a 'configs' directory and have your commits ignore them, while still not ignoring the example configs.

surfaace:
we are going to make changes to config.json and release_config.json. once those are updated, please review this PR again and make the updates necessary for it to work.

ColeGreenlee:
Even with changes, this PR shouldn't affect anything but help developers test on different configs. I can make changes accordingly once we merge the config files if it does affect anything

.gitignore
```diff
@@ -114,3 +114,8 @@ data/catch-ignore.yml
 release_config.json
 web/userdata.js
 location.json
+
+#Multiple config
+configs/
+!configs/config.json.example
+!configs/release_config.json.example
```
dmateusp: shouldn't the config examples be added to the commits ?

SteffanLong: That's what this does. Ignore the config directory, but allow the two example files.

ColeGreenlee: ^Yep!

surfaace: but you dont comit in the config and example files, you just delete them from directory with this PR

dmateusp: exactly ^    :p

ColeGreenlee: Oooops! I thought I had moved those two to the configs folder for his PR. If I did that could we move forward with this PR?

dmateusp: I think once you have both your files in the configs folder I don't see why we couldn't merge

dmateusp: @ColeGreenlee adding a warning for people so they move their config.json files to the new directory would be pretty good as well.

Like trying to open it on configs directory first and if not there adding a Warning print "config.json not found, please add one to the configs directory"

ColeGreenlee:
Okay, example files have been re-included. Was just a missing * form the .gitignore :+1:

.gitignore
ColeGreenlee: Added warnings in latest commit. Thanks for the suggestion!

ColeGreenlee:
Added the warnings as suggested. Is this PR good to go?

dmateusp:
waiting for a second review but it should be good to go !

eggins:
+1

ColeGreenlee:
Thanks!

douglascamata:
@ColeGreenlee can I ask for another PR with information about new folder for config files in README?

ColeGreenlee:
Yeah, I'd be more than happy to document this.

TSM-EVO:
Please do document this. What are the arguments for running configs from this /configs/ folder

PR #780, opened 2016-07-25
When possible uses a lucky egg

ethervoid:
Short Description: 

Let the user define if he want to use lucky eggs every 30 minutes. The worker could be used to activate for example inciense or other inventory elements

Status:

Right now is currently under development but I want to make the PR to show the works and don't duplicate efforts

### DIFF
--- a/pokemongo_bot/cell_workers/inventory_actions_worker.py
+++ b/pokemongo_bot/cell_workers/inventory_actions_worker.py
@@ -1,0 +1,47 @@
+import json
+import datetime
+import os.path
+
+from pokemongo_bot.human_behaviour import sleep
+from pokemongo_bot import logger
+
+
+class InventoryActionsWorker(object):
+
+    def __init__(self, bot):
+        self.config = bot.config
+        self.pokemon_list = bot.pokemon_list
+        self.api = bot.api
+        self.inventory_options_file = 'data/inventory_options'
+
+    def work(self):
+        if self.config.use_lucky_egg and (self.config.mode == "all" or
+                                          self.config.mode == "poke"):
+            self._use_lucky_eggs()
+
+    def _use_lucky_eggs(self):
+        lucky_egg_file = 'data/last-lucky-egg-use-{0}.json'.format(self.config.username)
+        if os.path.isfile(lucky_egg_file):
+            with open('data/last-lucky-egg-use-%s.json' %
+                    (self.config.username)) as f:
+                last_use_json = json.load(f)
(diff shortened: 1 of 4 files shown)


pokemongo_bot/cell_workers/inventory_actions_worker.py
```diff
@@ -0,0 +1,47 @@
...
+
+    def __init__(self, bot):
+        self.config = bot.config
+        self.pokemon_list = bot.pokemon_list
+        self.api = bot.api
+        self.inventory_options_file = 'data/inventory_options'
+
+    def work(self):
+        if self.config.use_lucky_egg and (self.config.mode == "all" or
+                                          self.config.mode == "poke"):
```
dmateusp: this suggests that someone doing evolution of a lot of pokemons for XP in "farm" mode and trying to use a lucky egg from your code will be denied. Changing the logic or explaining the feature on the readme seems important

pokemongo_bot/cell_workers/inventory_actions_worker.py
```diff
@@ -0,0 +1,47 @@
...
+            with open('data/last-lucky-egg-use-%s.json' %
+                    (self.config.username)) as f:
+                last_use_json = json.load(f)
+                self.last_lucky_egg_use = datetime.datetime.strptime(
+                    last_use_json['last_use'], '%Y-%m-%d %H:%M:%S.%f')
+        else:
+            # First time we let the last lucky egg use to minus 30 minutes to
+            # use lucky egg
+            self.last_lucky_egg_use = datetime.datetime.now() - datetime.timedelta(minutes=30)
+
```
dmateusp: I'm sure there is somewhere in the API a way of checking if a lucky egg is active, that way we could double check ?

ethervoid: You're right, this seems to exist here https://github.com/tejado/pgoapi/blob/master/pgoapi/protos/POGOProtos/Inventory_pb2.py#L341 . Thx for the heads up, better logic than datetime one.

pokemongo_bot/cell_workers/inventory_actions_worker.py
ethervoid: Good points, I'll check who to add it to the logic :)

dmateusp:
@ethervoid I'm closing the PR until you're happy with your new logic ? Feel free to open it again

ethervoid:
Sure, that was just a WIP to avoid duplication in functionality so other contributor don't start working on this kind of stuff :)

PR #802, opened 2016-07-25
Evolving based on IV and CP

guyz:
Short Description:

Instead of evolving pokemons by their CP, evolving is now done by IV first. To prevent situations where ridiculously low CP pokemons with high IVs are evolved, there is a new optional config value cp_min (default: 300). This is used as a threshold - first, all larger (than the threshold) CP pokemons are evolved based on their IV. Then, if possible, lower CP pokemons are evolved based on their CP alone, since presumably IV doesn't matter that much then.


guyz:
In the previous PR I was asked to modify config.json.example, but I see this has been removed from the dev branch?

Other than that this has been tested locally and is working.

SteffanLong:
This looks great, but I know we're trying to refactor the code into an event-driven worker-plugin system. Evolution will eventually be migrated into a worker, perhaps this PR should do that?

guyz:
Evolution is already its separate worker, so not sure how this applies? There is definitely refactoring to do to have the workers share some utility functionalities - but this is beyond just this worker. I'll do it myself if I get around to it soon, but that might not be the case.

Also, this change is fairly small but very useful. I've been using it for the past two days and it does a much better job compared to just CP.

solderzzc:
Please help to check the README, if feature list updated.

PR #813, opened 2016-07-25
Fixes stationary walks in SpiralNavigation

tstumm:
Short Description: 
Fixes stationary walks by reworking StepWalker
Fixes:
- 
- 
-


pokemongo_bot/step_walker.py
```diff
@@ -44,7 +44,8 @@ def step(self):
         # print 'distance'
         # print dist
 
-        if (self.dLat == 0 and self.dLng == 0) or dist < 15:
+        if (self.dLat == 0 and self.dLng == 0) or dist < self.speed:
```
TheSavior: I think `dist < self.speeed` isn't going to work correctly. Since `random_lat_long_delta()` returns something between -9.1 and 9.1, we really need something like `self.speed < dist + 9.1` Otherwise we could walk past and go on for ever.

PR #818, opened 2016-07-25
merge and refactor config files

SteffanLong:


johnw424:
Can we get an example of individual Pokemon capture logic in the new configs/config.json.example?

PR #708, opened 2016-07-24
Resolv submodule and install requirement error

surfaace:
Short Description: 
we can now pull down submodule as it was moved to external repo.

Fixes:
at least one travis-ci error, more to come

PR #843, opened 2016-07-25
Remove unused 'origin' submodule

alvarogzp:
As it is causing `git submodule init` and `git submodule update` to fail, because both origin and web submodules point to the same 'web' folder.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,6 +1,3 @@
-[submodule "origin"]
-	path = web
-	url = https://github.com/OpenPoGo/OpenPoGoWeb.git
 [submodule "web"]
 	path = web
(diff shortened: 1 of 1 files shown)


mercuriete:
Are you sure that you are not making a mistake?

submodule with ssh and not https leads in a bug ->  https://github.com/PokemonGoF/PokemonGo-Bot/pull/708/files

alvarogzp:
@mercuriete I deleted the 'origin' one because it has a more confusing name.
If ssh url cause travis to fail, I can change it to https.

alvarogzp:
@mercuriete fixed

surfaace:
already fixed

Issue #855, opened 2016-07-25
After evolving, program still tries to use old release_config.json file


perryhuynh:
### Expected Behavior
After evolving, program checks to see if the newly evolved pokemon need to be released

### Actual Behavior
After evolving, program checks to see if the newly evolved pokemon need to be released (possibly trying to look for the old `release_config.json` file and crashes?)

### Steps to Reproduce
Start bot, bot evolves pokemon then goes through release check and crashes.

### Other Information
OS: 
Git Commit: (run 'git log -n 1 --pretty=format:"%H"' and paste it here)

```
[19:15:45] Starting PokemonGo Bot....
[19:15:45] Scanning area for objects....
[19:15:45] Walking from (34.0092278, -118.4973045) to (34.0092278, -118.4973045) 0.00km
[19:15:47] Attempting to evolve all pokemons ...
[#] Successfully evolved Ekans with 206 cp!
[#] Successfully evolved Geodude with 60 cp!
[#] Evolved 2 pokemons! Checking if any of them needs to be released ...
Traceback (most recent call last):
  File "pokecli.py", line 223, in <module>
    main()
  File "pokecli.py", line 214, in main
    bot.take_step()
  File "/home/pi/PokemonGo-Bot/pokemongo_bot/__init__.py", line 41, in take_step
    self.work_on_cell(cell, location)
  File "/home/pi/PokemonGo-Bot/pokemongo_bot/__init__.py", line 145, in work_on_cell
    worker.work()
  File "/home/pi/PokemonGo-Bot/pokemongo_bot/cell_workers/evolve_all_worker.py", line 48, in work
    self._release_evolved(release_cand_list_ids)
  File "/home/pi/PokemonGo-Bot/pokemongo_bot/cell_workers/evolve_all_worker.py", line 71, in _release_evolved
    if self.should_release_pokemon(pokemon_name, pokemon_cp, pokemon_potential):
  File "/home/pi/PokemonGo-Bot/pokemongo_bot/cell_workers/evolve_all_worker.py", line 155, in should_release_pokemon
    if self._check_always_capture_exception_for(pokemon_name):
  File "/home/pi/PokemonGo-Bot/pokemongo_bot/cell_workers/evolve_all_worker.py", line 219, in _check_always_capture_exception_for
    always_capture_list = self._get_always_capture_list()
[...]

PR #859, opened 2016-07-25
Fixing evolve_all_worker release_config (issue #855)

ritashugisha:
Previously was trying to access config.release_config changed in commit 33a0ad6c6645415e7491b1d76fe664f40376fb16


SteffanLong:
Releasing pokemon needs to be converted into a worker anyway.

solderzzc:
good!

Issue #848, opened 2016-07-25
[Suggestion] Make captured pokemon message a different color


danielruswick:
I suggest making the captured pokemon message a different color from the rest of the log messages. That way, if i leave the script running and come back later, I can easily skim the output and see which pokemon were caught while I was away.

surfaace:
already blue color in dev branch

PR #862, opened 2016-07-25
#848 [Suggestion] Make captured pokemon message a different color

data2700:
Short Description: As suggested:  Added Cyan to the logger.py  plus changed captured worker to log in cyan so that all captures are easier to notice when looking at the long

Fixes:
- 
- 
-


earthchie:
Since some people's config aim for very high IV pokemon. Most of the time captured pokemons will be released right away. If you could improve this logic to display cyan color to the one that bot decided to keep, that would be great!

surfaace:
updates to gui has been made 4b371801f0d43cd737123589f52394843ad2faa1

PR #870, opened 2016-07-25
Increases pokestop movement efficiency

Cameronlund4:
Short Description: Makes the program move less distance while using pokestops, cutting down time

at the beginning. This should make walking times between pokestops much
less in certain cases. (For example, on the old loop, if fort A was
2.5mi North, fort B was 2.7 mi North, and fort C was 2.6mi South, the
program would go A->C->B, which is insanely inefficient)

Also, the self.position in __init__.py now stays up-to-date with the
actual current position. This was needed to allow for the more real-time
location checking.

### DIFF
--- a/pokemongo_bot/__init__.py
+++ b/pokemongo_bot/__init__.py
@@ -191,12 +191,19 @@ class PokemonGoBot(object):
                 forts.sort(key=lambda x: distance(self.position[
                            0], self.position[1], x['latitude'], x['longitude']))
 
-                for fort in forts:
+                while len(forts) > 0:
+                    # Sort all by distance from current pos- eventually this should
+                    # build graph & A* it
+                    forts.sort(key=lambda x: distance(self.position[
+                            0], self.position[1], x['latitude'], x['longitude']))
+                    fort = forts[0]
                     worker = MoveToFortWorker(fort, self)
                     worker.work()
-
+                    if fort in forts:
+                        forts.remove(fort)
                     worker = SeenFortWorker(fort, self)
                     hack_chain = worker.work()
+                    self.position = worker.position #Update our position to the bots position
                     if hack_chain > 10:
                         #print('need a rest')
                         break
--- a/pokemongo_bot/cell_workers/seen_fort_worker.py
+++ b/pokemongo_bot/cell_workers/seen_fort_worker.py
@@ -36,6 +36,7 @@ class SeenFortWorker(object):
             fort_name = 'Unknown'
         logger.log('Now at Pokestop: ' + fort_name + ' - Spinning...',
                    'cyan')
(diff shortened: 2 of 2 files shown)


Cameronlund4:
How can I tell why the travis build failed? It runs fine locally

theadriann:
Check the details of the build.
Currently it seems that not every needed module is installed and the build fails when it tries to import them.

Cameronlund4:
So is it an issue on my end, or on theirs?

surfaace:
import errors is on 'our' end, thats fine. bumping devs about review approval

skydrome:
pulled this change and seems its still pathing is random ways :/
http://i.imgur.com/lk5Vc0O.png

douglascamata:
Have my :+1:, sir

douglascamata:
@skydrome I think I know why:

```
for cell in cells:
            self.work_on_cell(cell, location)
```

Cells can have many forts, but they're already sorted by distance. But it will go to 1st fort, do the routing to every other. Next iteration of `for` comes, to go 2nd fort cell of the list and then again do the routing.

skydrome:
So do you think its better to use a low step count in an area with many pokestops for best results?

Cameronlund4:
@skydrome If it's completely ignoring them, it's likely an issue with whatever is finding the forts. The current system simply moves the part that orders their distances inside the loop, so it actively finds the closest stop, instead of the distance from the starting point

Cameronlund4:
Also, where are you reading that path from? If you're reading it from the list in order, then it's completely irrelevant, because the order changes after every stop. The path cannot be predetermined before execution (That is, unless you have a system that simulates how it would loop through when walking, which I doubt)

douglascamata:
@Cameronlund4 to be able to reorder the list of cells as the bot walks through them, it has to become an attribute of the bot, something like `self.current_cells = cells`. Then the workers code can manipulate those cells as necessary.

Cameronlund4:
I don't think the cells are an issue (It might be more efficient to work with them later, but I think the system I'm using now is fine). I just think for the map he used, he simply mapped out the list, which isn't how it actually moves between them. The list is irrelevant, it's the path the bot actually goes

earthchie:
@Cameronlund4 

don't worry, your code is pass with score 7.43/10
image

it's fail because the other files. 

E.g:
image

bstpierr:
Is this PR still valid? How well does it work on current dev HEAD?

douglascamata:
@bstpierr probably, but needs to fix conflicts.

Cameronlund4:
@douglascamata What conflicts?

TheSavior:
@Cameronlund4 We can't merge this PR because it has merge conflicts with dev.

PR #880, opened 2016-07-25
Keep best CP pokemon

koper89:
Short Description: 
Couldn't test it cause dev branch cannot find me any pokemons, but hopes it'll work.

### DIFF
--- a/pokemongo_bot/cell_workers/pokemon_catch_worker.py
+++ b/pokemongo_bot/cell_workers/pokemon_catch_worker.py
@@ -202,6 +202,38 @@ class PokemonCatchWorker(object):
                             break
         time.sleep(5)
 
+    def _check_if_higher_cp_pokemon(self, pokemon_name, value):
+	self.api.get_inventory()
+	response_dict = self.api.call()
+	return self._check_if_higher_cp_pokemon_all(pokemon_name, value, response_dict)
+
+    def _check_if_higher_cp_pokemon_all(self, pokemon_name, value, response_dict):
+	try:
+		reduce(dict.__getitem__, [
+			   "responses", "GET_INVENTORY", "inventory_delta", "inventory_items"], response_dict)
+	except KeyError:
+		pass
+	else:
+		for item in response_dict['responses']['GET_INVENTORY']['inventory_delta']['inventory_items']:
+			try:
+				reduce(dict.__getitem__, [
+					   "inventory_item_data", "pokemon"], item)
+			except KeyError:
+				pass
+			else:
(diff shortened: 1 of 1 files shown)


koper89:
if not self._check_if_higher_cp_pokemon(pokemon_name, cp):
 +		return True

in here should be False

Calcyfer:
The logic seems correct, I'll test it in my patched dev branch

pokemongo_bot/cell_workers/pokemon_catch_worker.py
```diff
@@ -324,8 +356,12 @@ def should_release_pokemon(self, pokemon_name, cp, iv, response_dict):
 
         if release_config.get('always_release', False):
             return True
-
-        release_cp = release_config.get('release_below_cp', 0)
+	
+	if release_config.get('keep_best_cp', True):
```
Calcyfer: Maybe we would want the default value to be False, so nothing changes for users who pull but don't update their release_config.json file?

Calcyfer:
Currently getting this error : 

`pokemon_catch_worker.py", line 338, in _get_catch_config_for
    catch_config = self.config.catch.get(pokemon)
`
`AttributeError: 'Namespace' object has no attribute 'catch'
`

pokemongo_bot/cell_workers/pokemon_catch_worker.py
koper89: Yep but in dev branch there's no more realeae congig as separate file

koper89:
Doesn't seem like issue with my addition

pokemongo_bot/cell_workers/pokemon_catch_worker.py
Calcyfer: You're right. This just caused quite a few errors :)

Calcyfer:
It's not, that's why I removed the comment - damn all these commits.

Works great on dev btw, after fixing the issues with the config created by other commits.

koper89:
Great to hear.

Nihisil:
👍  

I hope you have enough patience to merge all dev changes into your branch and that you will not give up with this pull-request.

slimjim2234:
LGTM!
:+1:

namelesswc:
Suggestion: It should keep lower cp pokemon in case it has very good iv otherwise I'm gonna loose all my perfekt pokemons.

koper89:
https://github.com/PokemonGoF/PokemonGo-Bot/pull/953
Moved to this pull.

PR #887, opened 2016-07-25
Clarifying the keep message appearing as duplicated

rrenato:
Short Description: 

Fixes:
- 
- 
- 

[2016-07-25 17:43:11] [x] Captured Spearow! [CP 147] [IV 0.82]
[2016-07-25 17:43:11] [x] Kept Captured Spearow! [CP 147]
Its better than:
[2016-07-25 17:43:11] [x] Captured Spearow! [CP 147] [IV 0.82]
[2016-07-25 17:43:11] [x] Captured Spearow! [CP 147]

### DIFF
--- a/pokemongo_bot/cell_workers/pokemon_catch_worker.py
+++ b/pokemongo_bot/cell_workers/pokemon_catch_worker.py
@@ -168,7 +168,7 @@ class PokemonCatchWorker(object):
                                             '[#] {} has been exchanged for candy!'.format(pokemon_name), 'green')
                                     else:
                                         logger.log(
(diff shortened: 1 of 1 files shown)


MikeDX:
We probably don't need to say Kept, but just not having the second "caught" line is enough since we aren't doing anything else (to say when we ship it off to be made into candy is a different event)

rrenato:
The reason I made this was because for me was not clear to see what I've kept so changing the message would help other people do a grep 'Kept' on the logs (I like to run the bot with a |tee -a log)

surfaace:
make new PR to dev branch please

PR #893, opened 2016-07-26
list pokemon to be transferred before initial_transfer

JohanJu:
Comment: I use more colors for different iv stages so the if statements can be reduced or new color added.


SteffanLong:
Initial transfer is being removed in favor of a release worker.

surfaace:
make new PR against dev branch

PR #904, opened 2016-07-26
Added bash scripts install.sh and run.sh to install dependencies and then run the bot, respectively.

Grace:
**Short Description:**
Included bash scripts to install dependencies, update submodule(s), and then run the bot. Additionally, a bug was fixed that broke the bot if 'release' or 'catch' keys were not found in config.json.

**Fixes:**
- install.sh added to automate installing dependencies for the bot
- run.sh added to easily run the bot with `./run.sh` instead of the full python command
- Bug was fixed that broke the bot if 'release' or 'catch' keys were not found in config.json


Grace:
I made another pull request for these changes, disregard this please. Thank you!

PR #920, opened 2016-07-26
Location file patch

widnyana:
Short Description: 

this patch fix issue for missing `location-username.json` and submodule init issue

Fixes:
- submodule init
- create location-%s.json if it doesnt exists

please review :)


solderzzc:

SteffanLong:
Seems overly complex. We don't need to check if the file exists. Just write it. If it doesn't exist, it will after you create it. This is how the rest of the files work.

widnyana:
@SteffanLong I've talked with `reaver01` in dev channel on slack, based on his suggestion i wrote this. :)

probably changing the file open mode from `w` to `w+` is the simplest way, any suggestion?

.gitmodules
```diff
@@ -1,3 +1,3 @@
 [submodule "web"]
 	path = web
-	url = https://github.com/OpenPoGo/OpenPoGoWeb.git
```
solderzzc: This line should not be touched.

widnyana: @solderzzc noted, please leave any suggestion here. 

I'll create another pull request with more clean commits and close this one.

solderzzc:
Sure, thanks a lot:)

PR #924, opened 2016-07-26
Linux/Mac needs sudo

maestroi:
Short Description: 
added sudo to installation guide
Fixes:
- sudo added for linux installation
- sudo added for Mac OSX installation

Sudo is needed to install some requirements

### DIFF
--- a/README.md
+++ b/README.md
@@ -105,7 +105,7 @@ $ git clone -b master https://github.com/PokemonGoF/PokemonGo-Bot
 $ cd PokemonGo-Bot  
 $ virtualenv .  
 $ source bin/activate  
-$ sudp pip install -r requirements.txt
+$ sudo pip install -r requirements.txt
 $ git submodule init
 $ git submodule update
 ```
--- a/README.md
+++ b/README.md
@@ -90,7 +90,7 @@ Please keep in mind that master is not always up-to-date whereas 'dev' is. In th
 ```
 $ git clone -b master https://github.com/PokemonGoF/PokemonGo-Bot  
 $ cd PokemonGo-Bot  
-$ pip install -r requirements.txt
+$ sudo pip install -r requirements.txt
 $ git submodule init
 $ git submodule update
 ```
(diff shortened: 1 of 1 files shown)


widnyana:
@maestroi not needed if you run it inside virtualenv,

instead of messing with your system-wide package I'd prefer using virtualenv.

maestroi:
@widnyana than maybe its good to add that to the linux instruction aswel ?

README.md
```diff
@@ -90,20 +90,22 @@ Please keep in mind that master is not always up-to-date whereas 'dev' is. In th
...
+Make sure you install the following first:
+[Requirements](https://github.com/PokemonGoF/PokemonGo-Bot/wiki/Installation)
 
 ```
 $ git clone -b master https://github.com/PokemonGoF/PokemonGo-Bot  
 $ cd PokemonGo-Bot  
 $ virtualenv .  
 $ source bin/activate  
-$ pip install -r requirements.txt
+$ sudp pip install -r requirements.txt
```
pulgalipe: You mistyped here, it's **sudo** not _sudp_

pulgalipe:
Agreed that using virtualenv is a safe choice, it lowers the risk of screw up your OS.
If someone would want to use multiple instances of bot by a script just make use of . notation from virtualenv.

Example:

```. ./bin/activate```

CapCap:
Linux/mac do NOT need sudo for running it- and it's a very bad idea to tell people they need it.

evrardjp:
👎 virtualenv is good enough.
The only change we could do is replace "(Optional)" by something like "(Recommended - optional)" in the docs

maestroi:
Changed the instructions to use virtualenv :+1:

douglascamata:
@maestroi please explain in the readme why virtualenv is required and it's happily merge it.

AnAkkk:
There is no need to use virtualenv or sudo on Linux. You should pass --user if you want to install dependencies for your user, like this:
pip install -r requirements.txt --user

douglascamata:
@AnAkkk virtualenv is not only about installing packages for the local user. It's about having many clean Python environments and being able to switch between them. Virtualenv is so useful that in Python 3.5 they embedded the idea into a `venv` module of the standard library.

DavidK1m:
Wouldn't "Recommended" sound better than "Recommend"?

maestroi:
@douglascamata its done i guess

douglascamata:
you have my +1

maestroi:
@douglascamata can you change the label ?

MaxLeiter:
:+1: Thanks!

PR #925, opened 2016-07-26
Fixing iv_display template KeyError

ritashugisha:
Was previously trying to access possibly missing keys in the pokemon[pokemon_data] dict (commit 7c1e6205fa60da91b306c322a1d376a949796eb7)


pokemongo_bot/cell_workers/pokemon_catch_worker.py
```diff
@@ -63,6 +58,12 @@ def work(self):
                                         pokemon['pokemon_data'][individual_stat] = 0
                                         continue
 
+                                iv_display = '{}/{}/{}'.format(
```
solderzzc: But will this fix such issue?

ritashugisha: Yes, the individual stats are set to 0 (line 58) if they key isn't found in the dict.

solderzzc:
Thanks, it was fixed on branch.

PR #929, opened 2016-07-26
Search and Capture while walking to Poke Stop & Initial transferring uses Release Criteria

joshho:
- Search and Capture while walking to Poke Stop 
- Initial transferring uses Release Criteria (instead of hard CP comparison)


pokemongo_bot/__init__.py
```diff
@@ -322,7 +330,7 @@ def catch_pokemon(self, pokemon):
         return_value = worker.work()
 
         if return_value == PokemonCatchWorker.BAG_FULL:
-            worker = InitialTransferWorker(self)
+            worker = TransferWorker(self)
```
solderzzc: Why need change name?

solderzzc:
This should be separated.
I think 'Initial transferring uses Release Criteria (instead of hard CP comparison)'  is good, and need README update.

pokemongo_bot/__init__.py
joshho: The TransferWorker also handles parts of the transfer task that happens after post-capture

pokemongo_bot/cell_workers/move_to_fort_worker.py
```diff
@@ -41,6 +43,8 @@ def work(self):
                 while True:
                     if step_walker.step():
                         break
+                    else:
+						self.bot._work_on_cell_catch(self.position)
```
TheSavior: We don't want workers calling back up to the bot like this

pokemongo_bot/spiral_navigator.py
```diff
@@ -22,9 +22,14 @@ def __init__(self, bot):
         self.origin_lon = self.bot.position[1]
         self._step_walker = None
 
-    def take_step(self):
+    def take_stepE(self):
```
TheSavior: Typo?

joshho: Not too familiar with Python, couldn't overload the method for some reason.

TheSavior:
I don't think this PR is what we want. I do think the separation of the method `_work_on_cell_catch` is a good one, but I don't think we want to flatten our logic instead of making it deeper. We definitely don't want to call work_on_cell_catch from inside a worker. 

Instead walking to pokestops should take a few steps towards the pokestop on each tick and return control to the event loop. That is the direction we went for the stepper and we should continue that direction instead of going the other way.

pokemongo_bot/cell_workers/move_to_fort_worker.py
joshho: What do you think if the _work_on_cell_catch logic was put away into a util of sorts?

pokemongo_bot/cell_workers/pokemon_catch_worker.py
```diff
@@ -82,7 +81,7 @@ def work(self):
                         
                         balls_stock = self.bot.pokeball_inventory()
                         while(True):
-
```
solderzzc: Code need to be clean up.

TheSavior:
@joshho Happy to elaborate more on the previous comment if you need it. I'm going to close this PR as this approach isn't compatible with the architecture we are moving towards. Please feel free to open another PR if need be.

PR #938, opened 2016-07-26
Mode/Cell Changes

heihachi:
Short Description: 
Should help fix the no catching pokemon after 1st/2nd pokestop. Modes can switch back to 'all' after farming 10+ pokeballs and losing 50 items in bags (recycle, login to app, or catching)
Fixes:
- Mode switching back and forth
- Cells update each tick and after a timer of 10 secs inbetween (this can be removed if theres no harm in doing constant scans)

### DIFF
--- a/pokemongo_bot/__init__.py
+++ b/pokemongo_bot/__init__.py
@@ -36,8 +36,9 @@ class PokemonGoBot(object):
         location = self.navigator.take_step()
         cells = self.find_close_cells(*location)
 
-        for cell in cells:
-            self.work_on_cell(cell, location)
+        self.work_on_cells(cells,location)
+        #for cell in cells:
+        #    self.work_on_cell(cell, location)
 
     def update_web_location(self, cells=[], lat=None, lng=None, alt=None):
         # we can call the function with no arguments and still get the position and map_cells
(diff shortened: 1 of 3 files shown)


heihachi:
Mainly need feedback as to if its working for you or if the issue is still present.

TheSavior:
I don't think we want this. This problem:
screen shot 2016-07-24 at 5 55 09 pm

is because of this code 
```
for fort in forts:
                    worker = MoveToFortWorker(fort, self)
                    worker.work()

                    worker = SeenFortWorker(fort, self)
                    hack_chain = worker.work()
```

It currently finishes looping through forts before it changes mode.

The solution we need to take is to have the worker only consider the first fort, which would then let it approach other forts on subsequent iterations. That would let it switch modes after the first fort that says it is full.

TheSavior:
@heihachi: Happy to elaborate more on the previous comment if you need it. I'm going to close this PR as this approach isn't compatible with the architecture we are moving towards. Please feel free to open another PR if need be.

heihachi:
@TheSavior will do thanks! and i can rework that if noone else is on it

heihachi:
Also should i make another PR with just the switching back and forth from all -> poke and all -> farm?

PR #939, opened 2016-07-26
SpiralNavigator rewrite, mode switch

tstumm:
Short Description: 
SpiralNavigator had code issues so I replaced it with some example code from pgoapi until we find a better solution.
Also I fixed fort-searching even if inventory is full

### DIFF
--- a/pokemongo_bot/__init__.py
+++ b/pokemongo_bot/__init__.py
@@ -62,7 +62,7 @@ class PokemonGoBot(object):
             status = map_objects.get('status', None)
             cells = map_objects['map_cells']
 
-            # insert detail info about gym to fort
+            #insert detail info about gym to fort
             for cell in cells:
                 if 'forts' in cell:
                     for fort in cell['forts']:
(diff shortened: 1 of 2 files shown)


pokemongo_bot/__init__.py
```diff
@@ -62,7 +62,7 @@ def update_web_location(self, cells=[], lat=None, lng=None, alt=None):
             status = map_objects.get('status', None)
             cells = map_objects['map_cells']
 
-            #insert detail info about gym to fort
+            # insert detail info about gym to fort
```
TheSavior: It's unfortunate that you have so many changes in this file. It makes it very difficult to read the diff here and likely that it will cause someone else merge conflicts.

TheSavior: In fact, this is probably why you now have merge conflicts.

TheSavior:
The changes to the SpiralWalker in this PR look good to me.

pokemongo_bot/__init__.py
```diff
@@ -212,6 +212,8 @@ def work_on_cell(self, cell, position):
                     if hack_chain > 10:
                         #print('need a rest')
                         break
+                    if self.config.mode == "poke":
```
TheSavior: This is a reasonable fix for now, but probably not the right long term fix. We should be removing this loop entirely so it instead only acts on one fort per event loop.

TheSavior:
:+1:

pokemongo_bot/__init__.py
tstumm: The problem is that both workers are not giving any informations back to the bot so they need to be refactored. This quickfix should work until that is being done.

PR #945, opened 2016-07-26
Support for non unicode location

Nihisil:
Short Description: 	Allow to set non unicode location, both as command argument and config value.

Fixes:
- Fixes for config file parsing and non unicode location
- Fixes for `-l` command argument and non unicode location
- Small style formatting changes in `pokecli.py`

How to reproduce crash on previous version?

Just try to set `-l "Красная площадь, Москва, город Москва"` or add same value to `location` in config.

### DIFF
--- a/pokecli.py
+++ b/pokecli.py
@@ -40,6 +40,7 @@ from pokemongo_bot import logger
 if sys.version_info >= (2, 7, 9):
     ssl._create_default_https_context = ssl._create_unverified_context
 
+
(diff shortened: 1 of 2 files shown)


Nihisil:
Non unicode location worked some days ago, and after last 100-200 commits it stopped to work.

I fixed it and hope it will not be changed anymore.

douglascamata:
extra :+1: for Python code style. I like it.

PR #959, opened 2016-07-26
Location file creating fix

Phil9l:
Fixes:
- Before this change I had to `touch location-%(username)s.json`, otherwise it it hadn't been created itself
- Replaced '/' in filenames with `os.path.join`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -337,6 +337,7 @@ If using multiple usernames format like this:
  * riberod07
  * th3w4y
  * Leaklessgfy
+ * phil9l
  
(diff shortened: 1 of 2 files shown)


bimahp:
can we just remove that "if" blocks ?

Phil9l:
@bimahp, well, we can do it, of course, but there can be a directory with the same name, so the exception will be raised.

solderzzc:
We don't accept PR on main, please submit to dev.
Thanks.

PR #968, opened 2016-07-26
Fixes web positioning

tstumm:
Short Description: 
Fixes crappy web position because web relies mainly on position information of bot while api positioning should be used.


douglascamata:
+1

douglascamata:
@tstumm just a tip to help using the api position everywhere:

```python
class PokemonGoBot

  @property
  def position(self):
    return (self.api._position_lat, self.api._position_lng)
```

This way every call to `self.bot.position` (in worker class) and `self.position` (in bot class) will automatically use the correct position from the api.

tstumm:
@douglascamata I didn't even know that python supports properties lol

PR #970, opened 2016-07-26
Added a progress bar for walking.

x1024:
There's now a TTY-only progress bar when walking from one place to another.

It doesn't show up when the program's output is piped to a file, or if it's run as a daemon.
I could use some help with testing on different OS/shell combinations.

### DIFF
--- a/pokemongo_bot/step_walker.py
+++ b/pokemongo_bot/step_walker.py
@@ -2,6 +2,18 @@ from math import sqrt
 
 from cell_workers.utils import distance, i2f
 from human_behaviour import random_lat_long_delta, sleep
+import sys
+
+
+def progress_bar(percentage):
+    percentage = min(100, max(0, percentage))
+    if not sys.stdout.isatty():
+        return
+    sys.stdout.write('\r')
+    # http://www.fileformat.info/info/unicode/char/2588/index.htm
+    msg = (u"[%-40s] %d%%" % (u"\u2588"*int(percentage*2//5), percentage))
+    sys.stdout.write(msg)
+    sys.stdout.flush()
 
 
 class StepWalker(object):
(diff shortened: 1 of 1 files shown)


tstumm:
Have you tested that? Cause im pretty sure it wont work, because StepWalker lives like only step long.

Otherwise that would be pretty cool.

x1024:
I have tested it, this is how it looks for me:

screenshot

Zgrkaralar:
I have tested it also, looks so cool 👍 
screen shot 2016-07-26 at 2 17 12 pm

tstumm:
Yep you're right, I misinterpreted the code... Looks very nice. Have you tried what happens if pokemon appears while walking?

x1024:
If you encounter something unexpected, there is no line break:

screenshot

I haven't yet thought too much about how to fix that. Maybe a simple "if we're in a tty, print a line break" in front of every line that can interrupt a walk?

solderzzc:
@x1024  Good suggestion, Let's merge here before conflict happen. save time for rebase.

KLVN:
Buggy on my Win 10 machine, maybe because of a different language format (german)?
`[ÔûêÔûêÔûêÔûêÔûêÔûêÔûêÔûêÔûêÔûêÔûêÔûêÔûêÔûêÔûê                         ] 39%`

x1024:
@KLVN it uses the "square block" character, for esthetic reasons. It could be changed to a simple "=" or "+" sign.
Perhaps someone can suggest a square block character that is locale-independent?

LuckyMe4Evers:
Screen is full of it
screenshot050

bimahp:
imho a simple "=" is sufficient for me

KLVN:
@x1024 Let's use "=" then. Isn't that beautiful as a full block, but it works. :)
pokegopath

JohnnyOmm:
I'm new to github. how do I download your step_walker.py

Issue #821, opened 2016-07-25
Not catching any pokemons


sffetlio:
Since I updated to dev, the bot is not catching any pokemons.
The log shows 20+ pokestops and no pokemons between them.

"mode": "all" in the config

### Other Information
OS: Ubuntu
fa9df3b05072432f3ca1f9d68a39cf086a4686eb

alvarogzp:
same here, only catches pokemons at the beginning, then it starts going to pokestops and forgets about pokemons

LeaklessGfy:
This is a critical bug that we are trying to fix at the moment ...
Not sure it will be fix in the next hours.

If you got a commit version, branch or tagged that work please refer the name here, like that we can try to move quickly.

Thanks :)

sffetlio:
I have 6e99169d963bd8637dfb55426650f2fe15e694bd and it's working there.

tstumm:
Does the problem still persist when using "poke" as farming mode?

gpires:
@tstumm: "poke" works, but "all" doesn't.

Commit: 32cefeed4760d6b5d2b3eb76d783d0f13d068e83

z4ppy:
Same here, last version (33a0ad6) and no capturing pokemon with "all".

josephmcox:
Even in mode poke not much is being caught. Could this be an issue with the new google directions commit?

josephmcox:
Confirmed if i checkout 335a0aa9a6e57fef17d0865b2f404e0ca785124b it is catching pokemon as notmal.  So something after that is causing this.

incenx:
@josephmcox but on #335a0aa , it has teleportation problem...

josephmcox:
Ahh didnt know that. I have been using it and not getting banned.... its been working great

leadboots5:
broken as of https://github.com/PokemonGoF/PokemonGo-Bot/pull/735 to help narrow it down to 7 or so builds

bstpierr:
I tried checking out 335a0aa and got softbanned after a few minutes due to teleporting between stops.

I've switched to using the master branch until this is resolved.

josephmcox:
Does master branch work for catching at a decent speed?

bstpierr:
It still seems to not see as many mons as I would expect, but at least it finds something between stops once in a while.

[...]

PR #979, opened 2016-07-26
[BUGFIX] Catch Pokemon while walking to fort

binarydepartment:
Short Description: 
Pokemon are now caught while walking to a fort. Please test. It's not pretty, but it should work. A proper refactor is needed to prevent something like this from happening in the future.

Fixes:
- #821

### DIFF
--- a/pokemongo_bot/__init__.py
+++ b/pokemongo_bot/__init__.py
@@ -22,7 +22,7 @@ from spiral_navigator import SpiralNavigator
 
 
 class PokemonGoBot(object):
-    
+
     @property
     def position(self):
         return (i2f(self.api._position_lat), i2f(self.api._position_lng), 0)
@@ -40,10 +40,13 @@ class PokemonGoBot(object):
 
     def take_step(self):
         location = self.navigator.take_step()
-        cells = self.find_close_cells(*self.position[0:2])
+        self.process_cells(work_on_forts=True)
 
+    def process_cells(self, work_on_forts=True):
+        location = self.position[0:2]
+        cells = self.find_close_cells(*location)
         for cell in cells:
-            self.work_on_cell(cell, location)
+            self.work_on_cell(cell, location, work_on_forts)
 
     def update_web_location(self, cells=[], lat=None, lng=None, alt=None):
         # we can call the function with no arguments and still get the position and map_cells
@@ -131,7 +134,7 @@ class PokemonGoBot(object):
         self.update_web_location(map_cells)
         return map_cells
 
-    def work_on_cell(self, cell, position):
+    def work_on_cell(self, cell, position, work_on_forts=1):
         # Check if session token has expired
         self.check_session(position)
 
@@ -201,8 +204,8 @@ class PokemonGoBot(object):
             for pokemon in cell['wild_pokemons']:
                 if self.catch_pokemon(pokemon) == PokemonCatchWorker.NO_POKEBALLS:
                     break
-        if (self.config.mode == "all" or
-                self.config.mode == "farm"):
+        if ((self.config.mode == "all" or
+                self.config.mode == "farm") and work_on_forts):
             if 'forts' in cell:
                 # Only include those with a lat/long
                 forts = [fort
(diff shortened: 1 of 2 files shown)


Mello-Yello:
Working for me :-)

registered99:
Need to fix the indentation

pokemongo_bot/__init__.py
```diff
@@ -34,10 +34,15 @@ def start(self):
 
     def take_step(self):
         location = self.navigator.take_step()
-        cells = self.find_close_cells(*location[0:2])
+        self.work_cells(1, self.position[0:2])
 
-        for cell in cells:
-            self.work_on_cell(cell, location)
+  def work_cells(self, work_on_forts=1, location=None):
```
douglascamata: Please, name `work_cells` better. I deserves a cleaner naming, because `work_on_cell` already exists and this coexistence is not intuitive. 

Also you should not defined `work_on_forts` default value with 1 if it's boolean. Use `True` or `False`.

pokemongo_bot/__init__.py
```diff
@@ -34,10 +34,15 @@ def start(self):
...
     def take_step(self):
         location = self.navigator.take_step()
-        cells = self.find_close_cells(*location[0:2])
+        self.work_cells(1, self.position[0:2])
 
-        for cell in cells:
-            self.work_on_cell(cell, location)
+  def work_cells(self, work_on_forts=1, location=None): 
+        if (location is None): 
+            location = self.position[0:2]
```
douglascamata: Please, don't get location from `self.position`, get if from `(self.api._position_lag, self.api._position_lng)`. Usage of `self.position` is being slowly eliminated from the project.

pokemongo_bot/__init__.py
```diff
@@ -34,10 +34,15 @@ def start(self):
 
     def take_step(self):
         location = self.navigator.take_step()
-        cells = self.find_close_cells(*location[0:2])
+        self.work_cells(1, self.position[0:2])
```
douglascamata: When you have a function like this, with named arguments, please use keywords to false them:

`self.work_cells(work_on_forts=1, location=self.position[0:2]`

Note: doesn't make any sense sending the location here is inside this method you have:

```
if (location is None): 
    location = self.position[0:2] 
```

It's better to simplify me method and remove the `position` parameters entirely.

binarydepartment: api._position_lng and api._position_lng are integers. Almost all methods of PokemonGoBot and some dependencies expect float.  What to do?

douglascamata: Use `i2f(self.api._position_lng)`. To avoid converting all the time, you can create a private method in the Bot main class, like:

```python
@property
def current_coordinates_as_float(self):
    return i2f(self.api._position_lat), i2f(self.api._position_lng) 
```

twaffle:
Manually merged in your changes, works well.  Only issue I have seen is a little spam as russling leaves may be triggered over and over again as the function is not used to be being called while moving.

douglascamata:
@binarydepartment you have merge conflicts now :( can you update your codebase from dev and fix them so I can merge this? Thanks!

binarydepartment:
Resolved conflicts. There is currently an error in the pokemon_catch_worker, I think it has been introduced by the lure changes.

douglascamata:
@binarydepartment I reverted the commit that introduced those errors.

binarydepartment:
Should be good to go now.

pokemongo_bot/cell_workers/move_to_fort_worker.py
```diff
@@ -41,6 +41,8 @@ def work(self):
                 while distance(i2f(self.api._position_lat), i2f(self.api._position_lng), lat, lng) > 10:
                     if step_walker.step():
                         break
+                    else:
+                        self.bot.process_cells(work_on_forts=False)
```
TheSavior: This looks really bad, workers should definitely not be calling methods like this on the bot. 

Especially this one. Workers should just return control to the event loop to run again. If they must, they should change a config property so they run differently next time. If this is done incorrectly, we can easily end up in an infinite loop and get a stack overflow error. This is very very scary.

binarydepartment: This would need quite some refactoring. Since the dev branch was pretty much broken for some time now, I submitted this quick & dirty fix. My python experience is limited to this project. I don't feel comfortable doing major changes.

PR #947, opened 2016-07-26
Fix config.json.pokemons.example

johnw424:
Short Description: 
Fix incorrect settings.

Fixes:
- release_under_cp -> release_below_cp
- release_under_iv -> release_below_iv
- cp_iv_logic -> logic

PR #980, opened 2016-07-26
Fix config.json.pokemons.example release config

alvarogzp:
Properties in the json were `release_under_cp` and `release_under_iv`, but code is searching for `release_below_cp` and `release_below_iv`, so this was not working if used by users.


alvarogzp:
Modified also evolve_all_worker to use `release_below_cp` and `release_below_iv` too.

There are no more places where `release_under_cp` and `release_under_iv` are used.

alvarogzp:
Same with cp_iv_logic to logic, as seen in #947

douglascamata:
Looks good to me.

solderzzc:
Thanks.

PR #995, opened 2016-07-26
Dump cells to enable more front-end features

Davideddu:
Short Description: 
It now dumps the cell list as a json into data/cells-$username.json, so that more front-ends can use this information (I'm working on a Kivy-based one).

### DIFF
--- a/README.md
+++ b/README.md
@@ -337,6 +337,7 @@ If using multiple usernames format like this:
  * riberod07
  * th3w4y
  * Leaklessgfy
+ * Davideddu
  
 -------
 ## Credits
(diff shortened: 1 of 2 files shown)


surfaace:
i assume cells becomes inactive after X amount of time, would i then suggest to add a checker to remove old cells during program execution

Davideddu:
Cells are provided by the API, I assume the API knows better what cells are inactive. Also, by using this patch with my program, I can tell you that this happens, as new pokestops appear automatically and "used" ones get marked.

There's no need to do any check, and, anyways, it's not related to what this pull-request adds. It just dumps the existing in-memory information to disk for other programs to use.

By the way, this is the program that uses it: https://github.com/Davideddu/PokeVy

solderzzc:
please sub pr to our dev branch, we need only PR to dev to make thing easy.
Let me close this one, waiting for PR on dev.

Davideddu:
Okay, thank you. I see there have been a lot of changes, I'll read the code and re-implement it.

Davideddu:
Submitted. #1019

PR #999, opened 2016-07-26
Item_filter format changed, supports itemcount

surfaace:
you can now limit how many items you want to keep


surfaace:
wrong branch

PR #1004, opened 2016-07-26
Broken connection fix

Phil9l:
Fixes:
- Added `reconnecting_timeout` to config params
- Added waiting for `reconnecting_timeout` minutes if NotLoggedInException is raised

### DIFF
--- a/pokecli.py
+++ b/pokecli.py
@@ -227,8 +227,8 @@ def main():
             # pokemon catched, etc.
 
         except NotLoggedInException:
-            logger.log('[x] Error while connecting to the server, please wait %s' % config.reconnecting_timeout, 'red')
-            time.sleep(reconnecting_timeout)
+            logger.log('[x] Error while connecting to the server, please wait %s minutes' % config.reconnecting_timeout, 'red')
+            time.sleep(config.reconnecting_timeout)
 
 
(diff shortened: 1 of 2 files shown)


surfaace:
a login reconnect when servers is not responding correctly alreade ixists, however when having a softban this could be useful althought i would set default value to 15-20 minutes just to made sure that it wont softban the ** out of you.

few others ned to have their input before it's approved

pokecli.py
```diff
@@ -201,19 +208,27 @@ def main():
...
+
+        except KeyboardInterrupt:
+            logger.log('[x] Exiting PokemonGo Bot', 'red')
+            finished = True
+            # TODO Add number of pokemon catched, pokestops visited, highest CP
+            # pokemon catched, etc.
+
+        except NotLoggedInException:
+            logger.log('[x] Error while connecting to the server, please wait %s minutes' % config.reconnecting_timeout, 'red')
+            time.sleep(config.reconnecting_timeout)
```
Fantastik71: time.sleep(X) sleeps X seconds.
but it's supposed to wait X minutes here, right?

solderzzc:
we don't do PR on the mater branch to make thing easier. Thanks for the PR :)

PR #1009, opened 2016-07-26
Catching pokemon from lures and API update

Calcyfer:
### Short Description: 
The script will now catch pokemons from stops with a lure.
To enable this, the API had to be updated since tejado's version misses methods and variables.

### Fixes:
- Pokemon from lures are now being caught
- A new version of the API has been added, allowing more features

### Important:
The API now resides in the `pgoapi` folder, **`src/pgoapi` should be removed** - I know I could have added it as a requirement, but I didn't want just one person to own that repository, slowing down development.
In the future, we might want to add the API back in a separate repo, but that's something the project members should decide on.

2016-07-26 17_31_46-c__windows_system32_cmd exe


surfaace:
please remove pgoapi as this is installed using 'pip install -r requirements.txt' and should not be in repo

Calcyfer:
Check my PR comment please, I've put it there for a reason. :)

solderzzc:
@Calcyfer  thanks for the PR, the feature is very needed, just we don't add PGOAPI here to prevent directly modification made upgrade unpossible, if it's a must change, please PR to pgoapi repo, then update here.
Close for now, waiting for changes.

Calcyfer:
Why the hell didn't I think of that.. ? :D 
Added the PR there and will update when it is merged.

https://github.com/tejado/pgoapi/pull/117

PR #1014, opened 2016-07-26
Add USR2 signal to reread the config

LIQRGV:
Short Description: 
Usage: kill -USR2 <pokecli PID>

Now we can send USR2 signal instead of restart the program

### DIFF
--- a/pokecli.py
+++ b/pokecli.py
@@ -25,6 +25,7 @@ OR OTHER DEALINGS IN THE SOFTWARE.
 Author: tjado <https://github.com/tejado>
 """
 
+import signal
 import os
 import re
 import json
@@ -174,7 +175,7 @@ def init_config():
             config.release_config.update(json.load(data))
 
     # create web dir if not exists
-    try: 
+    try:
         os.makedirs(web_dir)
     except OSError:
         if not os.path.isdir(web_dir):
@@ -185,6 +186,10 @@ def init_config():
 
     return config
 
+def reread_config(signum, frame):
+    print_yellow('reread config.json')
+    main()
+    idle = False
 
 def main():
     # log settings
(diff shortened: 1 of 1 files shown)


douglascamata:
LGTM

pokecli.py
```diff
@@ -206,15 +211,20 @@ def main():
...
+
     except KeyboardInterrupt:
         logger.log('[x] Exiting PokemonGo Bot', 'red')
         # TODO Add number of pokemon catched, pokestops visited, highest CP
         # pokemon catched, etc.
 
+signal.signal(signal.SIGUSR2, reread_config)
 
 if __name__ == '__main__':
-    main()
```
solderzzc: why we need a where here?
so are you working on a external signal control?

Zgrkaralar:
But my suggestion getting that signal name from config.json. Beucase i have 20 working bot if i use that when i send this signal it's will kill all @douglascamata @solderzzc

douglascamata:
@Zgrkaralar but why would you send that SIGNAL to all processes instead of just the process you want to reload the config?

pokecli.py
```diff
@@ -206,15 +211,20 @@ def main():
...
+    while True:
```
douglascamata: This is indeed weird. Please try to keep all necessary logic inside the `main` function (including loop) and remove this from the ` if __name__ == '__main__'` block. Make it as simple as possible, please.

LIQRGV:
Ah ya, the logic should implemented on main function.
Thanks for that suggestion.

Also add the code on `if __name__ == '__main__'` caused the application run forever.
My bad.

surfaace:
@LIQRGV please update files accordingly, also recreate PR to merge into dev branch

PR #1015, opened 2016-07-26
find_close_cells incorrectly receives target position instead of current one

alvarogzp:
The call to SpiralNavigator.take_step() is returning the next target position, and take_step in PokemonGoBot is using it to call find_close_cells.

Instead, find_close_cells should be called with self.position, that is, the current position.


solderzzc:
Finally, knew the reason of that position[2] exception.

PR #1017, opened 2016-07-26
Refactor Top-Level take_step to regain control after every tick movement

joshho:
Refactor Movement-to-Fort to give back control to TopLevel after each step 

Implement simple ignore encounter cache - those poke that are ignored on first pass, shouldn't show up on the screen a second time.


pokemongo_bot/__init__.py
```diff
@@ -58,6 +59,15 @@ def update_web_location(self, cells=[], lat=None, lng=None, alt=None):
                 cell_id=cellid
             )
             response_dict = self.api.call()
+            for i in range(1,3):
```
TheSavior: Why do we need this? It doesn't seem relevant to this PR

TheSavior: I think that if we wanted retry logic on the api, we would add it at the API layer instead of in this one instance where we call it.

pokemongo_bot/__init__.py
```diff
@@ -204,11 +236,15 @@ def work_on_cell(self, cell, position):
...
 
                 for fort in forts:
-                    worker = MoveToFortWorker(fort, self)
-                    worker.work()
-
-                    worker = SeenFortWorker(fort, self)
-                    hack_chain = worker.work()
+                    worker = MoveToFortWorker(fort, self, cell)
+                    hack_chain = None;
+                    while True:
```
TheSavior: There shouldn't be any loops here. `_work_on_cell` is called repeatedly and *is* the loop. This should only take a single step towards the nearest fort, or spin a single fort if it is near enough. Then we will go through _work_on_cell again and it will run for the next fort. That will allow us to catch pokemon between every fort.

TheSavior:
Merge conflicts

pokemongo_bot/cell_workers/move_to_fort_worker.py
```diff
@@ -19,17 +20,17 @@ def work(self):
...
         # print('Found fort {} at distance {}m'.format(fortID, dist))
-        logger.log('Found fort {} at distance {}'.format(
+        logger.log('Moving to fort {} at distance {}'.format(
             fortID, format_dist(dist, unit)))
-
         if dist > 10:
-            logger.log('Need to move closer to Pokestop')
             position = (lat, lng, 0.0)
-
+            self._is_moving = True
```
TheSavior: I don't think you read this from anywhere. Do you need this variable?

TheSavior:
This has been implemented and merged to Dev in https://github.com/PokemonGoF/PokemonGo-Bot/commit/9320e5cc461bf998517e6dcd456e2ec1ab4f9eff

PR #1025, opened 2016-07-26
Walk within range of stops rather than the center

### LINKED ISSUE #1018: Bot unnecessarily walks to center of Pokestops
This issue may also increase detectability since it's unlikely that users would walk to the exact center of every stop they visit.

Also, most stops are spinnable from the road or other more accessible place. Once we start following roads, this will need fixed.

### Expected Behavior

Bot should walk to anywhere within the range of the stop.

### Actual Behavior

Bot walks to exactly the center of a stop.

### Steps to Reproduce

Run the bot and observe its path.

bstpierr:
Used the same range of 10 that the stepper uses to deem when it’s close
enough.

This should fix #1018


TheSavior:
LGTM

douglascamata:
LGTM

PR #1026, opened 2016-07-26
Fix trying to recycle more items than we have

alvarogzp:
The call to `self.bot.item_inventory_count(item_id)` already returns collected items, so no need to sum them.

This is causing errors while recycling items.


solderzzc:
Nice catch !

PR #1038, opened 2016-07-26
Fix item_filter format change in config

Kinodim:
* change item_filter type and default value in config


surfaace:
doing commandline switch for item_filter is now to 'complex' and therefore should be removed as argument

PR #1054, opened 2016-07-26
Fix UnicodeEncodeError causing location not found

vieira:
Short Description: 
When there are special characters in initial location, e.g. "ç" a UnicodeEncodeError is raised and caught causing valid initial locations being reported as not parsable.

Fixes:
- UnicodeEncodeError causing location not found when initial location contains special chars, e.g. "ç".

### DIFF
--- a/pokemongo_bot/__init__.py
+++ b/pokemongo_bot/__init__.py
@@ -434,12 +434,11 @@ class PokemonGoBot(object):
 
         if self.config.location:
             try:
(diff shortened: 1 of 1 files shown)


douglascamata:
@vieira did you test this?

vieira:
Yes, I did. You can use e.g. "Praça do Comércio, Lisbon" as initial location to reproduce the problem and test the fix.

Calcyfer:
Tested and confirmed as working.

Issue #1046, opened 2016-07-26
Problem with location and move


mohakevin:
### Expected Behavior
Set location and begin to move

### Actual Behavior
Incorrect location or movement

### Steps to Reproduce
Log information:

[2016-07-26 20:17:59] [x] PokemonGO Bot v1.0
[2016-07-26 20:17:59] [x] Configuration initialized
[2016-07-26 20:17:59] [x] Coordinates found in passed in location, not geocoding.
[2016-07-26 20:17:59]
[2016-07-26 20:17:59] [x] Address found: 41.656144,-0.879090
[2016-07-26 20:17:59] [x] Position in-game set as: (41.656144, -0.87909, 0.0)
[2016-07-26 20:17:59]
...
[2016-07-26 20:18:01] [#] Scanning area for objects (1 / 25)
[2016-07-26 20:18:01] [#] Walking from (**2.03e-322, 0.0**) to (41.656144, -0.87909) for approx. **1113653.00 seconds**
[#] Attempting to evolve all pokemons ...

Some ideas what could be the problem?

Regards and thank you!

k3v1n-eu:
I have that problem as well since some minutes.

03t02:
Same problem...since some minutes too. I'm using master branch

karasube:
Same one.

config.json
```json
{
    "auth_service": "google",
    "username": "***",
    "password": "***",
    "location": "**coord**, ***coord***",
    "gmapkey": "***",
    "max_steps": 5,
    "mode": "all",
    "walk": 4.16,
    "debug": false,
    "test": false,
    "initial_transfer": 0,
    "location_cache": true,
    "distance_unit": "km",
    "item_filter": "101,102,103,104",
    "evolve_all": "NONE"
}
```

python --version 
`Python 2.7.9`

git pull
`Already up-to-date.`
(branch is master)

Linux Ubuntu 16.04

killerhis:
I'm having the same problem!

Critically:
Same problem, may be Niantic do something...

jovobe:
This is a problem caused by an update in the pgoapi from tejado. It can easily be fixed by pinning the pgoapi pip to the "v1.1.0" tag. Change the first line in the requirements.txt in this way:
`-e git+https://github.com/tejado/pgoapi.git@v1.1.0#egg=pgoapi`
and update your pip with
`pip install -r requirements.txt`

palakz:
[...]

Issue #1045, opened 2016-07-26
Problem with location with minus


olivierg13:
### Expected Behavior
Should work with location = "37.8059,-122.4289" for example

### Actual Behavior
Shows errors:
`Location Found: 37.8059,-122.4289
[13:15:39] [x] The location given using -l could not be parsed. Checking for a cached location.
[13:15:39] [x] Parsing cached location...
No cached Location. Please specify initial location.`

### Steps to Reproduce
Set up a normal location: "37.8059,122.4289"
It works.
Set up an inferior to zero location (US etc...): "37.8059,-122.4289"
It does not work

### Other Information
OS: MAC OS X 10.11.6
Git Commit: 3364aa39d50dbc7e34541007e66ed67ab4071c35

Thank you in advance :)

jovobe:
This is a problem caused by an update in the pgoapi from tejado. It can easily be fixed by pinning the pgoapi pip to the "v1.1.0" tag. Change the first line in the requirements.txt in this way:
`-e git+https://github.com/tejado/pgoapi.git@v1.1.0#egg=pgoapi`
and update your pip with
`pip install -r requirements.txt`

olivierg13:
Thanks @jovobe 
Unfortunately I got the following error after modification of the requirements.txt:
`Uninstalling six-1.4.1:
Exception:
Traceback (most recent call last):
  File "/Library/Python/2.7/site-packages/pip/basecommand.py", line 215, in main
    status = self.run(options, args)
  File "/Library/Python/2.7/site-packages/pip/commands/install.py", line 317, in run
    prefix=options.prefix_path,
  File "/Library/Python/2.7/site-packages/pip/req/req_set.py", line 736, in install
    requirement.uninstall(auto_confirm=True)
  File "/Library/Python/2.7/site-packages/pip/req/req_install.py", line 742, in uninstall
    paths_to_remove.remove(auto_confirm)
  File "/Library/Python/2.7/site-packages/pip/req/req_uninstall.py", line 115, in remove
    renames(path, new_path)
  File "/Library/Python/2.7/site-packages/pip/utils/__init__.py", line 267, in renames
    shutil.move(old, new)
  File "/System/Library/Frameworks/Python.framework/Versions/2.7/lib/python2.7/shutil.py", line 302, in move
    copy2(src, real_dst)
  File "/System/Library/Frameworks/Python.framework/Versions/2.7/lib/python2.7/shutil.py", line 131, in copy2
[...]

Issue #1039, opened 2016-07-26
Walking from (0.0, 0.0) to


Zgrkaralar:
### Expected Behavior
Normaly walk some near place. All my bot now showing this :(

### Actual Behavior
![screen shot 2016-07-26 at 10 55 43 pm](https://cloud.githubusercontent.com/assets/879674/17153231/5966e1d0-5384-11e6-990d-fe56e6fd6430.png)

### Other Information
OS: Ubuntu 14
Git Commit: ff665d625c028d490fe9d2fd4e6b0b8c10a2d9c9

kevinxh:
where is your start location?

Zgrkaralar:
@kevinxh 
![screen shot 2016-07-26 at 11 08 15 pm](https://cloud.githubusercontent.com/assets/879674/17153543/df41c292-5385-11e6-9a63-a1ab8f242e96.png)

LasseSkogland:
remove the `,17` at the end

Zgrkaralar:
@LasseSkogland same problem

LasseSkogland:
The bot does this when it receives a incorrect location, or bad coordinates.
Show me your config file(remove all credentials and gmaps key).

Zgrkaralar:
@LasseSkogland 
![conf](https://cloud.githubusercontent.com/assets/879674/17154062/213ab7f6-5388-11e6-88b4-789b122fb2ad.png)

Mello-Yello:
"location": "46.06XXXX,13.23XXXX"

```
[2016-07-26 22:11:51] [x] Starting PokemonGo Bot....
[2016-07-26 22:11:51] [#] Scanning area for objects (1 / 25)
[2016-07-26 22:11:51] [#] Walking from (2.27e-322, 6.4e-323) to (46.06XXXX, 13.23XXXX) for approx. 1270058.00 seconds
```

If I use a string like I used to do, I get this:
`[22:24:37] [x] The location given using -l could not be parsed. Checking for a cached location.`

fstepho:
Same here in master branch, but with source location (2.37e-322, 1e-323) 
it worked fine a couple hours ago 

`[2016-07-26 22:20:58] [#] Walking from (2.37e-322, 1e-323) to (48.863611, 2.326667) for approx. 1307205.00 seconds`

LasseSkogland:
I'm currently getting the same error.
Will look into it in more detail, please hold!

[...]

PR #1058, opened 2016-07-26
[Bugfix: Strange Location] Pins the pgoapi pip to v1.1.0 tag to fix f2i/i2f bug

jovobe:
Short Description: 
Due to an update in the pgoapi the f2i and i2f functions are not needed anymore. To fix this in the current code the fastest solution is to pin the pgoapi pip ti the v1.1.0 tag.
After this PR the users have to `pip install -r requirements.txt` again.

Fixes:
- Strange locations due to update in the pgoapi

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,4 +1,4 @@
--e git+https://github.com/tejado/pgoapi.git#egg=pgoapi
+-e git+https://github.com/tejado/pgoapi.git@v1.1.0#egg=pgoapi
 geopy==1.11.0
 protobuf==3.0.0b4
 requests==2.10.0


douglascamata:
`pgoapi` should have been pinned to an specific commit (I coded this pin), but someone removed the check. Did you test if tag `1.1.0` does work for us in every way we use the api?

douglascamata:
@jovobe I see that `pgoapi` got an updated version go the protobuf structures and changed code a bit in the last few hours. So i'm urgently merging this. Thanks!

jovobe:
No, I just started the bot and it seems to work properly.

jovobe:
This should fix #1046, #1045 and #1039.

LasseSkogland:
I removed all references to i2f and made a patch, bot works for now.
https://gist.github.com/LasseSkogland/f4c9fb3459d1ef5f4a2a6b842d6c4594

Zgrkaralar:
@LasseSkogland did you update dev version ?

LasseSkogland:
Yes, but couple seconds after i created the patch a new commit came in, somewhat invalidating the patch file..

Calcyfer:
@LasseSkogland Could you make a PR removing the i2f method from the code?

Zgrkaralar:
now dev version is working on me

LasseSkogland:
Yes dev version works, but it is a lazy fix. I will create a pull request and properly fix this issue.

Calcyfer:
Thanks Lasse, it's appreciated.

Zgrkaralar:
@LeaklessGfy yes not stable :(

PR #1061, opened 2016-07-26
Merge cells together to avoid staying in one cell too long

bstpierr:
This should help mitigate the issue where the bot travels to a stop
that is farther than a nearby one because the nearer one is in another
cell.

I also release control back to the make loop after catching any pokemon.

Example Path: 
screen shot 2016-07-26 at 2 44 22 pm

### DIFF
--- a/pokemongo_bot/__init__.py
+++ b/pokemongo_bot/__init__.py
@@ -236,10 +236,8 @@ class PokemonGoBot(object):
                            0], self.position[1], x['latitude'], x['longitude']))
 
                 # Move to and spin the nearest stop.
-                worker = MoveToFortWorker(forts[0], self)
-                worker.work()
-                worker = SeenFortWorker(forts[0], self)
-                worker.work()
+                MoveToFortWorker(forts[0], self).work()
+                SeenFortWorker(forts[0], self).work()
 
     def _setup_logging(self):
         self.log = logging.getLogger(__name__)
--- a/pokemongo_bot/__init__.py
+++ b/pokemongo_bot/__init__.py
@@ -43,7 +43,7 @@ class PokemonGoBot(object):
         location = self.navigator.take_step()
         cells = self.find_close_cells(*self.position[0:2])
 
-        cells = self.find_close_cells(*location)
+        cells += self.find_close_cells(*location)
 
         # Combine all cells into a single dict of the items we care about.
         forts = []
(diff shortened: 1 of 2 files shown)


pokemongo_bot/__init__.py
```diff
@@ -42,8 +43,22 @@ def take_step(self):
         location = self.navigator.take_step()
         cells = self.find_close_cells(*self.position[0:2])
 
+        cells = self.find_close_cells(*location)
```
lexxxel: I just stumbled over this PR and I'm not able to test it myself right now, but shouldn't it be a concatenation between the cells from line 44 (new) and line 46 (new) instead of a replacement?

Cameronlund4:
First of all, why remove these?
`if hack_chain > 10:		
                         #print('need a rest')		
                         break		
                     if self.config.mode == "poke":		
                         break`
Also, you forgot () after the last worker.work....
` # Move to and spin the nearest stop.
    worker = MoveToFortWorker(forts[0], self)
    worker.work()
    worker = SeenFortWorker(forts[0], self)
    worker.work`
Test code before pulling.

pokemongo_bot/__init__.py
```diff
@@ -210,22 +225,19 @@ def work_on_cell(self, cell, position):
...
                 forts.sort(key=lambda x: distance(self.position[
                            0], self.position[1], x['latitude'], x['longitude']))
 
-                for fort in forts:
-                    worker = MoveToFortWorker(fort, self)
-                    worker.work()
-
-                    worker = SeenFortWorker(fort, self)
-                    hack_chain = worker.work()
-                    if hack_chain > 10:
```
Cameronlund4: Why remove the hack_chain checks?

pokemongo_bot/__init__.py
```diff
@@ -210,22 +225,19 @@ def work_on_cell(self, cell, position):
...
-                        #print('need a rest')
-                        break
-                    if self.config.mode == "poke":
-                        break
+                # Move to and spin the nearest stop.
+                worker = MoveToFortWorker(forts[0], self)
+                worker.work()
+                worker = SeenFortWorker(forts[0], self)
+                worker.work
```
Cameronlund4: Forgot to actually call the work method... you just reference it

pokemongo_bot/__init__.py
bstpierr: It no longer goes through all the forts. It returns control back out to the main loop. Without this the bot wouldn't consider any new paths and just go through every fort in order of distance from the initial point.

This way it goes to the nearest, then re-assess were the nearest fort is.

TheSavior:
This looks great to me.

PR #1090, opened 2016-07-27
Set evolve speed in config

duhminick:
Short Description:  Allows the users to set the amount of time between evolves to simulate evolving on the app.

Fixes:
- Added `evolve_speed` to **config.json.example** and **config.json.pokemons.example**
- Added in to **evolve_all_worker.py**

If `evolve_speed` isn't in the config, it'll default to 3.7 (which is what it's set at right now)


solderzzc:
Could you please have the README.me updated.

solderzzc:
Good

xc-mezcal:
Plz add a random number to the evolve_speed

Like 20 will become 20 +- 2

Well, default 3.7 don't need that feather :D

duhminick:
@xc-mezcal The sleep method does add a bit to the number.

You can check it here if you want to see it for yourself: https://github.com/PokemonGoF/PokemonGo-Bot/blob/master/pokemongo_bot/human_behaviour.py#L8

PR #1059, opened 2016-07-26
Add metrics logging and output on close

bstpierr:
Output looks like the following:

[17:10:07] Exiting PokemonGo Bot
[17:10:07]
[17:10:07] Ran for 0:00:20
[17:10:07] Total XP Earned: 210  Averaging: 36534.47/h
[17:10:07] Travelled 0.01km
[17:10:07] Visited 0 stops
[17:10:07] Encountered 1 pokemon, 1 caught, 0 released, 0 evolved, 0 never seen before
[17:10:07] Threw 1 pokeball
[17:10:07] Earned 100 Stardust
[17:10:07]
[17:10:07] Highest CP Pokemon: Nidoran M [CP: 75] [IV: 1/10/5] Potential: 0.36
[17:10:07] Most Perfect Pokemon: Nidoran M [CP: 75] [IV: 1/10/5] Potential: 0.36

PR #1093, opened 2016-07-27
Track and log total distance walked

BrianDawson:
Short Description:  Tracks and logs total distance walked. Easy way to monitor how far you have walked, that way you know about when you can hatch your eggs.

### DIFF
--- a/pokemongo_bot/stepper.py
+++ b/pokemongo_bot/stepper.py
@@ -10,7 +10,7 @@ from s2sphere import CellId, LatLng
 from google.protobuf.internal import encoder
 
 from human_behaviour import sleep, random_lat_long_delta
-from cell_workers.utils import distance, i2f, format_time, convert
+from cell_workers.utils import distance, i2f, format_time, format_dist
 
 from pgoapi.utilities import f2i, h2f
 import logger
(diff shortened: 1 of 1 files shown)


pokemongo_bot/stepper.py
```diff
@@ -31,6 +30,7 @@ def __init__(self, bot):
         self.steplimit2 = self.steplimit**2
         self.origin_lat = self.bot.position[0]
         self.origin_lon = self.bot.position[1]
+        self.totalDistance = 0
```
MaxLeiter: this project generally uses _ instead of camelCasing

pokemongo_bot/stepper.py
```diff
@@ -90,6 +91,7 @@ def _walk_to(self, speed, lat, lng, alt):
             self.api.set_position(lat, lng, alt)
             self.bot.heartbeat()
             logger.log("[#] Finished walking")
+            logger.log('[#] Total distance walked: ' + str(convert(self.totalDistance, "m", "km")) + ' kilometers | ' + str(convert(self.totalDistance, "m", "mi")) + ' miles.')
```
MaxLeiter: Please have the unit displayed be the unit set in the config
https://github.com/PokemonGoF/PokemonGo-Bot/blob/master/config.json.example#L14

bstpierr:
This won't necessarily be correct. The game will discount any movement that is done too quickly. I already have a PR to track distance as well as many other stats: https://github.com/PokemonGoF/PokemonGo-Bot/pull/1059

BrianDawson:
@MaxLeiter Updated

solderzzc:
we don't accept PR on master branch to keep thing easy.

Issue #1031, opened 2016-07-26
Location caching doesn't work


luksireiku:
### Expected Behavior
To start from the location where the bot was last time when it was running.

### Actual Behavior
If there is location in config, starts always from that point, if not: `No cached Location. Please specify initial location.`

### Steps to Reproduce
Start the bot with and without location on config.

### Other Information
OS: 
Git Commit: (run 'git log -n 1 --pretty=format:"%H"' and paste it here)
3aca6d798f8a3e173dce0beb7b13c9505cc5b251

duhminick:
The quickest fix right now is to just create the file yourself.

If you're in Linux go to the web/ and data/ directories and type `touch location-YOURUSERNAME.json`

If you're on Windows go to the same two directories and create the file.

luksireiku:
I have the files created and still doesn't work, using master branch.

catchemnon:
not working on dev either 
Git Commit: 9575a1c10f64bbdc4dcbbdc0b4d3c19b80add8f9

rwohleb:
I'm on the master branch (3aca6d7) on OSX. As per issue #1037, I touched the missing last-location and location files and restarted the bot to get these files updating properly. Once I verified this was good I was able to update config.json and set location to false. Another restart and it was correctly pulling from the cache instead of the same initial location.

luksireiku:
Touched "last-location-MYUSERNAME.json" in data and set `location = false`, worked, but it should be written on the README how to make this working.

Kassadd:
I have in my last-location (updated properly after run the bot with a location): {"lat": 40.34413435159183, "lng": -3.829348779434767}, but when I set location : false, location-cache : true and I restart the bot, the log display;

Location Found: false
GeoPosition: (45.575847, 11.4561019, 0)

So I get a softban, because the bot starts in a very far location....

Any idea?

rwohleb:
Make sure that you have it as:
"location": false,

Not:
"location": "false",

skorokithakis:
This is because the code is trying to set the position in `__init__.py:407`:

```
self.position = (location_json['lat'], location_json['lng'], 0.0)
```

However, there's no setter for the `self.position` property, so this raises an exception. It has nothing to do with whether the files are there or not.

~~~
[04:49:53] PokemonGO Bot v1.0
[04:49:53] No config argument specified, checking for /configs/config.json
[04:49:53] Configuration initialized
[...]

PR #1100, opened 2016-07-27
Fix Location caching doesn't work (#1031)

kbinani:
Short Description: 

Fix `can't set attribute` exception here:

```
self.position = (location_json['lat'],
                 location_json['lng'], 0.0)
```

Fixes:
- #1031


skorokithakis:
I can confirm that this fixes the problem.

louisgv:
LGTM as well.

solderzzc:
Good catch.

PR #1101, opened 2016-07-27
when filter set to 0, it will failed

wdjwxh:
Short Description: 

Fixes:
when filter set to zero. it will failed.


solderzzc:
Good catch..

Issue #948, opened 2016-07-26
Bot crashes after a few hours of uptime


TheGoldenXY:
### Expected Behavior
The bot shouldn't crash at all. It should be stable and run 24/7 without crashes.

### Actual Behavior
The bot crashes after a few hours uptime. It's crashing at different times. Sometimes 30 mins, 2 hours, etc etc.

http://m.imgur.com/rZ0Z2D5

### Steps to Reproduce
Use the latest version and run the bot.

### Other Information
OS: Ubuntu 14.04
Git Commit: (run 'git log -n 1 --pretty=format:"%H"' and paste it here)
3aca6d798f8a3e173dce0beb7b13c9505cc5b251

MeGaPk:
+1
I have same exception:
```
Traceback (most recent call last):
  File "./pokecli.py", line 220, in <module>
    main()
  File "./pokecli.py", line 211, in main
    bot.take_step()
  File "/home/server/PokemonGo-Bot/pokemongo_bot/__init__.py", line 36, in take_step
    self.stepper.take_step()
  File "/home/server/PokemonGo-Bot/pokemongo_bot/stepper.py", line 62, in take_step
    self._work_at_position(position[0], position[1], position[2], True)
  File "/home/server/PokemonGo-Bot/pokemongo_bot/stepper.py", line 139, in _work_at_position
    self.bot.work_on_cell(cell, position, pokemon_only)
  File "/home/server/PokemonGo-Bot/pokemongo_bot/__init__.py", line 93, in work_on_cell
    worker.work()
  File "/home/server/PokemonGo-Bot/pokemongo_bot/cell_workers/move_to_fort_worker.py", line 30, in work
    self.stepper._walk_to(self.config.walk, *position)
  File "/home/server/PokemonGo-Bot/pokemongo_bot/stepper.py", line 88, in _walk_to
    alt, False)
  File "/home/server/PokemonGo-Bot/pokemongo_bot/stepper.py", line 102, in _work_at_position
    response_dict = self.api.call()
  File "/home/server/PokemonGo-Bot/src/pgoapi/pgoapi/pgoapi.py", line 81, in call
    response = request.request(api_endpoint, self._req_method_list, player_position)
  File "/home/server/PokemonGo-Bot/src/pgoapi/pgoapi/rpc_api.py", line 98, in request
    raise NotLoggedInException()
pgoapi.exceptions.NotLoggedInException
```

tb87:
+1
[...]

PR #1115, opened 2016-07-27
Added multiaccount automation scripts

ElGullo:
I wrote a couple of bash scripts to automate multiaccount botting with screen.
Although totally not well written i found them really handy, expecially when running the bot on a vps.

Basically this adds the ability to start and stop multiple bots at a time in separate screen sessions (and with separate configs).
detailed features are described in the how to file.

I'm not sure if this file hierarchy is correct and if the how to should go into the readme or not, please let me know.

The loop feature is taken from #948, i know the relogin issue and other crashes were solved in recent commits but still i think its handy for unexpected crashes when the bot is left unattended.

### DIFF
--- a/multiaccount_tools/how_to_configure
+++ b/multiaccount_tools/how_to_configure
@@ -1,0 +1,40 @@
+How to configure the multibot scripts
+
+first you need to install screen.
+how to depends on your package manager (assuming you are running linux).
+
+then change the parameters in EACH file to suit your preferences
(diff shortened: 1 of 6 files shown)


solderzzc:
We should not encourage people run multiple accounts.

MeGaPk:
imho, strange solution, and more hardcoded..
I think better write in python and launch in sub proccess. 
For multiaccount, use config1, config2 and etc...

PR #1124, opened 2016-07-27
Update _get_catch_config in pokemon_catch_worker

xc-mezcal:
Short Description: 

It should return the setting given by "any" in the catch_config file, instead of return {} for a "unspecified" pokemon.

Fixes:
delete the second  if condition


xc-mezcal:
@solderzzc What is a Travis-ci/pr build?  Why it is failed 0.0

PR #1130, opened 2016-07-27
Spawnpt change

dorklord:
Short Description: Update vars to match response dict + handle negative coords 

Fixes:
- response comes back with spawn_point_id in place of spawnpoint_id
- negative coords cause i2f to error so catch it without breaking


solderzzc:
We don't merge PR on master to keep thing easy. Please commit to dev if need.

PR #1136, opened 2016-07-27
Fixes

smallsweet:
Short Description: 
some fixes so the metrics will be printed correctly at the end of the run


TheSavior:
Thanks!

eggins:
:+1:

PR #1141, opened 2016-07-27
Allow comments in JSON files

Shoh:
Short Description: Adding ability to allow comments in JSON files by using jsmin to remove comments when reading JSON files in.  Makes it easier for people to know how to setup the config files.


TheSavior:
I'm not sure we should change this from a regular json file. I could imagine a world where the webui or some other tool want's to use these. Lots of people generate these json and parse them for other things (like running 1000s of bots). A proprietary json format feels like a pain.

Shoh:
What do you mean exactly?  It's still a regular json file - it's just enabling the use of regular JS comments inside it.  It's allowed because jsmin strips out all of the comments before reading it.  So if we want to read it in other places then we just have to pass it through jsmin first.  So I don't understand what you mean by "change from a regular json file" - it's still a regular json file in the end?

surfaace:
im actually for this change as it helps the 'brick'-users to understand the config. With that said, having a non standard json configuration file is not good either, but if you can do comments like // and /* */ i dont see an issue with this.

surfaace:
:+1:

douglascamata:
@Shoh users should learn how to configure the bot in a wiki, not in a JSON file. This is what wikis are for.

Phil9l:
Why not to use python file as config? Like `settings.py` in django or `config.py` in Flask?
So it would be something like this:
```python
AUTH_SERVICE = 'google'
USERNAME = '...'
PASSWORD = '...'
...
# Some comment are, obviously, ok here
DEBUG = False
TEST = False
```

And in `pokecli.py` something like this:
```python
import settings

if settings.DEBUG:
    pass
```

douglascamata:
@Phil9l Python file as config is a much better suggestion!

Phil9l:
@douglascamata btw, it would also allow config file separation.
Something like this in `settings.py`:
```python
try:
    from pokemon_settings import *
except:
    pass

AUTH_SERVICE = 'google'
USERNAME = '...'
...
```

douglascamata:
@Phil9l I asked some opinions in Slack and I think we're sticking with JSON and will update the wiki with instructions.

configs/config.json.example
```diff
@@ -1,39 +1,131 @@
...
     "max_steps": 5,
+
+    /*  modes:
+            "all"   - catch pokemon + farm pokestops
+            "poke"  - start out only catching pokemon
+            "farm"  - start out only farming pokestops
+    */
     "mode": "all",
+
+    // Walk speed - 4.16 is the max speed the bot can walk and still hatch eggs
```
Ryurem: The max walking speed for egg hatching ist 14.
I would also add that it measures in km/h

douglascamata: This speed is in meter per seconds. It takes a lot of time to actually walk because the bot does API calls and other stuff that takes some time.

Shoh: @Ryurem I wasn't the one who stated that 4.16 is the max speed for hatching eggs.  I just put it here because that is what others have said and what is documented elsewhere.  If it's wrong then we can change it.

Shoh:
@douglascamata I agree that we shouldn't need this and it should all be in the wiki.  The fact is though that people are too damn lazy and will rather just pop into chat and ask "how do I set the bot to evolve my pokemon?" rather than going and reading the wiki.  I only decided to add this because I'm tired of people coming in and asking how to configure the bot.  With this, they have no damn excuse because it's right in front of their damn face when they open the config file.

Comments in the JSON file, while not acceptable as part of the standard, IS acceptable when it is minified.  Douglas Crockford (the creator of JSON) says so himself: https://plus.google.com/+DouglasCrockfordEsq/posts/RK8qyGVaGSr

> Suppose you are using JSON to keep configuration files, which you would like to annotate. Go ahead and insert all the comments you like. Then pipe it through JSMin before handing it to your JSON parser.﻿

"Pipe it through JSMin before handing it to your JSON parser."  Which is EXACTLY what is being done here.  Hate the idea of comments in JSON?  I don't know why.  If it doesn't change the parts that matter (the actual data), I don't know why you would be against it.

mercuriete:
I dont feel that nor adding comments to a json neither using a python file are good choices...
If you want to have comments in config files.. maybe using yaml format is more human readable.

I still prefer json as it is now in master but is a personal opinion.

Keep going with pokemon go, you are an awesome team :+1: 
Cheers!

Shoh:
If people are still totally against comments in the config JSON (even though it doesn't affect the bot itself in any way, shape, or form because those comments never even make it past the minifier calls when reading in the JSON file in the initial setup of the bot), the what about having the comments/description of each config setting placed in a config_readme.txt or something?  That way people don't need to go looking too far to find out how to configure the bot.

configs/config.json.example
Ryurem: @douglascamata 
Analyse the source code if you don't believe me

solderzzc:
@Shoh  Your contribution is very welcome, we are currently working on a Desktop version to easy the configuration effort :) https://github.com/PokemonGoF/PokemonGo-Bot-Desktop

douglascamata:
I'm sorry, but comment in the JSON files will never be accepted. You can start a new issue to discuss about better documentation for configuration files or change or format (comments only allowed if format natively support it).

PR #1142, opened 2016-07-27
Refactor: Catch visible pokemon worker

TheSavior:
Refactoring this stuff out into a worker instead of being in the init.


tstumm:
👍

PR #1019, opened 2016-07-26
Dump cells to enable custom front-end functionalities

Davideddu:
It now dumps the cell list as a json into data/cells-$username.json, so that more front-ends can use this information (I'm working on a [Kivy-based one](https://github.com/Davideddu/PokeVy)).

PR #1145, opened 2016-07-27
Dump cells to enable custom front-end functionalities

surfaace:
Fixed merge issues for #1019

It now dumps the cell list as a json into data/cells-$username.json, so
that more front-ends can use this information (I'm working on a
Kivy-based one).


solderzzc:
👍

PR #1146, opened 2016-07-27
Adding a WorkerResult and the MoveToFortWorker only takes a single step

TheSavior:


pokemongo_bot/__init__.py
```diff
@@ -186,8 +189,12 @@ def work_on_cell(self, cell, position, work_on_forts=1):
...
                 if len(forts) > 0:
                     # Move to and spin the nearest stop.
-                    MoveToFortWorker(forts[0], self).work()
-                    SeenFortWorker(forts[0], self).work()
+                    if MoveToFortWorker(forts[0], self).work() == WorkerResult.RUNNING:
+                        return
+                    if SeenFortWorker(forts[0], self).work() == WorkerResult.RUNNING:
+                        return
+
+        self.navigator.take_step()
```
TheSavior: This is probably the biggest change in this PR. We now take a step on each tick only if the other things we do on that step are all done. This is different from the way it was before where we took a step with the navigator on *every* tick.

solderzzc:
👍

tstumm:
👍

PR #1159, opened 2016-07-27
Moving these flags into the workers. Make them run on each tick

TheSavior:


pokemongo_bot/__init__.py
```diff
@@ -159,17 +159,13 @@ def work_on_cell(self, cell, position, work_on_forts=1):
         # Check if session token has expired
         self.check_session(position)
 
-        if self.config.initial_transfer:
```
TheSavior: This flag was already inside of InitialTransferWorker so it was being checked twice.

tstumm:
👍

douglascamata:
:+1:

PR #1160, opened 2016-07-27
Removing duplicated release logic from catching pokemon

TheSavior:
This logic is an exact copy paste from the ReleasePokemonWorker. Since that will be called on every tick as of https://github.com/PokemonGoF/PokemonGo-Bot/pull/1159, then we don't need this logic here.


tstumm:
Seems legit.
👍

douglascamata:
👍

PR #1182, opened 2016-07-27
Initial Transfer filter

Witzbold:
Short Description: 
It would be nice to have a filter for the initial transfer function. Up to now one can only go down to 1 pokemon that is kept. It would be nice to be able to change this value to 2 which will keep 2. 
Furthermore it would be nice to be able to use the feature on specific pokemon. For example one wants to keep his Weedles with 10-300 CP but also wants to release his Pigeys with 300-500 CP


surfaace:
do i understand this correctly.. you want to merge from our master branch to our devbranch?
also readme in dev branch has been majorly updated making this readme obsolete

surfaace:
all included in this PR is either merged at some point, or is modified to work in the same way as suggested here

PR #1185, opened 2016-07-27
problem solved for not going to poketop

anonymouslulz22:
Short Description: 
fix for not going to the pokestop
Fixes:
- 
- 
- 

so today i pulled the latest dev and tried to run and saw that the bot is just finding a fort and not moving to it! here is a fix for it and i have tested it and its working again

### DIFF
--- a/pokemongo_bot/cell_workers/move_to_fort_worker.py
+++ b/pokemongo_bot/cell_workers/move_to_fort_worker.py
@@ -41,13 +41,13 @@ class MoveToFortWorker(object):
                 )
 
 
-           step_walker = self._step_walker
+        step_walker = self._step_walker
             if step_walker == True:
                 self.api.player_update(latitude=lat, longitude=lng)
                 response_dict = self.api.call()
                 logger.log('Arrived at Pokestop')
                 return WorkerResult.SUCCESS
-            elif step_walker == False:
+                elif step_walker == False:
                 return WorkerResult.RUNNING
 
         logger.log('Arrived at Pokestop')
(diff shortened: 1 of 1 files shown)


gpires:
line 44 is not indented.

anonymouslulz22:
yea! let me check out

tstumm:
Your indention is wrong

gpires:
```
        step_walker = self._step_walker
        if step_walker == True:
            self.api.player_update(latitude=lat, longitude=lng)
            response_dict = self.api.call()
            logger.log('Arrived at Pokestop')
            return WorkerResult.SUCCESS
        elif step_walker == False:
            return WorkerResult.RUNNING
```

Like this?

pokemongo_bot/cell_workers/move_to_fort_worker.py
```diff
@@ -41,7 +41,13 @@ def work(self):
                 )
 
 
-            if not self._step_walker.step():
+        step_walker = self._step_walker
+            if step_walker == True:
+                self.api.player_update(latitude=lat, longitude=lng)
```
douglascamata: please just add this `self.api.player_update` calls  after each step is taken by the `StepWalker` and revert all the other changes in this file.

pokemongo_bot/cell_workers/move_to_fort_worker.py
```diff
@@ -41,8 +41,16 @@ def work(self):
                 )
 
 
-            if not self._step_walker.step():
+            step_walker = self._step_walker
+                if step_walker == True:
+                self.api.player_update(latitude=lat, longitude=lng)
```
douglascamata: don't know if you saw my comment, but please move the `self.api.player_update` call to after the `StepWalker` finishes moving and leave this worker untouched.

tstumm:
This does not fix the problem. We just found the problematic code and fixed it. PR is coming.

anonymouslulz22:
ok!

PR #1188, opened 2016-07-27
Revert "Removing duplicated release logic from catching pokemon"

surfaace:
Reverts PokemonGoF/PokemonGo-Bot#1160

Reverted because without this pokemon isnt released if having evolve_captured : false.
until seperate release_pokemon_worker is done code needs to be in

### DIFF
--- a/pokemongo_bot/cell_workers/pokemon_catch_worker.py
+++ b/pokemongo_bot/cell_workers/pokemon_catch_worker.py
@@ -209,9 +209,52 @@ class PokemonCatchWorker(object):
                                             logger.log(
                                             'Failed to evolve {}!'.format(pokemon_name))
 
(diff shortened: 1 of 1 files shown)


douglascamata:
@surfaace There is the `InitialReleaseWorker` that can be updated and renamed to `ReleaseWorker`

douglascamata:
I don't think reverting was a good idea. Improving `InitialReleaseWorker` is.

surfaace:
until InitialReleaseWorker is fixed, we should be able to transfer away pokemons based on filter when catched, else we causes us a 'bug'

douglascamata:
@surfaace so why not fix the bug instead of reverting a good refactor?

surfaace:
it wasnt a bug before you removed the code :) modifying workers now

PR #1194, opened 2016-07-27
combine configs

VictorChen:
Don't think we need `config.json.example` AND `config.json.pokemons.example` since they are basically identical (except one has a list of release pokemons)? Let's combine them instead?


MaxLeiter:
:-1: people will probably `cp config.json.example config.json`; we don't want them to have to remove all the unnecessary Pokemon lines

surfaace:
:-1:

PR #1202, opened 2016-07-27
Update README.md

fdelia:
Short Description: 
Extended the configuration for the nginx server and web client. For beginners, just to help them through all steps.


KLVN:
#1161
Maybe you should move this to the wiki.

surfaace:
:-1: Web folder is not apart of this repo so we only need to reference where it is and they will have to handle howto config

Which has already been done here: https://github.com/PokemonGoF/PokemonGo-Bot/pull/1202

PR #1205, opened 2016-07-27
Randomize `normalized_reticle_size` and `spin_modifier` parameter

kbinani:
Short Description: 
---
When calling `catch_pokemon` API, parameter `normalized_reticle_size` is fixed to `1.950` and also `spin_modifier` is fixed to `1`.

This PR makes them random.

Fixes:
---
- `normalized_reticle_sizes` and `spin_modifier` are fixed to constant values

### DIFF
--- a/CONTRIBUTORS.md
+++ b/CONTRIBUTORS.md
@@ -42,3 +42,4 @@
  * SpaceWhale
  * klingan
  * reddivision
+ * kbinani
--- a/configs/config.json.example
+++ b/configs/config.json.example
@@ -37,6 +37,8 @@
     "longer_eggs_first": true,
     "evolve_captured": false,
     "release_pokemon": true,
+    "catch_randomize_reticle_factor": 1.0,
+    "catch_randomize_spin_factor": 1.0,
     "catch": {
       "any": {"catch_above_cp": 0, "catch_above_iv": 0, "logic": "or"},
       "// Example of always catching Rattata:": {},
--- a/configs/config.json.pokemon.example
+++ b/configs/config.json.pokemon.example
@@ -36,6 +36,8 @@
     "longer_eggs_first": true,
     "evolve_captured": false,
     "release_pokemon": true,
+    "catch_randomize_reticle_factor": 1.0,
+    "catch_randomize_spin_factor": 1.0,
     "catch": {
       "any": {"catch_above_cp": 0, "catch_above_iv": 0, "logic": "or" },
 
--- a/pokecli.py
+++ b/pokecli.py
@@ -324,6 +324,22 @@ def init_config():
         type=int,
         default=10,
     )
+    add_config(
+        parser,
+        load,
+        long_flag="--catch_randomize_reticle_factor",
+        help="Randomize factor for pokeball throwing accuracy (DEFAULT 1.0 means no randomize: always 'Excellent' throw. 0.0 randomizes between normal and 'Excellent' throw)",
+        type=float,
+        default=1.0
+    )
+    add_config(
+        parser,
+        load,
+        long_flag="--catch_randomize_spin_factor",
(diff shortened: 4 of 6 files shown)


MikeDX:
This should be a configurable settings, since some people (like me) prefer to have the normalised value and get the xp bonus every catch

kbinani:
```
    ...
    "evolve_captured": false,
    "humanize_catch_parameters": true,
    "catch": {
    ...
```

I'm trying to add `humanize_catch_parameters` to `config.json`.
If this is set to `true`, `normalized_reticle_size` and `spin_modifier` will be randomized.

How about the default value?

Likael:
Rather than a true/false, wouldn't be better to fill a % for how much randomness you want ?
Something like that :

```
    ....
    catch_randomize_reticule_factor : 0.8
    catch_randomize_spin_chance : 0.6
    catch_randomize_hit_chance : 1.0
    ....
```
In the example above. you would be able to randomize the reticule between 80% et 100% of the optimal value (1.950?), and have 60% chance to do a spin hit.
The last parameter isn't implemented yet, but that would be nice to be able to miss sometimes too, to completely simulate a human behaviour. 
That would allow more control over the hit ratio, and add more variety amongst all bots.

kbinani:
I've updated the code based on @Likael 's idea (before EDIT2, sorry).

I've checked actual (in game) value of `normalize_reticle_size` and `spin_modifier`.
It seems `1 <= normalized_reticle_size <= 1.950` and `0 <= spin_modifier <= 1`.

Config parameters looks like this:

```
    "catch_randomize_reticule_factor": 1.0,
    "catch_randomize_spin_factor": 1.0,
```

catch_randomize_reticule_factor:
---
* `1.0` (default, and max value) means NO randomize: always 'Excellent' throw.
* `0.0` (min) means randomize between normal and 'Excellent' throw.

catch_randomize_spin_factor:
---
* `1.0` (default, and max value) means NO randomize: always perfect 'Super Spin' curve ball.
* `0.0` (min) means randomize between normal (not curved) ball between 'Super Spin' curve ball.

I didn't write code about`hit_pokemon` and is fixed to `1`, so Pokeballs will always hit Pokemon at this time.

TheSavior:
Can we update this to use the new config approach?

kbinani:
Updated to use `add_config` in `pokecli.py`

TheSavior:
:+1:

kbinani:
```
encounter_id: (...MASKED...)
pokeball: ITEM_POKE_BALL
normalized_reticle_size: 1.01213991642
spawn_point_id: "(...MASKED...)"
hit_pokemon: true
spin_modifier: 0.187229141593
```

This is an actual in-game, captured value of `CatchPokemonMessage`, when I had thrown almost not curved, and bad throw (but hit the pokemon)

kbinani:
Rebased to `dev` again

kbinani:
Rebased to `dev` again

TheSavior:
👍

douglascamata:
👍

Kassius509:
Should add to the Wiki documentation. I would but its locked. 

Thx

>` **catch_randomize_reticule_factor:**`
> 
> 1.0 (default, and max value) means NO randomize: always 'Excellent' throw.
> 0.0 (min) means randomize between normal and 'Excellent' throw.
>` **catch_randomize_spin_factor:**`
> 
> 1.0 (default, and max value) means NO randomize: always perfect 'Super Spin' curve ball.
> 0.0 (min) means randomize between normal (not curved) ball between 'Super Spin' curve ball.

mjmadsen:
This doesn't seem to be working still. I always get the curve (or nice throw) bonus but rarely get anything beyond that.

    "catch_randomize_reticle_factor": 1.0,
    "catch_randomize_spin_factor": 1.0,

kbinani:
"Catch pokemon api seems didn't throw excellent ball after server update?" https://github.com/tejado/pgoapi/issues/157

Same issue reported to upstream

PR #1217, opened 2016-07-27
Update the location & location_cache logic

xc-mezcal:
Short Description: 

Previously, the location_cache can't be used, even people set location_cache to TRUE. Everytime, it starts from the initial position.

Fixes:
-
Now the bot check the initial position first. 
Then check if people set location_cache. If so, read it and update it, otherwise use the initial location.
If there is no initial location and no location cache, print msg and exit.


bstpierr:
👍

theadriann:
Looking good. (didn't test it tho)

douglascamata:
:+1:

PR #1231, opened 2016-07-27
Reduce number of inventory calls

bstpierr:
There’s way too many API calls for the inventory, so I added a cached
response that’s used until someone invalidates it (like when
catching/throwing and spinning stops).

I also removed the calls to get_player that were completely unused and the useless get_inventory in the heartbeat.


solderzzc:
Let merge your code, many thanks.

PR #1237, opened 2016-07-27
Only work on forts when there is space in the bag. Do not switch mode

TheSavior:


pokemongo_bot/__init__.py
```diff
@@ -44,9 +44,9 @@ def start(self):
         random.seed()
 
     def take_step(self):
-        self.process_cells(work_on_forts=True)
```
TheSavior: work_on_forts was no longer set to False anywhere, since some of the changes yesterday.

pokemongo_bot/__init__.py
```diff
@@ -174,8 +174,11 @@ def work_on_cell(self, cell, position, work_on_forts=1):
         if worker.work() == WorkerResult.RUNNING:
             return
 
-        if ((self.config.mode == "all" or
-                self.config.mode == "farm") and work_on_forts):
+
+        number_of_things_gained_by_stop = 5
```
bstpierr: What does this variable do?

pokemongo_bot/__init__.py
```diff
@@ -174,8 +174,11 @@ def work_on_cell(self, cell, position, work_on_forts=1):
...
+
+        if ((self.get_inventory_count('item') < self._player['max_item_storage'] - 5) and
```
bstpierr: It's actually possible for stops to give out more than 5 items. I've seen it give out 6 in the actual game.

bstpierr:
You still get XP for spinning, even if there's no room in your bag. I'm not sure completely avoiding spinning a fort is necessary.

pokemongo_bot/__init__.py
TheSavior: Just make it clear why the number is there.

pokemongo_bot/__init__.py
TheSavior: Yeah, I think I've seen someone say 7 on reddit. I'd be happy to set it to that, but I was kinda trying to go with the average case instead of the best case.

TheSavior:
@bstpierr Yeah, you do get xp for spinning. However, our code right now says that we should spin forts until you have no more space for items, and then you should switch to catching pokemon until you have no more balls, and then it switches back to only capturing forts. 

This is a great starting point to enable the bot to handle doing both and manage its space properly. I'm sure we will end up adding configuration for people to spin forts regardless and not catch any pokemon at all if they want.

Let's start here though.

bstpierr:
I'm not sure what issue this is attempting to solve. A lot of these sorts of decisions will be fixed when we add behavior trees.

Any one else have opinions on this PR?

TheSavior:
We want to move towards behavior trees. This is part of that.

This is improving the logic for when to spin forts that enables us to run in a behavior tree style, every tick. Versus what happens now where we set a config variable and go down a completely different path to only catching pokemon or only spinning forts.

This will let us catch pokemon if we have pokebag space, and spin forts if we have item space.

This makes the behaviors actually self contained.

bstpierr:
👍

douglascamata:
:+1:

PR #1240, opened 2016-07-27
Config parsing clean-ups

cwild:
Quick bit of house keeping

Fixes:
- Do not use __dict__ attribute directly on config
- Use dict.get() with default param instead of existence checks


bstpierr:
👍 I like it!

douglascamata:
:+1:

PR #1270, opened 2016-07-27
Only catch Exception subclasses

cwild:
Do not use "except:" as this is unlikely to always behave as desired!

Fixes:
- Catch "Exception"s and let things like SystemExit, KeyboardInterrupt bubble up

### DIFF
--- a/pokemongo_bot/__init__.py
+++ b/pokemongo_bot/__init__.py
@@ -379,7 +379,7 @@ class PokemonGoBot(object):
 
                 if item_id in items_stock:
                     items_stock[item_id] = item_count
-            except:
+            except Exception:
                 continue
         return items_stock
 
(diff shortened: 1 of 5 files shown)


douglascamata:
@cwild errr.... every exception inherits from `Exception`. The bad behaviours are catching `Exception` and not exactly the type of exception you expect, and having a fucking lot of code in one `try...except` block.

cwild:
Not every exception inherits from Exception, you can see the inheritance hierarchy on this page:
https://docs.python.org/2/library/exceptions.html

This is why ```except:``` is bad:
https://docs.python.org/2.7/howto/doanddont.html#except

If you try to raise a SIGINT during the ```try... except:...``` then it will swallow the resulting ```KeyboardInterrupt``` and won't close as expected.

douglascamata:
@cwild I see.

douglascamata:
👍

surfaace:
:+1:

PR #1281, opened 2016-07-27
Transfer poke worker

surfaace:
pulled out catched pokemon and initial transfer to seperate worker


eggins:
:+1:

pokemongo_bot/cell_workers/pokemon_catch_worker.py
```diff
@@ -213,53 +216,11 @@ def work(self):
...
-                                            Set(id_list2) - Set(id_list1))
-                                        if len(pokemon_to_transfer) == 0:
-                                            raise RuntimeError(
-                                                'Trying to transfer 0 pokemons!')
-                                        self.transfer_pokemon(pokemon_to_transfer[0])
-                                        self.bot.metrics.released_pokemon()
-                                        logger.log(
-                                            '{} has been exchanged for candy!'.format(pokemon_name), 'green')
+                                    worker = PokemonTransferWorker(self.bot)
+                                    worker.release_catched_pokemon(pokemon_name, pokemon_to_transfer[0], cp, pokemon_potential)
```
douglascamata: This instantiation is wrong in 2 ways:

1. You're transferring the captured pokemon without any verification. 
2. The event loop already has a PokemonTransferWorker.work call, so no need to release pokemon in here.

TheSavior: It makes me so happy that we are trying to enforce doing these things in the event loop instead of nested deep. Thank you @douglascamata

PR #1285, opened 2016-07-27
Added handling 'None' handling

kabadisha:
Fixes:
- Some times response_gym_details['responses'] is empty and when this happens the bot would crap out. This defaults to an empty set instead and so avoids the failure.

Error experienced on and fix tested on OSX only.

### DIFF
--- a/pokemongo_bot/__init__.py
+++ b/pokemongo_bot/__init__.py
@@ -100,7 +100,8 @@ class PokemonGoBot(object):
                                                      gym_latitude=fort.get('latitude'),
                                                      gym_longitude=fort.get('longitude'))
                             response_gym_details = self.api.call()
-                            fort['gym_details'] = response_gym_details['responses']['GET_GYM_DETAILS']
+                            if response_gym_details['responses'] is not None:
+                                fort['gym_details'] = response_gym_details['responses']['GET_GYM_DETAILS']
 
         user_data_cells = "data/cells-%s.json" % (self.config.username)
         with open(user_data_cells, 'w') as outfile:
--- a/pokemongo_bot/__init__.py
+++ b/pokemongo_bot/__init__.py
@@ -100,8 +100,7 @@ class PokemonGoBot(object):
                                                      gym_latitude=fort.get('latitude'),
                                                      gym_longitude=fort.get('longitude'))
                             response_gym_details = self.api.call()
-                            if response_gym_details['responses'] is not None:
-                                fort['gym_details'] = response_gym_details['responses']['GET_GYM_DETAILS']
+                            fort['gym_details'] = response_gym_details.get('responses', {}).get('GET_GYM_DETAILS', None)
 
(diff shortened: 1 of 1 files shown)


pokemongo_bot/__init__.py
```diff
@@ -100,7 +100,8 @@ def update_web_location(self, cells=[], lat=None, lng=None, alt=None):
                                                      gym_latitude=fort.get('latitude'),
                                                      gym_longitude=fort.get('longitude'))
                             response_gym_details = self.api.call()
-                            fort['gym_details'] = response_gym_details['responses']['GET_GYM_DETAILS']
+                            if response_gym_details['responses'] is not None:
+                                fort['gym_details'] = response_gym_details['responses']['GET_GYM_DETAILS']
```
douglascamata: `fort['gym_details'] = response_gym_details.get('responses', {}).get('GET_GYM_DETAILS', None)`

kabadisha: @douglascamata: Ah, thanks! Unfortunately, having never looked at Python source before 20 minutes ago, I can only guess that your comment shows a way to set a default value of an empty result set? Is that right?

douglascamata: Yup... example:

```python
my_dict = {'hello': 1}

my_dict.get('hello', 123)
>>> 1

my_dict.get('bye', 'not found')
>>> not found
```

Got how it works?

kabadisha: Schweet. Got it - thanks buddy. Will update shortly (just testing)

douglascamata:
@kabadisha test for a while to check if it doesn't cause any issues, please.

kabadisha:
@douglascamata No problemo - will leave running over night. So far so good though. Have never been able to get the bot to run for more than 1 or two minutes at a time before, but have been running for nearly 40 minutes so far with the original 'if' fix. Running with your suggestion now for 10 minutes so far and all seems good. Will report back in the morning.

douglascamata:
@kabadisha thanks! If more ppl help testing and approves I might merge before you wake.

kabadisha:
I just got the error again... So maybe response_gym_details can be null as well?

kabadisha:
I think maybe self.api.call() is returning null sometimes. It happens quite frequently for me - I get errors like this:
PokemonGo-Bot/pokemongo_bot/__init__.py", line 144, in find_close_cells
    map_objects = response_dict.get('responses', {}).get('GET_MAP_OBJECTS', {})
AttributeError: 'NoneType' object has no attribute 'get'

Thoughts?

kabadisha:
Maybe it happens when the servers are busy and the request times out?

tstumm:
The problem is, that `response_dict` is null

kabadisha:
Yes - exactly. The previous line is: response_dict = self.api.call()
Which makes me think that self.api.call() is returning null.

tstumm:
This is something that should be handled by PGoApi. Well it they won't fix it we will have to.

PR #1286, opened 2016-07-27
Fixed seperate transfer bug in worker

surfaace:
Now checks for conditions before transfering


pokemongo_bot/cell_workers/pokemon_catch_worker.py
```diff
@@ -213,53 +216,13 @@ def work(self):
...
-                                            raise RuntimeError(
-                                                'Trying to transfer 0 pokemons!')
-                                        self.transfer_pokemon(pokemon_to_transfer[0])
-                                        self.bot.metrics.released_pokemon()
-                                        logger.log(
-                                            '{} has been exchanged for candy!'.format(pokemon_name), 'green')
+                                            
+                                    worker = PokemonTransferWorker(self.bot)
+                                    if worker.should_release_pokemon(pokemon_name, cp, pokemon_potential):
+                                        worker.release_catched_pokemon(pokemon_to_transfer[0])
```
douglascamata: @surfaace this shouldn't be here at all. This worker is already instantiated in the event loop at every tick. You don't need to call it here, it will be called after every small step the bot takes.

pokemongo_bot/cell_workers/pokemon_transfer_worker.py
```diff
@@ -75,7 +71,7 @@ def get_pokemon_potential(self, pokemon_data):
         for individual_stat in iv_stats:
             try:
                 total_iv += pokemon_data[individual_stat]
-            except Exception:
+            except:
```
douglascamata: @surfaace just one last detail to be good to go for me: put back `except Exception` here.

surfaace: good spot

TheSavior:
@surfaace Can you add a thing to the code that parses the config and throws if initial_transfer is set and tells people what to set it to? It would save us a lot of headaches answering questions.

pokemongo_bot/cell_workers/pokemon_catch_worker.py
```diff
@@ -192,6 +194,10 @@ def work(self):
                                 if status is 1:
 
                                     id_list2 = self.count_pokemon_inventory()
+                                    pokemon_to_transfer = list(Set(id_list2) - Set(id_list1))
+                                    if len(pokemon_to_transfer) == 0:
+                                        raise RuntimeError(
```
jtdroste: Maybe it's because I don't have a lot of state on this, but why are we raising a runtime error?  Shouldn't the bot just "return" in this case?

I don't think workers should cause the bot to exit

jtdroste:
👍

douglascamata:
:+1:

pokemongo_bot/cell_workers/pokemon_catch_worker.py
```diff
@@ -34,9 +34,11 @@ def work(self):
             if self.response_key in response_dict['responses']:
                 if self.response_status_key in response_dict['responses'][self.response_key]:
                     if response_dict['responses'][self.response_key][self.response_status_key] is 7:
-                        if self.config.initial_transfer:
+                        if self.config.release_pokemon:
                             logger.log('Pokemon Bag is full!', 'red')
-                            return PokemonCatchWorker.BAG_FULL
+                            worker = PokemonTransferWorker(self)
```
TheSavior: I thought we weren't going to do this?

PR #1289, opened 2016-07-27
Allow google analytics collect bot event.

solderzzc:
Short Description: 
Allow google analytics collect bot event.
Forbid personal information to be sent.


TheSavior:
This is a really scary and I imagine controversial thing to add. We are essentially gathering data on everyone using the bot and their usage. I think it would be valuable to do something like this so we can measure how the bot is being used, but we need to absolutely have a config flag that will let people disable it.

pokemongo_bot/central_ai/bot_event.py
```diff
@@ -0,0 +1,20 @@
+# -*- coding: utf-8 -*-
+from UniversalAnalytics import Tracker
+
+class BotEvent(object):
```
TheSavior: Should this be named Analytics?

douglascamata:
@solderzzc there's one guy implementing the Event/Signal system for logging and web socket and he's almost done, let's w8 for this system to merge this PR. It will be much easier.

solderzzc:
@douglascamata  Sure, let's wait.
@TheSavior This is from https://github.com/analytics-pros/universal-analytics-python
To convert Google analytics JS version into HTTP post.

douglascamata:
@solderzzc I agree with @TheSavior. If 100% need a config to disable data collection and if we set default option to collect data, we need to give users a BIG WARNING.

solderzzc:
@douglascamata  Sure, will work on it.

pokecli.py
```diff
@@ -176,6 +176,13 @@ def init_config():
         type=float,
         default=15.0
     )
+    parser.add_argument(
+        "-hr",
+        "--health_record",
+        help="Send anonymous bot event to GA for bot heath record, furthur ML will depend on it. Set \"health_record\":false if you need disable it.",
+        type=bool,
+        default=True
```
jtdroste: This should be defaulted to false.  We should encourage people to opt-in, but it should never be the default for tracking.

eggins: heath > health

jtdroste:
We should leave any collecting of data defaulted to off.  **Highly** against defaulting this to true (aka on by default).

solderzzc:
@jtdroste  we have a big warning (in red)  and two seconds waiting for exiting. Do you think that's reasonable for that?

solderzzc:
Should I use 'git rebase' or 'git merge' dev to solve conflict?

solderzzc:
Shouldn't use git rebase.

Issue #1290, opened 2016-07-27
TypeError: argument of type 'NoneType' is not iterable


james26s:
About an hour into the bot running, I get this error.  Any ideas?

```
File "pokecli.py", line 221, in <module>
    main()
  File "pokecli.py", line 212, in main
    bot.take_step()
  File "/Users/user/Downloads/PokemonGo-Bot/pokemongo_bot/__init__.py", line 36, in take_step
    self.stepper.take_step()
  File "/Users/user/Downloads/PokemonGo-Bot/pokemongo_bot/stepper.py", line 62, in take_step
    self._work_at_position(position[0], position[1], position[2], True)
  File "/Users/user/Downloads/PokemonGo-Bot/pokemongo_bot/stepper.py", line 139, in _work_at_position
    self.bot.work_on_cell(cell, position, pokemon_only)
  File "/Users/user/Downloads/PokemonGo-Bot/pokemongo_bot/__init__.py", line 96, in work_on_cell
    hack_chain = worker.work()
  File "/Users/user/Downloads/PokemonGo-Bot/pokemongo_bot/cell_workers/seen_fort_worker.py", line 31, in work
    if 'responses' in response_dict \
TypeError: argument of type 'NoneType' is not iterable
```

kennycoc:
I created a PR that should fix this. after it's accepted run `git pull` and you should be good.

TheGoldenXY:
Please upgrade to the latest `dev` branch by running this command:
```
git reset --hard
git pull -f
```

Then run this command:
`pip install --upgrade -r requirements.txt`

If issue persists, open a new issue.

/closed

PR #1304, opened 2016-07-28
Added None Check to API Response

kennycoc:
Short Description: 
We don't check that we actually receive any data from the api when we request the fort details.

I added a check
Fixes:
-  #1290 
- 
-


tstumm:
We don't merge into master

PR #1310, opened 2016-07-28
More randomized PokeStop pattern

Viper896:
Short Description: 

Added a random cool down time of between 5 and 15 minutes when visiting a poke stop.  Helps create more random pokestop pathing.

### DIFF
--- a/configs/config.json.example
+++ b/configs/config.json.example
@@ -5,17 +5,21 @@
     "location": "SOME_LOCATION",
     "gmapkey": "GOOGLE_MAPS_API_KEY",
     "max_steps": 5,
-    "mode": "all",
+    "catch_pokemon": true,
+    "spin_forts": true,
+    "add_fort_cooldown": {
+    "add_cooldown_time": "random",
+    "cd_min_delta": 300,
+    "cd_max_delta": 900
+    },
     "walk": 4.16,
     "action_wait_min": 1,
     "action_wait_max": 4,
     "debug": false,
     "test": false,
+    "health_record": true,
     "location_cache": true,
     "distance_unit": "km",
-    "pokestop_cooldown": "300",
-    "cooldown_min": 300,
-    "cooldown_max": 900,
     "reconnecting_timeout": 15,
     "item_filter": {
       "1": { "keep" : 100 },
@@ -29,6 +33,8 @@
     "evolve_speed": 20,
     "cp_min": 300,
     "use_lucky_egg": false,
+    "hatch_eggs": true,
+    "longer_eggs_first": true,
     "evolve_captured": false,
     "release_pokemon": true,
     "catch": {
@@ -39,6 +45,10 @@
     "release": {
       "any": {"release_below_cp": 0, "release_below_iv": 0, "logic": "or"},
       "// Example of always releasing Rattata:": {},
-      "// Rattata": { "always_release" : true }
+      "// Rattata": { "always_release" : true },
+      "// Example of keeping stronger (based on CP) Pidgey:": {},
+      "// Pidgey": { "keep_best_cp" : true },
+      "// Example of keeping stronger (based on IV) Zubat:": {},
+      "// Zubat": { "keep_best_iv" : true }
     }
 }
--- a/pokecli.py
+++ b/pokecli.py
@@ -84,11 +84,16 @@ def init_config():
         default=False
     )
     parser.add_argument(
-        "-m",
-        "--mode",
-        help="Farming Mode",
-        type=str,
-        default="all"
+        "--catch_pokemon",
+        help="Enable catching pokemon",
+        type=bool,
+        default=True
+    )
+    parser.add_argument(
+        "--spin_forts",
+        help="Enable Spinning Pokestops",
+        type=bool,
+        default=True
     )
     parser.add_argument(
         "-w",
@@ -176,6 +181,13 @@ def init_config():
         type=float,
         default=15.0
     )
+    parser.add_argument(
+        "-hr",
+        "--health_record",
+        help="Send anonymous bot event to GA for bot health record. Set \"health_record\":false if you need disable it.",
+        type=bool,
+        default=True
+    )
 
     # Start to parse other attrs
     config = parser.parse_args()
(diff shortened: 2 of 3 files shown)


douglascamata:
@Viper896 make the amount of time to ignore configurable in `config.json` (default should be 5 minutes), please. Also request for someone to update wiki for you when you finish.

Viper896:
@douglascamata Sorry for all the sloppy git merges, still learning.   working code for the user selectable pokestop cooldown timers has been updated.

pokecli.py
```diff
@@ -195,6 +195,21 @@ def init_config():
...
+    if 'cooldown_min' in load:
+        config.cooldown_min = load['cooldown_min']
+    else:
+        config.cooldown_min = 300
+
+    if 'cooldown_max' in load:
+        config.cooldown_max = load['cooldown_max']
+    else:
+        config.cooldown_max = 900
+
```
douglascamata: Please, use this pattern: `load.get('parameter_you_want', 'default_parameter_value')`

pokemongo_bot/cell_workers/seen_fort_worker.py
```diff
@@ -1,7 +1,13 @@
 # -*- coding: utf-8 -*-
 
 import time
+<<<<<<< HEAD
+import json
+import random
+=======
+import random
```
douglascamata: Merge conflict was committed here ^

pokemongo_bot/cell_workers/seen_fort_worker.py
```diff
@@ -73,6 +79,21 @@ def work(self):
...
+                userpokestop_cooldown = (time.time() + int(self.config.pokestop_cooldown)) * 1000
+            else:
+                userpokestop_cooldown = (time.time() + 300) * 1000
+
+=======
+            ran_timer = random.randint(300, 600)
+            self.bot.fort_timeouts[self.fort["id"]] = (time.time() + ran_timer) * 1000 #set fort cooldown for random time between 5 and 15 min.
+            
+>>>>>>> refs/remotes/origin/patch-2
             if spin_result == 1:
```
douglascamata: and here

pokemongo_bot/cell_workers/seen_fort_worker.py
```diff
@@ -97,10 +118,15 @@ def work(self):
...
+<<<<<<< HEAD
+                    
+                pokestop_cooldown = userpokestop_cooldown
+=======
 
-                pokestop_cooldown = spin_details.get(
-                    'cooldown_complete_timestamp_ms')
+                pokestop_cooldown = self.bot.fort_timeouts[self.fort["id"]]
+>>>>>>> refs/remotes/origin/patch-2
                 self.bot.fort_timeouts.update({self.fort["id"]: pokestop_cooldown})
```
douglascamata: and here

pokemongo_bot/cell_workers/seen_fort_worker.py
```diff
@@ -139,7 +164,10 @@ def work(self):
                     'chain_hack_sequence_number']
             else:
                 logger.log('Possibly searching too often - taking a short rest :)', 'yellow')
-                self.bot.fort_timeouts[self.fort["id"]] = (time.time() + 300) * 1000  # Don't spin for 5m
+<<<<<<< HEAD
+                self.bot.fort_timeouts[self.fort["id"]] = pokestop_cooldown
+=======
+>>>>>>> refs/remotes/origin/patch-2
```
douglascamata: and here

Viper896:
Should be good now... working locally and passed trav-ci checks

pokemongo_bot/cell_workers/seen_fort_worker.py
```diff
@@ -73,6 +75,15 @@ def work(self):
 
             spin_details = response_dict['responses']['FORT_SEARCH']
             spin_result = spin_details.get('result', -1)
+            userpokestop_coolwon = 0
```
Surceis: misprint

Viper896: Thanks,  kind of feel stupid for missing that one.

configs/config.json.example
```diff
@@ -13,6 +13,9 @@
     "test": false,
     "location_cache": true,
     "distance_unit": "km",
+    "pokestop_cooldown": "300",
```
douglascamata: Pokestop cooldown is always 300 seconds (5 minutes). Why would we let users customise this? It doesn't make sense.

pokemongo_bot/cell_workers/seen_fort_worker.py
```diff
@@ -97,10 +108,9 @@ def work(self):
                         logger.log('- ' + str(item_count) + "x " + item_name + " (Total: " + str(self.bot.item_inventory_count(item_id)) + ")", 'yellow')
                 else:
                     logger.log("[#] Nothing found.", 'yellow')
-
-                pokestop_cooldown = spin_details.get(
-                    'cooldown_complete_timestamp_ms')
```
douglascamata: You're ignoring the cooldown that the game is giving you, why? Shouldn't you get this cooldown and add a random delta to it?

pokemongo_bot/cell_workers/seen_fort_worker.py
```diff
@@ -119,8 +129,7 @@ def work(self):
             elif spin_result == 2:
                 logger.log("[#] Pokestop out of range")
             elif spin_result == 3:
-                pokestop_cooldown = spin_details.get(
-                    'cooldown_complete_timestamp_ms')
+                pokestop_cooldown = userpokestop_cooldown
```
douglascamata: Same thing as my other comment about ignoring pokestop cooldown the server is telling you.

configs/config.json.example
Viper896: It does make sense.  The area where I run my bot has Pokestops evenly spaced and close together.  So what ends up happening is my just goes around in a circle, I end up with a ton of items and very few pokemon.  By setting the timeout to be longer, it forces the bot to find a fort that is not in timeout.

__init__.py: line 190: forts = filter(lambda x: x["id"] not in self.fort_timeouts, forts)

This allows the user to create a larger search area as the bot will now have to find forts farther away because it thinks all the ones close are in timeout.

douglascamata: @Viper896 you already have better parameters down here to do this: 

```
 "cooldown_min": 300,
 "cooldown_max": 900,
```

It should be clear in code that default cooldown is 5 minutes and we add a random delta because we want. 

I would even rename and simplify these keys to:

```
 "pokestop_cd_delta_min": 300,
 "pokestop_cd_delta_max": 600,
```

This would mean: for every pokestop spun sum randint(300, 600) to the server-given cooldown.

samliu:
Hi there, FWIW if you're trying to emulate natural movement you shouldn't use uniform random, you should sample times from an exponential distribution

https://en.wikipedia.org/wiki/Poisson_point_process

But this is a good start

PR #1334, opened 2016-07-28
Replacing config.mode with capture_pokemon and spin_forts

TheSavior:
This also changes some logic. Previously, when you ran out of pokeballs, we would start farming pokestops and never catch anymore pokemon again. Then once we had a full bag, the program would crash. That doesn't happen anymore.

### DIFF
--- a/pokecli.py
+++ b/pokecli.py
@@ -84,11 +84,16 @@ def init_config():
         default=False
     )
     parser.add_argument(
-        "-m",
-        "--mode",
-        help="Farming Mode",
-        type=str,
-        default="all"
+        "--catch_pokemon",
+        help="Enable catching pokemon",
+        type=bool,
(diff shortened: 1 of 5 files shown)


solderzzc:
👍

Nihisil:
What will hapen if `self.config.catch_pokemon = True` and `items_stock[pokeball] is 0`? Maybe we should handle it

TheSavior:
@Nihisil, this shouldn't change that at all. The catch_pokemon_worker should have a check internally for whether there are even any items for it to try to catch pokemon. The fort workers do that currently. We should make sure the pokemon catch workers do that after this PR

Nihisil:
@TheSavior yeah, makes sense

eggins:
Merge failed
:+1:

PR #1369, opened 2016-07-28
fixed transfer worker api crash

surfaace:
try catch exception to prevent bot from crashing


douglascamata:
@surfaace please avoid using `try...catch` for dict like this. Use the `.get` method.

douglascamata:
👍

tstumm:
👍

PR #1403, opened 2016-07-28
Value in the config files were not correctly set

DayBr3ak:
Short Description: 

every value matching a False ie (0, None, or False) were ignored. We only want the None value to be matched


surfaace:
:+1:

douglascamata:
:-1:

douglascamata:
I'm fixing the whole config load, it was messy very error prone.

DayBr3ak:
I close this pr then ?

PR #1421, opened 2016-07-28
Bulk processing.

alxnegrila:
Short Description: 
Check that bot(s) are frequently running.
Add new users easy.
Restart running bots easy.


solderzzc:
We shouldn't encourage user to start multiple instance.
👎

TheSavior:
I agree with that and think we also don't want to switch to a user's fork of the web module. 

👎

PR #1459, opened 2016-07-28
[FEATURE] Api Wrapper to handle connection issues

DayBr3ak:
Short Description: 

I'm trying to handle connections issues with this wrapper.
I've been testing it for an hour for now. Maybe you want to review/test this yourself.
For now it just does some retries (5) on the request if it failed.
And throw an error if it fails instead of an error like `__getitem__ on bool ` ... etc

I'll update this when I've done enough testing

### DIFF
--- a/pokemongo_bot/api_wrapper.py
+++ b/pokemongo_bot/api_wrapper.py
@@ -1,0 +1,58 @@
+# api_wrapper.py
+
+from pgoapi import PGoApi
+from pgoapi.exceptions import NotLoggedInException
+import logger
+
+class ApiWrapper(object):
+    def __init__(self, api):
+        self._api = api
+        self.reset_auth()
+
+    def reset_auth(self):
+        self._api._auth_token = None
+        self._api._auth_provider = None
+        self._api._api_endpoint = None
+
+    # wrap api methods
+    def _can_call(self):
+        if not self._api._req_method_list:
+            raise RuntimeError('Trying to call the api without setting any request')
+        if self._api._auth_provider is None or not self._api._auth_provider.is_login():
+            logger.log('Not logged in!', 'red')
+            raise NotLoggedInException()
+        return True
+
+    def call(self):
+        if not self._can_call():
+            return False
+
+        api_req_method_list = self._api._req_method_list
+        result = None
+        max_retry = 5 # TODO configuration param
+        try_cnt = 0
+        while True:
+            self._api._req_method_list = api_req_method_list # api internally clear this field after a call
+            result = self._api.call()
+            if result is None:
+                try_cnt += 1
+                logger.log('Server seems to be busy or offline - try again - {}/{}'.format(try_cnt, max_retry), 'red')
+                if try_cnt >= max_retry:
+                    raise ServerBusyOrOfflineException()
+                sleep(10)
+            else:
+                break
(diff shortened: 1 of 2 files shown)


pokemongo_bot/api_wrapper.py
```diff
@@ -0,0 +1,55 @@
...
+        try_cnt = 0
+        while True:
+            self._api._req_method_list = [req_method for req_method in api_req_method_list] # api internally clear this field after a call
+            result = self._api.call()
+            if result is None:
+                try_cnt += 1
+                logger.log('Server seems to be busy or offline - try again - {}/{}'.format(try_cnt, max_retry), 'red')
+                if try_cnt >= max_retry:
+                    raise ServerBusyOrOfflineException()
+                sleep(10)
```
cwild: Where are you importing sleep from?

DayBr3ak: good catch

pokemongo_bot/api_wrapper.py
```diff
@@ -0,0 +1,55 @@
...
+        if not self._can_call():
+            return False
+
+        api_req_method_list = self._api._req_method_list
+        result = None
+        max_retry = 5 # TODO configuration param
+        try_cnt = 0
+        while True:
+            self._api._req_method_list = [req_method for req_method in api_req_method_list] # api internally clear this field after a call
+            result = self._api.call()
```
cwild: Worth wrapping in a ```try...except``` block perhaps? Exceptions can be raised from here still.

DayBr3ak: What else can be raised here if I may ask?

pokemongo_bot/api_wrapper.py
douglascamata: holy shit, please do not wait 10 seconds... 1 second is good enough.

DayBr3ak: done

pokemongo_bot/api_wrapper.py
cwild: Well I am not familiar with the underlying API. I'd say whenever you call a function, there's a risk of an Exception being raised (maybe not even in the current API release). So it's a matter of deciding if you want to harden your API wrapper to handle those errors or not.

Any of the core maintainers care to share their opinions?

DayBr3ak: You're right, api.call use request.request so I could check for those exceptions

douglascamata:
:+1:

TheSavior:
I wonder if this kind of retry logic should be moved upstream. @tejado, do you have any thoughts?

TheSavior:
👍 

[Approved with PullApprove](https://pullapprove.com/PokemonGoF/PokemonGo-Bot/pull-request/1459/?utm_source=github-pr&utm_medium=comment-badge&utm_campaign=PokemonGoF/PokemonGo-Bot)

tejado:
Yes, this would be a fantastic idea. Already had similar thoughts. I will implement it also in pgoapi 1.2.0

douglascamata:
While @tejado doesn't release `pgoapi` 1.2.0 we keep this wrapper.

douglascamata:
👍

PR #1460, opened 2016-07-28
[FEATURE] Added logging function on files

BoGnY:
Short Description:

Added the function to write every bot output on .txt files
So, if the bot crashes, it is possible to read what it has previously done until the crash!


DayBr3ak:
Why not using python api `logging` with file handles ?

BoGnY:
I do not know much about Python programming, but the API's documentation of logging not seem to me very suitable for this purpose..
With 3 lines of code I wrote the logger, with logging's API seemed more intricate for the purpose..

Certainly it can be improved by those who have more knowledge than me..

It is a very basic code, but it does something that I consider fundamental, since the bot crashes every 10/15 minutes and I never know what it has done!

pokemongo_bot/logger.py
```diff
@@ -8,17 +9,25 @@
...
         'cyan': '96m'
     }
-    if color not in colorHex:
-        print('[' + time.strftime("%H:%M:%S") + '] '+ string)
-    else:
-        print('[' + time.strftime("%H:%M:%S") + '] ' + u'\033['+ colorHex[color] + string.decode('utf-8') + '\033[0m')
-    if lcd:
-        if(string):
-            lcd.message(string)
+	if color not in colorHex:
```
DayBr3ak: you used tabs instead of spaces

DayBr3ak:
fair enough. But you should at least review your code. python files are commonly used with the `with open() as fd` statement
and your changed indentations with tabulations where it was not needed

cwild:
Also... ```except:``` made a comeback and we only just cleared all of those up.
I'd be more in favour of the logging module.

pokemongo_bot/logger.py
```diff
@@ -8,17 +9,25 @@
     lcd = False
 
 def log(string, color = 'white'):
-    colorHex = {
+	# Write bot output also in a log file
+	try:
+		logs_file = open('logs/' + time.strftime("%Y%m%d") + '.txt', 'a')
+		logs_file.write('[' + time.strftime("%H:%M:%S") + '] ' + string + '\n')
+		logs_file.close()
+	except:
```
douglascamata: Don't do this. Let exceptions be raised if any error happens on logging to file, please.

douglascamata:
We're merging today an event system that will be used for logging and sending messages through web socket server. Then we will drop this custom log function in favour of the `logging` module, because being able to properly have different log levels is much more important than having colours.

BoGnY:
Indentations are changed because previously indentations was made with space, and not tab.. So python crash for wrong indentations..
I commits your suggestions..
However, if in the next days you will implement new logging system, is great! ;)

pokemongo_bot/logger.py
BoGnY: Added raise on exception ;)

logs/20160728.txt
```diff
@@ -0,0 +1,2 @@
+[19:07:38] PokemonGO Bot v1.0
+[19:07:38] Testing log file
```
VirtualSatai: Remove this log file and add all further log files to the .gitignore.

douglascamata:
Did you read my comment? I'm almost finishing on the EventManager for logs and web socket. Entire log system will be rewritten using `logging` module.

BoGnY:
Yes but I had understood that as long as you you don't ended up the new EventManager for logs that could be used.. Then was replaced by your..

However, no problem.. I hold the code in my fork and use mine, until new EventManager will be implemented.

Issue #1456, opened 2016-07-28
Evolving not working


vicch:
### Expected Behavior
Evolve pokemons on startup as configured.

### Actual Behavior
Not evolving at all. Wasted a lucky egg.

### Steps to Reproduce
```
"evolve_all": "Caterpie,Weedle,Pidgey,Rattata,Zubat,Eevee",
"evolve_cp_min": 10,
"evolve_speed": 15,
```

### Other Information
OS: Windows 7
Git Commit: 0a9444bd0da806e2a765af43b4de874146d5faf2

aTadTasty:
it's prob because of server lagging right now. can't even log in for me.

http://downdetector.com/status/pokemon-go

vicch:
@aTadTasty Login is fine for me. The bot just completely omits evolving step and goes on spinning and catching.

sangwon04:
@vicch honestly, if you're going to risk losing an egg you might as well do the evolving yourself.

aTadTasty:
@vicch have you tried updating the requirements after the pull? `sudo pip install -r requirements.txt`

vicch:
@aTadTasty Yes, `git pull` and reinstalled the requirements several times, no luck.

aTadTasty:
duplicate of https://github.com/PokemonGoF/PokemonGo-Bot/issues/1335 and https://github.com/PokemonGoF/PokemonGo-Bot/issues/1339 i guess we just gotta wait for a new resolve

vicch:
@sangwon04 NSFW

vicch:
Closed as duplicate of #1335 and #1339.

Vvkmnn:
Hey, random question. How did you close your own issue?

I can't get it to work for some reason.

sangwon04:
@vicch very understandable. Perhaps hold off on using the egg/evolve till after work then hah.

vicch:
@Vvkmnn There's a button on the left of "Comment".

Vvkmnn:
@vicch ...Wow, I'm having a slow day.

Thanks.

Issue #1339, opened 2016-07-28
Not evolving at start


Steefj3:
### Expected Behavior
Bot should evolve pokemon at start

### Actual Behavior
It used to do this, but doesnt anymore. Only (not at start) one if evolve_captured is set. I have set evolve_all with specific pokemon.

### Other Information
Latest dev branch
OS: 
Git Commit: (run 'git log -n 1 --pretty=format:"%H"' and paste it here)

JulioMansilla:
Please refer to Issue #1335 .

TheGoldenXY:
This issue has already been fixed.

Please upgrade to the latest `dev` branch by running this command:
```
git reset --hard
git pull -f
```

Then run this command:
`pip install --upgrade -r requirements.txt`

If issue persists, open a new issue.

/closed

Issue #1335, opened 2016-07-28
Lucky eggs are not being used, evolve doesn't work. 


JulioMansilla:
### Expected Behavior

"use_lucky_egg": true, will use lucky egg
"evolve_all": "all", will evolve all pokemons
### Actual Behavior

Lucky egg is not sued
Evolution process doesn't start
### Steps to Reproduce

set config file to:
"use_lucky_egg": true,
"evolve_all": "all",

load bot, no egg will be used, no pokemons will be evolved.
### Other Information
OS: Mac
Git Commit: 4c46ad7e6b41f4d66400aee82be469cdbbb26f80

leadboots5:
I have the same experience. I will also note that "evolve_all" doesn't attempt to do any evolving on script start up.
15e0161d32c7a78c230c6e75cfd6867663c985d1

JulioMansilla:
@leadboots5, I have the same issue with evolving too.

leovarmak:
@JulioMansilla Same issue here too. The lucky egg is not being used

beedy01:
evolve doesnt run properly, so am i, i got that issue..

JulioMansilla:
I've edited the issue to reflect the evolve issue too.

Vvkmnn:
I've also got this issue on e25a7091f5f26eef17bc38f6f90df26ad199ffbc but there was a PR to possibly solve it: #1256

It's still waiting on some testing before it gets merged.

TheGoldenXY:
This issue has already been fixed.

Please upgrade to the latest `dev` branch by running this command:
```
git reset --hard
git pull -f
```

Then run this command:
`pip install --upgrade -r requirements.txt`

If issue persists, open a new issue.

/closed

PR #1469, opened 2016-07-28
Fixing evolve worker being skipped problem

vicch:
Short Description: 
`EvolveAllWorker._should_run()` logic was wrong, causing evolve worker not running at startup.

Fixes:
- Fixed `EvolveAllWorker._should_run()`

### DIFF
--- a/pokemongo_bot/cell_workers/evolve_all_worker.py
+++ b/pokemongo_bot/cell_workers/evolve_all_worker.py
@@ -53,17 +53,19 @@ class EvolveAllWorker(object):
 
     def _should_run(self):
         # Will skip evolving if user wants to use an egg and there is none
-        skip_evolves = False
 
-        if self.config.evolve_all:
-            return skip_evolves
+        if not self.config.evolve_all:
+            return False
 
         # Pop lucky egg before evolving to maximize xp gain
         use_lucky_egg = self.config.use_lucky_egg
         lucky_egg_count = self.bot.item_inventory_count(Item.ITEM_LUCKY_EGG.value)
 
-        if use_lucky_egg and lucky_egg_count > 0:
-            logger.log('Using lucky egg ... you have {}'
+        if not use_lucky_egg: # If not using lucky eggs, skipping checks below
+            return True
+
+        if lucky_egg_count > 0:
+            logger.log('Using lucky egg... you have {}'
                        .format(lucky_egg_count))
             response_dict_lucky_egg = self.bot.use_lucky_egg()
             if response_dict_lucky_egg and 'responses' in response_dict_lucky_egg and \
@@ -75,13 +77,12 @@ class EvolveAllWorker(object):
                                .format(lucky_egg_count-1), 'green')
                 else:
                     logger.log('Failed to use lucky egg!', 'red')
-                    skip_evolves = True
-        elif use_lucky_egg: #lucky_egg_count is 0
-            # Skipping evolve so they aren't wasted
-            logger.log('No lucky eggs... skipping evolve!', 'yellow')
-            skip_evolves = True
+                    return False
 
-        return skip_evolves
+        else: # lucky_egg_count is 0
+            # Skipping evolve so pokemon aren't wasted
+            logger.log('No lucky eggs... skipping evolve!', 'yellow')
+            return False
 
     def _release_evolved(self, release_cand_list_ids):
         response_dict = self.bot.get_inventory()


vicch:
Should be able to solve #1335 , #1339 , #1456

aradhell:
not waiting for evolve_speed ?

vicch:
There is a `sleep()`, but the last time I tried, it didn't seem to be working. No idea. This could be a whole other issue.
https://github.com/vicch/PokemonGo-Bot/blob/b13da71f428cf64904f0a6ee4d250387ee5b9063/pokemongo_bot/cell_workers/evolve_all_worker.py#L165

douglascamata:
tests, please.

stefansaraev:
works fine for me.

thomaswelton:
Doesn't work. When set to use a luck egg the lucky egg is applied but the evloves are skipped.
I had to update the config so lucky egg was false then restart so it evolved.
Also there is no code that checks to see if a luck egg is already applied. If the bot is restarted and a luck egg is active then the evolves should run.

thomaswelton:
That being said thanks for the pull request, it allowed me to evolve all the the 200 pidgeys I had collected previously for and allowed me to level up

vicch:
@thomaswelton 200! That's pretty aggressive. You are welcome. Batch evolving is indeed vital for quick leveling. Hope this could get completely fixed soon.

pokemongo_bot/cell_workers/evolve_all_worker.py
```diff
@@ -53,17 +53,19 @@ def work(self):
...
         # Pop lucky egg before evolving to maximize xp gain
         use_lucky_egg = self.config.use_lucky_egg
         lucky_egg_count = self.bot.item_inventory_count(Item.ITEM_LUCKY_EGG.value)
 
-        if use_lucky_egg and lucky_egg_count > 0:
-            logger.log('Using lucky egg ... you have {}'
+        if not use_lucky_egg: # If not using lucky eggs, skipping checks below
+            return True
+
+        if lucky_egg_count > 0:
```
codybaldwin: This will attempt to use a lucky egg every tick.

pokemongo_bot/cell_workers/evolve_all_worker.py
```diff
@@ -75,13 +77,12 @@ def _should_run(self):
...
-        elif use_lucky_egg: #lucky_egg_count is 0
-            # Skipping evolve so they aren't wasted
-            logger.log('No lucky eggs... skipping evolve!', 'yellow')
-            skip_evolves = True
+                    return False
 
-        return skip_evolves
+        else: # lucky_egg_count is 0
+            # Skipping evolve so pokemon aren't wasted
+            logger.log('No lucky eggs... skipping evolve!', 'yellow')
```
codybaldwin: This message will be displayed every tick if the user has no lucky eggs

vicch: Even when `"use_lucky_egg": false` in config?

codybaldwin: updated original comment

pokemongo_bot/cell_workers/evolve_all_worker.py
vicch: I suppose yes, if you set `"use_lucky_egg": true` and you actually have lucky eggs.

pokemongo_bot/cell_workers/evolve_all_worker.py
vicch: OK. Isn't that what it is supposed to do? The idea is to skip evolving if you want to use lucky eggs but don't have any, otherwise the Pidgeys you stacked will be wasted.

codybaldwin: I'm talking about the fact that this message will be displayed every second or so (technically every tick)

pokemongo_bot/cell_workers/evolve_all_worker.py
codybaldwin: so every 30 mins it will use a lucky egg.  This doesn't seem ideal.  Not only that, currently it will send the API request every second (technically tick) which is not good at all.

pokemongo_bot/cell_workers/evolve_all_worker.py
vicch: I see. Well I'm afraid this fix is not meant to cover that. My intention was only to make batch evolving work again. You can set `"use_lucky_egg": false` to avoid the messages I guess.

vicch:
Already fixed and merged in #1541

PR #1475, opened 2016-07-28
typo

fgaz:


douglascamata:
master is in frozen state and will soon be updated with dev's code, we do not care about this, sorry.

PR #1478, opened 2016-07-28
Auto-Login every 1 minute after get disconnected

muhammetkucuk:
Hello, it's quite possible you can get disconnected in any time.
Here is a shell script for solution.
For running in Linux: 

sudo chmod 755 start.sh
./start.sh

That's all. You should kept this script in the main directory of your bot.

### DIFF
--- a/start.sh
+++ b/start.sh
@@ -1,0 +1,49 @@
+#!/bin/bash
+while (true)
+do
+virtualenv .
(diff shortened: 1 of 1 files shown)


douglascamata:
@muhammetkucuk I'm sorry but this is not the correct solution, we need something much more elegant than this. Closing.

PR #1481, opened 2016-07-28
Fix randomly keyerror and try again

majorvin:
Time to time the server returns success empty response and the script times out.


douglascamata:
This is not the proper fix. Something much better has a PR open and is under tests.

Issue #1466, opened 2016-07-28
[DEV] invalid <lambda> value:


anovosiolov:
./run.sh
[21:28:46] PokemonGO Bot v1.0
usage: pokecli.py [-h] [-cf CONFIG] [-a AUTH_SERVICE] [-u USERNAME]
                  [-p PASSWORD] [-l LOCATION] [-lc LOCATION_CACHE]
                  [--catch_pokemon CATCH_POKEMON] [--spin_forts SPIN_FORTS]
                  [-w WALK] [-k GMAPKEY] [-ms MAX_STEPS] [-rp RELEASE_POKEMON]
                  [-d DEBUG] [-t TEST] [-du DISTANCE_UNIT] [-ev EVOLVE_ALL]
                  [-ecm EVOLVE_CP_MIN] [-ec EVOLVE_CAPTURED]
                  [-le USE_LUCKY_EGG] [-rt RECONNECTING_TIMEOUT]
                  [-hr HEALTH_RECORD]

> `pokecli.py: error: argument -l/--location: invalid <lambda> value: u'\u05d0\u05e7\u05e1\u05d5\u05d3\u05d5\u05e1 \u05d0\u05e9\u05d3\u05d5\u05d3'
`

### Steps to Reproduce
Unicode location in config:
like:  
"location": "אקסודוס אשדוד",

Worked in commit: 4c46ad7e6b41f4d66400aee82be469cdbbb26f80
### Other Information
OS: Ubuntu 14.04
Git Commit: 34caa002a464b7665cdcfb531ea3d302632898ac

th3w4y:
try below workaround



"location": "\\u05d0\\u05e7\\u05e1\\u05d5\\u05d3\\u05d5\\u05e1 \\u05d0\\u05e9\\u05d3\\u05d5\\u05d3"


In [1]: import json

In [2]: json.dumps({'location':"אקסודוס אשדוד"})
Out[2]: '{"location": **"\\u05d0\\u05e7\\u05e1\\u05d5\\u05d3\\u05d5\\u05e1 \\u05d0\\u05e9\\u05d3\\u05d5\\u05d3"**}'

In [3]:

anovosiolov:
So to put "location": "\u05d0\u05e7\u05e1\u05d5\u05d3\u05d5\u05e1 \u05d0\u05e9\u05d3\u05d5\u05d3" in config file ? 
Gives me the same result:
[...]

PR #1503, opened 2016-07-28
Use type unicode for argument location

vieira:
Short Description: 
Fix for issue with invalid <lambda> value for location argument, e.g.: invalid <lambda> value: u'Pra\xe7a' when it contains special characters like "ç".

Fixes:
- issue with invalid <lambda> value for location argument with special chars, e.g. "ç"

### DIFF
--- a/pokecli.py
+++ b/pokecli.py
@@ -151,7 +151,7 @@ def init_config():
         short_flag="-l",
         long_flag="--location",
         help="Location",
-        type=unicode,
+        type=parse_str,
         default=''
     )
     add_config(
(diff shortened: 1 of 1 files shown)


douglascamata:
You didn't test using both command line and JSON location.

vieira:
@douglascamata Thanks for spotting! Should be working with both command line and JSON file now.

vieira:
Should fix issue #1466.

douglascamata:
@vieira my code does the same thing as yours...

douglascamata:
Please rename the function to something more clear: `parse_utf8_str` or `parse_unicode_str`.... dunno which one fits best.

vieira:
The problem is easy to reproduce, just set initial location to e.g., "Praça do Comércio, Lisboa" and you will see the difference.

vieira:
@douglascamata Ok, will do.

TheSavior:
👍

PR #1508, opened 2016-07-28
[FEATURE] Releasing soft ban after jumping between long distances

weknin:
Please delete, became a mess. reopening.


solderzzc:
You should use dev branch as your code base.

Issue #1504, opened 2016-07-28
Server seems to be busy or offline - try again and again


thehawkes:
dev updated ~5 mins ago.
every 2. output is 
`Server seems to be busy or offline`
```
[00:28:43] Server seems to be busy or offline - try again - 1/5                              [
[00:28:46] [x] Found fort 555832c91442475e96bc886c10e3e9f1.16 at distance 0.07km              
[00:28:46] [x] Need to move closer to Pokestop                                                
[00:28:50] [x] Found fort 555832c91442475e96bc886c10e3e9f1.16 at distance 0.07km              
[00:28:50] [x] Need to move closer to Pokestop                                                
[00:28:51] Server seems to be busy or offline - try again - 1/5                               
[00:28:55] [x] Found fort 555832c91442475e96bc886c10e3e9f1.16 at distance 0.06km              
[00:28:55] [x] Need to move closer to Pokestop                                                
[00:28:56] Server seems to be busy or offline - try again - 1/5                               
[00:29:00] [x] Found fort 555832c91442475e96bc886c10e3e9f1.16 at distance 0.06km              
[00:29:00] [x] Need to move closer to Pokestop                                                
[00:29:02] Server seems to be busy or offline - try again - 1/5                               
[00:29:05] [x] Found fort 555832c91442475e96bc886c10e3e9f1.16 at distance 0.05km              
[00:29:05] [x] Need to move closer to Pokestop                                                
[00:29:07] Server seems to be busy or offline - try again - 1/5                               
[00:29:10] [x] Found fort 555832c91442475e96bc886c10e3e9f1.16 at distance 0.05km              
[00:29:10] [x] Need to move closer to Pokestop                                                
[00:29:12] Server seems to be busy or offline - try again - 1/5                               
[00:29:16] Something rustles nearby!                                                          
[00:29:16] A Wild Drowzee appeared! [CP 386] [Potential 0.49]                                 
[00:29:16] IV [Stamina/Attack/Defense] = [6/10/6]                                             
[00:29:19] Catch Rate with normal Pokeball is low (40.01%). Throwing Razz Berry... (1 left!)  
[00:29:19] Catch Rate with normal Pokeball has increased to 60.02%                            
[00:29:19] Using Pokeball (chance: 60.02%)... (77 left!)                                      
[00:29:19] [-] Attempted to capture Drowzee - failed.. trying again!                          
[...]

PR #1510, opened 2016-07-28
suppress the first "unimportant" api messages

thehawkes:
Short Description: 

suppress the first "unimportant" api messages "Server seems to be busy or offline"

Fixes:
- #1504


DayBr3ak:
Seems good for me

TheSavior:
Throttling instead.

Issue #1187, opened 2016-07-27
[REQUEST] Softban lift


kotecekk:
So, as some of us know, there is an easy way to lift softban:

walk to a pokestop, spin it and close it. You won’t be getting any rewards from it, however, doing so up to 40 times (or until you finally receive the rewards) will remove your soft ban.

This feature should be implemented.

svista:
Great

DimaVIII:
This is not true! It the soft ban is depending on your "jump" distance and can remain from 5 min to 4 hours. (Experts in bans, got a lot this and learn how to jump without getting a softban, but I will not post it public.)
The solution for the bot will be better like this:
If you don't get a rewards from pokestops (at least 2 pokestops need to be checked) logout for 5 min and try again.

weknin:
I implemented what @svista said. Yet, didn't test if it works.

PP19:
Dont implement that feature, it wont remove your softban, spinning 40 times just takes long enough to let it expire.

kotecekk:
Spinnig 40 times will remove your soft ban.

Whosdaman:
I just tested this myself and it is 100% accurate. Softban was completely lifted and activity returned to normal after spinning and exiting a Pokestop around 40 times.

PP19:
This "feature" sounds like "Autoban"

IQNeoXen:
Works perfectly well for me. Should be implemented imo.

Whosdaman:
It is exactly 40 spins on the same pokestop. On the 40th spin, the pokestop will give you the items. We need to have this implemented asap in my opinion.

VictorChen:
You know what else removes the soft ban? Just waiting a few minutes and restarting the bot...

wandering7man:
Dont work if u across 5000km in 1 sec.

MikeDX:
Is this legit? I feel like I need to test it. Would be great for when I want to teleport to the states to catch that elusive pokemon that's just appeared...

IQNeoXen:
@wandering7man it actually does... just try it out ^^

solderzzc:
Wow, is that real?  CC to @douglascamata @TheSavior

MikeDX:
There is definitely something to this.

I had a bot running in new york, and decided to move it to australia after I caught two Tauro so that I could capture something new and try this out.

[...]

Issue #1321, opened 2016-07-28
[Suggestion] If you do this can release from soft ban


yidongw:
If you spin the pokestop for 40 times, you are good to go.

I already confirmed. It will be a great feature to add.

Also a friend told me if you throw 40 pokeballs, you get released too. I haven't tested yet. I'm not sure its throw and missing pokemon for 40 times or I have to hit pokemon 40 times.

yidongw:
Even for the same pokestop

aitzol:
I have tested and it works. Dirty patch attached
[patch.diff.txt](https://github.com/PokemonGoF/PokemonGo-Bot/files/388225/patch.diff.txt)

PR #1513, opened 2016-07-28
[FEATURE] Releasing account from soft ban

weknin:
Short Description: 
After jumping between distances, account will be soft-banned.
This feature will remove the ban.

Fixes:
Issue #1187, #1321

### DIFF
--- a/CONTRIBUTORS.md
+++ b/CONTRIBUTORS.md
@@ -40,3 +40,4 @@
  * surfaace
  * surceis
  * SpaceWhale
+ * Weknin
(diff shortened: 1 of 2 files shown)


krook1024:
This'd be so useful!

DayBr3ak:
Shouldn't you add some sleep feature to not spam the server ? Or did I miss it ? If that's the case my bad. 
Seems like a great feature for the travelers :)

douglascamata:
Has conflicts with dev and have a much more easy fix: just remove the RuntimeError being raised in line 118 of SeenFortWorker (original file from dev).

douglascamata:
👎

douglascamata:
#1522 has much better implementation!

Issue #1245, opened 2016-07-27
Walk in circle


DavidTai1:
### Expected Behavior
walk around

### Actual Behavior
walk in circle

### Steps to Reproduce
starting position: -35.289139, 149.132514

I have kept a file for recording the pokstops that it visited(in pic).
![1](https://cloud.githubusercontent.com/assets/12397480/17188143/e21ee92a-5409-11e6-8bc6-58cc59b478c5.png)
![b8de7a5a-9674-4273-bbae-2c55e5e18cf3](https://cloud.githubusercontent.com/assets/12397480/17188150/e766ce5c-5409-11e6-814a-37f8c614cbac.png)

### Other Information
OS: MAC OS X 10.11.5
Git Commit: I download zip file to replace the old files and remove old pyc. The file downloaded is from commit 3a18aa57dfb67e2e09c207087cad6e5aa1867112

DavidTai1:
Time was: [14:30:04][14:31:30][14:34:05][14:37:24]  / [14:39:49][14:41:06] ect....

DavidTai1:
I think it is because the 5min has past after spin the first stop, and at that time the first stop is also the closest stop to spin. I modified spin cooldown to 10min and it walks to other stops. 
Suggestions: Maybe keep a routing table for pokstops within a limited area?

binarydepartment:
Can confirm:
![circle](http://pix.toile-libre.org/upload/original/1469647181.png)

DavidTai1:
@binarydepartment You can try to change cooldown time for spinning. Mine is about 10min ([14:39:49] - [14:30:04])and it works pretty well now but this is not a good solution might fail in your case.

DavidTai1:
@Philipp59  dev 3a18aa57dfb67e2e09c207087cad6e5aa1867112
max_steps: 5
mode:all

T3hKonijn:
same bug for me
http://i.imgur.com/KbqFkJz.png
mode: all
walk 4.16
max_steps 5
branch dev
commit 15a034427d805e0bacb79fc0a1f3c96a6c93f0e3

javiercravioto:
Where do you modify spin cooldown?

DavidTai1:
@javiercravioto .../pokemongo_bot/cell_workers/seen_fort_worker.py    line 120, change 300s to the seconds you want. remember to remove file seen_fort_worker.pyc

javiercravioto:
[...]

PR #1515, opened 2016-07-28
Added circle avoidance

MFizz:
Short Description: 
Avoides circles by cashing the last X (specified in config) spun pokestops and filtering them. 
Fixes:
#1245

### DIFF
--- a/configs/config.json.example
+++ b/configs/config.json.example
@@ -32,6 +32,8 @@
     "longer_eggs_first": true,
     "evolve_captured": false,
     "release_pokemon": true,
+    "avoid_circles": false,
+    "max_circle_size": 10,
     "catch": {
       "any": {"catch_above_cp": 0, "catch_above_iv": 0, "logic": "or"},
       "// Example of always catching Rattata:": {},
--- a/configs/config.json.example
+++ b/configs/config.json.example
@@ -32,8 +32,10 @@
     "longer_eggs_first": true,
     "evolve_captured": false,
     "release_pokemon": true,
-    "avoid_circles": false,
-    "max_circle_size": 10,
+    "spin_forts": {
+      "avoid_circles": false,
+      "max_circle_size": 10
+    },
     "catch": {
       "any": {"catch_above_cp": 0, "catch_above_iv": 0, "logic": "or"},
       "// Example of always catching Rattata:": {},
--- a/pokecli.py
+++ b/pokecli.py
@@ -315,6 +315,9 @@ def init_config():
     config.hatch_eggs = load.get("hatch_eggs", True)
     config.longer_eggs_first = load.get("longer_eggs_first", True)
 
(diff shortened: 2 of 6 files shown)


pokecli.py
```diff
@@ -315,6 +315,9 @@ def init_config():
     config.hatch_eggs = load.get("hatch_eggs", True)
     config.longer_eggs_first = load.get("longer_eggs_first", True)
 
+    config.avoid_circles = load.get("avoid_circles", False)
+    config.max_circle_size = load.get("max_circle_size", 20)
+
```
douglascamata: @MFizz this is not the default we add configuration. Check `add_config` calls.

configs/config.json.example
```diff
@@ -32,6 +32,8 @@
     "longer_eggs_first": true,
     "evolve_captured": false,
     "release_pokemon": true,
+    "avoid_circles": false,
+    "max_circle_size": 10,
```
douglascamata: What you think about moving these parameter to the `spin_forts` key? They seem out of context here.

```
'spin_forts': {
  'avoid_circles': false,
  max_circle_size: 10
}
```

Makes much more sense for me!

pokemongo_bot/__init__.py
```diff
@@ -38,6 +38,7 @@ def __init__(self, config):
         self.metrics = Metrics(self)
         self.latest_inventory = None
         self.cell = None
+        self.path = [None] * config.max_circle_size
```
douglascamata: `self.path` should be renamed to something like `self.last_visited_forts`

pokecli.py
MFizz: Changed to add_config

configs/config.json.example
MFizz: I like the additional context. What do you think about changing forts to pokestop in the configs to avoid confusion?
`'spin_pokestops': {
  'avoid_circles': false,
  max_circle_size: 10
}`

pokemongo_bot/__init__.py
MFizz: Changed it to `recent_forts`, or do we need it more explicit.

douglascamata:
:+1:

TheSavior:
:+1:

solderzzc:
👍

solderzzc:
This is super good ~

configs/config.json.example
```diff
@@ -32,6 +32,10 @@
     "longer_eggs_first": true,
     "evolve_captured": false,
     "release_pokemon": true,
+    "spin_forts": {
```
vicch: There is already key `spin_forts`, don't they conflict? Shouldn't `avoid_circles` and `max_circle_size` be on the same level as `spin_forts` ? (Sorry if I'm wrong, not a Python man)

Issue #1491, opened 2016-07-28
KeyError: 'GET_INVENTORY'


Lunik:
### Actual Behavior
````
Traceback (most recent call last):
  File "./pokecli.py", line 357, in <module>
    main()
  File "./pokecli.py", line 68, in main
    bot.tick()
  File "/home/lunik/PokemonGo-Bot/pokemongo_bot/__init__.py", line 64, in tick
    if worker(self).work() == WorkerResult.RUNNING:
  File "/home/lunik/PokemonGo-Bot/pokemongo_bot/cell_workers/recycle_items_worker.py", line 14, in work
    item_count_dict = self.bot.item_inventory_count('all')
  File "/home/lunik/PokemonGo-Bot/pokemongo_bot/__init__.py", line 361, in item_inventory_count
    'GET_INVENTORY']['inventory_delta']['inventory_items']
KeyError: 'GET_INVENTORY'
```
### Steps to Reproduce
run `./run.sh` in `dev` branch

### Other Information
OS: Ubuntu
Git Commit: c11af09b64df4be77257207dfd147649489c776e

banksymate:
I'm getting the same.
`Traceback (most recent call last):
  File "pokecli.py", line 357, in <module>
    main()
  File "pokecli.py", line 68, in main
    bot.tick()
  File "/root/pgo/dev/PokemonGo-Bot/pokemongo_bot/__init__.py", line 64, in tick
    if worker(self).work() == WorkerResult.RUNNING:
  File "/root/pgo/dev/PokemonGo-Bot/pokemongo_bot/cell_workers/recycle_items_worker.py", line 14, in work
    item_count_dict = self.bot.item_inventory_count('all')
  File "/root/pgo/dev/PokemonGo-Bot/pokemongo_bot/__init__.py", line 361, in item_inventory_count
    'GET_INVENTORY']['inventory_delta']['inventory_items']
KeyError: 'GET_INVENTORY'
`

d0t1q:
I can confirm and replicate this issue

./pokecli.py -cf conf.json 
[21:24:04] PokemonGO Bot v1.0
[21:24:04] Configuration initialized
[21:24:04] [x] Coordinates found in passed in location, not geocoding.
[21:24:04]
[21:24:04] Location Found: **
[21:24:04] GeoPosition: (**)
[21:24:04]
[21:24:04] Attempting login to Pokemon Go.
[21:24:05] Login to Pokemon Go successful.
[...]

PR #1519, opened 2016-07-29
Added retries for get_inventory if no GET_INVENTORY in response.

Adastras:
Short Description: Performs an exponential backoff retry on api.get_inventory() if response['responses']['GET_INVENTORY'] doesn't exist. Should clear up most KeyError: 'GET_INVENTORY' issues.

Also fixed pokemon_transfer_worker.py to use the bot.get_inventory() (thus using this code) instead of a direct api get_inventory().

Fixes: #1491
- 
- 
-

### DIFF
--- a/pokemongo_bot/__init__.py
+++ b/pokemongo_bot/__init__.py
@@ -302,8 +302,16 @@ class PokemonGoBot(object):
 
     def get_inventory(self):
         if self.latest_inventory is None:
-            self.api.get_inventory()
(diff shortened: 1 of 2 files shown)


DayBr3ak:
Actually #1459 kinda do that but for all api calls, not just inventory

solderzzc:
Added while inside main while logic will cause reading issue.
Thanks for PR anyway.

PR #1253, opened 2016-07-27
Try to split player service (To review)

LeaklessGfy:
Short Description: 
Try to make services inside this mess (first incrementation)

PR #1522, opened 2016-07-29
Added softban unban feature

MikeDX:
Short Description: 

Upon detection of a softban, bot will stop wasting balls and berries on pokemon and travel to a pokestop to try to unban itself.

Fixes:

Softban from travelling too far / quickly
Some small grammar changes (seem -> seems)

### DIFF
--- a/pokemongo_bot/__init__.py
+++ b/pokemongo_bot/__init__.py
@@ -14,7 +14,7 @@ from pgoapi import PGoApi
 from pgoapi.utilities import f2i
 
 import logger
-from cell_workers import SpinNearestFortWorker, CatchVisiblePokemonWorker, PokemonCatchWorker, SeenFortWorker, MoveToFortWorker, PokemonTransferWorker, EvolveAllWorker, RecycleItemsWorker, IncubateEggsWorker
+from cell_workers import SpinNearestFortWorker, CatchVisiblePokemonWorker, PokemonCatchWorker, SeenFortWorker, MoveToFortWorker, PokemonTransferWorker, EvolveAllWorker, RecycleItemsWorker, IncubateEggsWorker, SoftBanWorker
 from cell_workers.utils import distance, get_cellid, encode, i2f
 from human_behaviour import sleep
 from item_list import Item
@@ -35,6 +35,7 @@ class PokemonGoBot(object):
         return self.api._position_lat, self.api._position_lng, 0
 
     def __init__(self, config):
+        self.softbanned = False
         self.config = config
         self.fort_timeouts = dict()
         self.pokemon_list = json.load(open(os.path.join('data', 'pokemon.json')))
@@ -77,6 +78,7 @@ class PokemonGoBot(object):
         self.check_session(self.position[0:2])
 
         workers = [
+            SoftBanWorker,
             IncubateEggsWorker,
             PokemonTransferWorker,
             EvolveAllWorker,
--- a/pokemongo_bot/cell_workers/catch_visible_pokemon_worker.py
+++ b/pokemongo_bot/cell_workers/catch_visible_pokemon_worker.py
@@ -15,7 +15,7 @@ class CatchVisiblePokemonWorker(object):
         self.position = bot.position
 
     def work(self):
-        if not self.config.catch_pokemon:
+        if not self.config.catch_pokemon or self.bot.softbanned:
             return
 
         if 'catchable_pokemons' in self.cell and len(self.cell['catchable_pokemons']) > 0:
--- a/pokemongo_bot/cell_workers/move_to_fort_worker.py
+++ b/pokemongo_bot/cell_workers/move_to_fort_worker.py
@@ -23,7 +23,8 @@ class MoveToFortWorker(object):
         dist = distance(self.position[0], self.position[1], lat, lng)
 
         if dist > 10:
-            logger.log('Moving towards fort {}, {} left'.format(fortID, format_dist(dist, unit)))
+            if self.bot.softbanned == False:
+                logger.log('Moving towards fort {}, {} left'.format(fortID, format_dist(dist, unit)))
 
             step_walker = StepWalker(
                 self.bot,
@@ -35,5 +36,8 @@ class MoveToFortWorker(object):
             if not step_walker.step():
                 return WorkerResult.RUNNING
 
-        logger.log('Arrived at Pokestop')
+        if self.bot.softbanned == False:
+            logger.log('Arrived at Pokestop')
+
         return WorkerResult.SUCCESS
+
--- a/pokemongo_bot/cell_workers/pokemon_catch_worker.py
+++ b/pokemongo_bot/cell_workers/pokemon_catch_worker.py
@@ -122,9 +122,12 @@ class PokemonCatchWorker(object):
                                     logger.log('Catch Rate with normal Pokeball has increased to {}%'.format(success_percentage))
                                 else:
                                     if response_dict['status_code'] is 1:
-                                        logger.log('Fail to use berry. Seem like you are softbanned.', 'red')
+                                        logger.log('Failed to use berry. Seems like you are softbanned.', 'red')
+                                        logger.log('Heading to nearest Pokestop to attempt unban.', 'red')
+                                        self.bot.softbanned = True
+                                        break
                                     else:
-                                        logger.log('Fail to use berry. Status Code: {}'.format(response_dict['status_code']),'red')
+                                        logger.log('Failed to use berry. Status Code: {}'.format(response_dict['status_code']),'red')
 
                             # change ball to next tier if catch rate is too low
                             current_type = pokeball
--- a/pokemongo_bot/cell_workers/seen_fort_worker.py
+++ b/pokemongo_bot/cell_workers/seen_fort_worker.py
@@ -28,6 +28,7 @@ class SeenFortWorker(object):
     def work(self):
         lat = self.fort['latitude']
         lng = self.fort['longitude']
+        ret = 0
 
         self.api.fort_details(fort_id=self.fort['id'],
                               latitude=lat,
@@ -40,25 +41,11 @@ class SeenFortWorker(object):
             fort_name = fort_details['name'].encode('utf8', 'replace')
         else:
             fort_name = 'Unknown'
-        logger.log('Now at Pokestop: ' + fort_name,
+        if self.bot.softbanned == False:
+            logger.log('Now at Pokestop: ' + fort_name,
                    'cyan')
-        if self.config.catch_pokemon and 'lure_info' in self.fort:
-            # Check if the lure has a pokemon active
-            if 'encounter_id' in self.fort['lure_info']:
-                logger.log("Found a lure on this pokestop! Catching pokemon...", 'cyan')
-
-                pokemon = {
-                    'encounter_id': self.fort['lure_info']['encounter_id'],
-                    'fort_id': self.fort['id'],
-                    'latitude': self.fort['latitude'],
-                    'longitude': self.fort['longitude']
-                }
-
-                self.catch_pokemon(pokemon)
-
-            else:
-                logger.log('Found a lure, but there is no pokemon present.', 'yellow')
-            sleep(2)
+        else:
+            logger.log('Using Pokestop: ' + fort_name + ' to try to remove ban', 'red')
 
         logger.log('Spinning ...', 'cyan')
 
@@ -74,15 +61,18 @@ class SeenFortWorker(object):
             spin_details = response_dict['responses']['FORT_SEARCH']
             spin_result = spin_details.get('result', -1)
             if spin_result == 1:
-                logger.log("Loot: ", 'green')
                 experience_awarded = spin_details.get('experience_awarded',
                                                       False)
                 if experience_awarded:
+                    logger.log("Loot: ", 'green')
                     logger.log(str(experience_awarded) + " xp",
                                'green')
+                    if self.bot.softbanned == True:
+                       logger.log("Softban removed!", "green") 
+                       self.bot.softbanned = False
 
                 items_awarded = spin_details.get('items_awarded', False)
-                if items_awarded:
+                if items_awarded and self.bot.softbanned == False:
                     self.bot.latest_inventory = None
                     tmp_count_items = {}
                     for item in items_awarded:
@@ -96,7 +86,8 @@ class SeenFortWorker(object):
                         item_name = self.item_list[str(item_id)]
                         logger.log('- ' + str(item_count) + "x " + item_name + " (Total: " + str(self.bot.item_inventory_count(item_id)) + ")", 'yellow')
                 else:
-                    logger.log("[#] Nothing found.", 'yellow')
+                    if self.bot.softbanned == False:
+                        logger.log("[#] Nothing found.", 'yellow')
 
                 pokestop_cooldown = spin_details.get(
                     'cooldown_complete_timestamp_ms')
@@ -107,17 +98,10 @@ class SeenFortWorker(object):
                         format_time((pokestop_cooldown / 1000) -
                                     seconds_since_epoch)))
 
-                if not items_awarded and not experience_awarded and not pokestop_cooldown:
-                    message = (
-                        'Stopped at Pokestop and did not find experience, items '
-                        'or information about the stop cooldown. You are '
-                        'probably softbanned. Try to play on your phone, '
-                        'if pokemons always ran away and you find nothing in '
-                        'PokeStops you are indeed softbanned. Please try again '
-                        'in a few hours.')
-                    raise RuntimeError(message)
+                if self.bot.softbanned == False and not items_awarded and not experience_awarded and not pokestop_cooldown:
+                    logger.log("Seems you are softbanned, trying to fix...", "red")
+                    self.bot.softbanned = True
 
-                self.bot.recent_forts = self.bot.recent_forts[1:] + [self.fort['id']]
             elif spin_result == 2:
                 logger.log("[#] Pokestop out of range")
             elif spin_result == 3:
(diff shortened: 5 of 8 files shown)


douglascamata:
@MikeDX what about extracting softban unban feature into a worker that runs every tick, checks `self.bot.softbanned` and run if it is true? Let's make things more modular!

pokemongo_bot/__init__.py
```diff
@@ -26,6 +26,8 @@
 
 class PokemonGoBot(object):
 
+    softbanned = False
```
DayBr3ak: place this in the __init__ method ?

pokemongo_bot/cell_workers/seen_fort_worker.py
```diff
@@ -108,14 +108,19 @@ def work(self):
...
+                    logger.log("Seems you are softbanned, trying to fix...", "red")
+
+                    for spinstop in range (0, 50):
+                        self.api.fort_search(fort_id=self.fort['id'],
+                            fort_latitude=lat,
+                            fort_longitude=lng,
+                            player_latitude=f2i(self.position[0]),
+                            player_longitude=f2i(self.position[1]))
+                        response_dict = self.api.call()
+
```
DayBr3ak: add a sleep ? seems much more human and less botty :)

MikeDX:
@douglascamata So, to do this:

if bot banned:
walk to pokestop
spin spin spin
set banned = false?

resume normal operation

douglascamata:
SoftBanFixWorker:
```
if bot banned:
    walk to pokestop (using MoveToFortWorker)
    spin once (using SeenFortWorker)
    if still banned:
        return WorkerState.RUNNING (so tick is immediately finished and next one starts)
```

Add this detail to SeenFortWorker
```
if got loot:
  set banned = false
```

And this SoftBanFixWorker should be added as first element of the worker list, to run first.

MikeDX:
Ok, makes sense. should the unban worker be in the root, or as a cell worker? (guessing root)

MikeDX:
@douglascamata How would you suggest we keep the nearest fort between ticks, since currently it is doing its job but moving between forts to try to fix each tick which isn't what we want.

MikeDX:
New worker created, removed lots of output when iterating through the spins. Works well, although takes about a minute to complete. 

Also checks for lures after the ban is lifted

MikeDX:
Resolved conflicts

wilsonyqm:
Can't wait to see this PR merged

MatthewARoy:
Ah excellent, I really hope this works as expected. It will be a very nice feature to auto-unban.
Thanks guys!

douglascamata:
@MikeDX actually we want it to recalculate the nearest fort at every tick. It should work without `self.bot.nearest_fort`.

MikeDX:
Using nearest every tick moves to the next one instead of spinning the same stop.

MikeDX:
rebased to dev + squashed again.

douglascamata:
@MikeDX but `SpinNearestFort` worker won't execute at all while you return `WorkerResult.RUNNING`, because you are the first worker

MikeDX:
I'm not calling that method or using SpinNearestFort anymore - it doesnt need it. The first time the bot runs, banned = false

then it goes through the list, eventually calling SpinNearestFortWorker which *does* set the nearest_fort value. Even if CatchVisiblePokemonWorker sets the ban to true before SpinNearest runs, SpinNearestFortWorker will still execute and find a stop regardless.

douglascamata:
@MikeDX hum, I see now. I think of another option other than the attribute `self.nearest_fort`: do your own fort search.

As simple as seen in `SpinNearestFortWorker`:

```
def nearest_pokestop(self):
    if 'forts' in self.bot.cell:
        forts = [fort
            for fort in self.bot.cell['forts']
             if 'latitude' in fort and 'type' in fort]
        forts.sort(key=lambda x: distance(self.position[
                       0], self.position[1], x['latitude'], x['longitude']))
        return forts[0]
    return None
```

douglascamata:
#1578 makes it even better! Fort search is not in `self.bot` and it doesn't include any filter. Perfect for you!

MikeDX:
@douglascamata Would have been handy to have that answer, when I asked the question 7 hours ago...

douglascamata:
#1578 `self.bot.get_forts` include ALL nearby forts, regardless of any filter. That's all you need.

TheSavior:
That function should probably be pulled out as well, we use it in a bunch of places. 

Side note, @douglascamata I think we need a new place for those types of functions that aren't the bot class.

douglascamata:
@TheSavior yes, we need refactor like #1253. But who wants to test/merge? :P

MikeDX:
@douglascamata I have updated the bot to get its own fort position, but since that function is not yet merged, I have replaced it with my own for now.

douglascamata:
@MikeDX ok... I do a quick update of your code in dev after we merge #1578

douglascamata:
👍

douglascamata:
@TheSavior give a thumbs up here... I update the SoftBanWorker to use your utility functions after we merge our PR.

pokemongo_bot/__init__.py
```diff
@@ -77,6 +78,7 @@ def tick(self):
         self.check_session(self.position[0:2])
 
         workers = [
+            SoftBanWorker,
```
TheSavior: This should probably be down the list, right before catching pokemon. Since the bot can still transfer and evolve while banned.

TheSavior:
👍

pokemongo_bot/__init__.py
douglascamata: @TheSavior but which pokemon would you transfer or evolve if you cannot catch?

pokemongo_bot/cell_workers/catch_visible_pokemon_worker.py
```diff
@@ -15,7 +15,7 @@ def __init__(self, bot):
         self.position = bot.position
 
     def work(self):
-        if not self.config.catch_pokemon:
+        if not self.config.catch_pokemon or self.bot.softbanned:
```
TheSavior: we shouldn't need to change this

TheSavior:
:-1:

pokemongo_bot/cell_workers/move_to_fort_worker.py
```diff
@@ -23,7 +23,8 @@ def work(self):
         dist = distance(self.position[0], self.position[1], lat, lng)
 
         if dist > 10:
-            logger.log('Moving towards fort {}, {} left'.format(fortID, format_dist(dist, unit)))
+            if self.bot.softbanned == False:
```
TheSavior: We shouldn't be changing this either.

pokemongo_bot/cell_workers/catch_visible_pokemon_worker.py
douglascamata: why not?

douglascamata: it will waste pokeballs if we don't do this

pokemongo_bot/cell_workers/move_to_fort_worker.py
douglascamata: this one I agree

pokemongo_bot/cell_workers/soft_ban_worker.py
```diff
@@ -0,0 +1,31 @@
...
+        self.running = False
+        self.bot = bot
+    def work(self):
+       if self.bot.softbanned == True:
+           nearest_fort = self.nearest_pokestop()
+           MoveToFortWorker(nearest_fort, self.bot).work()
+           SeenFortWorker(nearest_fort, self.bot).work()
+
+           if self.bot.softbanned == True:
+               return WorkerResult.RUNNING
```
TheSavior: This will keep anything after it from running in the tick. This means that NO other worker needs to know about whether we are soft banned. They should not have any check in it at all.

pokemongo_bot/cell_workers/soft_ban_worker.py
```diff
@@ -0,0 +1,31 @@
...
+
+
+class SoftBanWorker(object):
+    def __init__(self, bot):
+        self.running = False
+        self.bot = bot
+    def work(self):
+       if self.bot.softbanned == True:
+           nearest_fort = self.nearest_pokestop()
+           MoveToFortWorker(nearest_fort, self.bot).work()
```
TheSavior: You don't need this. The only way to get into the softbanned mode is to have spun a fort and not gotten anything. That means you are close enough to spin it again.

pokemongo_bot/cell_workers/soft_ban_worker.py
```diff
@@ -0,0 +1,31 @@
...
+           SeenFortWorker(nearest_fort, self.bot).work()
+
+           if self.bot.softbanned == True:
+               return WorkerResult.RUNNING
+ 
+    def nearest_pokestop(self):
```
TheSavior: This doesn't seem right...

pokemongo_bot/cell_workers/soft_ban_worker.py
```diff
@@ -0,0 +1,31 @@
+# -*- coding: utf-8 -*-
+
+from pokemongo_bot.cell_workers.utils import distance
+from pokemongo_bot.worker_result import WorkerResult
+from pokemongo_bot.cell_workers import SeenFortWorker
```
TheSavior: You can combine these imports

pokemongo_bot/cell_workers/seen_fort_worker.py
```diff
@@ -74,15 +61,18 @@ def work(self):
...
             spin_result = spin_details.get('result', -1)
             if spin_result == 1:
-                logger.log("Loot: ", 'green')
                 experience_awarded = spin_details.get('experience_awarded',
                                                       False)
                 if experience_awarded:
+                    logger.log("Loot: ", 'green')
                     logger.log(str(experience_awarded) + " xp",
                                'green')
+                    if self.bot.softbanned == True:
```
TheSavior: The only thing that should set us to *not* be softbanned anymore should be the softbanned worker.

pokemongo_bot/__init__.py
TheSavior: That's a good point, at least with the current behavior of the workers we have.

pokemongo_bot/cell_workers/catch_visible_pokemon_worker.py
TheSavior: It will never get to this worker since the softbanned worker returns RUNNING

pokemongo_bot/cell_workers/soft_ban_worker.py
MikeDX: No, because if a catch worker sets banned to true (if a berry fails) then spinworker needs to know.

pokemongo_bot/cell_workers/soft_ban_worker.py
douglascamata: no no, he can't... all these 3 imports are importing from different inner-modules.

pokemongo_bot/cell_workers/seen_fort_worker.py
MikeDX: How would that even work?

pokemongo_bot/cell_workers/soft_ban_worker.py
MikeDX: You can, but it's pretty horrible coding practice and certainly not what PEP recommends.

TheSavior: Line 5 and 6 both import from `pokemongo_bot.cell_workers`

pokemongo_bot/cell_workers/soft_ban_worker.py
MikeDX: Not true. You can be softbanned by throwing a berry and failing, or even if a pokemon ran away at 100% catch rate.

pokemongo_bot/cell_workers/move_to_fort_worker.py
MikeDX: This is about the only merge comment I agree with!

pokemongo_bot/cell_workers/soft_ban_worker.py
MikeDX: I've no idea what part you are even referring to. This is all based on suggestions from @douglascamata after the initial (small) commit was rejected for not being modular.

pokemongo_bot/cell_workers/soft_ban_worker.py
douglascamata: catch worker will never set banned to true, because we only know we are banned when we spin pokestop

pokemongo_bot/cell_workers/seen_fort_worker.py
douglascamata: Using `return WorkerResult.SUCCESS`

pokemongo_bot/cell_workers/pokemon_catch_worker.py
```diff
@@ -122,9 +122,12 @@ def work(self):
                                     logger.log('Catch Rate with normal Pokeball has increased to {}%'.format(success_percentage))
                                 else:
                                     if response_dict['status_code'] is 1:
-                                        logger.log('Fail to use berry. Seem like you are softbanned.', 'red')
+                                        logger.log('Failed to use berry. Seems like you are softbanned.', 'red')
+                                        logger.log('Heading to nearest Pokestop to attempt unban.', 'red')
+                                        self.bot.softbanned = True
+                                        break
```
TheSavior: Why isn't this a return? If we are softbanned, what else is there to do in the catch worker?

pokemongo_bot/cell_workers/soft_ban_worker.py
MikeDX: If this is the case why does the berry fail and show a softban message?

pokemongo_bot/cell_workers/soft_ban_worker.py
MikeDX: line 6 isnt even in your comment

MikeDX:
@TheSavior @douglascamata Can I suggest one of you take what I've done and do all the testing and re-implementation of what I have done since you don't agree with any of this pull request.

pokemongo_bot/cell_workers/catch_visible_pokemon_worker.py
MikeDX: If there are more than one pokemon within range, yes it will

pokemongo_bot/cell_workers/soft_ban_worker.py
MikeDX: line 6 isnt even in your comment

MatthewARoy:
Thanks for working on it @MikeDX
Have we done any testing yet? I've seen it on a few websites but wanted to make sure with more thorough testing. If you get a branch in working order I'll check it out, make a new account and softban it a few times to test

MikeDX:
I did a lot of testing yesterday and spent many hours on it getting it to work as it should and bent it to fit "the rules"

Since closing this pull request, dev has changed multiple times and a lot has changed meaning most if not all would have to be re-done. I made a lot of changes to rebase with dev, and it still isn't working properly. One worker returns "11" or "0" - whatever that means, certainly not a SUCCESS or RUNNING value. 

As with my previous pull request, I'm not really willing to waste any more time on it for it to be glossed over and nitpicked. See above comments on the code - almost every single one was based upon a misunderstanding of what the problem was and how to fix it.

I have participated in multiple open source projects in the past and I have never seen such a crazy set of rules for merging code before - it is totally hit and miss here. I see garbage merged into the code that breaks the whole system, and even with syntax errors with tabulation, yet when I submit working, tested code, it is pulled apart for not being done in a certain way - and then nitpicked again by someone who did not even understand the problem in the first place. Frustrating and absolutely crazy.

TheSavior:
@MikeDX, I hear you and I definitely agree, I did not understand the problem and your solution as a whole while I was reviewing it. Our goal on this project is to be as inclusive as possible and to help contributors feel as good as possible contributing. We failed on this one, and that is entirely my fault.

We definitely want to get this feature in, it is valuable to many people using this bot. I understand the lack of interest you have iterating on this more without a clear path to success. Recognizing that, we will probably need to spend our time doing it. Because you deleted your branch, we will probably have to pull out the functionality from the diff itself. 

We are overwhelmed with pull requests and tend to do a first pass on reviews just pattern matching on the common issues we've seen. Typically that includes things like: implementing a feature that requires changes to lots of workers, workers calling other workers, PRs that say they do one thing and then actually change a bunch of other things as well. Unfortunately, I pattern matched this PR to the first case, which was not fair to this PR and the work you've done. When spending more time looking at it and making sure I understand the things actually being done, it makes more sense. 

Again, this is my mistake and thus I need to remedy that by getting this functionality added. I hope this experience won't completely put you off from this project, we always want to value quality contributions and quality contributors. Thank you for contributing this PR.

TheSavior:
@MikeDX Here is a follow up PR. Please take a look, your code influenced the implementation. https://github.com/PokemonGoF/PokemonGo-Bot/pull/1724

MikeDX:
@TheSavior I appreciate what you are saying, but pull request 1724 is shockingly bad and @douglascamata didn't even follow his own advice that he told me to do after my original commit!

I really hope you get your act(s) together on this project. It showed so much promise but arrogance and a certain level of ignorance has definitely put me off almost entirely now which for me is a real shame as there is absolutely no motivation to work on a feature or bug fix since submissions are so poorly vetted.

MikeDX:
See issue #2049. 

As I said before, I have contributed to many projects both open source and within professional environments and this is easily the worst managed one I have ever come across.

DayBr3ak:
"I have contributed so many time, I'm the bestttttt" Who is arrogant now ?

PR #1529, opened 2016-07-29
Added logging for missed API calls

demgod:
Short Description: 
Identify which calls we missed and must retry, maybe make this debug only?

Fixes:
- Lists API calls that were attempted

### DIFF
--- a/pokemongo_bot/api_wrapper.py
+++ b/pokemongo_bot/api_wrapper.py
@@ -2,6 +2,7 @@
 
 from pgoapi import PGoApi
 from pgoapi.exceptions import NotLoggedInException
+from POGOProtos.Networking.Requests_pb2 import RequestType
 from human_behaviour import sleep
 import logger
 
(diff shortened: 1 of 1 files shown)


pokemongo_bot/api_wrapper.py
```diff
@@ -63,6 +64,7 @@ def call(self, max_retry=5):
             if not self._is_response_valid(result, request_callers):
                 try_cnt += 1
                 logger.log('Server seems to be busy or offline - try again - {}/{}'.format(try_cnt, max_retry), 'red')
+                logger.log('Attempted to call: {}'.format(", ".join([RequestType.Name(isinstance(req_method, dict) and req_method.keys()[0] or req_method) for req_method in api_req_method_list])), 'red')
```
DayBr3ak: You could have used `request_callers` which basically is what you are trying to log (example: `["GET_GYM_DETAILS", "GET_INVENTORY"]`

TheSavior:
I believe this is less valuable now that we understand the problem a little better and have throttling. Also, we can get this information by turning on the DEBUG config flag.

Issue #1309, opened 2016-07-28
[Important REQUEST] evolve_captured with list


DimaVIII:
### Expected Behavior
evolve_captured should be working exactly like the evolve_all.
With parameter "true", "Pidgey,Weedle" (list), "false".

This function is build to gain extra XP and work fine but if you don't want to lose all your candy's for special Pokemon you can't use it.
For example if you want to farm candy's for Charmander to evolve later a Charizard.

Of course you can work with "pokemon_to_transfer" but the bot will than end up fill up your Pokemon bag.

**I think this is a the most important function for this bot.**

DimaVIII:
https://github.com/PokemonGoF/PokemonGo-Bot/issues/1076

TheGoldenXY:
Issue has been resolved.

Please upgrade to the latest `dev` branch by running this command:
```
git reset --hard
git pull -f
```

Then run this command:
`pip install --upgrade -r requirements.txt`

If issue persists, open a new issue.

/closed

PR #1532, opened 2016-07-29
"evolve_captured" is now using a list instead of a boolean

### LINKED ISSUE #1076: evolve_captured in config.json does not use the specified evolve list
### Expected Behavior
Bot also checks against list of pokemon to evolve

### Actual Behavior
Bot only using `pokemon_to_transfer`

### Steps to Reproduce
Bot captures pokemon and evolves

### Other Information
OS: 
Git Commit: `db5b719dba2faebf9401b929811d0308e2561317`

Excerpt from `config.json`
`"evolve_all": "Pidgey,Weedle,Rattata,Caterpie,Geodude,Ekans,Zubat",`
Log:
```
[22:53:57] Found fort 5420f52faf294cdcbba72166743ef45b.16 at distance 0.02km
[22:53:57] Need to move closer to Pokestop
[                                        ] 0%[22:54:01] Something rustles nearby!
[22:54:01] A Wild Sandshrew appeared! [CP 290] [Potential 0.53]
[22:54:01] IV [Stamina/Attack/Defense] = [15/2/7]
[22:54:05] Using Pokeball (chance: 38.66%)... (0 left!)
[22:54:06] Captured Sandshrew! [CP 290] [15/2/7]
[22:54:06] Sandshrew has been evolved!
```

valerian:
"evolve_captured" is now using a list instead of a boolean, working the same way as "evolve_all"

Fixes #1076 and #1309

### DIFF
--- a/configs/config.json.example
+++ b/configs/config.json.example
@@ -30,7 +30,7 @@
     "use_lucky_egg": false,
     "hatch_eggs": true,
     "longer_eggs_first": true,
-    "evolve_captured": false,
+    "evolve_captured": "NONE",
     "release_pokemon": true,
     "spin_forts": {
       "avoid_circles": false,
--- a/configs/config.json.pokemon.example
+++ b/configs/config.json.pokemon.example
@@ -30,7 +30,7 @@
     "use_lucky_egg": false,
     "hatch_eggs": true,
     "longer_eggs_first": true,
-    "evolve_captured": false,
+    "evolve_captured": "NONE",
     "release_pokemon": true,
     "spin_forts": {
       "avoid_circles": false,
--- a/pokecli.py
+++ b/pokecli.py
@@ -342,6 +342,13 @@ def init_config():
             ' Set these to true or false and remove "mode" from your configuration')
         return None
 
+    if (config.evolve_captured
+        and (not isinstance(config.evolve_captured, str)
+             or str(config.evolve_captured).lower() in ["true", "false"])):
+        parser.error('"evolve_captured" should be list of pokemons: use "all" or "none" to match all ' +
+                     'or none of the pokemons, or use a comma separated list such as "Pidgey,Weedle,Caterpie"')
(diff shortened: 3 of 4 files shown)


douglascamata:
Did you test both in CLI and JSON?

valerian:
>Did you test both in CLI and JSON?

yes

configs/config.json.example
```diff
@@ -30,7 +30,7 @@
     "use_lucky_egg": false,
     "hatch_eggs": true,
     "longer_eggs_first": true,
-    "evolve_captured": false,
```
TheSavior: What behavior should happen for people who haven't updated their config? If false won't work the same as "NONE" then I think we should have a check and if it is a boolean raise an exception to let users know the configuration has changed.

valerian:
Added a parse error for when `evolve_catched` is not a string, or is a boolean-looking string:

> "evolve_captured" should be list of pokemons: use "all" or "none" to match all or none of the pokemons, or use a comma separated list such as "Pidgey,Weedle,Caterpie"

configs/config.json.example
valerian: `false` will do the same as `"none"`, but `true` will not do the same as `"all"`. Added a parse error to cover that situation, as requested.

douglascamata:
👍

DimaVIII:
https://github.com/PokemonGoF/PokemonGo-Bot/issues/1643

bigkraig:
evolve_captured should be a list in the config example:
```
    "evolve_captured": ["NONE"],
```

valerian:
@bigkraig I just followed the exact same logic as `evolve_all` which is currently like that in the config:

```
    "evolve_all": "NONE",
```

valerian:
@DimaVIII this is not relevant to this pull request, I did only change the conditions for evolving and the config parsing, the rest was untouched.

PR #1536, opened 2016-07-29
Hotfix api_wraper.py responed server error msg for the first try| Optimized the berry_inventory function called times

xc-mezcal:
Short Description: 

*************************************This is wrong guess but a quite close fix **********************
Some time, the `result = self._api.call()` is getting response but not filled with full response msg. Then the check begins,

**************************************The True may be ************************************
We believe that server now detects massive requests, so we put a force sleep after an API call. 

So the speed of making api request is reduced to hot fix this problem
*********************************************************************************************

This makes the every first response check failed (because it's still point to empty) and second response succeed. 

Fixes:
Add a wait time to "wait" the result to be "fully filled" solve this problem for me.

### DIFF
--- a/pokemongo_bot/cell_workers/pokemon_catch_worker.py
+++ b/pokemongo_bot/cell_workers/pokemon_catch_worker.py
@@ -78,6 +78,9 @@ class PokemonCatchWorker(object):
                             return False
 
                         items_stock = self.bot.current_inventory()
+                        berry_id = 701 # @ TODO: use better berries if possible
+                        berries_count = self.bot.item_inventory_count(berry_id)
+                        
                         while True:
                             ## pick the most simple ball from stock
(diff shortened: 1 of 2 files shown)


pokemongo_bot/cell_workers/pokemon_catch_worker.py
```diff
@@ -78,6 +78,9 @@ def work(self):
                             return False
 
                         items_stock = self.bot.current_inventory()
+                        berry_id = 701 # @ TODO: use better berries if possible
```
TheSavior: This change seems irrelevant to this PR.

xc-mezcal: I don't know how git hub PR works.

But it automatically added my second PR to this one. lol

PR #1541, opened 2016-07-29
[FIX] evolve_all and use_lucky_egg

codybaldwin:
Short Description: 
fixed issues: #1335 , #1339 , #1456
Fixes:
- Evolve_all will now work again
- Use_lucky_egg is also fixed.  Will only run on startup (not continuously attempt to use egg and spam message)
- Slight optimization to EvolveAllWorker._should_run() so that it requires less logic to reach a decision
- Added tick count to bot that can be used in workers. This will allow the worker to modify its behavior and add flexibility to the workers.  ex: run on first tick, run every other tick 'if tick_count%2 is 0', etc

### DIFF
--- a/pokemongo_bot/__init__.py
+++ b/pokemongo_bot/__init__.py
@@ -39,6 +39,7 @@ class PokemonGoBot(object):
         self.latest_inventory = None
         self.cell = None
         self.recent_forts = [None] * config.max_circle_size
+        self.tick_count = 0
 
     def start(self):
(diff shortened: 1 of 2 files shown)


codybaldwin:
As I stated in PR #1469 the implementation should only run the use_lucky_egg logic once at the beginning.  Messages and all related luck egg logic will only run once

solderzzc:
Procedure is more clear.

PR #1546, opened 2016-07-29
Update README.md

arlber:
Short Description: 

Fixes:
- 
- 
- 

For Windows virtualenv creates a batch file to modify the shell prompt in the Scripts folder

### DIFF
--- a/README.md
+++ b/README.md
@@ -157,11 +157,13 @@ $ git submodule update
 $ git clone -b dev https://github.com/PokemonGoF/PokemonGo-Bot
 $ cd PokemonGo-Bot
 $ virtualenv .
-$ Scripts\activate
+$ source bin/activate
 $ pip install -r requirements.txt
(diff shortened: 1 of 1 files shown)


maestroi:
For dev it needs to be cerified that it only needed for windows that linux still needs the other comment, also needs clarification

NamPNQ:
@maestroi:
@arlber is correct, follow tut in [virtualenv doc](https://virtualenv.pypa.io/en/stable/userguide/)

arlber:
@maestroi you are correct, I went back and made another commit clarifying the development section. It may make sense to change the development section to say something along the lines of "same procedure as above but change 'master' to 'branch'" to reduce the amount of text

maestroi:
@arlber Sounds good now

arlber:
Installation instructions removed from the file and moved to the wiki.

Issue #1544, opened 2016-07-29
Flag `spin_forts` doesn't work


jasonliu119:
I tried to debug it and print `str(config.spin_forts)` in `pokecli.py`, it says `{u'avoid_circles': False, u'max_circle_size': 10}`. What is wrong?

### Expected Behavior
when I config spin_forts to false, it should not spin the forts.

### Actual Behavior

when I config spin_forts to false, it still spins the forts.

### Steps to Reproduce
just try set spin_forts to false

### Other Information
OS: Ubuntu 14.04
Git Commit: (run 'git log -n 1 --pretty=format:"%H"' and paste it here)
6c6fc59de9ea02a2079cb7548f80937f65894088

banksymate:
Please specify the OS and Git Commit

PR #1547, opened 2016-07-29
Fix circle_avoidance

MFizz:
PR #1553 offers the same fix while keeping the possibility to have nested keywords
Short Description: 
spin_forts was assigned twice in config, removed nested keywords

Fixes:
#1544 
circle avoidance now works as intended

PR #1553, opened 2016-07-29
Fix/alt circle avoidance

MFizz:
Short Description: 
alternative fix to PR #1547 while keeping the possibility to have nested keywords

spin_forts was assigned twice in config

Fixes:
#1544 
circle avoidance now works as intended

### DIFF
--- a/configs/config.json.example
+++ b/configs/config.json.example
@@ -6,7 +6,6 @@
     "gmapkey": "GOOGLE_MAPS_API_KEY",
     "max_steps": 5,
     "catch_pokemon": true,
-    "spin_forts": true,
     "walk": 4.16,
     "action_wait_min": 1,
     "action_wait_max": 4,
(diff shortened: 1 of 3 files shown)


douglascamata:
There is more elegant solution to this by modifying `add_config` to make another parameter, maybe named `embedded_in`, to store this kind of information. 

I don't think replacing a simple `dict.get` with a recursive search function is something good in terms of code complexity. And it also hides from `add_config` logic that the key is inside another. 

Your code is prone to bugs like, another key in the dict root has a subkey with the same name. You will never search in the correct place because you take the first subkey you find that matches the name you want.

MFizz:
Agreed. My idea originally was to keep the key unique regardless of depth, since cli/ flag functionality would dictate it anyway and to use upper levels only for grouping logically. But I can see how this feels dirty/ hackish. My problem now is just how to deal with the json - cli duality and that long_flag is used as attribute name and clie flag.

PR #1559, opened 2016-07-29
annoying bug

gregtampa:
Short Description: 
temporary workaround for a log that spams the bot
Fixes:
-  spams in log due to server latency
- 
- 

Added if try_cnt > 1: to remove the annoying count in logs due to server latency
or something
The response that comes back is "52" from api.


tvb:
➕1

PR #1560, opened 2016-07-29
Change MoveToFort text to not send 2 lines and have a more humanreadable

heihachi:
Short Description: 
Changed the way MoveToFort was outputting.

PR #1566, opened 2016-07-29
Reduce log spam when moving towards forts

TheSavior:
Proposing this solution instead of #1560 

screen shot 2016-07-29 at 12 48 55 am


douglascamata:
👍

heihachi:
Looks good to me! In mine i wanted it to output every 0.1km or as close as possible but that is loads better

PR #1568, opened 2016-07-29
Moved egg hatching to incubation worker, added feedback for users

reddivision:
Short Description: 
whatever I wrote last time + added pokemon name/stats to hatched egg output + misc fixes for things pointed out in CR
Fixes:
- incubation code

(If you experience weird behavior and haven't already, please `rm -rf src/` and `pip install -r requirements.txt` kthx)

### DIFF
--- a/CONTRIBUTORS.md
+++ b/CONTRIBUTORS.md
@@ -41,3 +41,4 @@
  * surceis
  * SpaceWhale
  * klingan
+ * reddivision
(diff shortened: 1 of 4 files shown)


TheSavior:
This looks good to me. I only had a few comments from your last PR and it looks like you addressed them here.

👍

douglascamata:
👍

PR #1571, opened 2016-07-29
Update pokemon_transfer_worker.py

HannibalChew:
Grammar correction.

### DIFF
--- a/pokemongo_bot/cell_workers/pokemon_transfer_worker.py
+++ b/pokemongo_bot/cell_workers/pokemon_transfer_worker.py
@@ -60,7 +60,7 @@ class PokemonTransferWorker(object):
                                                                           best_pokemon['cp'],
                                                                           best_pokemon['iv']), 'green')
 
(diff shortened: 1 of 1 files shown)


Zgrkaralar:
👍

pokemongo_bot/cell_workers/pokemon_transfer_worker.py
```diff
@@ -60,7 +60,7 @@ def work(self):
                                                                           best_pokemon['cp'],
                                                                           best_pokemon['iv']), 'green')
 
-                        logger.log("Exchange {} pokemon(s)".format(len(all_pokemons)), "green")
+                        logger.log("Exchanging {} pokemon".format(len(all_pokemons)), "green")
```
TheSavior: It should really say `Transferring`, right?

Zgrkaralar: Maybe **Transferring** is better cuz on game always saying transfer transfer 💃

maestroi:
seems good, i think there is no clarification is needed because in the game it is called transfer

Issue #1570, opened 2016-07-29
Bot walking slow


Bashin:
### Expected Behavior
Setting in config "walk": 4.16 (m/s)
Should take about 50 - 60 seconds to get to a pokestop at a distance of 0.2 km

### Actual Behavior
Moving to pokestop at a speed of ~0.9 (m/s) - took 220 seconds for 0.2 km!
Pathing looks great though! :-)

### Steps to Reproduce
Just run the bot and compare times and distances traveled

### Other Information
OS: Win10
Git Commit: annoying bug (#1559) - installed from scratch one hour ago

BartKoppelmans:
The speed was changed to be km/h. So you are walking at 4.16km/h. Set it to 15 to have some more speed!

Bashin:
Great i'm an idiot :-D
Maybe the default value should be changed to a higher value

PR #1574, opened 2016-07-29
Increase default walking speed

Bashin:
Short Description: 
Walking speed was changed from m/s to km/h, so i updated the config files to match the old default speed.

### DIFF
--- a/configs/config.json.example
+++ b/configs/config.json.example
@@ -8,7 +8,7 @@
     "catch_pokemon": true,
     "websocket_server": false,
     "spin_forts": true,
-    "walk": 4.16,
+    "walk": 15,
     "action_wait_min": 1,
     "action_wait_max": 4,
     "debug": false,
--- a/configs/config.json.pokemon.example
+++ b/configs/config.json.pokemon.example
@@ -7,7 +7,7 @@
     "max_steps": 5,
     "catch_pokemon": true,
     "spin_forts": true,
-    "walk": 4.16,
+    "walk": 15,
     "action_wait_min": 1,
     "action_wait_max": 4,
     "debug": false,


Jojo1910:
Isn´t 14 km/h the perfect speed for eggs?

Bashin:
Maybe.. I just converted 4.16 m/s to 15 km/h (multiplied by 3.6)

douglascamata:
@Bashin StepWalker consider the speed in `m/s`, because our `distance` functions returns in meters:

```
distance(
            self.initLat,
            self.initLng,
            destLat,
            destLng
        )
```

Bashin:
Are you sure about that? -> #1570 
Wiki says km/h too

Nihisil:
I put 14 to the config and bot started to walk really fast (to compare with previous walk speed).

And with this speed eggs still hatching.

Bashin:
I put 15 to the config and if you compare distance traveled and time it has to be km/h.
Egg hatching works for me too

MikeDX:
If you set a higher speed, it may still not work depending upon location - also the egg hatching distance count may not match the actual distance the bot has travelled.

Using a higher value in speed in one location allowed eggs to be hatched at a fairly regular place, but in another, the bot ran for a while and the counter did not go up, with the only difference in config being the location.

douglascamata:
We can't exactly tell how fast the bot is walking, because we make some API calls and have to sleep a of of time, so it's clearly visible that we're NEVER walking at the speed we think to.

Don't think of walking speed as a unit (m/s or km/h), just know that bigger is faster and smaller if slower.

Bashin:
Ok, but the default value is really slow (i get about 4 km/h after a lot of calculating)

douglascamata:
@Bashin 4 km/h is a very human-like walking. We want to provide a default value close to human walking, so it's good.

maestroi:
don't increase default walking speed because in europe i you walk to fast it does not hatch, 4km/h is pretty much a good speed, also there is no use to go faster.

douglascamata:
Closing because the default value should definitely be human-like.

Bashin:
It works for me in europe.. I think you could double the default value without getting any problems and it's still a very human-like value.

maestroi:
normal average walking speed is like 5km/h so would be very strange if u walk like 8 that's pretty quick for walking 6 hours straight especially if you look at a phone screen.

A7F:
i set the speed on 13 and still hatch eggs... i still don't get wether is in km/h or m/s btw works (Europe, Italy)
If the speed is intended in km/h it's still reasonable just because you can move by bike

Bashin:
I think its very reasonable too, but it's ok for me if the value remains as it is (maybe round it to 4 or 5). The wiki should be updated anyway. I'll set walk to 10 in my config as it is working for me an there is still a safety margin up to 14 or 15.
Any chance to get editor rights for the wiki?

PR #1585, opened 2016-07-29
Add a progress bar when walking

klingan:
Short Description: Add a progress bar

Fixes:
- Add a progress bar
- TODO: make scale a config value


douglascamata:
Too much log pollution and you use a global 😞

PR #1588, opened 2016-07-29
Added docker-compose easy provisioning

ogir-ok:
Short Description: 
run it with just `docker-compose up`
Fixes:
- added little changes to Dockerfile
- created docker-compose.yml


maestroi:
don't know docker well but does it run pip as root?
or dus it use env because sometimes i get error on the pgoapi if run without env, it needs root

ogir-ok:
When you use docker, you run all stuff from root. As far as it's isolated, virtualenv is not needed.

maestroi:
@ogir-ok thnx for the clarification, than the change sounds good to me! +1

Dockerfile
```diff
@@ -3,4 +3,5 @@ FROM python:2.7-onbuild
 RUN apt-get update \
     && apt-get install -y python-protobuf
```
NamPNQ: Don't needed, pip install is enought

docker-compose.yml
```diff
@@ -0,0 +1,9 @@
+bot:
+  build: .
+  volumes:
+    - .:/web
+  ports:
+    - "8800:80"
+  command: /bin/bash -c "cd /web/;
```
NamPNQ: Why you need pip install again, it already in dockerfile, i suggest use image `python:onbuild` instead of custom `Dockerfile`

ogir-ok: Requirements can change between your two docker-compose up. 
Rebuild image every time is waste of time.

Dockerfile
ogir-ok: This was in initial Dockerfile. So I don't know if I can remove it

bigkraig: you can remove it, I've been removing it from mine during testing. it runs after pip has run and built everything

docker-compose.yml
bigkraig: I also volume out the configs and data directories. configs so i can inject my config.json and data so if i start a new container i keep my last location

ogir-ok: This does too.
Whole directory is volume

Dockerfile
ogir-ok: Removed

joergpatz:
Generally I would recommend to run a single process inside the docker container to follow docker's (container) [philosophy](https://docs.docker.com/engine/userguide/eng-image/dockerfile_best-practices/#/run-only-one-process-per-container).

And the great use case of docker-compose is to define the services to be started #1669.

ogir-ok:
```
docker exec -it pokemongobot_bot_1 bash                                                      22:12
root@7ee6ee1c686a:/usr/src/app# ps ax                                                                                                             
  PID TTY      STAT   TIME COMMAND
    1 ?        Ss     0:00 /bin/bash -c cd /web/; pip install -r requirements.txt; python pokecli.py
   23 ?        S      0:00 python pokecli.py
   73 ?        Ss     0:00 bash
   79 ?        R+     0:00 ps ax

```

There is just one process

docker-compose.yml
joergpatz: How often do package requirements change? For my point of view I see no sense to do a `pip install` on every container run.

ogir-ok: It changes. And somebody will spend time to find why this is not starting. 
Install latest requirements is always good. 
While person can forget to run `docker-compose build`

Dockerfile
joergpatz: I added this to the Dockerfile because Protobuf 3 library is declared as an installation requirement, see Wiki: [https://github.com/PokemonGoF/PokemonGo-Bot/wiki/Installation#requirements-click-each-one-for-install-guide](https://github.com/PokemonGoF/PokemonGo-Bot/wiki/Installation#requirements-click-each-one-for-install-guide)

ogir-ok: 1. It works without it. 
2. It's in requirements.txt. Probably error is on some platforms without it's installation.

docker-compose.yml
NamPNQ: Easy to update dependencies without rebuild `docker-compose exec bot pip install -r requirements.txt`

gregtampa:

ogir-ok:
Not needed anymore as far as some kind of docker-compose provided in dev

PR #1589, opened 2016-07-29
Fix circle_avoidance

MFizz:
Short Description: 

- spin_forts was assigned twice in config

- fort related arguments are now embedded in `fort` keyword

- added `embedded` flag for parser, in order to handle nested keywords

Fixes:
#1544 
circle avoidance now works as intended

### DIFF
--- a/configs/config.json.example
+++ b/configs/config.json.example
@@ -7,7 +7,6 @@
     "max_steps": 5,
     "catch_pokemon": true,
     "websocket_server": false,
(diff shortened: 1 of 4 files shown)


douglascamata:
This has another issue: inner keys lacks any reference in the same to outter ones and this is bad. I have one that will handle it properly. Will open PR for u to see.

douglascamata:
Have a look #1590

PR #1590, opened 2016-07-29
Consolidate similar meaning configuration keys properly inside another key

douglascamata:


pokecli.py
```diff
@@ -182,10 +182,11 @@ def init_config():
     add_config(
         parser,
         load,
-        long_flag="--spin_forts",
+        long_flag="--spin",
```
TheSavior: wouldn't this be --forts_spin based on your logic below?

douglascamata: hum, let me try something

douglascamata: @TheSavior if in `add_config` I rewrite the long_flag like this:

```
if embedded_in:
    long_flag = "{}_{}".format(embedded_in, long_flag)
```

What you think

TheSavior: What if we scrap all of our cli arguments and instead dynamically let people override anything in the config.json, supporting deeply nested objects like so:

```
--username my_username --forts.avoid_circles false --location_cache true
```

DisasteR: just go for long_flag="--forts_spin" for now.

rewriting all cli arguments management is a little out of this PR scope no ?

TheSavior: Yeah, out of scope. I've been teasing @douglascamata about this change for a few days. It was a proposal from a few days ago that solved some very specific problems that we just never seemed to implement: https://github.com/PokemonGoF/PokemonGo-Bot/issues/573#issuecomment-234757535

TheSavior:
:+1:

TheSavior:
:+1:

PR #1591, opened 2016-07-29
CatchVisiblePokemonWorker now catches pokemon from lures

TheSavior:
Moving this logic out of the SeenFortWorker


douglascamata:
👍

PR #1610, opened 2016-07-29
Fix: response_dict received as none from api

oalmali:
Short Description: 
When api response is none for response_dict it raises typeError so none check is required
Fixes:
- TypeError when response_dict is none
- 
-


douglascamata:
This fix is `pgoapi`'s responsibility. And the bug doesn't happen anymore, lol.

PR #1616, opened 2016-07-29
add ignore_list to config

dhluong90:
Short Description: 

I just want to add ignore_lis to config, when some body don't want to catch a list of pokemons, they can ignore. 

Please teach me if any wrong. 
Thansk.


douglascamata:
`master` is feature frozen until we have a stable version of the `dev`

PR #1647, opened 2016-07-29
allows the use of comments in json config file

mako-taco:


douglascamata:
We already discussed about this with other contributors and we decided not to allow comments in JSON, because it breaks the format and we don't want that. JSON is for pure data.

PR #1664, opened 2016-07-29
Fix transfer worker not triggered for last pokemon

bimahp:
Short Description: 
If there is only one pokemon and there is no duplicate, release validation will skip that pokemon

Fixes:
- Transfer worker not triggered for last pokemon


TheSavior:
:+1:

Simple enough

PR #1666, opened 2016-07-29
the MoveToFortWorker should always go to the nearest fort

bigkraig:
Short Description: 
get_nearest_fort assumes that the list of forts is sorted by distance but its not unless order_by_distance is set to True.

Fixes:
- added order_by_distance=True param to get_forts


TheSavior:
👍 

Whoops.

PR #1671, opened 2016-07-29
Show Pokestop names

cwild:
Restore ability to show Pokestop names

- Configured option, off by default
- Each fort is queried once per run and the result is cached

### DIFF
--- a/pokecli.py
+++ b/pokecli.py
@@ -341,6 +341,8 @@ def init_config():
     config.hatch_eggs = load.get("hatch_eggs", True)
     config.longer_eggs_first = load.get("longer_eggs_first", True)
 
+    config.forts_show_name = load.get('forts', {}).get('show_name', False)
+
     if config.auth_service not in ['ptc', 'google']:
         logging.error("Invalid Auth service specified! ('ptc' or 'google')")
         return None
--- a/pokecli.py
+++ b/pokecli.py
@@ -324,6 +324,15 @@ def init_config():
         type=int,
         default=10,
     )
+    add_config(
+        parser,
+        load,
+        short_flag="-sn",
+        long_flag="--forts.show_name",
+        help="Toggle displaying the name of the Pokestops as you pass by.",
+        type=bool,
+        default=True,
+    )
 
     # Start to parse other attrs
     config = parser.parse_args()
@@ -340,8 +349,6 @@ def init_config():
 
     config.hatch_eggs = load.get("hatch_eggs", True)
     config.longer_eggs_first = load.get("longer_eggs_first", True)
-
-    config.forts_show_name = load.get('forts', {}).get('show_name', False)
 
     if config.auth_service not in ['ptc', 'google']:
         logging.error("Invalid Auth service specified! ('ptc' or 'google')")
--- a/pokecli.py
+++ b/pokecli.py
@@ -300,6 +300,15 @@ def init_config():
     add_config(
         parser,
         load,
+        short_flag="-bf",
+        long_flag="--softban_fix",
+        help="Fix softban automatically",
+        type=bool,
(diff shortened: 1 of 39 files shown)


pokecli.py
```diff
@@ -341,6 +341,8 @@ def init_config():
     config.hatch_eggs = load.get("hatch_eggs", True)
     config.longer_eggs_first = load.get("longer_eggs_first", True)
 
+    config.forts_show_name = load.get('forts', {}).get('show_name', False)
```
douglascamata: Please use the convention of calling `add_config` to set up this. Check how other parameters that are inside the `forts` group are used.

pokemongo_bot/cell_workers/seen_fort_worker.py
```diff
@@ -31,6 +33,16 @@ def work(self):
         lat = fort['latitude']
         lng = fort['longitude']
 
+        if getattr(self.bot.config, 'forts_show_name', False) == True:
```
TheSavior: We don't put default configuration here, only when we parse the configuration. I also happen to think this should be True by default.

pokemongo_bot/cell_workers/seen_fort_worker.py
```diff
@@ -138,3 +150,23 @@ def get_fort_in_range(self):
             return fort
 
         return None
+
+    def get_fort_details(self, fort_id, latitude, longitude):
```
TheSavior: This method that does caching probably shouldn't be here. We have more than one place that looks at fort information, so we should save that data and provide a function that everything can use. We should pull this out into __init__.py, or ideally a new file where we can put methods like this.

cwild: Thanks for the feedback. I will work on this further tomorrow.

cwild:
@TheSavior Where would you prefer I move the caching functionality and fort_details lookup?

TheSavior:
For now, let's pull that caching stuff into the bottom of the __init__.py and we can improve it later.

@douglascamata and I don't actually think this needs to be configurable. Let's just change the default and *always* show the fort name instead of the ID. Should make this simpler. :)

cwild:
@TheSavior Can you let me know your thoughts on the latest commits please?

TheSavior:
@cwild this is looking great. Can you remove the configuration part of this feature? Let's make this the standard behavior. We like this enough that we want to always show the fort name.

pokemongo_bot/cell_workers/catch_lured_pokemon_worker.py
```diff
@@ -30,14 +30,10 @@ def get_lured_pokemon(self):
...
-
-        self.api.fort_details(fort_id=fort['id'],
+        details = fort_details(self.bot, fort_id=fort['id'],
                               latitude=fort['latitude'],
                               longitude=fort['longitude'])
-
-        response_dict = self.api.call()
-        fort_details = response_dict.get('responses', {}).get('FORT_DETAILS', {})
-        fort_name = fort_details.get('name', 'Unknown').encode('utf8', 'replace')
+        fort_name = details.get('name', 'Unknown').encode('utf8', 'replace')
```
TheSavior: You could probably use this same line in seen_fort_worker. You have a try catch there right now.

TheSavior:
Looks like you have merge conflicts. We are so close to getting this merged! :)

douglascamata:
w8ing for merge with dev

Zgrkaralar:
After this merge showing this problem

screen shot 2016-07-30 at 3 51 33 am

cwild:
Apologies for breakage. I see this has been rectified already by @douglascamata , thanks.

PR #1674, opened 2016-07-29
fixes Polyline class to handle a case in which google is returning only one point

th3w4y:
Short Description: 

Fixes:
- Polyline class  should handle the case in which google returning only one point https://github.com/th3w4y/PokemonGo-Bot/issues/27
- PolylineStepWalker walks for only one seconds https://github.com/th3w4y/PokemonGo-Bot/issues/28


douglascamata:
👍

th3w4y:
One more fix
https://github.com/th3w4y/PokemonGo-Bot/issues/28

PR #1682, opened 2016-07-29
[DEV FEATURE] Test framework .... beginning

DayBr3ak:
Short Description: 

Trying to add unit testing. So it's a draft for now, I just used ApiWrapper as an example to show you what I've done and get feedbacks. And more importantly, should I continue with this style ?

I added a .travis.yml entry, I'm not sure if it is going to work, let's see


DayBr3ak:
I get a `can't import pythongo_api.api_wrapper`, trying to fix this right now

douglascamata:
@DayBr3ak mocking everything is not the solution for testing tools that depends on external services. In fact, we will never be able to test properly our code because everything depends a lot on too many complex variables.

DayBr3ak:
@douglascamata  Can't you valid behaviors even with mocked objects ?

douglascamata:
@DayBr3ak you can validate behaviour, but good behaviour is not enough to ensure the bot works. We would need to update mocks when responses changes and it would be a pain in the ass. And we will have more code creating mocks than tests, much more.

DayBr3ak:
That's sad then, can't really test properly with the pokemongo servers. An alternative (which is just a dream) would be a sandbox pokemongo server where we would just run our tests against

Well even if this is not applicable for the bot, small modules could benefit from unit testing like api_wrapper for ex.

douglascamata:
> Well even if this is not applicable for the bot, small modules could benefit from unit testing like api_wrapper for ex.

I agree. :D

DayBr3ak:
@douglascamata I was wondering, I added a test for the StepWalker and I noticed if the distance was too big, it would not "step" at all (i.e stay in place)
Is this intended ?

CapCap:
@douglascamata you can totally mock this up like @DayBr3ak is doing- and this is exactly the way testing should be done. Yeah, it's a pain if the API changes- but that's how you write software that you can be at least somewhat sure works....

DayBr3ak:
Does this have 0 interest? Should I drop it?

douglascamata:
@DayBr3ak I do have interest. But we're moving to behaviour trees in a few days and a lot will change.... good thing is that we will be able to test behaviour separately and it will be great. Do you mind w8ing for it to help us?

DayBr3ak:
@douglascamata I don't mind at all.

DayBr3ak:
I rebased to current dev branch.
I also suppressed the travis entry for vanilla python unittest, as `nosetest` find those tests automatically.

Should I test in the file like did @TheSavior or I can keep my test classes in the ./tests folder

TheSavior:
I haven't gone through this code very much, just a cursory glance. My thoughts:

I think there is a lot we can test in this bot. For example, as you say, the steppers can be tested that they walk an expected path and call the api to update location with the right data. 

In regards to things that parse api responses, I think those functions should be pulled out so that other parts of the bot can use them without having to duplicate that logic. Those individual functions can never truly know if they will work when they are running, as @douglascamata is concerned about. However, that is a class of things that will always have problems. I think we can get some value out of having tests for those types of functions. If the api response *does* change and it fails in production, we can get the new response and update our test to make sure it works in all the cases. 

In regards to unittest and what I have now, I would *definitely* not assume that I know what I'm doing in python. Haha. But we should make sure we are using the same test framework for everything. If `nosetest` is the right approach, we should change the tests that I wrote to use that. I also don't particularly have any thoughts on where the files should live.

All that being said, I defer this pull request to @douglascamata's approval, he is significantly more familiar with python and some of the choices here than I am.

DayBr3ak:
@TheSavior I got rid of nosetest, I jumped on it because it seemed nice, but in retrospective, most people will be familiar with unittest rather than nosetest.

TheSavior:
Sounds good.

I think it is typically considered bad form to change app code when adding tests. It makes it hard to validate the changes. Is it possible for you to add these tests without actually changing the code under test?

DayBr3ak:
I'm on it !

DayBr3ak:
One of the reason I changed app code is one of my test was failing because of a "bug" that doesn't seem to appear "yet" in the app code. So I was raising a RuntimeError when that was happening. Now that test is in auto skip. I'll make a pr if this code is merged

run_tests.py
```diff
@@ -0,0 +1,7 @@
+#!/usr/bin/env python
+import unittest
+
+loader = unittest.TestLoader()
+tests = loader.discover(start_dir='.', pattern='*_test.py')
```
TheSavior: I think this can be done with command line flags. I'm not sure of best practices.

tests/api_wrapper_test.py
```diff
@@ -0,0 +1,108 @@
...
+        self.api.get_inventory(test='awesome')
+        self.api.fort_details()
+
+        good_return_value = {'responses': {'GET_INVENTORY': {}, 'FORT_DETAILS': {}}, 'status_code': 0}
+        self._api.call = MagicMock(return_value=good_return_value)
+
+        result = self.api.call()
+        self.assertEqual(result, good_return_value)
+
+    @unittest.skipIf(SKIP_TIMED, "Please install module 'timeout_decorator'")
```
TheSavior: I'm not sure why we skip these and have the if/else on whether the module is installed. If it is in `requirements.txt` then it will be installed by travis and thus exist

run_tests.py
DayBr3ak: I was thinking of this because it's easier to tweak in the future, but sure I can revert

tests/api_wrapper_test.py
DayBr3ak: oops you didn't see that wrong branch !

TheSavior:
This is looking much better

DayBr3ak:
I have a test failing which is directly related to #1825. Can we merge this pr ? @douglascamata @TheSavior . Thanks :)

douglascamata:
@DayBr3ak just merged #1825

DayBr3ak:
@douglascamata  thanks !

TheSavior:
:+1:

douglascamata:
👍

PR #1683, opened 2016-07-29
Making the SpiralNavigator a worker

TheSavior:
This will make it easier for people to configure what navigation scheme they want. 

The list of workers will end up being defined in people's configuration files once we get the behavior tree stuff. People will be able to put whatever they want in whatever order they want. 

And they we can support people creating new repos that export a plugin that we can load and they can use those additional custom workers!


douglascamata:
@TheSavior you're failing pylint, you criminal!

th3w4y:
Spiral Navigator  does not return WorkerResult.RUNNING to the tick...?

PR #1688, opened 2016-07-29
Added Caching Framework

demgod:
Short Description: Rough framework for caching, further iterations will place objects in their own databases that are maintained for the session.

Caches responses (i.e. GET_MAP_OBJECTS) and stores the resulting json data in a CacheObject for later use. For example:
`player_cache = Cache.get('GET_PLAYER')`
`print(player_cache.data) # prints the json data associated with the GET_PLAYER response`
`print(player_cache.time_stamp) # prints the unix time stamp of when we received this data`
`print(player_cache.age()) # prints the number of seconds that have passed since we received it`
`print(player_cache.is_outdated(60)) # return true if its age is > 60 seconds, otherwise false`
`

Fixes:
- Needlessly requesting updated information on Inventory, Pokemon, Gyms, Forts, etc. when nothing has changed.

### DIFF
--- a/pokemongo_bot/cache.py
+++ b/pokemongo_bot/cache.py
@@ -1,0 +1,50 @@
+import logger
+import time
+
+class CacheObject(object):
+    
+    def __init__(self, name, data):
+        self.name = name
+        self.data = data
+        self.time_stamp = int(time.time())
+        
+    def age(self):
+        return int(time.time()) - self.time_stamp
+        
+    def is_outdated(self, max_age=60):
+        return max_age <= self.age()
+        
+class Cache(object):
+    
+    database = {}
+    
+    @staticmethod
+    def process_response(response):
+        responses = response['responses']
+        for type in responses:
+            Cache.set(type, responses[type])
+    
+    @staticmethod
+    def get(type):
+        return Cache.database[type]
+        
+    @staticmethod
+    def set(type, data):
+        Cache.database[type] = CacheObject(type, data)
+        
+    @staticmethod
+    def remove(type):
+        del Cache.database[type]
+        
+    @staticmethod
+    def clear():
+        Cache.database = {}
+        
+    @staticmethod
+    def print_ages():
+        db = Cache.database
+        for type in db:
+            cache_object = db[type]
+            logger.log("[CACHE] {} - {}".format(type, cache_object.age()), 'yellow')
+            
+    
--- a/pokemongo_bot/cache.py
+++ b/pokemongo_bot/cache.py
@@ -3,8 +3,8 @@ import time
 
 class CacheObject(object):
     
-    def __init__(self, name, data):
-        self.name = name
+    def __init__(self, response_type, data):
+        self.response_type = response_type
         self.data = data
         self.time_stamp = int(time.time())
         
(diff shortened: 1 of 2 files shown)


pokemongo_bot/cache.py
```diff
@@ -0,0 +1,50 @@
...
+        return max_age <= self.age()
+        
+class Cache(object):
+    
+    database = {}
+    
+    @staticmethod
+    def process_response(response):
+        responses = response['responses']
+        for type in responses:
```
douglascamata: find a better name for this, `type` is a built-in function.

douglascamata:
@demgod explain better how to system works. Sounds good but... how the hell does it work?

pokemongo_bot/cache.py
demgod: How do you feel about name?

douglascamata: response_type

douglascamata:
In fact, we just need a system to manipulate the inventory, I mean:

- make just one inventory call per tick
- provide the workers with an object to update the inventory as they use or drop items
- profit!

DayBr3ak:
That's an awesome idea @douglascamata

demgod:
In theory, since we're always provided in the response with items we receive, we should only ever request the inventory once and then modify the inventory as need when we expend or receive items. Additionally a cache for forts, gyms, and pokemon would allow for referencing game objects that we've seen in the past. For example, it currently seems like we request updated fort/gym information just about every second.

douglascamata:
@demgod you're not going to the root of the problem, you're just adding some "make-up". 

I already found that we do one `GET_GYM_DETAILS` and one `FIND_MAP_OBJECTS` every tick to send info to web ui (won't be needed with my event system already merged). 

One `GET_PLAYER`, `GET_INVENTORY` and `CHECK_AWARDED_BADGES` is just the heartbeat, which NEEDS to fire these requests. 

Inventory we already cache in a different way and update the cache throughout workers.

demgod:
@douglascamata I get that it needs a heartbeat, but I can't image the application does it that frequently. Considering the stress already on their server I'd imagine they try to optimize as many requests as they can (haven't checked it out, just basing the notion on basic design philosophy). The root of the problem as I see it is that there is currently no uniformed way for workers/modules to share requested data between each other. If worker A needs to perform request x and worker b needs the same data it has to perform the request a second time completely unnecessarily.

douglascamata:
@demgod there is shared data. We access inventory in many others, even tho we don't many repeated inventory calls. The heartbeat are the same requests the phone does....

klingan:
@douglascamata we can still do the heartbeat but ignore the data it returns for objects that are supposed to be static like forts. This will prevent bugs with e.g. walking if the API stutters.

demgod:
@douglascamata I know the calls are the same, what I'm saying is the frequency is different. For example, we request the full inventory more frequently than the application. As far as the shared data is concerned the inventory system works fine, for inventory related information. My point with this is that instead of making data structures for each individual request type, centralize it into one generic cache system and build off of that. So everything uniformly draws from one location.

TheSavior:
It sounds like we are taking a different approach to inventory caching. For more information, see this PR. Thanks for contributing!

PR #1698, opened 2016-07-29
Revert "Making the SpiralNavigator a worker"

douglascamata:
Reverts PokemonGoF/PokemonGo-Bot#1683

It causes a walking loop because we are creating a new instance of the navigator every single time and it doesn't work like that.


th3w4y:
was it because Spiral Navigator was not returning WorkerResult.RUNNING to the tick...?

PR #1720, opened 2016-07-29
Decrease logging spam by showing walker logs temporarily in console

klingan:
Short Description: Show walking logs only temporarily at the bottom line to prevent a log spammed with this. 

Fixes:
- Walking logs do not spam the logfile anymore

### DIFF
--- a/pokemongo_bot/cell_workers/move_to_fort_worker.py
+++ b/pokemongo_bot/cell_workers/move_to_fort_worker.py
@@ -56,7 +56,7 @@ class MoveToFortWorker(object):
             if not step_walker.step():
                 return WorkerResult.RUNNING
 
-	logger.log('Arrived at pokestop.')
+        logger.log('Arrived at pokestop.')
         return WorkerResult.SUCCESS
 
     def get_nearest_fort(self):
--- a/pokemongo_bot/cell_workers/move_to_fort_worker.py
+++ b/pokemongo_bot/cell_workers/move_to_fort_worker.py
@@ -1,3 +1,4 @@
+import sys
 from utils import distance, format_dist, i2f
 from pokemongo_bot.constants import Constants
 from pokemongo_bot.human_behaviour import sleep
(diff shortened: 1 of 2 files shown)


douglascamata:
@klingan this makes very hard to spot bugs like this: https://github.com/PokemonGoF/PokemonGo-Bot/issues/1715

klingan:
Correct, seen that happen too when the distance suddenly went up again. 
But easier to spot other bugs that get hidden because of tons of walking loglines ;)

Might be a solution in the middle to add an extra logline when the bots starts moving to a new fort?

douglascamata:
@klingan can u keep at least 5 last movements in the bottom?

klingan:
Not sure if that's easy to make
What I did now is : 

[00:37:00] Moving to new fort: 0af0ee29e6644b25994f52dfc1bd7819.12
Moving towards fort 0af0ee29e6644b25994f52dfc1bd7819.12, 40.82m left



[00:37:00] Moving to new fort: 0af0ee29e6644b25994f52dfc1bd7819.12
[00:37:16] Arrived at pokestop.
[00:37:16] Lured pokemon at fort f9667177ec9744909531dc86f2303ae0.16

It will log the Moving to new fort whenever the fortID changes during walking

douglascamata:
@TheSavior please review this (I think you won't like....)

douglascamata:
@TheSavior you think we should create a Storage class for workers that need to storage some data for next tick?

pokemongo_bot/cell_workers/move_to_fort_worker.py
```diff
@@ -31,6 +32,10 @@ def work(self):
         lat = nearest_fort['latitude']
         lng = nearest_fort['longitude']
         fortID = nearest_fort['id']
```
douglascamata: please, `fort_id` instead of `forID`.

klingan: Not my code but I will update it anyway :+1:

TheSavior:
I haven't run this, so I'm surely ignorant here, but why wouldn't this have the same problem as we had with the progress bar? Where if something else happens, then it prints at the end of the previous line and we get weird log results?

douglascamata:
@klingan upload a screenshot showing before/after your changes, please?

klingan:
I can make a screencast if you want?

douglascamata:
@klingan even better!

klingan:
[http://screencast.com/t/JsLTQfXSWQ](http://screencast.com/t/JsLTQfXSWQ)

TheSavior:
Thanks for that screencast. I think there is a lot of value right now in having these in the logs. Especially since we have been iterating so much on these walkers. I might be more up for this in the future if user's could configure which kinds of messages they want to see. I think that is out of scope for now though. 

I'm a 👎  on this for now.

douglascamata:
@klingan I'm going to start work on event system usage for logs right now we will have properly using `logging ` for multiple level. You will be able to make the bot more silent if you want.

TheSavior:
This might make more sense now that we have the event system. If you feel like there is still work that needs to be done, please open another PR.

PR #1724, opened 2016-07-29
SoftBan Worker

douglascamata:
Fixes SoftBan!

### DIFF
--- a/pokemongo_bot/cell_workers/pokemon_catch_worker.py
+++ b/pokemongo_bot/cell_workers/pokemon_catch_worker.py
@@ -122,6 +122,7 @@ class PokemonCatchWorker(object):
                                 else:
                                     if response_dict['status_code'] is 1:
                                         logger.log('Fail to use berry. Seem like you are softbanned.', 'red')
+                                        self.bot.softban = True
                                     else:
                                         logger.log('Fail to use berry. Status Code: {}'.format(response_dict['status_code']),'red')
 
@@ -171,6 +172,8 @@ class PokemonCatchWorker(object):
                                 if status is 3:
                                     logger.log(
                                         'Oh no! {} vanished! :('.format(pokemon_name), 'red')
+                                    if success_percentage == 100:
+                                        self.softban = True
                                 if status is 1:
                                     self.bot.metrics.captured_pokemon(pokemon_name, cp, iv_display, pokemon_potential)
 
@@ -180,6 +183,7 @@ class PokemonCatchWorker(object):
                                         pokemon_potential,
                                         iv_display
                                     ), 'blue')
+                                    self.bot.softban = False
 
                                     if (self.config.evolve_captured
                                         and (self.config.evolve_captured[0] == 'all'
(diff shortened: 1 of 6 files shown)


pokemongo_bot/cell_workers/soft_ban_worker.py
```diff
@@ -0,0 +1,59 @@
...
+
+    def work(self):
+        if not self.should_run():
+            return
+
+        forts = self.bot.get_forts(order_by_distance=True)
+
+        if len(forts) == 0:
+            logger.log('Found no forts to reset softban, skipping...', 'red')
+            return
```
TheSavior: The intent is that it would continue to the next worker if it doesn't have a fort available?

pokemongo_bot/cell_workers/soft_ban_worker.py
```diff
@@ -0,0 +1,59 @@
...
+        else:
+            logger.log('Starting 50 spins...')
+            for i in xrange(50):
+                if (i + 1) % 10 == 0:
+                    logger.log('Spin #{}'.format(str(i+1)))
+                self.spin_fort(forts[0])
+            self.softban = False
+            logger.log('Softban should be fixed.')
+
+    def spin_fort(self, fort):
```
TheSavior: This could get pulled out. Seems like there are a couple of places that would be able to use this function.

TheSavior:
:+1:

pokemongo_bot/cell_workers/soft_ban_worker.py
douglascamata: Yes, because we need to walk randomly until we can find a fort to walk to.

pokemongo_bot/cell_workers/soft_ban_worker.py
douglascamata: Definitely we should start service classes to wrap some API functions. I created #1726 for us to keep track of this.

solderzzc:
👍

MatthewARoy:
Excellent, excited for this feature.

pokemongo_bot/cell_workers/pokemon_catch_worker.py
```diff
@@ -185,6 +186,8 @@ def work(self):
                                 if status is 3:
                                     logger.log(
                                         'Oh no! {} vanished! :('.format(pokemon_name), 'red')
+                                    if success_percentage == 100:
+                                        self.softban = True
```
wilsonyqm: @douglascamata , I am not sure this works. It's better be `self.bot.softban = True`

pokemongo_bot/cell_workers/soft_ban_worker.py
```diff
@@ -0,0 +1,61 @@
...
+            if forts[0]['id'] in self.bot.fort_timeouts:
+                del self.bot.fort_timeouts[forts[0]['id']]
+            return WorkerResult.RUNNING
+        else:
+            logger.log('Starting 50 spins...')
+            for i in xrange(50):
+                if (i + 1) % 10 == 0:
+                    logger.log('Spin #{}'.format(str(i+1)))
+                self.spin_fort(forts[0])
+            self.softban = False
```
wilsonyqm: This too.

JohnnyOmm:
Error in softban_worker

`cannot import name MoveToFortWorker`

MikeDX:
You disappoint me @douglascamata  - you took the easy way out with this. Was it because doing it with one spin per step required too many changes / too difficult to implement? I could have quite easily done it this way in a much shorter time - but you have implemented this in almost exactly the same way (except being in a worker) to my original commit.

I also see this pull request has several errors which other people have reported.

I really don't see how you can approve shoddy hacks like this, and not eat your own dog food. 

0/10

PR #1733, opened 2016-07-29
Updated item_filter in config.json.example to use item names instead item id's

Astralestialux:
Short Description: 
Someone in the slack channel hit a full inventory, asked how to throw away items. This makes it more obvious in the config, as item_filter in config.json.example is now more human readable by using item names instead of id's

Fixes:
- config.json.example updated
- code to deal with changes in recycle_items_worker.py


configs/config.json.example
```diff
@@ -22,12 +22,12 @@
     "distance_unit": "km",
     "reconnecting_timeout": 15,
     "item_filter": {
-      "1": { "keep" : 100 },
-      "101": { "keep" : 10 },
-      "102": { "keep" : 30 },
-      "103": { "keep" : 30 },
-      "201": { "keep" : 30 },
-      "701": { "keep" : 100 }
+      "Pokeball":       { "keep" : 100 },
```
TheSavior: We should raise an exception if there is something in this list that isn't a known item. Otherwise people will have a typo here. Raising that exception will save us having to answer lots of support issues.

Astralestialux: Added a check before it even connects to server (for faster feedback), hence the reason for the json.load

Astralestialux: fixing commit..

pokecli.py
```diff
@@ -394,6 +394,12 @@ def init_config():
         parser.error("--catch_randomize_spin_factor is out of range! (should be 0 <= catch_randomize_spin_factor <= 1)")
         return None
 
+    # item list config verification
+    item_list = json.load(open(os.path.join('data', 'items.json')))
+    for config_item_name, bag_count in config.item_filter.iteritems():
+        if config_item_name not in item_list.viewvalues():
+            raise Exception('Config.json error - item "' + config_item_name + '" does not exist (spelling mistake?)')
+
```
Astralestialux: Okay it has the check very early on before server connection (for fast feedback) hence the reason for the json.load

Astralestialux: Fixed conflicts.

TheSavior:
It looks like you lost the changes to recycle_items_worker

pokecli.py
TheSavior: Great. This should probably use 

```
parser.error("YOUR STRING")
return None
```

instead of raising though.

TheSavior:
:+1:

douglascamata:
@Astralestialux please, verify somehow that the item name is not a number to help ppl migrate their configs.

TheSavior:
@douglascamata He already has an error message if one of the items isn't a string that matches an item in the items.json file

douglascamata:
👍

PR #1751, opened 2016-07-30
Don't try to release pokemons in forts

Nihisil:
Without this check transfer worker will try to release pokemon in fort each tick (without success), so it is not right behaviour.

### DIFF
--- a/pokemongo_bot/cell_workers/pokemon_transfer_worker.py
+++ b/pokemongo_bot/cell_workers/pokemon_transfer_worker.py
@@ -93,6 +93,11 @@ class PokemonTransferWorker(object):
                 continue
 
             pokemon_data = pokemon['inventory_item_data']['pokemon_data']
+
+            # pokemon in fort, so we cant transfer it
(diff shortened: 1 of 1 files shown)


douglascamata:
@Nihisil nice one, dude! I'd would give you 10 👍 if I had 10 thumbs.

douglascamata:
@Nihisil shouldn't this verification be present in every release method tho?

Nihisil:
@douglascamata not sure about it. Currently I'm checking this in the method that build a list of pokemons that should be transfered. And we have release logic only in transfer worker, so it not should be a problem.

Actually we have one more release call in the evolve worker `self.api.release_pokemon(pokemon_id=pid)`, but this shit should be refactored.

douglascamata:
@Nihisil but is it the only method used to get pokemon to release?

Nihisil:
@douglascamata yep. We build list of pokemons to release in `_release_pokemon_get_groups` (in this place I have added the check). Later we release pokemons by conditions from this list.

In other part of project we don't have release logic, except evolve worker.

douglascamata:
Ok ok. 👍

PR #1756, opened 2016-07-30
[FEATURE DEV] Runtime Destination change though a json config file

asaf400:
# Feature: Runtime Destination change though json.config file #

New destination.json in config directory, 
when it has a value, the bot will walk towards the destination (respects global speed)


pokemongo_bot/__init__.py
```diff
@@ -28,6 +28,22 @@
 
 
 class PokemonGoBot(object):
+
+    WORKERS = [
```
Nihisil: I think it wasn't properly merged.

They just have removed `WORKERS` https://github.com/PokemonGoF/PokemonGo-Bot/commit/5eef040f2b31f1ab70d106f39165610c03f7542c

PR #1762, opened 2016-07-30
Fixed missing location for player on map

Stasmo:
Short Description: 

Fixes:
- Player was missing from map because location files weren't being written unless they already existed.


surfaace:
this is already fixed in dev

PR #1768, opened 2016-07-30
Runtime Destination change though json.config file

asaf400:
# Feature: Runtime Destination change though json.config file #

New destination.json in config directory, 
when it has a value, the bot will walk towards the destination (respects global speed)

## Config File: ##

made get_pos_by_name a public function
Also made a PEP8 Fix for the Constants

### DIFF
--- a/pokemongo_bot/__init__.py
+++ b/pokemongo_bot/__init__.py
@@ -36,6 +36,8 @@ class PokemonGoBot(object):
         cell_workers.EvolveAllWorker,
         cell_workers.RecycleItemsWorker,
         cell_workers.CatchVisiblePokemonWorker,
+        cell_workers.ShouldMoveToPositionWorker,
+        cell_workers.MoveToPositionWorker,
         cell_workers.SeenFortWorker,
         cell_workers.MoveToFortWorker,
         cell_workers.CatchLuredPokemonWorker,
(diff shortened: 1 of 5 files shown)


ewold:
👍 Great, just what i needed. Hope this gets merged soon.

TheSavior:
The work that @douglascamata is to use websockets to support manual intervention of the bot, such as changing location like this. It will also support all sorts of other configuration and being able to be manipulated by a web ui. My guess is that is the right generic approach for this, but I'll let @douglascamata provide his thoughts.

asaf400:
@TheSavior Yes, I am aware of that, Note that my code includes 2 workers.

1st worker trying to parse the config json file each tick,
If it detects a proper location, it sets Bot's global instance variable 'destination_location' to the Lat,Lng

2nd worker is listening for changes in the bot's variable 'destination_location', and walks there with the walking speed configured for the bot, and also, because the way workers are 'working' by an array order, it catches pokemon along the way so this can be used regardless of the 1st worker

k4n30:
@asaf400 this merge is severly outdated. How do you want to proceed?

TheSavior:
@douglascamata How do we want to proceed with this based on the socket server?

asaf400:
I Can rebase the changes to latest dev, and refacot the code to make it compatible with the new event log \ tasks which are the major changes made to the code since this PR

solderzzc:

PR #1769, opened 2016-07-30
fixes move distance to be related to time elapsed

### LINKED ISSUE #1586: [DEV] Slow walking speed
### Expected Behavior
Bot walks like a normal human

### Actual Behavior
```
[05:23:52] Moving towards fort 41fa8df35089444994e59fcadcd3d080.16, 0.04km left
[05:24:42] Moving towards fort 41fa8df35089444994e59fcadcd3d080.16, 0.04km left
[05:25:32] Moving towards fort 41fa8df35089444994e59fcadcd3d080.16, 0.03km left
[05:26:22] Moving towards fort 41fa8df35089444994e59fcadcd3d080.16, 0.03km left
[05:27:13] Moving towards fort 41fa8df35089444994e59fcadcd3d080.16, 0.02km left
[05:28:02] Moving towards fort 41fa8df35089444994e59fcadcd3d080.16, 0.02km left
```
Which I assume is crawling (0,468km/h)...

### Steps to Reproduce
From the config:
```
"walk": 4.46,
"distance_unit": "km",
```

### Other Information
OS: Ubuntu 14.02
Git Commit: a1433f235c0afc00a15d79ff99711c0cd520dc7f

mako-taco:
Short Description: 
The problem with walking speed is that we're not taking into account time elapsed since the last step of the stepper.

I also simplified the math of step_walker a little bit, because we had calculated these nice components to add to lat/lng, but then recalculated them each step.  

A new StepWalker is also no longer created each tick of the state machine, and instead is recycled.

Fixes #1586

### DIFF
--- a/pokemongo_bot/step_walker.py
+++ b/pokemongo_bot/step_walker.py
@@ -34,9 +34,6 @@ class StepWalker(object):
             self.dLat = (dest_lat - init_lat) / int(self.steps)
             self.dLng = (dest_lng - init_lng) / int(self.steps)
 
-    def stop(self):
-        self.lastTime = None
-
     def step(self):
         # Since speed is m/s, we need to scale up our distance to move by
         # the number of seconds elapsed since we last moved
@@ -46,15 +43,22 @@ class StepWalker(object):
             time_scale = current_time - self.lastTime
             self.lastTime = current_time
 
-        if (self.dLat == 0 and self.dLng == 0) or self.dist < self.speed:
+        if (self.dLat == 0 and self.dLng == 0) or self.dist < (self.speed * time_scale):
             self.api.set_position(self.destLat, self.destLng, 0)
             self.lastTime = None
             return True
 
-        cLat = self.bot.position[0] + (self.dLat * time_scale) + random_lat_long_delta()
-        cLng = self.bot.position[1] + (self.dLng * time_scale) + random_lat_long_delta()
+        c_lat = self.bot.position[0] + (self.dLat * time_scale) + random_lat_long_delta()
+        c_lng = self.bot.position[1] + (self.dLng * time_scale) + random_lat_long_delta()
 
-        self.api.set_position(cLat, cLng, 0)
+        self.api.set_position(c_lat, c_lng, 0)
+        self.dist = distance(
+            c_lat,
+            c_lng,
+            self.destLat,
+            self.destLng
+        )
+
         self.lastTime = current_time
         self.bot.heartbeat()
 
--- a/pokemongo_bot/step_walker.py
+++ b/pokemongo_bot/step_walker.py
@@ -10,11 +10,11 @@ class StepWalker(object):
         self.bot = bot
         self.api = bot.api
 
-        self.initLat, self.initLng = self.bot.position[0:2]
+        init_lat, init_lng = self.bot.position[0:2]
 
         self.dist = distance(
-            self.initLat,
-            self.initLng,
+            init_lat,
+            init_lng,
             dest_lat,
             dest_lng
         )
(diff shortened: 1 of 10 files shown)


pokemongo_bot/step_walker.py
```diff
@@ -24,36 +24,42 @@ def __init__(self, bot, speed, dest_lat, dest_lng):
...
 
-        totalDLat = (self.destLat - self.initLat)
-        totalDLng = (self.destLng - self.initLng)
-        magnitude = self._pythagorean(totalDLat, totalDLng)
-        unitLat = totalDLat / magnitude
-        unitLng = totalDLng / magnitude
-
-        scaledDLat = unitLat * self.magnitude
-        scaledDLng = unitLng * self.magnitude
+        c_lat = self.bot.position[0] + (self.dLat * time_scale) + random_lat_long_delta()
```
th3w4y: totally off calculation you are walking with 60km/s

```
In [54]: %cpaste
Pasting code; enter '--' alone on the line to stop or use Ctrl-D.
:import time
import haversine
orig_lat, orig_lng = 0.1700271, 0.518072999999998
dest_lat, dest_lng = 0.1800271, 0.518072999999998
time_scale = 1
lastTime = time.time()
time.sleep(6)
time_scale = time.time() - lastTime
time_scale
haversine.haversine((orig_lat, orig_lng),(dest_lat, dest_lng))
c_lat = orig_lat + dest_lat* time_scale
c_lng = orig_lng + dest_lng* time_scale
haversine.haversine((orig_lat, orig_lng),(c_lat, c_lng))::::::::::::
:--
Out[54]: 366.1980195401676

In [55]: 
```

mako-taco:
The couple other places where `StepWorker` is used should probably be updated, but this won't break them (they just wont get any of the benefits)

buettner123:
applied this as an patch to my local dev branch. it works much smoother now and xp/h just doubled from 2000 to 4000 (which is still not the best but much better than before :) )

mako-taco:
that's really low... I get about 18k/hour in a densely populated area

th3w4y:
What i like the most is that StepWalker is no longer instantiated... move_to_fort.step() on every bot.tick()
This is exactly what is needed (or similar) for the PolylineWalker()

skorokithakis:
@mako-taco Could that be because of your config? Can you post it?

CapCap:
@mako-taco Can you resolve the conflicts? This is great

mako-taco:
Working on getting a similar concept working in spiral nav, then i'll merge in dev

xc-mezcal:
Cool patch!

skorokithakis:
Can we get this merged ASAP? Currently walking is completely broken (bot stays in the same place).

mako-taco:
Talked with @douglascamata and he likes walking how it is. For those of you using this branch, I'd recommend checking out master and multiplying your walk speed by 10x.  Then just hope that no one improves the speed of individual workers, because its going to dramatically change how you walk.

skorokithakis:
The problem here is that walking is completely broken, the Bot walks randomly or stays in place. I was hoping this PR would fix that.

skorokithakis:
Also, @douglascamata, I thought that this PR means that walking currently reports a wildly wrong speed, is that not the case? If I specify 12 km/h and the actual speed is 1.2 km/h, that's not a matter of preference, it's objectively a bug.

mako-taco:
the `walk` variable isn't really a speed, it's a distance which is walked every tick. the ticks happen at non-regular intervals, so while at one point, when the bot was doing less, `walk` was _almost_ in m/s., but now that the bot does so much between movements, it  doesnt really represent a speed anymore. So don't think of it as 12km/h or 12m/s, just 12 meters per step.

skorokithakis:
Ah, that's too bad. As you say, it should take into account the time that's passed since the last step and multiply by the speed to come up with the distance. I guess I'll adjust to taste...

PR #1770, opened 2016-07-30
Optimized berry_count in Pokemon_catch_worker

xc-mezcal:
Short Description: 
Every single try, worker will call the get inventory api to check berry amount. This should be reduced.

Fixes:
Move the berry_count check before the while and use a local variable to keep track of it
- 
-

### DIFF
--- a/pokemongo_bot/cell_workers/move_to_fort_worker.py
+++ b/pokemongo_bot/cell_workers/move_to_fort_worker.py
@@ -2,7 +2,7 @@ from pokemongo_bot import logger
 from pokemongo_bot.constants import Constants
 from pokemongo_bot.step_walker import StepWalker
 from pokemongo_bot.worker_result import WorkerResult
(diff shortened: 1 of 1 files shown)


xc-mezcal:
I also want to purpose to updating the catch pokemon logic of the bot. Maybe add a feature to capture VIP pokemon

pokemongo_bot/cell_workers/pokemon_catch_worker.py
```diff
@@ -77,6 +77,8 @@ def work(self):
                             return False
 
                         items_stock = self.bot.current_inventory()
+                        berry_id = 701  # @ TODO: use better berries if possible
+                        berries_count = self.bot.item_inventory_count(berry_id)
```
kbinani: Possibly `berries_count = items_stock[berry_id]` should also work here

Issue #1752, opened 2016-07-30
Popping a lucky egg before evolving all


Donwulff:
### Expected Behavior
Bot pops lucky egg before starting to evolve pokemon.

### Actual Behavior
Bot doesn't pop lucky egg before starting to evolve pokemon.

### Steps to Reproduce
Set...
"evolve_all": "NONE",
"use_lucky_egg": false,
Run bot

### Other Information
dev branch

### Fix:
evolve_all_worker.py:
```python
# Evolve all is used - Don't run after the first tick or if the config flag is false
        if self.bot.tick_count is not 1 or not self.bot.config.use_lucky_egg:
            return True
```
tick_count is increased by 1 before running each worker, therefore the first tick is 1, not 0.

codybaldwin:
Thank you, a recent change to when the tick count is incremented broke this code as you pointed out. I'll submit a PR for the fix.

PR #1774, opened 2016-07-30
[FIX] use_lucky_egg

codybaldwin:
Short Description: 

Fixes:
- Issue: #1752 
- Updated the tick_count check to account for the changes to when the tick_count is incremented
-  use_lucky_egg will now work

Short Description: 
 Since the tick_count is incremented prior to running the workers, the
tick_count will be 1 on the first tick


leovarmak:
@codybaldwin When can we use this lucky_egg feature on Master ?

PR #1775, opened 2016-07-30
fix #1766

### LINKED ISSUE #1766: dev branch crash in incubate_eggs_worker
Expected behaviour is that the bot doesn't crash. :) Actual behaviour:
[06:05:41] Moving towards fort 71db06db5f9447bba4a35b33d3a2eaf2.16, 0.47km left
[06:05:46] Moving towards fort 71db06db5f9447bba4a35b33d3a2eaf2.16, 0.46km left
  File "pokecli.py", line 450, in <module>
    main()
  File "pokecli.py", line 68, in main
    bot.tick()
  File "C:\Miniconda2\envs\PokemonGo-Bot-dev\pokemongo_bot\__init__.py", line 84, in tick
  File "C:\Miniconda2\envs\PokemonGo-Bot-dev\pokemongo_bot\cell_workers\incubate_eggs_worker.py", line 29, in work
    self._hatch_eggs()
  File "C:\Miniconda2\envs\PokemonGo-Bot-dev\pokemongo_bot\cell_workers\incubate_eggs_worker.py", line 134, in _hatch_eggs
    for pokemon in pokemon_data:
UnboundLocalError: local variable 'pokemon_data' referenced before assignment

Just above it looks to be try-except statement for assigning to pokemon_data. It's been over a decade since I've had to know Python intimately, so I'm not certain which circumstances could cause it to become unassigned. But it seems like either that except-branch needs to skip the whole following code, or the try branch needs to assign to temporary and assign it over to pokemon_data only if that succeeds.

CapCap:
Short Description: 
dont use variables before you declare them

Fixes:
- #1766

### DIFF
--- a/pokemongo_bot/cell_workers/incubate_eggs_worker.py
+++ b/pokemongo_bot/cell_workers/incubate_eggs_worker.py
@@ -127,6 +127,7 @@ class IncubateEggsWorker(object):
         xp = result.get('experience_awarded', 0)
         sleep(self.hatching_animation_delay)
         self.bot.latest_inventory = None
+        pokemon_data = []
         try:
             pokemon_data = self._check_inventory(pokemon_ids)
         except:


Nihisil:
It is not fix, because `pokemon_data` not suppose to be empty.

So, something crashed inside `def _check_inventory()` and we need to fix it.

Nihisil:
So, we need to find out why it crashed and fix it.

Also, as back up plan just in case we need to set `pokemon_data = []` as you did.

CapCap:
yes, absolutely there is another problem: but there's no reason to try/catch if youre guaranteed to fail anyway

Nihisil:
https://github.com/PokemonGoF/PokemonGo-Bot/pull/1777 more global fix

PR #1777, opened 2016-07-30
Egg Incubation - IV fix and UnboundLocalError fix

reddivision:
Short Description: 
Some small fixes to incubate eggs worker for egg hatching, as well as egg hatching detection
Fixes:
- IVs out of order compared to webui - they now match A/D/S format (added calculated decimal value as well for consistency)
Sample of new output with cp %
- mitigation for pokemon_data sometimes not initialized due to bad `_check_inventory` call
- custom log message and early return for conditions where hatched pokemon not received from api or identified
- added temp lists to check inventory call to mitigate incubator list duplication seen for some users

PR #1802, opened 2016-07-30
fix UnboundLocalError

meska:
Short Description: 

Fixes:
- 
- 
- 

UnboundLocalError: local variable 'pokemon_data' referenced before assignment

### DIFF
--- a/pokemongo_bot/cell_workers/incubate_eggs_worker.py
+++ b/pokemongo_bot/cell_workers/incubate_eggs_worker.py
@@ -130,7 +130,7 @@ class IncubateEggsWorker(object):
         try:
             pokemon_data = self._check_inventory(pokemon_ids)
         except:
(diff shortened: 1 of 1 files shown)


reddivision:
this is already being fixed: https://github.com/PokemonGoF/PokemonGo-Bot/pull/1777

pokemongo_bot/cell_workers/incubate_eggs_worker.py
```diff
@@ -130,7 +130,7 @@ def _hatch_eggs(self):
         try:
             pokemon_data = self._check_inventory(pokemon_ids)
         except:
-            pass  # just proceed with what we have
+            pokemon_data = [] # just proceed with what we have
```
reddivision: this would just break the results printing

PR #1807, opened 2016-07-30
Upgrade the capture logic for VIP pokemons!

xc-mezcal:
Short Description: 

This Pull basically upgrade the capture logic of the bot.
It has the following features:

1.Add VIP pokemon (means: force to use the best ball & berry to maximize the capture rates!)
(Can config it in configure file)

2.Add the logic for bot to always save the Ultra balls excpet for VIPs and pokemon with low capture rates (rare pokemon!) even we feed them with berry! 
**(Utral ball is so rare, so let's don't use it for catching a pidgey!)
***(Always use berry before trying to eventually use an ultra ball!)

3.Use berry first, if we have a lot of berry. (Hint: berry has a roughly equal value to a Great ball)

### DIFF
--- a/configs/config.json.example
+++ b/configs/config.json.example
@@ -57,5 +57,28 @@
       "// any": {"keep_best_iv": 3},
       "// Example of keeping the 2 strongest (based on CP) and 3 best (based on IV) Zubat:": {},
       "// Zubat": {"keep_best_cp": 2, "keep_best_iv": 3}
-    }
+    },    
+    "vips" : {
+         "Any pokemon put here directly force to use Berry & Best Ball to capture, to secure the capture rate!": {},
+        "any": {"catch_above_cp": 1200, "catch_above_iv": 0.9, "logic": "or" },
+        "Lapras": {},
+        "Moltres": {},
+        "Zapdos": {},
+        "Articuno": {},
+
+        "// S-Tier pokemons (if pokemon can be evolved into tier, list the representative)": {},
+        "Mewtwo": {},
+        "Dragonite": {},
+        "Snorlax": {},
+        "// Mew evolves to Mewtwo": {},
+        "Mew": {},
+        "Arcanine": {},
+        "Vaporeon": {},
+        "Gyarados": {},
+        "Exeggutor": {},
+        "Muk": {},
+        "Weezing": {},
+        "Flareon": {}
+        
+    } 
 }
--- a/configs/config.json.pokemon.example
+++ b/configs/config.json.pokemon.example
@@ -285,5 +285,28 @@
       "Raichu": { "release_below_cp": 708, "release_below_iv": 0.8, "logic": "and" },
       "Cloyster": { "release_below_cp": 717, "release_below_iv": 0.8, "logic": "and"},
       "Mr. Mime": { "release_below_cp": 650, "release_below_iv": 0.8, "logic": "and" }
-    }
+    },
+    "vips" : {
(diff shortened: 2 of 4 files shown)


configs/config.json.example
```diff
@@ -57,5 +57,28 @@
       "// any": {"keep_best_iv": 3},
       "// Example of keeping the 2 strongest (based on CP) and 3 best (based on IV) Zubat:": {},
       "// Zubat": {"keep_best_cp": 2, "keep_best_iv": 3}
-    }
+    },    
+    "vips" : {
+         "Any pokemon put here directly force to use Berry & Best Ball to capture, to secure the capture rate!": {},
+        "any": {"catch_above_cp": 1200, "catch_above_iv": 0.9, "logic": "or" },
```
k4n30: Should this be:
`"any": {"catch_above_cp": 0, "catch_above_iv": 0, "logic": "or"}, `

solderzzc:
This is really nice, do you have some test result or screen shot?

configs/config.json.example
xc-mezcal: @k4n30 Nope. Remember, The logic "any" here is to secure to capture VIP pokemon!

So anything above 1200 CP or 0.9 IV, will force the bot to use a better strategy to capture!

IF you put 0 CP or 0 IV, your best ball will ran out of stock soon!

xc-mezcal:
@solderzzc 
Yep, I have tested for most cases.
Cause a VIP pokemon is hard to encounter with. I will put some weak one there pretend to be a VIP to easily show you the test result :D

xc-mezcal:
Trying to capture a fake VIP Zubat. But saddly he vanished
vip_zubat_test-

solderzzc:
Super good @!@

jbennett94:
For me, this doesn't use ultra balls even though I have them. Will throw a razz berry and a great ball.

xc-mezcal:
@jbennett94  Can you be more specific? 

It will try to save ultra ball, but bot will still use ultraball if 

1. It's not VIP pokemon (either listed or CP above 1200 or IV above 0.9)

2. berry + great ball still have a low catch rate (<0.3)

Donwulff:
I like what's been done with the logic here very much. However, possibly because my bags are already full of Great Balls from the previous logic that would never use them while common pokemons are escaping me, I'd still like to be able to configure it to use more Great Balls.It seems like either a single setting like "capture_target_difficulty" (any simpler name?) instead of 0.35 would be handy, or even better

xc-mezcal:
0.35 is the previously logic for all balls

0.5 is the previously logic for berry

We probably can introduce one more configuration in config file. But that will make end-user more confused :D  

You can definitely config your file and tune it by yourself :D

configs/config.json.example
k4n30: Roger, sorry the logic in my head was messed up

jbennett94:
@xc-mezcal since writing this post, I paid more attention to the predicted capture rates and assume when I saw the VIP pokemon a razz + great ball will have given a high enough capture rate the program deemed high enough

xc-mezcal:
@jbennett94 
You can modify it at line around 156 in pokemon_catch_worker.py
`if catch_rate[pokeball-1] < 0.35 and items_stock[current_type] > 0:` 
(just before the second berry throw check, 0.35-> some higher value will use the great ball more often. I kept 0.35 from the previous ball picking logic :D)
 to see what suits you best and share your experience with us.

Issue #1814, opened 2016-07-30
Decode ASCII Error when display fort names in some lanaguages


Viper896:
Needs Verification:

Traceback (most recent call last):
  File "pokecli.py", line 464, in <module>
    main()
  File "pokecli.py", line 68, in main
    bot.tick()
  File "/Users/mike/repos/fun/PokemonGo-Bot/pokemongo_bot/__init__.py", line 84, in tick
    if worker.work() == WorkerResult.RUNNING:
  File "/Users/mike/repos/fun/PokemonGo-Bot/pokemongo_bot/cell_workers/move_to_fort_worker.py", line 43, in work
    logger.log('Moving towards fort {}, {} left'.format(fort_name, format_dist(dist, unit)))
  File "/Users/mike/repos/fun/PokemonGo-Bot/pokemongo_bot/logger.py", line 21, in log
    print('[' + time.strftime("%H:%M:%S") + '] ' + string)
  File "/Users/mike/repos/fun/PokemonGo-Bot/lib/python2.7/codecs.py", line 369, in write
    data, consumed = self.encode(object, self.errors)
UnicodeDecodeError: 'ascii' codec can't decode byte 0xce in position 31: ordinal not in range(128)
```

From Slack Chat:
It's probably due to having forts names in a different language (greek in my case)

commit: 2f277fd7d08ff9a7ad45cf46e102a00db7ba9fdf
>Display Fort Name instead of Fort ID in Log (#1801)

**Comment by hklcf** [comment]
same here

![image](https://cloud.githubusercontent.com/assets/410302/17269526/7459d62c-567d-11e6-87d4-a5a3611ded58.png)

2f277fd7d08ff9a7ad45cf46e102a00db7ba9fdf

**Comment by CapCap** [comment]
```
import sys  

reload(sys)  
sys.setdefaultencoding('utf8')
```

**Comment by hklcf** [comment]
@CapCap waiting for your PR :+1:

**Comment by xorinzor** [comment]
Confirmed, having this issue too.

```
Traceback (most recent call last):
  File "pokecli.py", line 464, in <module>
    main()
  File "pokecli.py", line 68, in main
    bot.tick()
  File "C:\Users\xorinzor\Desktop\PokemonGo-Bot\pokemongo_bot\__init__.py", line 84, in tick
    if worker.work() == WorkerResult.RUNNING:
[...]

PR #1817, opened 2016-07-30
fix unicode issues

CapCap:
Short Description: 

Fixes:
- #1814 
- 
-

### DIFF
--- a/pokecli.py
+++ b/pokecli.py
@@ -32,11 +32,6 @@ import logging
 import os
 import ssl
 import sys
-
-# this is required to fix unicode errors in python 2.x
-reload(sys)
-sys.setdefaultencoding('utf8')
-
 import time
 from datetime import timedelta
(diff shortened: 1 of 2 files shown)


vieira:
The problem is that the log function is sometimes called with an already decoded utf8 str while some other times it is called with an unicode sequence that needs to be decoded before print. 

```
def log(string, color='white'):
    try:
        string = string.decode('utf8')
    except UnicodeEncodeError:
        pass

<rest of the original code>
```

This is enough to fix the problem and make log work with both str and unicode sequences.

DayBr3ak:
190640f06fc623f804c554ce1ac9264c9fa8b22d makes this irrelevant

vieira:
@DayBr3ak still getting errors if the initial location contains any special characters.

vieira:
#1825 should fix the issue with the initial location.

Issue #1654, opened 2016-07-29
IOError: [Errno 24] Too many open files


JaisDK:
### Expected Behavior
Not crash

### Actual Behavior
Crash

### Steps to Reproduce
Run bot and have the map open for a while

### Other Information
OS: Windows 10
Browser: Chrome v51.0.2704.103
Python 2.7.12

### Full stack trace
[16:24:14] Moving towards fort ..., 0.17km left
Traceback (most recent call last):
  File "pokecli.py", line 408, in <module>
  File "pokecli.py", line 68, in main
  File "C:\xampp\htdocs\pgobot\pokemongo_bot\__init__.py", line 92, in tick
  File "C:\xampp\htdocs\pgobot\pokemongo_bot\cell_workers\move_to_fort_worker.py", line 53, in work
  File "C:\xampp\htdocs\pgobot\pokemongo_bot\step_walker.py", line 61, in step
  File "C:\xampp\htdocs\pgobot\pokemongo_bot\__init__.py", line 499, in heartbeat
  File "C:\xampp\htdocs\pgobot\pokemongo_bot\__init__.py", line 158, in update_web_location
IOError: [Errno 24] Too many open files: u'data/cells-[username].json'

mrseeker:
Sounds like files are not properly getting closed by the bot. Can you include the git revision that made this happen?

Surceis:
Ok, i also got this a few times :smile:

OS Windows 7 x64
two different PC

commit c5483349e02fc0b1c8c8ed31411ae8ca71ec4177
```shell
[15:29:51] Something rustles nearby!
[15:29:52] A Wild Gastly appeared! [CP 98] [Potential 0.67]
[15:29:52] IV [Stamina/Attack/Defense] = [5/13/12]
Traceback (most recent call last):
  File "pokecli.py", line 408, in <module>
  File "pokecli.py", line 68, in main
  File "E:\pokego\pokemongo_bot\__init__.py", line 92, in tick
  File "E:\pokego\pokemongo_bot\cell_workers\catch_visible_pokemon_worker.py", line 37, in work
  File "E:\pokego\pokemongo_bot\cell_workers\catch_visible_pokemon_worker.py", line 76, in catch_pokemon
  File "E:\pokego\pokemongo_bot\cell_workers\pokemon_catch_worker.py", line 80, in work
  File "E:\pokego\pokemongo_bot\__init__.py", line 377, in current_inventory
[...]

PR #1820, opened 2016-07-30
Too many open files error and high IO;

rguedes:
Short Description: 
Too many open files error and high IO

Fixes:
- Fix: "IOError: [Errno 24] Too many open files"

### DIFF
--- a/pokemongo_bot/__init__.py
+++ b/pokemongo_bot/__init__.py
@@ -157,6 +157,7 @@ class PokemonGoBot(object):
         user_data_cells = "data/cells-%s.json" % (self.config.username)
         with open(user_data_cells, 'w') as outfile:
             json.dump(cells, outfile)
+            outfile.close()
 
         user_web_location = os.path.join('web', 'location-%s.json' % (self.config.username))
         # alt is unused atm but makes using *location easier
@@ -168,6 +169,7 @@ class PokemonGoBot(object):
                     'alt': alt,
                     'cells': cells
                     }, outfile)
+                outfile.close()
         except IOError as e:
             logger.log('[x] Error while opening location file: %s' % e, 'red')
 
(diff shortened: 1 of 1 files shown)


Nihisil:
Hm, they should to be automatically closed, when you use `with open()`. Did you test it?

rguedes:
Yes and the errors stopped

DayBr3ak:
of course, `with` statement handle the closing of files automatically even when execeptions are raised. This pr should have no effect on the problem at all

rguedes:
Apparently not close or not work well, at least in windows...

Nihisil:
`with` will close files when GC will do the job. So, probably for some configuration it is too much time between GC runs, and for them we need to close files manually.

DayBr3ak:
well if it its a GC problem, at least get rid of the with statement

pokemongo_bot/__init__.py
```diff
@@ -157,6 +157,7 @@ def update_web_location(self, cells=[], lat=None, lng=None, alt=None):
         user_data_cells = "data/cells-%s.json" % (self.config.username)
         with open(user_data_cells, 'w') as outfile:
             json.dump(cells, outfile)
+            outfile.close()
```
cwild: ```with``` closes outfile automatically.

DayBr3ak:
and to be fair I'm pretty sure with does't work with GC it's just sugar for a __enter__ and __exit__ method.

makistsantekidis:
@Nihisil @DayBr3ak I think that when using a `with` statement the file is closed immediately after exiting the `with` context, since the __exit__ function of `open` is called which makes sure to close the file.

If you want you can confirm this by adding 

```
print outfile.closed
```
after the `with` contexts

Surceis:
this https://github.com/PokemonGoF/PokemonGo-Bot/issues/1654

douglascamata:
#1654 figures out the source of the problem, closing this and adding #1654 to the approved issue list.

PR #1824, opened 2016-07-30
Make logger work with both str and unicode

vieira:
Short Description: 
log is sometimes called with str, other times with unicode sequences. This change makes it work for both cases.

Fixes:
- Unicode decode errors at Pokestops with special chars (see #1814)

### DIFF
--- a/pokemongo_bot/logger.py
+++ b/pokemongo_bot/logger.py
@@ -10,6 +10,11 @@ except Exception:
 
 
 def log(string, color='white'):
+    try:
+        string = string.decode('utf8')
+    except UnicodeEncodeError:
+        pass
+    
     color_hex = {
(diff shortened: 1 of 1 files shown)


DayBr3ak:
I tried your fix. Text is white now, no colors. And I get spammed by empty lines.

DayBr3ak:
Sorry about before, copy pasta fail.
Works good so far

DayBr3ak:
190640f06fc623f804c554ce1ac9264c9fa8b22d makes this irrelevant

vieira:
@DayBr3ak Nope, I am still getting errors if the initial location contains any special characters.

vieira:
#1825 should fix the issue with the initial location.

PR #1825, opened 2016-07-30
Log location as a str to fix crash on start when initial location contains special chars

vieira:
Short Description: 

We already have location as an encoded str in location_str and most calls to log pass a string as argument, so this is a bit more consistent. Only tested on top of #1824, please do not merge without it.

Fixes:
- Nothing, this should be equivalent but a bit more consistent.


vieira:
Now that 190640f was merged this PR is needed when the initial location contains special characters, e.g. "ã", otherwise the bot won't start.

Fixes:
- crash on start when initial location contains special characters, e.g. "ç", "ã", etc

Issue #1823, opened 2016-07-30
'pokemon_data' referenced before assignment in incubate_eggs_worker


JohanJu:
### Expected Behavior
pokemon_data should be assigned

### Actual Behavior
File "/home/john/PokemonGo-Bot-dev/pokemongo_bot/cell_workers/incubate_eggs_worker.py", line 29, in work
    self._hatch_eggs()
  File "/home/john/PokemonGo-Bot-dev/pokemongo_bot/cell_workers/incubate_eggs_worker.py", line 134, in _hatch_eggs
    for pokemon in pokemon_data:
UnboundLocalError: local variable 'pokemon_data' referenced before assignment

### Code
```
        try:
            pokemon_data = self._check_inventory(pokemon_ids)
        except:
            pass  # just proceed with what we have
        for pokemon in pokemon_data:
```
Maybe pokemon_data wasn't assigned before except 

### Steps to Reproduce
run with 
"hatch_eggs": true,
"longer_eggs_first": true,
for some time

### Other Information
OS: Linux 4.4 manjaro kde
Git Commit: sorry downloaded zip, but probably 79152bc65f8194ee9eb591fa657b99423771409e on dev

chadhao:
Happens on Ubuntu 14.04.4 LTS as well. Same configuration
"hatch_eggs": true,
"longer_eggs_first": true,

**Traceback**
  File "pokecli.py", line 464, in <module>
    main()
  File "pokecli.py", line 68, in main
    bot.tick()
  File "/root/pokemon/bot01/pokemongo_bot/__init__.py", line 84, in tick
    if worker.work() == WorkerResult.RUNNING:
  File "/root/pokemon/bot01/pokemongo_bot/cell_workers/incubate_eggs_worker.py", line 29, in work
    self._hatch_eggs()
  File "/root/pokemon/bot01/pokemongo_bot/cell_workers/incubate_eggs_worker.py", line 134, in _hatch_eggs
    for pokemon in pokemon_data:
UnboundLocalError: local variable 'pokemon_data' referenced before assignment

bagwaa:
I also have this issue, trying to run without `hatch_eggs` at the moment to see if its 100% that

YuxuanChen:
Try replacing the word `pass` with `return`, see if that works.

PR #1826, opened 2016-07-30
Fix 'pokemon_data' referenced before assignment in incubate_eggs_worker (#1823)

DisasteR:
Short Description: 
'pokemon_data' referenced before assignment in incubate_eggs_worker

Fixes:
- 'pokemon_data' referenced before assignment in incubate_eggs_worker #1823
- 
-


reddivision:
already fixed in #1777

PR #1838, opened 2016-07-30
Stale location cache check

cwild:
If specifying a new location, don't mandate that location_cache be turned off explicitly.

Fixes:
- Change of --location will work without having to turn location cache off


cwild:
Opening a new PR as the comments above relate to other commits

cwild:
See #1932 .. comment history on this PR is weird.

PR #1842, opened 2016-07-30
Added purple colour

fr3ky:
Short Description: 
Added purple colour to logger for the ones prefer PowerShell.

### DIFF
--- a/pokemongo_bot/logger.py
+++ b/pokemongo_bot/logger.py
@@ -13,6 +13,7 @@ def log(string, color = 'white'):
         'green': '92m',
         'yellow': '93m',
         'blue': '94m',
+		'purple': '95m',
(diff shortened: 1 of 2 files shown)


cwild:
How is it for ones who prefer something when there is no config option?

pokemongo_bot/logger.py
```diff
@@ -13,6 +13,7 @@ def log(string, color = 'white'):
         'green': '92m',
         'yellow': '93m',
         'blue': '94m',
+		'purple': '95m',
```
cwild: You have indented with tabs instead of spaces.

fr3ky:
I noticed the indentation due to tab... purple seems to work well in PS and cmd too.

PR #1848, opened 2016-07-30
Softban Bypass

jadbalout:
Short Description: 
It has been a rumor that the softban can be bypassed by spinning the same fort 40 times. I have tried this and it works.
Enhancements:
- Softban Bypass implemented.

Further on, we could add mode options to ignore the pokemon until soft ban removed, but that is up to you guys since some did not like that idea.


LasseSkogland:
This has already been added to the latest commit on dev branch.

solderzzc:
Yes @LasseSkogland , let me close this commit.
Thanks for the PR.

PR #1850, opened 2016-07-30
Added nickname templating and associated worker

reddivision:
Short Description: 
Users can now specify a template (using python Formatter syntax) to nickname pokemon in their inventory. The worker for this will process the user's entire inventory each tick, but will only update pokemon that require an update to match the new template (i.e. only caught/hatched pokemon after the first tick).

Use a blank template (`""`) to revert all pokemon to their original names.

Valid names in templates are:
- `name` = pokemon name
- `id` = pokemon type id (e.g. 1 for Bulbasaurs)
- `cp` = pokemon's CP
- `iv_attack` = pokemon's attack IV
- `iv_defense` = pokemon's defense IV
- `iv_stamina` = pokemon's stamina IV
- `iv_ads` = pokemon's IVs in X/X/X format (matches web UI format -- A/D/S)
- `iv_sum` = pokemon's IVs as a sum (e.g. 45 when 3 perfect 15 IVs)
- `iv_pct` = pokemon's IVs as a percentage (0-100)

Examples: 
- `{name}_{iv_pct}` => `Mankey_69`
- `{iv_pct}_{iv_ads}` => `91_15/11/15`
sample

### DIFF
--- a/configs/config.json.example
+++ b/configs/config.json.example
@@ -12,6 +12,12 @@
         "type": "CollectLevelUpReward"
       },
       {
+        "type": "NicknamePokemon",
+        "config": {
+          "nickname_template": "{iv_pct}_{iv_ads}"
+        }
+      },
+      {
         "type": "IncubateEggs",
         "config": {
           "longer_eggs_first": true
--- a/pokemongo_bot/cell_workers/incubate_eggs.py
+++ b/pokemongo_bot/cell_workers/incubate_eggs.py
@@ -106,15 +106,13 @@ class IncubateEggs(BaseTask):
                         "used": False
                     })
                 elif 'is_egg' not in pokemon and pokemon['id'] in lookup_ids:
-                    matched_pokemon.append({
-                        "pokemon_id": pokemon.get('pokemon_id', -1),
-                        "cp": pokemon.get('cp', 0),
+                    matched_pokemon.append(pokemon.update({
                         "iv": [
                             pokemon.get('individual_attack', 0),
                             pokemon.get('individual_defense', 0),
                             pokemon.get('individual_stamina', 0)
                         ]
-                    })
+                    }))
                 continue
             if "player_stats" in inv_data:
                 self.km_walked = inv_data.get("player_stats", {}).get("km_walked", 0)
@@ -147,7 +145,7 @@ class IncubateEggs(BaseTask):
             for pokemon in pokemon_data:
                 # pokemon ids seem to be offset by one
                 if pokemon['pokemon_id']!=-1:
-                    pokemon['name'] = self.bot.pokemon_list[(pokemon['pokemon_id']-1)]['Name']
+                    pokemon['name'] = self.bot.pokemon_list[(pokemon.get('pokemon_id')-1)]['Name']
                 else:
                     pokemon['name'] = "error"
         except:
@@ -160,7 +158,7 @@ class IncubateEggs(BaseTask):
         for i in range(len(pokemon_data)):
             logger.log("-"*30,log_color)
             logger.log("[!] Pokemon: {}".format(pokemon_data[i]['name']), log_color)
-            logger.log("[!] CP: {}".format(pokemon_data[i]['cp']), log_color)
+            logger.log("[!] CP: {}".format(pokemon_data[i].get('cp',0)), log_color)
             logger.log("[!] IV: {} ({:.2f})".format("/".join(map(str, pokemon_data[i]['iv'])),(sum(pokemon_data[i]['iv'])/self.max_iv)), log_color)
             logger.log("[!] XP: {}".format(xp[i]), log_color)
             logger.log("[!] Stardust: {}".format(stardust[i]), log_color)
--- a/pokemongo_bot/cell_workers/nickname_pokemon.py
+++ b/pokemongo_bot/cell_workers/nickname_pokemon.py
@@ -1,0 +1,88 @@
+from pokemongo_bot import logger
+from pokemongo_bot.human_behaviour import sleep
+from pokemongo_bot.cell_workers.base_task import BaseTask
+
+class NicknamePokemon(BaseTask):
+    def initialize(self):
+        self.template = self.config.get('nickname_template','').lower().strip()
+        if self.template == "{name}":
+            self.template = ""
+    
+    def work(self):
+        try:
+            inventory = reduce(dict.__getitem__, ["responses", "GET_INVENTORY", "inventory_delta", "inventory_items"], self.bot.get_inventory())
+        except KeyError:
+            pass
+        else:
+            pokemon_data = self._get_inventory_pokemon(inventory)
+            for pokemon in pokemon_data:
+                self._nickname_pokemon(pokemon) 
+            
+    def _get_inventory_pokemon(self,inventory_dict):
+        pokemon_data = []
+        for inv_data in inventory_dict:
+            try:
+                pokemon = reduce(dict.__getitem__,['inventory_item_data','pokemon_data'],inv_data)
+            except KeyError:
+                pass
+            else:
+                if not pokemon.get('is_egg',False):
+                    pokemon_data.append(pokemon)
+        return pokemon_data
+        
+    def _nickname_pokemon(self,pokemon):
+        """This requies a pokemon object containing all the standard fields: id, ivs, cp, etc"""
+        new_name = ""
+        instance_id = pokemon.get('id',0)
+        if not instance_id:
+            logger.log("Pokemon instance id returned 0. Can't rename.",'red')
+            return
+        id = pokemon.get('pokemon_id',0)-1
+        name = self.bot.pokemon_list[id]['Name']
+        cp = pokemon.get('cp',0)
+        iv_attack = pokemon.get('individual_attack',0)
+        iv_defense = pokemon.get('individual_defense',0)
+        iv_stamina = pokemon.get('individual_stamina',0)
+        iv_list = [iv_attack,iv_defense,iv_stamina]
+        iv_ads = "/".join(map(str,iv_list))
+        iv_sum = sum(iv_list)
+        iv_pct = "{:0.0f}".format(100*iv_sum/45.0)
+        log_color = 'red'
+        try:
+            new_name = self.template.format(name=name,
+                                    id=id,
+                                    cp=cp,
+                                    iv_attack=iv_attack,
+                                    iv_defense=iv_defense,
+                                    iv_stamina=iv_stamina,
+                                    iv_ads=iv_ads,
+                                    iv_sum=iv_sum,
+                                    iv_pct=iv_pct)[:12]
+        except KeyError as bad_key:
+            logger.log("Unable to nickname {} due to bad template ({})".format(name,bad_key),log_color)
+        if pokemon.get('nickname', "") == new_name:
+            return
+        self.bot.api.nickname_pokemon(pokemon_id=instance_id,nickname=new_name)
+        response = self.bot.api.call()
+        sleep(1.2)
+        try: 
+            result =  reduce(dict.__getitem__, ["responses", "NICKNAME_POKEMON"], response)
+        except KeyError:
+            logger.log("Attempt to nickname received bad response from server.",log_color)
+            if self.bot.config.debug:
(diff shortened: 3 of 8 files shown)


pokemongo_bot/cell_workers/utils.py
```diff
@@ -144,3 +145,57 @@ def print_yellow(message):
 
 def print_red(message):
     print(u'\033[91m' + message.decode('utf-8') + '\033[0m')
+
+def nickname_pokemon(bot,pokemon):
+    """This requies a pokemon object containing all the standard fields: id, ivs, cp, etc"""
+    template = bot.config.nickname_template.lower()
+    if template == "{name}":
```
cwild: What if the Pokemon has already been nicknamed? Do you know the existing name at this point?

reddivision: Except for the case where someone's renaming their entire inventory (which won't be common, I imagine), the pokemon can't have been nicknamed already. In the case where someone wants to "force" a revert to original name, they could use a hack (since I don't trim whitespace) `{name} `. That said, I can refactor all this if there's a desire for it.

To answer your other question, yes, we can know a pokemon is nicknamed by the presence of a `nickname` key in the `pokemon_data` dictionary.

cwild: Forgive me as I don't know what the pokemon variable looks like at this point, but are you saying ```pokemon.get('nickname')``` exists at this point of the code?

reddivision: in a manner of speaking, yes. the `pokemon` object may or may not have a `'nickname'` key in it, depending on if it's been nicknamed previously or not.

solderzzc:
This is great, could it also be set to an 'NickName' ?
The usage is:
Eevee evolution is usually random. Unless you change the name of your Eevee to Sparky, Pyro or Rainer. Sparky allows your Eevee to evolve into Jolteon, Pyro allows it to evolve into Flareon and Rainer allows it to evolve into Vaporeon.

solderzzc:
Please also open issue to notify botConfig
https://github.com/PokemonGoF/botConfig

configs/config.json.example
```diff
@@ -35,6 +35,8 @@
     "use_lucky_egg": false,
     "hatch_eggs": true,
     "longer_eggs_first": true,
+    "nickname_template": "{name}_{iv_pct}",
+    "nickname_on_startup": false,
```
solderzzc: "nickname_on_startup" was not mentioned in pokecli.py.

douglascamata:
@reddivision please make the nicknaming settings all be embedded inside a key in the config, like the `forts` key embed many other keys. Look for their `add_config` call in `pokecli.py` to learn how to do it.

Philipp59:
recommend to change also the order in catch_worker from

                                iv_display = '{}/{}/{}'.format(
                                    individual_stamina,
                                    individual_attack,
                                    individual_defense

in 
iv_display = '{}/{}/{}'.format(
individual_attack,
individual_defense
individual_stamina,

tested your pull request works like a charm :) 
but i found it quite confusing that your logical and good order from attack/Defense/Stamina didnt match with the catch worker ordering.

also "{name}" in config  resets name to default works also very good.

one problems I noticed.

1. place for letters
There are only 12 letters for nickname so that "{name}""{iv_pct}" doesn't fit for all pokemons 
e.g.:
[17:12:26] Nickname set successfully! Scyther is now Scyther_0.98
[17:12:28] Nickname set successfully! Dratini is now Dratini_0.58
[17:12:30] Nickname set successfully! Electabuzz is now Electabuzz_0

So it would be better to set the default config to "{iv_pct}_{iv_ads}"

Therefore change in utils.py
iv_pct = "{:.2f}".format(iv_sum/45.0) 
to
iv_pct = "{:2f}".format(iv_sum/45.0)*100

with that wie change 0.93 in 93 which is saving 2 letters of space 

to that per default config Pokemon will be names in for e.g. "93_14/14/14" or 100_15/15/15 would fit also

Philipp59:
getting this error now when egg is ready 

```
  File "pokecli.py", line 475, in <module>
    main()
  File "pokecli.py", line 68, in main
    bot.tick()
  File "/Users/XXX/Desktop/bots/PokemonGo-BotFix/pokemongo_bot/__init__.py", line 84, in tick
  File "/Users/XXX/Desktop/bots/PokemonGo-BotFix/pokemongo_bot/cell_workers/incubate_eggs_worker.py", line 29, in work
    self._hatch_eggs()
  File "/Users/XXX/Desktop/bots/PokemonGo-BotFix/pokemongo_bot/cell_workers/incubate_eggs_worker.py", line 134, in _hatch_eggs
    pokemon['name'] = self.bot.pokemon_list[(pokemon.get('pokemon_id',0)-1)]['Name']
AttributeError: 'NoneType' object has no attribute 'get'
```

Working with 9b2450a7d81b33e0b46cf355c3d4ac9c82e92cfa

douglascamata:
@reddivision if you have the `NicknameAll` behaviour in the beginning of the tick, you don't need to try to set nicknames on pokemons after capture or egg hatches. Also, to make the `NicknameAll` behaviour don't make a request for every single pokemon that was already renamed, you can check the current pokemon name and the new name it would have. If they match, you can skip the pokemon.

bigkraig:
nickname_on_startup doesn't work

pokemongo_bot/cell_workers/nickname_pokemon.py
```diff
@@ -0,0 +1,88 @@
+from pokemongo_bot import logger
+from pokemongo_bot.human_behaviour import sleep
+from pokemongo_bot.cell_workers.base_task import BaseTask
+
+class NicknamePokemon(BaseTask):
+    def initialize(self):
+        self.template = self.config.get('nickname_template','').lower().strip()
+        if self.template == "{name}":
+            self.template = ""
```
reddivision: This might seem unintuitive, but setting the nickname to blank (like in the app) will actually revert the pokemon back to its original name.  So, we set the template to blank to achieve this behavior when we anticipate that is what the user wants.

pokemongo_bot/cell_workers/incubate_eggs.py
```diff
@@ -106,15 +106,13 @@ def _check_inventory(self, lookup_ids=[]):
                         "used": False
                     })
                 elif 'is_egg' not in pokemon and pokemon['id'] in lookup_ids:
-                    matched_pokemon.append({
-                        "pokemon_id": pokemon.get('pokemon_id', -1),
-                        "cp": pokemon.get('cp', 0),
+                    matched_pokemon.append(pokemon.update({
```
TheSavior: Why does a PR that is adding a task for renaming pokemon need to modify the incubate eggs task?

pokemongo_bot/cell_workers/pokemon_catch_worker.py
```diff
@@ -52,9 +50,9 @@ def work(self):
                                 individual_defense = pokemon_data.get("individual_defense", 0)
 
                                 iv_display = '{}/{}/{}'.format(
-                                    individual_stamina,
                                     individual_attack,
-                                    individual_defense
+                                    individual_defense,
```
TheSavior: It seems like the changes to this file aren't relevant either

Philipp59: its a nice feature to resort the Bot ordering of IV from SAD to ADS so its in the same order like the renaming.
It would be quite irritating if a pokemon appears with 4/10/15 (SAD) and then its renamed in 10/15/4 (ADS).
Therefore not change the order in general from SAD (why?) to a more logic on (ADS)

pokemongo_bot/cell_workers/incubate_eggs.py
reddivision: It was initially a change required for an earlier iteration of this to work (as it required that pokemon objects passed in maintained their standard keys). As you can see, instead of passing out an object with only a subset of keys, it maintains the original object and just appends new calculated data. I opted to leave it in because I feel this is still a better approach in case the egg incubation worker wants to utilize other keys in the pokemon object. At the very least, it doesn't break anything.

pokemongo_bot/cell_workers/pokemon_catch_worker.py
reddivision: The reason that the renaming worker uses ADS format is because that is also used in the web UI. We're using this as an opportunity to make the pokemon catch worker output consistent with this format as well.

pokemongo_bot/cell_workers/incubate_eggs.py
TheSavior: If it isn't *needed* by this PR, can you undo these changes? It makes it much scarier to review and accept these PRs since it is hard to tell what the impact across the whole project is going to be. Keeping changes small and focused on a single worker makes our lives way easier. If we need these changes down the road, I'd feel more comfortable making them when it actually impacts something.

pokemongo_bot/cell_workers/nickname_pokemon.py
douglascamata: Please add this behaviour explanation as a comment in the code.

douglascamata:
I'd like to reiterate my suggestion:

> @reddivision if you have the NicknameAll behaviour in the beginning of the tick, you don't need to try to set nicknames on pokemons after capture or egg hatches (they will be renamed in the next tick). Also, to make the NicknameAll behaviour don't make a request for every single pokemon that was already renamed, you can check the current pokemon name and the new name it would have. If they match, you can skip the pokemon.

You can add one config to the worker to make it only rename pokemon on start up, to make it even less expensive, and one config to force rename all pokemon regardless of current name.

reddivision:
@douglascamata your initial suggestion you quoted was that nickname all shouldn't be implemented in catching/hatch methods but should just work as part of the chain (it now does) and that it should check names before submitting request to ensure it doesn't name the same again (it also now does that). Your suggestions here are completely different. To address them: 

What is the use case that someone will want to rename their entire inventory on startup and not after catching/hatching? Is that really in any way practical/useful? The current implementation is that the nicknames of all pokemon will be uniform at all times. 

Also, all pokemon will be renamed regardless of their current name unless it is already the same. With that, what is the point in allowing them to force a rename? It's just going to name it what it already is.

douglascamata:
> What is the use case that someone will want to rename their entire inventory on startup and not after catching/hatching? Is that really in any way practical/useful?

@reddivision when we evolve pokemon, for example, does it handle properly?

reddivision:
@douglascamata yes, it should. the next tick it would recalculate the pokemon's new nickname and it would then be using the pokemon's new id/name in the nickname calculation. it would detect that the nickname does not match the current nickname and fire a request.

douglascamata:
👍 

LGTM

TheSavior:
👍

bixuanzju:
Hi, I have the following config
```
      {
        "type": "NicknamePokemon",
        "config": {
            "nickname_template": "{name}_{iv_ads}"
        }
      },
```
After running, I notice that not every pokemon is renamed as per the config
screen shot 2016-08-01 at 8 24 43 pm
What does `15/` for example mean? The other two are zero?

Philipp59:
change to  **"nickname_template": "{iv_pct}_{iv_ads}"**

bixuanzju:
@Philipp59 yeah, that works, but I want the pokemon's original name plus its ads, could that be done?

Philipp59:
As I wrote above the letter space is limited to 12 so you can't take name+ads.
Why you need the original name of the pokemon ? oO

bixuanzju:
Ah, I see, no particular reason, the current config is fine, thanks!

reddivision:
@bixuanzju check out [the wiki](https://github.com/PokemonGoF/PokemonGo-Bot/wiki/Configuration-files#pokemon-nicknaming) for additional usage information

vdubus:
I would suggest to format the `iv_pct` output on 3 digit so that we can easily sort pokemons by name.

```python
iv_pct = "{:03.0f}".format(100*iv_sum/45.0)
```

Maybe some other template can also gain something from that.

configs/config.json.example
danielsdian: Any news about renaming only the Eevees?

PR #1874, opened 2016-07-30
Fixed a small bug when user didn't have a VIP config!

xc-mezcal:
Short Description: 
This fix a small bug when user didn't update their config file for VIP pokemon setting (leave it for empty)

Fixes:
Return false if there is no such a setting instead of crashing.
- 
-


xc-mezcal:
@solderzzc need to add 2 lines there for a small bug

mercuriete:
I hit that bug.
I dont have vip pokemon.

solderzzc:
sure, thanks a lot.
I was thinking this part :)

mercuriete:
nice!

PR #1748, opened 2016-07-30
Unban Immediately When You Fly

wilsonyqm:
Short Description: 
  Since the SoftBan can be resolved, trainer is able to fly to some specific location to catch legendary Pokemon. In order to do so,  you need to encounter the Pokemon first, and fly back to the home location you have set.

Only thing we need to do: change the location, set your home location where you cannot get banned in `config.json`, run the program.  Set `'home_location' = "home, location"`,  in `configs/config.json`.

Better Close 
```json
      {
        "type": "FollowSpiral"
      }
```
Fixed:
`self.softban` -> `self.bot.softban`, to make sure softban works on Bot Object.

PR #1887, opened 2016-07-30
Draft feature to snipe pokemon

enjoy2000:
Short Description: 
Add draft commit with hope that u guys can give me some advices.
Always got softbanned when trying to snipe pokemon.

### DIFF
--- a/pokecli.py
+++ b/pokecli.py
@@ -264,6 +264,15 @@ def init_config():
     add_config(
         parser,
         load,
+        short_flag="-spl",
+        long_flag="--snipe_list",
+        help="File containing locations to snipe",
+        type=str,
+        default=None
+    )
+    add_config(
+        parser,
+        load,
         short_flag="-d",
         long_flag="--debug",
         help="Debug Mode",
--- a/pokemongo_bot/cell_workers/snipe_pokemon.py
+++ b/pokemongo_bot/cell_workers/snipe_pokemon.py
@@ -23,14 +23,19 @@ class SnipePokemon(object):
         # process additional configs
 
     def work(self):
+        if self.bot.tick_count is not 1:
+            return
+
         while True:
             location = raw_input('Enter snipe location: ')
             location = location.replace(' ', '')
 
-            if len(location) > 0:
+            pattern = "^(\-?\d+(\.\d+)?),\s*(\-?\d+(\.\d+)?)$"
+            if re.match(pattern, location):
                 self.snipe_pokemon(location)
             else:
-                break   
+                logger.log('Wrong format location!', 'red')
+                break
 
 
     def snipe_pokemon(self, location, delay=2): 
@@ -94,6 +99,8 @@ class SnipePokemon(object):
             self.api.set_position(*prevPosition)
             time.sleep(delay)
             self.heartbeat()
+
+            return None
 
         catchWorker = PokemonCatchWorker(catch_pokemon, self.bot)
         
--- a/pokemongo_bot/cell_workers/snipe_pokemon.py
+++ b/pokemongo_bot/cell_workers/snipe_pokemon.py
@@ -5,23 +5,16 @@ import sys
 import time
 
 from . import PokemonCatchWorker
+from base_task import BaseTask
 from pokemongo_bot import logger
 from utils import distance
 
-class SnipePokemon(object):
+class SnipePokemon(BaseTask):
 
-    def __init__(self, bot, config):
-        self.bot = bot
-        self.api = bot.api
-        self.config = config
-        self.pokemon_list = bot.pokemon_list
-        self.position = bot.position
-
-        self._process_config()
-
-    def _process_config(self):
-        # process additional configs
-        return
+    def initialize(self):
+        self.api = self.bot.api
+        self.pokemon_list = self.bot.pokemon_list
+        self.position = self.bot.position
 
     def work(self):
         if self.bot.tick_count is not 1:
--- a/pokemongo_bot/cell_workers/snipe_pokemon.py
+++ b/pokemongo_bot/cell_workers/snipe_pokemon.py
@@ -1,0 +1,111 @@
+import json
+import os
+import re
+import sys
+import time
+
+from . import PokemonCatchWorker
+from pokemongo_bot import logger
+from utils import distance
+
+class SnipePokemon(object):
+
+    def __init__(self, bot, config):
+        self.bot = bot
+        self.api = bot.api
+        self.config = config
+        self.pokemon_list = bot.pokemon_list
+        self.position = bot.position
+
+        self._process_config()
+
+    def _process_config(self):
+        # process additional configs
+
+    def work(self):
+        while True:
+            location = raw_input('Enter snipe location: ')
+            location = location.replace(' ', '')
+
+            if len(location) > 0:
+                self.snipe_pokemon(location)
+            else:
+                break   
+
+
+    def snipe_pokemon(self, location, delay=2): 
+        prevPosition = self.bot.position
+
+        # Check if session token has expired
+        self.bot.check_session(self.bot.position[0:2])
+        self.bot.heartbeat()
+
+        # Teleport to location
+        logger.log('Teleport to location..', 'green')
+        latitude, longitude = location.split(',')
+        self.api.set_position(float(latitude), float(longitude), 0)
+        self.cell = self.bot.get_meta_cell()
+
+        catch_pokemon = None
+        if 'catchable_pokemons' in self.cell and len(self.cell['catchable_pokemons']) > 0:
+            logger.log('Something rustles nearby!')
+            # Sort all by distance from current pos- eventually this should
+            # build graph & A* it
+            self.cell['catchable_pokemons'].sort(
+                key=
+                lambda x: distance(self.position[0], self.position[1], x['latitude'], x['longitude']))
+
+
+            user_web_catchable = 'web/catchable-%s.json' % (self.bot.config.username)
+            for pokemon in self.cell['catchable_pokemons']:
+                with open(user_web_catchable, 'w') as outfile:
+                    json.dump(pokemon, outfile)
+
+                with open(user_web_catchable, 'w') as outfile:
+                    json.dump({}, outfile)
+
+            catch_pokemon = self.cell['catchable_pokemons'][0]
+
+            # Try to catch VIP pokemon
+            for pokemon in self.cell['catchable_pokemons']:
+                pokemon_num = int(pokemon['pokemon_id']) - 1   
+                pokemon_name = self.pokemon_list[int(pokemon_num)]['Name']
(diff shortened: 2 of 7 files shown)


douglascamata:
What you mean by snipe?

enjoy2000:
Ah that word from `.net` version.

People can use trick to teleport to pokemon coords and encounter it then teleport back. 
After that we can catch it normally without softban.

DayBr3ak:
Yes apparently you have to start the fight and teleport back before throwing the first pokeball

wilsonyqm:
Can you take a look at this PR https://github.com/PokemonGoF/PokemonGo-Bot/pull/1748

enjoy2000:
image
```./pokecli.py --snipe=True```
First run and it success.

But maybe need more refactoring work, please help to review.

File: pokecli.py [hunk 1]
```diff
@@ -104,6 +108,34 @@ def report_summary(bot):
     if metrics.most_perfect is not None:
         logger.log('Most Perfect Pokemon: {}'.format(metrics.most_perfect['desc']), 'cyan')
 
+def snipe():
```
TheSavior: This shouldn't be here. Now that the bot is fully configurable by the user in config.json, I'd recommend creating a SnipePokemon worker. Then a user could specify just that in their config.json and get the behavior you want. There should generally be no reason this file needs to be modified anymore.

pokemongo_bot/__init__.py
```diff
@@ -89,6 +90,89 @@ def tick(self):
             if worker.work() == WorkerResult.RUNNING:
                 return
 
+        self.navigator.take_step()
+        
+        self.tick_count +=1
+
+    def snipe(self, location, delay=2):
```
TheSavior: This shouldn't be here either, for the same reason as my comment above.

enjoy2000:
@TheSavior Updated, please have a look again. Thanks.

nikhil-pandey:
Awesome!
image

nikhil-pandey:
Processing snipe targets from a configuration file on each tick would be awesome.

pokemongo_bot/cell_workers/snipe_pokemon.py
```diff
@@ -0,0 +1,119 @@
...
+import os
+import re
+import sys
+import time
+
+from . import PokemonCatchWorker
+from pokemongo_bot import logger
+from utils import distance
+
+class SnipePokemon(object):
```
TheSavior: This should extend BaseTask which gives you bot and config for free. Take a look at the other workers for an example

TheSavior:
@douglascamata Please review this

pokemongo_bot/cell_workers/snipe_pokemon.py
```diff
@@ -0,0 +1,112 @@
...
+    def initialize(self):
+        self.api = self.bot.api
+        self.pokemon_list = self.bot.pokemon_list
+        self.position = self.bot.position
+
+    def work(self):
+        if self.bot.tick_count is not 1:
+            return
+
+        while True:
```
TheSavior: This makes me super uncomfortable. I understand why it is here and what you are trying to do, but this is super yucky as a worker to have a while true that waits for user input. Perhaps @douglascamata's socket.io approach will be what we need to support this.

nikhil-pandey: @TheSavior Or we could use a configuration file with list of snipe targets and process them one by one and remove the targets as they are processed. Would be usable even when bot is running.

pokemongo_bot/cell_workers/snipe_pokemon.py
```diff
@@ -0,0 +1,112 @@
...
+                lambda x: distance(self.position[0], self.position[1], x['latitude'], x['longitude']))
+            catch_pokemon = self.cell['wild_pokemons'][0]
+
+        if not catch_pokemon:
+            logger.log('No pokemon found!', 'yellow')
+            time.sleep(delay)
+            # go back 
+            self.api.set_position(*prevPosition)
+            time.sleep(delay)
+            self.heartbeat()
```
nikhil-pandey:
```
AttributeError: 'SnipePokemon' object has no attribute 'heartbeat'
```

Should be `self.bot.heartbeat()`

enjoy2000:
@TheSavior @douglascamata @nikhil-pandey is something like `watchdog` & a config file is ok?

TheSavior:
I think I defer to @douglascamata here. I'd want to make sure what we do is compatible with the websockets stuff which is probably coming in the next day or two. It is also possible he would want to merge something like that in the mean time and then he can go back and update it. Not sure, I'll wait for him.

Nihisil:
I'm pretty sure we don't need to have user inputs in bot. It should be configured only from configs, because in that case we have ability to run it automatically by provision tools.

nikhil-pandey:
@Nihisil Right. I think we should use a configuration file with a list of destinations to snipe. The destinations are removed once they are processed. I don't exactly know what the plan is with websockets. I assume it will be used to perform actions from the web later on through API calls? If so, same config file can be updated later during API calls.

nikhil-pandey:
@douglascamata Could you please review this?

douglascamata:
Please put the snipe coordinates inside the worker config parameter.

douglascamata:
Also please consider adding the pokemon name along with each coordinate. This way snipping is more effective because it will search only for the pokemon you want to snipe and not for everything there..

Kassius509:
@Nihisil @nikhil-pandey @douglascamata  Are you guys suggesting we add the sniping coordinates to the already jam packed config file? If so I would have to disagree with you.

I believe the use case intended was to allow sniping of pokemon while the application was already running, allowing you to pass individual locations as you discover them without having to close the application then open, alter, and save the config.json and again relaunch the application each time. 

I like the idea of adding the pokemon name along with the coordinate, but I would suggest this remain an optional feature.

`./pokecli.py —snipe=True —location=1.232,1.232 —"Dragonite"`

enjoy2000:
@Kassius509 Maybe an option for config file location is what u need. You can share it among your Docker instances.

Nihisil:
@Kassius509 all flags in config.json can be overwrote by console arguments, it is a main approach (at least we try to do it).

BartKoppelmans:
Just an idea, maybe we could work together with [this site](http://www.pokespawns.be/) to use their database for sniping. That way we can just set sniping to true and automatically snipe the rare pokemon that get listed there.

nelsyeung:
I've tried this and it doesn't seem to work at all.
```
[12:02:13]  Found location: 40.773115629, -73.980903625
[12:02:14]  Teleport to location..
[12:02:15]  No pokemon found!
```

I've basically tested this with common pokemon that are scanned from [PokemonGo-Map](https://github.com/AHAAAAAAA/PokemonGo-Map), I used their provided coordinates with this and every one will return "No pokemon found!".

I've also tried the coordinates provided by [Rare Pokemon Spawns](http://www.pokespawns.be/), then confirm whether the Pokemon actually exists by putting the location into the [PokemonGo-Map](https://github.com/AHAAAAAAA/PokemonGo-Map), finally input the location to snipe_list.json, but again, always, "No pokemon found!".

What am I missing with this? If my coordinates are not specific enough, is it not possible to have a configuration to expand the searching for the teleported part? That might cause softban but you can have a brute force way of teleporting to and back for the search.

enjoy2000:
@nelsyeung fixed, because api was updated

enjoy2000:
@nelsyeung consider using this #1992 or checkout master branch on [my fork](https://github.com/enjoy2000/PokemonGo-Bot)

drees5:
i get this error
./pokecli.py --snipe=true
[15:11:45]  PokemonGO Bot v1.0 
[15:11:45]  No config argument specified, checking for /configs/config.json 
usage: pokecli.py [-h] [-cf CONFIG] [-a AUTH_SERVICE] [-u USERNAME]
                  [-ws WEBSOCKET.SERVER_URL]
                  [-wss WEBSOCKET.START_EMBEDDED_SERVER]
                  [-wsr WEBSOCKET.REMOTE_CONTROL] [-p PASSWORD] [-l LOCATION]
                  [-lc LOCATION_CACHE] [--forts.spin FORTS.SPIN] [-w WALK]
                  [-k GMAPKEY] [-d DEBUG] [-t TEST] [-du DISTANCE_UNIT]
                  [-ec EVOLVE_CAPTURED] [-rt RECONNECTING_TIMEOUT]
                  [-hr HEALTH_RECORD] [-ac FORTS.AVOID_CIRCLES]
                  [-mcs FORTS.MAX_CIRCLE_SIZE]
                  [--catch_randomize_reticle_factor CATCH_RANDOMIZE_RETICLE_FACTOR]
                  [--catch_randomize_spin_factor CATCH_RANDOMIZE_SPIN_FACTOR]
                  [--map_object_cache_time MAP_OBJECT_CACHE_TIME]
pokecli.py: error: unrecognized arguments: --snipe=true
(i have downloaded your fork)

JohnnyOmm:
What do I type in to snipe a pokemon or add the location? theres no example besides this
`   {
        "type": "SnipePokemon"
      },`

nelsyeung:
@JohnnyOmm If you use his master branch, you simply need to add that to the task and it will prompt you to enter a location. However, I myself can't really find pokemon that way reliably. I would also recommend #1992 if you want sniping.
@drees5 For his master branch, you don't need to specify a command line argument, you simply add `"type": "SnipePokemon"` to your `tasks`. If you are using his feature branch, you do the same but also you have to cp the `snipe_list.example.json` file to `snipe_list.json` then you run `python pokecli.py -spl configs/snipe_list.json` to make it work.

enjoy2000:
Remove argument `—snipe=True` pls, just need to add task `SnipePokemon` in `config.json`

Kozeo:
How do I test this?

JohnnyOmm:
@nelsyeung I used the regular dev version and im getting this error when I add it to config
`pokemongo_bot.tree_config_builder.ConfigException: No worker named SnipePokemon defined
`

tolinwei:
Just found this PR, since [PokemonGo-Map](https://github.com/AHAAAAAAA/PokemonGo-Map) has been ceased, is this still working? If yes, what is the source of the teleport list – I read the previous comments and haven't found any here

nelsyeung:
@tolinwei Check out this version of the [PokemonGo-Map](https://github.com/PokemonGoMap/PokemonGo-Map). If you want rare spawns try [PokeSpawns](http://pokespawns.be). Also check out [MoveToMapPokemon](https://github.com/PokemonGoF/PokemonGo-Bot/blob/dev/docs/configuration_files.md#sniping-movetolocation) feature that's already out for the less specific sniping feature.

k4n30:
Should this be closed as the code it's based off is pretty old, and the movetomap task config option exists until #3672 is implemented?

TheSavior:
Yeah, probably a good choice.

Thanks @enjoy2000, it will help provide us some base code to look at when building a new version.

PR #1899, opened 2016-07-30
Possibility to set another config with run script

03t02:


03t02:
Where can I add my name in "contributors" ? @CapCap thanks

PR #1904, opened 2016-07-30
Configuring tasks in the config.json

TheSavior:


TheSavior:
Declining for now. We need to do some more renames and refactors first.

PR #1917, opened 2016-07-30
Improved Path Navigator, Now Supports geopositioning resolution

asaf400:
Short Description: 
Refactored Path Navigator, Now Supports geopositioning resolution

path config json file format changed, new format allows string to use with geopositioning but also support a standard 'lat, lng'


douglascamata:
👍

PR #1923, opened 2016-07-30
Do a Heartbeat even when a step was short

mhdasding:
Short Description: 
When you supply a path for the PathNavigator and the waypoints are too dense the web-location json won't be updated.

Fixes:
- Do a heartbeat even when the StepWalker made a short step

### DIFF
--- a/pokemongo_bot/step_walker.py
+++ b/pokemongo_bot/step_walker.py
@@ -39,6 +39,7 @@ class StepWalker(object):
     def step(self):
         if (self.dLat == 0 and self.dLng == 0) or self.dist < self.speed:
             self.api.set_position(self.destLat, self.destLng, 0)
+            self.bot.heartbeat()
             return True
 
         totalDLat = (self.destLat - self.initLat)


douglascamata:
Is it really necessary?

mhdasding:
i use 13.5 km/h walk speed which makes it a lot easier to get caught in that if statement and it won't update the web ui location never again when the waypoints are a bit dense. I don't see a reason not to do a heartbeat in thise case.

douglascamata:
@mhdasding I have my speed set to 80 and I don't get any softban while botting.

mhdasding:
@douglascamata surely you won't get softbanned for using 80 as speed, but i doubt your eggs are hatching while speeding like this.

But that's not the issue here, the issue is that the web ui does not update when using dense gpx paths. That's why i would like to add this heartbeat.

douglascamata:
@mhdasding yes, they are.... because our speed is not m/s or km/h, it's an empirical number. We ALWAYS walks in small steps.

douglascamata:
update_web_location is going away with the websocket events I'll implement tomorrow tho

mhdasding:
Simple question then: do you want a heartbeat to happen on each bot.tick?

douglascamata:
@mhdasding not at every tick, it's a waste of time

mhdasding:
ok then this can be closed, i hope your websocket implementation will trigger the web location on each tick or at least with a lower boundary tick rate.

The info in pokecli.py line 201 
`"Walk instead of teleport with given speed (meters per second, e.g. 2.5)"`
might need to be updated if it's an empirical number.

PR #1932, opened 2016-07-30
Stale location cache

cwild:
If specifying a new location, don't mandate that location_cache be turned off explicitly.

Fixes:
- Change of --location will work without having to turn location cache off


douglascamata:
👍

TheSavior:
Great! 

👍

PR #1934, opened 2016-07-30
Fix Grammer Needs -> Need

helloworld:
Short Description: 

Fixes:
- "Checking if any of them needs to be released" -- changed "needs" to "need"
- 
-


TheSavior:
Thanks for the PR. Unfortunately, we don't merge PRs to master. If you could recreate this PR to dev (it may already be fixed!) that would be greatly appreciated.

PR #1937, opened 2016-07-30
updates spiral navs shape to be more... spiraly

mako-taco:
Short Description: 
The spiral navigator's points are definitely not spiral-shaped.

After patch:
image


douglascamata:
@mako-taco the SpiralNavigator does something like a squared-spiral... and we like it (I guess). So do you mind creating another Navigator do implement this perfect spiral route?

mako-taco:
oh, if it's working as intended I'll just close this. nbd

douglascamata:
@mako-taco you can feel free to create navigators with different routes, tho.

PR #1980, opened 2016-07-31
Added more accurate random and its model file.

kanemasa1987:
Short Description: 
Added and modified random functions to simulate a human player better by using other random generator.
However, parameters for log-normal are tricky (ex: for average 50ms and 20ms std mu=-3.00, sigma=0.132,) so I have added pre-calculated model file and its calculation code.

Fixes:
- Modified jitter to use log-normal instead of uniform random. (http://stats.stackexchange.com/questions/1315/how-do-i-figure-out-what-kind-of-distribution-represents-this-data-on-ping-respo)
- Modified gps error to use gaussian instead of uniform random.
- Added human reflex rng based on http://www.humanbenchmark.com/tests/reactiontime/statistics
- Added human reflex to sleep function.
- Parameter calculations are done this way (https://books.google.co.jp/books?id=CvUvxpSe8NoC&pg=PA436&lpg=PA436&dq=jitter+model+lognormal&source=bl&ots=5U6R6DHuXx&sig=zPKFk-lAEVjEXSjoW52zOutOj_A&hl=ja&sa=X&ved=0ahUKEwj7icbYupzOAhVKq48KHf89D-QQ6AEIJjAB#v=onepage&q=jitter%20model%20lognormal&f=false)

"Updated Fixes":
- Added somewhat okay rng for reticle and spin
- Added ballthrow delay function

Time and reflex distributions
time

reticle size distributions
reticle

spin distributions
spin


pokemongo_bot/human_behaviour.py
```diff
@@ -1,17 +1,175 @@
...
+        output_file_path='lognormal_connection.model'):
+    '''
+    Precalc parameters for lognormal ping model.
+    '''
+    def calc_lognormal_ping_param(
+            target_mu, target_std, init_mu=0.1, init_std=0.1, debug=False):
+        # calculate mu and sighma for lognormal
+        import numpy as np
+        import theano
+        import theano.tensor as T
```
DayBr3ak: Those aren't dependencies available

pokemongo_bot/human_behaviour.py
```diff
@@ -1,17 +1,175 @@
...
+    error = min(max(-radius, error), radius)
+    return error
+
+
+# Humanized `normalized_reticle_size` parameter for `catch_pokemon` API.
+# 1.0 => normal, 1.950 => excellent
+def normalized_reticle_size(factor):
+    minimum = 1.0
+    maximum = 1.950
+    return uniform(
```
DayBr3ak: Why not use your RNG for this too ?

kanemasa1987: Gps errors should be approximated by gaussian.

I'm not quite sure for reticle and spin rngs. 
Speaking of myself, I try to get simple 'Great' and don't throw spin balls because I'm bad at it.
However, that might be my preference and most of the users might be acting otherwise.
Besides, I'm also not certain what distributions are proper at this point.

It should be other than uniform to prevent us from getting banned in the future, but I decided to keep it as it is for now, before going deeper and messing around with the codes.

DayBr3ak:
Very nice work!

pokemongo_bot/human_behaviour.py
```diff
@@ -1,17 +1,219 @@
...
+    '''
+    Chances to skip the reticle should be constant,
+    so the wait time before throwing is binomial,
+    given that the monster does not interrupt... <- TODO
+    '''
+    for trial in range(10):
+        if random() < success_prob:
+            break
+    return trial
+    time.sleep(1.8*(trial+random()))
```
cwild: This sleep will never execute.

kanemasa1987: Oh, I forgot to remove the checking code! :bow:

kanemasa1987:
Wrong destination, closing and sending request again.

PR #1986, opened 2016-07-31
Re-enable item ID's in the item_filter.

Jasperrr91:
Short Description: 
Allows users to enter both item ID's and item names in the item_filter.

Fixes:
- Re-allow item ID's in item_filter


pokemongo_bot/cell_workers/recycle_items.py
```diff
@@ -7,6 +7,7 @@ def __init__(self, bot, config):
         self.bot = bot
 
     def work(self):
+
```
douglascamata: remove this empty line, dude...

PR #1989, opened 2016-07-31
Let the user know that the maps api key exceeded its limits.

wcastello:
Short Description: When the maps api key exceeds the maximum number of requests geopy.exc.GeocoderQuotaExceeded is raised, but main() on pokecli.py won't catch it and will just raise a not very meaninful NotLoggedInException(). This change lets the user know it's time to get a new key. 

This: 
screen shot 2016-07-30 at 9 00 21 pm

Becomes this:
screen shot 2016-07-30 at 9 00 39 pm

### DIFF
--- a/pokecli.py
+++ b/pokecli.py
@@ -36,6 +36,7 @@ import time
 from datetime import timedelta
 from getpass import getpass
 from pgoapi.exceptions import NotLoggedInException
+from geopy.exc import GeocoderQuotaExceeded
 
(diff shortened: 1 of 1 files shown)


solderzzc:
Better warning style :)

solderzzc:
We can change it a little bit to let user use the (xx.xxx, xx,xxx) format.

TheSavior:
:+1:

surfaace:
:+1:

PR #2000, opened 2016-07-31
Foundation for remote control of the bot over websocket

douglascamata:


pokemongo_bot/websocket_remote_control.py
```diff
@@ -0,0 +1,39 @@
...
+
+    def __init__(self, bot):
+        self.bot = bot
+        self.host, port_str = self.bot.config.websocket_server_url.split(':')
+        self.port = int(port_str)
+        self.sio = SocketIO(self.host, self.port)
+        self.sio.on('bot:process_request', self.on_remote_command)
+
+    def start(self):
+        self.thread.start()
```
TheSavior: is `self.thread` defined?

pokemongo_bot/websocket_remote_control.py
```diff
@@ -0,0 +1,39 @@
...
+        return self
+
+    def process_messages(self):
+        while True:
+            self.sio.wait(5)
+
+    def on_remote_command(self, command):
+        name = command['name']
+        command_handler = getattr(self, name)
```
TheSavior: There should probably be a check that this is defined. Otherwise it seems like a security risk.

pokemongo_bot/event_handlers/logging_handler.py
```diff
@@ -4,7 +4,10 @@
 
 
 class LoggingHandler(EventHandler):
-    def handle_event(self, event, sender, level, data):
+    def handle_event(self, event, sender, level, formatted_msg, data):
```
TheSavior: What do we need this `formatted_msg` for? It seems like it would make logging more complicated if things are sending event, formatted_msg and data.

pokemongo_bot/websocket_remote_control.py
douglascamata: oh shit, I deleted a line

pokemongo_bot/event_handlers/logging_handler.py
douglascamata: because we might want to have logger messages with real text instead of just parameters for events

pokemongo_bot/websocket_remote_control.py
```diff
@@ -0,0 +1,49 @@
...
+        self.sio.on('bot:process_request', self.on_remote_command)
+        self.thread = threading.Thread(target=self.process_messages)
+
+    def start(self):
+        self.thread.start()
+        return self
+
+    def process_messages(self):
+        while True:
+            self.sio.wait(5)
```
TheSavior: are you missing a line in here? something to call on_remote_command or something?

douglascamata: `self.sio.wait(5)` will process messages based on the callback I registered:

`self.sio.on('bot:process_request', self.on_remote_command)`

solderzzc:
👍

TheSavior:
:+1:

PR #2002, opened 2016-07-31
Always run Bot, if something goes wrong, just run again...

robertpro:
Short Description: 

Sometimes bot dies if the user lost connection (pgoapi.exceptions.**ServerBusyOrOfflineException**) , so just run again :)

### DIFF
--- a/run.sh
+++ b/run.sh
@@ -15,4 +15,7 @@ else
     fi
 fi
 
-python pokecli.py --config ${config}
+while [ 1 ]; do  # Always run, if something goes wrong, just run again...
+    python pokecli.py --config ${config}
+    sleep 30
+done
--- a/run_forever.sh
+++ b/run_forever.sh
@@ -15,4 +15,7 @@ else
     fi
 fi
 
-python pokecli.py --config ${config}
+while [ 1 ]; do  # Always run, if something goes wrong, just run again...
+    python pokecli.py --config ${config}
+    sleep 30
+done


douglascamata:
Add this to another file with a proper name, like run_forever.sh

robertpro:
@douglascamata can I add a new param ? something like.. 

./run.sh -f/--forever ?

Zgrkaralar:
My forever.sh like this i'm using ubuntu for this project.
```

exec 1> >(logger -s -t $(basename $0)) 2>&1

cd /var/www/PokemonGo-Bot/

if ps aux | grep "pokecli.py -cf configs/yourconfig.json" | grep -v grep; then
    echo Found Bot
else
    echo Not Found Bot
    echo Starting Now
    nohup python /var/www/PokemonGo-Bot/pokecli.py -cf configs/yourconfig.json &
    echo Started
fi

```
I'm testing with 9 bots working perfect.

I create one cron every minutes checking **running my forever.sh** this checking first **if ps aux | grep "pokecli.py -cf configs/youconffile"**

For seeing log i'm using this command
**tail -f grep forever /var/log/syslog**

nikhil-pandey:
Or you could use something like `supervisor` to run the bot.
Here's what my config looks like
```
[program:pgo-user1]
process_name=%(program_name)s
command=python /home/pi/bot/pokecli.py --config ./configs/user1.json
directory=/home/pi/bot
environment=PATH="/home/pi/bot/bin"
autostart=true
autorestart=true
user=pi
numprocs=1
redirect_stderr=true
stdout_logfile=/home/pi/bot/logs/user1.log
```

Clownvin:
I just modified the code ever so slightly so that on any exception other than KeyboardInterrupt, it'll just restart from the beginning. So far works perfectly fine.

Pacane:
I think this would be better. (Taken from my own script)

```bash
#!/usr/bin/env bash

# Starts PokemonGo-Bot
config=""

if [ ! -z $1 ]; then
    config=$1
else
    config="./configs/config.json"
    if [ ! -f ${config} ]; then
        echo -e "There's no ./configs/config.json file"
        echo -e "Please create one or use another config file"
        echo -e "./run.sh [path/to/config/file]"
        exit 1
    fi
fi

until (python pokecli.py -cf ${config}); do
    echo "App crashed with exit code $?.  Respawning.." >&2
    sleep 1
done
```

robertpro:
@douglascamata I just created that new file :)

HecticSeptic:
I just use this:

```
import subprocess
import time
import random

bots = ["config", "configForBot2", "etc"] ## each .json config file
times = [10, 30, 30, 120] ## min and max for poll and restart delay

def newBot(bot): ## this should only be done if we're sure it isn't running
    return subprocess.Popen(["python.exe", "pokecli.py", "-cf", "./configs/"\
                             + bot + ".json"], creationflags=\
                            subprocess.CREATE_NEW_CONSOLE)

sp = [newBot(bot) for bot in bots]

while (True):
    for bot in bots:
        i = bots.index(bot)
        poll = sp[i].poll()
        if (poll is not None):
            print bot + " appears to have stopped. Code: " + str(poll)
            print "waiting a bit before restarting..."
            time.sleep(random.randint(times[0], times[1]))
            print bot + ": restarting " + str(sp[i].pid)
            sp[i].kill() ## in case needed, spawning dupes will be bad
            sp[i] = newBot(bot)
        else: print bot + " still in use."
    time.sleep(random.randint(times[2], times[3]))
    ## wait a bit, if a bot died, resurrect it
````

**Comment by nanogirl21** [comment]
I use this to automatically restart my bot if it crash. 

```
#!/usr/bin/env bash

# Assumes the config file is created and named config.json

# Starts PokemonGo-Bot After Crash
config=""
loops=0

if [ ! -z $1 ]; then
    config=$1
else
    config="./configs/config.json"
    if [ ! -f ${config} ]; then
        echo -e "There's no ./configs/config.json file"
        echo -e "Please create one or use another config file"
        echo -e "./run.sh [path/to/config/file]"
        exit 1
    fi
fi
while true
do
  SECONDS=0
  python pokecli.py --config ${config}
  ((loops++))
  echo -e
  echo -e "RESTARTING"
  echo -e "Loop Iterations: " $loops
  duration=$SECONDS
  echo -e "Runtime was $(($duration / 60)) minutes and $(($duration % 60)) seconds"
  echo -e
  rm web/location*.json
  sleep 1
  
done
```

matejvelikonja:
and if you are using docker, you can just use restart:always

RussellNS:
I don't know if this is helpful for Windows users, but this command at a CMD prompt will simply re-run the bot, over and over, regardless of how it exited:

`for /L %x in (1,0,10) do python pokecli.py -cf .\configs\config.json`

bibibkamal:
I'm sorry but I'm new at this. How to do this? i have no idea to put this into my bot. Pls help :)

RussellNS:
@bibibkamal

I believe the issue that spawned this thread of responses has been fixed.  So short answer is, you shouldn't have to do these things.  However, if you're referencing the code from the previous users, then they are talking about writing a script outside the PokemonGo-Bot code.  Then they are using that external script to run the PokemonGo-Bot.  However, the 1-liner that I posted wasn't a script.  It was just a single line to be used to start the PokémonGo-Bot.  The logic in that 1 line was re-run the PokemonGo-Bot anytime it exited.

Again, you shouldn't need to do any of these things now.  Hope your PokemonGo experience is awesome!

middleagedman:
Real simple
while true; do python2 pokecli.py --config ./configs/config.json; echo "Bot Failed, restarting" | mail -s"Restart" someemail@somewhere.com; sleep 60; done

Nihisil:
Current version of run.sh is doing pretty same thing.

https://github.com/PokemonGoF/PokemonGo-Bot/blob/dev/run.sh

So, I will close this PR. @robertpro if you think that your version is better, you are welcome to send a new PR with your fixes :)

PR #2007, opened 2016-07-31
Tasks now extend a base task

TheSavior:


solderzzc:
Classify is very clear. 
👍

PR #2018, opened 2016-07-31
Workaround Empty Forts

loklam:
Short Description: Sometimes, the API return empty forts. This causes move_to_fort fail. To workaround this, return prev forts, if API return empty forts.

Fixes:
- 
- 
-

### DIFF
--- a/pokemongo_bot/__init__.py
+++ b/pokemongo_bot/__init__.py
@@ -111,7 +111,7 @@ class PokemonGoBot(object):
 
         if (len(forts) == 0):
             forts = oldForts
-            logger.log('using old forts list')
+#            logger.log('using old forts list')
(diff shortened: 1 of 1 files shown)


Sammy-L:
I've tested this code. It works very well. Niantic changed the search respond rate to about 4s since last patch. So the fort scan nearly always return empty result. Saving the old result does help for situations while server refuse to give response.

kloz8879:
I Have also tested this and it works great! Thank you.
2016-08-01_07h15_08

z4ppy:
Work for me !

loklam:
This has been fixed at https://github.com/PokemonGoF/PokemonGo-Bot/pull/2269

PR #2032, opened 2016-07-31
Add option to spin fort even if inventory is full

kbinani:
Short Description:
---
Added new option `move_to_fort_even_if_full_inventory` which allows bot to move to forts and spin it for XP, even if inventory is full.

Fixes:
---
- Bot cannot earn XP by spinning forts in case inventory is full.
- Minor resource leak in `pokemongo_bot#__init__`, by using `with` for `json.load`.
- Minor logger bug: unexpected space char included in colored log message.

### DIFF
--- a/configs/config.json.example
+++ b/configs/config.json.example
@@ -74,6 +74,7 @@
     "evolve_captured": "NONE",
     "catch_randomize_reticle_factor": 1.0,
     "catch_randomize_spin_factor": 1.0,
+    "move_to_fort_even_if_full_inventory": false,
     "catch": {
       "any": {"catch_above_cp": 0, "catch_above_iv": 0, "logic": "or"},
       "// Example of always catching Rattata:": {},
--- a/configs/config.json.path.example
+++ b/configs/config.json.path.example
@@ -37,7 +37,10 @@
         "type": "CatchLuredPokemon"
       },
       {
-        "type": "SpinFort"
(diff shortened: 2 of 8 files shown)


configs/config.json.example
```diff
@@ -74,6 +74,7 @@
     "evolve_captured": "NONE",
     "catch_randomize_reticle_factor": 1.0,
     "catch_randomize_spin_factor": 1.0,
+    "move_to_fort_even_if_full_inventory": false,
```
TheSavior: Can you make this a task level configuration like you did for SpinFort?

kbinani: This option is added here because it is used from both `SpinFort` and `MoveToFort` tasks (by calling `PokemonGoBot#should_move_to_fort_for_loot` method).
Is there better way to share a config between tasks?

TheSavior: Hmm. Got it.

Perhaps the tasks don't actually need to share the configuration. SpinForts could have a `min_space_required` config that could be set to 0 and defaults to the 5 it is at now. It actually looks like you pretty much have that with the `spin_even_if_inventory_full` boolean. Although it doesn't look like that is actually used. In that case, the change to spin would be completely decoupled from moving to forts.

kbinani:
Close this PR.
I am :+1:  to https://github.com/PokemonGoF/PokemonGo-Bot/pull/2188

PR #2044, opened 2016-07-31
Move follow path task config

achretien:
Short Description:

We can now config FollowPath task in the task config node


TheSavior:
Thank you very much for this PR @achretien!

PR #2065, opened 2016-07-31
Output pokemon move name when catching

enjoy2000:
Short Description: 

Output pokemon move name when catching


knobunc:
Looks good to me!  Nice idea too.

surfaace:
:+1:

surfaace:
same as #2117, but that's more detailed

Nihisil:
In the current dev version we have information about pokemon moves inside Pokemon object.

So, if you want to print it when you catch a pokemon, you can do it easily with a new PR.

PR #2073, opened 2016-07-31
Better enforce rules about Pokemons to retain.

giomasce:
Rules about maximum CP and IV to transfer Pokemons are applied
also when keeping the best ones.

### DIFF
--- a/pokemongo_bot/cell_workers/transfer_pokemon.py
+++ b/pokemongo_bot/cell_workers/transfer_pokemon.py
@@ -41,7 +41,12 @@ class TransferPokemon(BaseTask):
                                 all_pokemons.remove(pokemon)
                                 best_pokemons.append(pokemon)
 
-                    if best_pokemons and all_pokemons:
+                    transfer_pokemons = [pokemon for pokemon in all_pokemons
+                                         if self.should_release_pokemon(pokemon_name,
+                                                                        pokemon['cp'],
+                                                                        pokemon['iv'])]
+
+                    if transfer_pokemons:
                         logger.log("Keep {} best {}, based on {}".format(len(best_pokemons),
                                                                          pokemon_name,
                                                                          order_criteria), "green")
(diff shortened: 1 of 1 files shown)


douglascamata:
@giomasce did you test?

giomasce:
I am using this change right now (three parallel bots) and it appears to work. It does what I expect it to do.

douglascamata:
👍

solderzzc:
👍

matrix14:
Can you explain how to use it? 

I want to have all pokemons above 500 CP, but when I have pokemon under 500CP i want to keep best.
Is it possible?

giomasce:
@matrix14 Yes, that is how it is meant to be used.I have this release rule:

    "any": {"release_below_cp": 400, "release_below_iv": 0.8, "logic": "or", "keep_best_iv": 2},

It means: always keep the best two Pokemons by IV. Release the others, but only if they are below 400 CP or 0.8 IV.

mzygmunt:
@giomasce I think the logic of this feature should be:
Release all Pokes if they are below 400 CP or 0.8 IV and keep only 2 by IV from the rest.

valerian:
@mzygmunt @giomasce 
The problem is that it's hard to interpret by simply reading the rule if you don't know how it's supposed to work in detail, and that's why I was told that such a feature wouldn't be done yet.
And then it was done anyway, and with a bug :/

mzygmunt:
@valerian Those features are important for account with 25 levels or above, when we want to catch a lot of good Pokes while XP'ing. I know that, the features might be difficult to be done without config complication, but I think, is worth spending it for bot progress.

bixuanzju:
@mzygmunt Reading from [here](https://github.com/PokemonGoF/PokemonGo-Bot/blob/dev/pokemongo_bot/cell_workers/transfer_pokemon.py#L17), @matrix14 is right about interpreting the logic: when you have both ""release_below_x" and "keep_best_x", the bot filters the best pokemons to keep, and then for the rest, apply the logic of "release_below_x" to decide the set of pokemons to release.

PR #2105, opened 2016-07-31
[FIX-Config] 'use_lucky_egg' should not be true in the exemple config file

FilipiMor:
Short Description: 
Just to avoid people complaining, and really this config is so long that some times it's easy to lose track of a undesirable value.
Fixes:
- 'use_lucky_egg' should not be true in the exemple config file
- changed 'use_lucky_egg' to false in all config exemple files.
-


TheSavior:
:+1:

PR #2113, opened 2016-07-31
Fix using -1 pokeball to catch. Add a logic if user don't set VIP pokemons, then don't save their Ultra ball.

xc-mezcal:
Short Description: 

Fix using -1 pokeball to catch. Add a logic if user don't set VIP pokemons, then don't save their Ultra ball.

Fixes:
1.Check the stock before throwing a ball [fix the -1 pokeball throwing]

2.Check if user set VIPs pokemons in config, if not, don't try to save Ultra Ball, just get them all

### DIFF
--- a/pokecli.py
+++ b/pokecli.py
@@ -382,6 +382,9 @@ def init_config():
     config.action_wait_min = load.get('action_wait_min', 1)
     config.raw_tasks = load.get('tasks', [])
     config.vips = load.get('vips',{})
+    config.vip_strategy = True
+    if not config.vips:
+        config.vip_strategy = False
 
     if len(config.raw_tasks) == 0:
         logging.error("No tasks are configured. Did you mean to configure some behaviors? Read https://github.com/PokemonGoF/PokemonGo-Bot/wiki/Configuration-files#configuring-tasks for more information")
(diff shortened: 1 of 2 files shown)


xc-mezcal:
Got to go, let me know if you find any bugs.

douglascamata:
What if there is no VIP but I still want to save ultraballs for higher CP/IV pokemon?

xc-mezcal:
@douglascamata The problem you describes actually match exactly for setting {any: cp_above xxx, iv above xxx, 'or' } in VIP section.

douglascamata:
@xc-mezcal move the vip strategy parameter to a config of the CatchPokemon behaviour, like the IncubateEgg behaviour does.

xc-mezcal:
@douglascamata  Will do this later, currently im out now

Kassius509:
What if I only want to save a portion of my ultra balls, like 30, to catch VIP's and the rest could be used at will?

k4n30:
@xc-mezcal this merge is severly outdated. How do you want to proceed? Some of your functionality is already present in the current dev code

xc-mezcal:
@k4n30  This merge should be closed cause of its outdated. :)

PR #2114, opened 2016-07-31
Show pokemon moves when catching

iananass:
Example:
A Wild Staryu appeared! [CP 515] [Potential 0.67] [Quick Attack(7.51 DPS)] [Power Gem(13.79 DPS)]


iananass:
Some merge issue, need to recreate commit, sorry

PR #2117, opened 2016-07-31
Show pokemon moves when catching

iananass:
Example:
A Wild Staryu appeared! [CP 515] [Potential 0.67] [Quick Attack(7.51 DPS)] [Power Gem(13.79 DPS)]


TheSavior:
Why is this important? It doesn't effect the bot catching it, right?

Nihisil:
I think this PR is allow to see pokemon attacks, the first and the second with dps. It can be useful to know what is attack dps for your pokemon, when you will fight a gym in real life, but maybe it should be added to the web version?

kanemasa1987:
It should be useful for rename/transfer/gym battles in the future.

Nihisil:
@kanemasa1987 gym battles will not be implemented in this bot

kanemasa1987:
oh, then maybe rename and transfer decision making etc.

iananass:
This feature have is only UI. I think that DPS is very important
As written above in future it could be very useful into transfer decision making. Because pokemon with low CP you can always power up and get strong pokemon. You can do nothing with pokemon which have weak attack to make it strong.

TheSavior:
I don't think changing the log really matters, people are unlikely to see something show up in there log and manually do something with that pokemon. I actually think it is a negative change to the current log statement as it doesn't matter to anyone when catching pokemon.

I do think have the JSON blob is valuable for other workers like transferring or evolving pokemon. I'd be happy to merge that JSON blob so that we can find a more positive way to use this data.

My preference would be for us to not change the log message on catch from where it is now though.

Nihisil:
In the current dev version we have information about pokemon moves inside Pokemon object.

So, if you want to print it when you catch a pokemon, you can do it easily with a new PR.

PR #2130, opened 2016-07-31
Fix throw type always normal when trying to catch pokemon

Likael:
Short Description: 
Fix the catch pokemon call to match the API one.

Fixes:
- Throw are now rightly typed, i.e. Excellent, Great, Nice, etc...


xc-mezcal:
Thx for fixing!

Issue #2104, opened 2016-07-31
[dev] handle_soft_ban.py error | one argument missing in __init__()


toremielck:
Traceback (most recent call last):
  File "pokecli.py", line 494, in <module>
    main()
  File "pokecli.py", line 70, in main
    bot.tick()
  File "/Users/toremielck/Desktop/PokemonGo-Bot-kajoschooPTC/pokemongo_bot/__init__.py", line 85, in tick
    if worker.work() == WorkerResult.RUNNING:
  File "/Users/toremielck/Desktop/PokemonGo-Bot-kajoschooPTC/pokemongo_bot/cell_workers/handle_soft_ban.py", line 31, in work
    MoveToFort(self.bot).work()
TypeError: __init__() takes exactly 3 arguments (2 given)

bigkraig:
this should fix it
```
diff --git a/pokemongo_bot/cell_workers/handle_soft_ban.py b/pokemongo_bot/cell_workers/handle_soft_ban.py
index 0cfa700..fd12b13 100644
--- a/pokemongo_bot/cell_workers/handle_soft_ban.py
+++ b/pokemongo_bot/cell_workers/handle_soft_ban.py
@@ -28,7 +28,7 @@ class HandleSoftBan(BaseTask):
         )
 
         if fort_distance > Constants.MAX_DISTANCE_FORT_IS_REACHABLE:
-            MoveToFort(self.bot).work()
+            MoveToFort(self.bot, self.bot.config).work()
             self.bot.recent_forts = self.bot.recent_forts[0:-1]
             if forts[0]['id'] in self.bot.fort_timeouts:
                 del self.bot.fort_timeouts[forts[0]['id']]
```

toremielck:
How do i apply the fix?

bigkraig:
sorry, i think its this. open up the file and change the line with a - to look like the line with the +.

```
diff --git a/pokemongo_bot/cell_workers/handle_soft_ban.py b/pokemongo_bot/cell_workers/handle_soft_ban.py
index 0cfa700..f112310 100644
--- a/pokemongo_bot/cell_workers/handle_soft_ban.py
+++ b/pokemongo_bot/cell_workers/handle_soft_ban.py
@@ -28,7 +28,7 @@ class HandleSoftBan(BaseTask):
         )
 
         if fort_distance > Constants.MAX_DISTANCE_FORT_IS_REACHABLE:
-            MoveToFort(self.bot).work()
[...]

PR #2136, opened 2016-07-31
fix for #2104

bigkraig:
Short Description: 
Fix TypeError
```
File "pokecli.py", line 494, in 
main()
File "pokecli.py", line 70, in main
bot.tick()
File "/Users/toremielck/Desktop/PokemonGo-Bot-kajoschooPTC/pokemongo_bot/init.py", line 85, in tick
File "/Users/toremielck/Desktop/PokemonGo-Bot-kajoschooPTC/pokemongo_bot/cell_workers/handle_soft_ban.py", line 31, in work
MoveToFort(self.bot).work()
TypeError: init() takes exactly 3 arguments (2 given)
```
Fixes:
- Worker needs the config param

### DIFF
--- a/pokemongo_bot/cell_workers/handle_soft_ban.py
+++ b/pokemongo_bot/cell_workers/handle_soft_ban.py
@@ -28,7 +28,7 @@ class HandleSoftBan(BaseTask):
         )
 
         if fort_distance > Constants.MAX_DISTANCE_FORT_IS_REACHABLE:
-            MoveToFort(self.bot).work()
+            MoveToFort(self.bot, self.config).work()
             self.bot.recent_forts = self.bot.recent_forts[0:-1]
             if forts[0]['id'] in self.bot.fort_timeouts:
                 del self.bot.fort_timeouts[forts[0]['id']]


pokemongo_bot/cell_workers/handle_soft_ban.py
```diff
@@ -28,7 +28,7 @@ def work(self):
         )
 
         if fort_distance > Constants.MAX_DISTANCE_FORT_IS_REACHABLE:
-            MoveToFort(self.bot).work()
+            MoveToFort(self.bot, self.config).work()
```
TheSavior: If MoveToFort takes configuration, this will cause problems since it won't be getting the correct config values. self.config is specific to the task. This is why tasks shouldn't be calling other tasks.

@douglascamata Any ideas on a fix?

douglascamata: the `config` here needs to have the same structure of the `config` hash in JSON

TheSavior: Agreed, I'm not sure how we fix this though. HandleSoftBan doesn't (and shouldn't) know what the config is for MoveToFort. Is there any way we can refactor HandleSoftBan to not require anything about MoveToFort? Or do we just say screw it create the config that MoveToFort wants (ignoring however the user might have configured that task) and use it?

douglascamata: `HandleSoftBan` will decide how it wants to move to the fort.

SMWARREN: See Issue: #1522 

I just replaced (self.bot, self.config) with (self, self.bot).

TheSavior:
This has been fixed via another PR.

PR #2137, opened 2016-07-31
Warns if there aren't sufficient space left for loot.

leoi11:
Short Description: 
When space is insufficient, it sends out requests to Niantic even faster than before
becuase tasks are querying Niantic and doing noops. More server busy (error 52)
errors appear in the log without an explanation of what's actually
going on.

Fixes:
- MoveToFort, SpinFort, etc terminate silently
- Users do not notice that MoveToFort/SpinFort is skipping due to the lack of space.
- Causing error code 52 as a indirect result
- Misleading error message (Server busy)

### DIFF
--- a/pokemongo_bot/cell_workers/move_to_fort.py
+++ b/pokemongo_bot/cell_workers/move_to_fort.py
@@ -8,9 +8,10 @@ from utils import distance, format_dist, fort_details
 
 class MoveToFort(BaseTask):
     def should_run(self):
-        if not self.bot.has_space_for_loot():
+        has_space_for_loot = self.bot.has_space_for_loot()
+        if not has_space_for_loot:
             logger.log("Not moving to any forts as there aren't enough space. You might want to change your config to recycle more items if this message appears consistently.", 'yellow')
-        return (self.bot.has_space_for_loot()) or self.bot.softban
+        return has_space_for_loot or self.bot.softban
 
     def work(self):
         if not self.should_run():
--- a/pokemongo_bot/cell_workers/move_to_fort.py
+++ b/pokemongo_bot/cell_workers/move_to_fort.py
@@ -8,6 +8,8 @@ from utils import distance, format_dist, fort_details
 
 class MoveToFort(BaseTask):
     def should_run(self):
+        if not self.bot.has_space_for_loot():
+            logger.log("Not moving to any forts as there aren't enough space. You might want to change your config to recycle more items if this message appears consistently.", 'yellow')
         return (self.bot.has_space_for_loot()) or self.bot.softban
 
     def work(self):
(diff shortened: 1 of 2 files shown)


pokemongo_bot/cell_workers/recycle_items.py
```diff
@@ -29,6 +29,9 @@ def work(self):
                 else:
                     logger.log("-- Failed to discard " + item_name, 'red')
 
+        if not self.bot.has_space_for_loot:
+            logger.log("-- Inventory is still lacking space after recycling.. You might want to fix your config to recycle more things.", 'yellow')
```
TheSavior: I don't think we need the `-- ` and you probably don't want the double period between your sentences.

douglascamata:
@leoi11 isn't better to warn if inventory is fall in places where inventory matters, like SpinFort?

douglascamata:
@leoi11 can you do the same for pokemon bag size and CatchPokemon, please?

TheSavior:
The goal isn't to make the bot fix itself, the goal is to inform users that their configuration is broken. In that case I think this should be an error in SpinFort and CatchPokemon which will keep the task from running altogether.

pokemongo_bot/cell_workers/move_to_fort.py
```diff
@@ -8,6 +8,8 @@
 
 class MoveToFort(BaseTask):
     def should_run(self):
+        if not self.bot.has_space_for_loot():
+            logger.log("Not moving to any forts as there aren't enough space. You might want to change your config to recycle more items if this message appears consistently.", 'yellow')
```
TheSavior: I think you can just return False after this line in both workers and you'll have the behavior we are looking for.

leoi11:
Are you guys okay if we address the case for CatchPokemon in a separate PR?

douglascamata:
👍

leoi11:
Addressed @TheSavior's https://github.com/PokemonGoF/PokemonGo-Bot/pull/2137#discussion_r72910050 with https://github.com/PokemonGoF/PokemonGo-Bot/pull/2137/commits/38a82a5845b6829fd23ed91fd8338b742dcf91a4.

I prefer
```
        has_space_for_loot = self.bot.has_space_for_loot()
        if not has_space_for_loot:
            logger.log("Not moving to any forts as there aren't enough space. You might want to change your config to recycle more items if this message appears consistently.", 'yellow')
        return has_space_for_loot or self.bot.softban
```
over
```
        if not self.bot.has_space_for_loot():
            logger.log("Not moving to any forts as there aren't enough space. You might want to change your config to recycle more items if this message appears consistently.", 'yellow')
            return True
        return self.bot.softban
```
because I can read all the conditions in one (last) line, but I'm open to suggestions.

TheSavior:
:+!:

douglascamata:
👍

Issue #2070, opened 2016-07-31
New behaviour: total amount of pokeballs.


Breakster:
I added the following behaviour. It allows you to specify an overall amount of balls to keep. If you exceed the amount, the bot first tries to ditch pokeballs, then greatballs and so on. See below for an example. If anyone is interested i can submit a pull request.

[16:28:53] Now at Pokestop: ####
[16:28:53] Spinning ...
[16:28:53] Loot:
[16:28:53] 50 xp
[16:28:53] - 1x Pokeball (Total: 101)
[16:28:53] - 1x Greatball (Total: 33)
[16:28:53] - 1x Razz Berry (Total: 10)
[16:28:53] PokeStop on cooldown. Time left: 4 minutes, 59 seconds
[16:28:56] Keeping 150 balls in total
[16:28:56] -- Recycled 2x Pokeball

knobunc:
YES PLEASE!  Same for potions, etc.!

Parth6182:
Can you also help me keep a specific amount of greatballs? the same way it keeps only 100 pokemballs ( i.e.. whats the code for a great ball and ultra ball )

PR #2140, opened 2016-07-31
Adding an Anyball item to the release config

Breakster:
Short Description: 
An extra item type is added: "Anyball". This allow a 'keep' amount of anyball. When the total amount of balls exceeds this number, firstly pokeballs will be removed, then greatballs, and so on. (Also see issue #2070)

Fixes:
- 
- 
-

### DIFF
--- a/configs/config.json.example
+++ b/configs/config.json.example
@@ -62,7 +62,7 @@
     "distance_unit": "km",
     "reconnecting_timeout": 15,
     "item_filter": {
-      "Pokeball":       { "keep" : 100 },
+      "Anyball":        { "keep" : 150 },
       "Potion":         { "keep" : 10 },
       "Super Potion":   { "keep" : 20 },
       "Hyper Potion":   { "keep" : 30 },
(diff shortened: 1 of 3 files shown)


douglascamata:
👍 

@TheSavior have a look at this. LGTM,

klingan:
This would certainly minimalize the issue where you stop moving to forts if your limits arent reached yet but your bag is full nevertheless.

TheSavior:
👍

klingan:
So what happened to this? 

pokemongo_bot.tree_config_builder.ConfigException: item Anyball does not exist, spelling mistake? (check for valid item names in data/items.json)

Breakster:
So this one was reverted because there was an error when the key was not present. How to proceed? I still think it's a good feature. Also, it can be extended for potions and heals.

TheSavior:
@Breakster, yeah, a new PR would be great fixing it in a way that doesn't break other things would be great. See the comments on the revert: https://github.com/PokemonGoF/PokemonGo-Bot/pull/2166

Breakster:
I'm on it!

PR #2143, opened 2016-07-31
Allow for 3-7 decimal points in latitude / longitude

L422Y:
The current regular expression doesn't allow for less than six decimal points, update regex.


TheSavior:
We don't accept pull requests to master. Please change the destination to dev.

PR #2145, opened 2016-07-31
Quickspin implementation.

UnicoRush:
Quickspin allows to unban character.


douglascamata:
This feature is already present in the `HandleSoftBan` task.

PR #2150, opened 2016-07-31
Added check for valid keep_best_iv amount

iGaskin:
Short Description: 
Some folks have accidentally released high IV pokemon due to invalid "keep best IV" amount in their config file

Fixes:
- 
- 
-


Donwulff:
I imagine you meant to check & complain about "less than 1", not greater than 1.

PR #2153, opened 2016-07-31
[FIX] Cluster Selection

MFizz:
Short Description: 
added secondary criteria (lure_expires_in) for cluster selection, so it doesn't jump from cluster to cluster when they're equally large 

Fixes:
At this time the bot can switch destiny from tick to tick when there are several max clusters equally large

### DIFF
--- a/configs/config.json.cluster.example
+++ b/configs/config.json.cluster.example
@@ -1,4 +1,10 @@
-"tasks": [
+{
+    "auth_service": "google",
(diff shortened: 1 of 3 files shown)


TheSavior:
:+1:

TheSavior:
It looks like it is still jumping between clusters and walking long distances for clusters of the same size.

MFizz:
Can you show me some logs?

PR #2161, opened 2016-08-01
Logger tidy up

fr3ky:
Minor logger changes. 
- Forts are now called PokeStops
- Some number formatting
- Fort names instead of IDs
- Purple color for catching Pokemon [works well both on black/blue background]
- Latest SubProject commit (OpenPoGo Web UI)


douglascamata:
I'm rewritting logs with events now. Won't need this.

PR #2173, opened 2016-08-01
Replace all `logger.log` calls with events!

douglascamata:
All events generate log entries and also websocket messages!

## Future logs

Logs in tasks should use the `self.emit_event` inherited from `BaseTask`. Try to avoid adding many events with information that doens't matter. Remember to register your events in  `pokemongo_bot/__init__.py` where the events are registered. 

## Websocket messages

Run `./pokecli.py --show_events` to see all websocket events the bot will emit and the information they give.

## Attention!

No more `logger.log` call in the tasks. Everything that is logged should be a registered event and then tasks will emit them. 

## Config to start websocket server

This config will start the embedded websocket server automatically with the bot at `localhost` and on port `4000`. It will also enable remote control to get player information.

```
"websocket": {
        "start_embedded_server": true,
        "server_url": "localhost:4000",
        "remote_control": true
    },
```

## Multibot instructions

Multibot needs just one websocket server. So you can set `start_embedded_server` to false and use the utility script included to start a websocket server:

```
./ws_server.py --host localhost --port 4000
```

## Remote control example: get player information:

```
def print_response(msg):
    print msg

socketio.emit('remote:send_request', {'name': 'get_player_info', 'account': 'bot@email.com'})
soccketio.on('get_player_info:bot@email.com', print_response)
```


douglascamata:
Need to add something, w8.

eggins:
:+1:

douglascamata:
There are still a few events to add. I'll update tomorrow.

solderzzc:
Nice, still no Conflict. 👍

TheSavior:
Tests are failing. :)

douglascamata:
@TheSavior I can confirm it's problem with float number precision. My OS X machine and my Debian VM have different coordinates is result for clique algorithm (this is where test fails).

TheSavior:
@douglascamata If that is the case, I believe we have a test util for verifying that two floats are "close enough". Perhaps we should be using that wherever the test is currently failing.

douglascamata:
@TheSavior problem is that 2 floats being GPS coordinates "close enough" floats can be very distant from each other.

TheSavior:
I believe it takes an argument for you to define what is *close enough*.

MFizz:
`unittest.assertAlmostEqual` always rounds to a given decimal point, so it could be tricky to find a reliable parameter. To be really safe you could use smth like `assertTrue(utils.distance() < x)` with x = 1m.

pokemongo_bot/cell_workers/pokemon_catch_worker.py
```diff
@@ -1,11 +1,11 @@
 # -*- coding: utf-8 -*-
 
 import time
-from pokemongo_bot import logger
 from pokemongo_bot.human_behaviour import (normalized_reticle_size, sleep,
                                            spin_modifier)
+from pokemongo_bot.cell_workers.base_task import BaseTask
 
-class PokemonCatchWorker(object):
+class PokemonCatchWorker(BaseTask):
```
TheSavior: PokemonCatchWorker is *not* a worker. It is essentially an API helper method that other workers use. It needs to be pulled out of this folder, but that shouldn't be done in this PR. This line definitely needs to get changed back though.

pokemongo_bot/cell_workers/catch_visible_pokemon.py
```diff
@@ -9,17 +8,26 @@
...
             # build graph & A* it
             self.bot.cell['catchable_pokemons'].sort(
                 key=
-                lambda x: distance(self.bot.position[0], self.bot.position[1], x['latitude'], x['longitude']))
+                lambda x: distance(self.bot.position[0], self.bot.position[1], x['latitude'], x['longitude'])
+            )
 
-            user_web_catchable = 'web/catchable-%s.json' % self.bot.config.username
             for pokemon in self.bot.cell['catchable_pokemons']:
-                with open(user_web_catchable, 'w') as outfile:
```
TheSavior: Removing this means that the web uis will be completely unusable until they support the events. I think we should leave these lines in to support the old UIs until we have a new UI that works with the event system.

douglascamata: ooops, this was a mistake :x

TheSavior:
:+1:

solderzzc:
👍

TheSavior:
:+1:

SpencerCarstens:
By "Multibot" do you mean running more than one bot, or some feature I'm not aware of / can't find?

douglascamata:
@SpencerCarstens I mean running more than one instance of the bot and just one websocket server.

TCB13:
@douglascamata thanks for the great additions... but I noticed this also kills colored output on the Terminal... Any notes on this?

douglascamata:
@TCB13 I might do a fix when we got updated protobuf for bots.... I need something smart, because I'll add config to each task to be able to disable theirs logs or log to a file.

TCB13:
@douglascamata I'm aware that this commit lays a lot of ground for web sockets but there are some folks that want to keep it more simple and just run it on the terminal... without any web UI etc. Not having colored output makes it harder.

douglascamata:
@TCB13 I know people will complain about this as soon as they see everything in black and white.... we will think about this later.

SpencerCarstens:
@douglascamata Was this idea ever flushed out anymore, or did it get put on hold due to the API debacle? There are a lot of pull requests still open over in the Go-Map project so I wanted to see what the plan was before I start rearranging deck chairs on the Titanic.

douglascamata:
@SpencerCarstens the feature is implemented in the bot and we have 3 people working on the new websocket-based UI. It will be released soon. What does the Go-Map project has to do with this anyway?

SpencerCarstens:
@douglascamata 

Sorry, need coffee - I meant [PokemonGo-Web](https://github.com/PokemonGoF/PokemonGo-Web).

I would love to help, what project are they working on the UI in?

douglascamata:
@SpencerCarstens dunno exactly where they are hosting it for now. But @brantje and @Jasperrr91 are working on it with Angular 1.5.

SpencerCarstens:
@douglascamata Perfect! I can certainly be a resource on that.

PR #2182, opened 2016-08-01
vary items keep by +- 10%

gweedo767:
Varies id_filter_keep by 10% of config value on each call to make tracking harder


1cu:
There Should be a config option for the variance and to turn this feature off.

Issue #2180, opened 2016-08-01
Not spinning/moving any forts as there aren't enough space.


Thieund:
### Expected Behavior
Normal behavior
### Actual Behavior
There are many "Server seems to be busy or offline - try again - 1/5"
And app keeps posting "Not spinning/moving (to) any forts as there aren't enough space."

### Other Information
OS: Windows server 2012.
P/s: Working fine and no error/warning in windows 10 with the SAME configuration.

Anakin5:
Bot should spin whether we have space or not since it always rewards xp. It does not make sense to stop walking toward forts and stop spinning them if there is not space in  inventory.

aTadTasty:
agreed with @Anakin5 this is prob just a bug?

PR #2188, opened 2016-08-01
Make SpinFort spin even when bag is full

clavat:
Short Description: 
Spin fort reward XP
Fixes:
- #2180 

Remove space check before spinning.
Spin fort still usefull because it rewars XP (#2180)


CapCap:
Can you have this be a config option?

TheSavior:
Yeah, this needs to be a configuration option.

pokecunt:
download

clavat:
I guess someone can do this in a better way than me.
I'm not familiar with Python, so i can't do a great job !

PR #2193, opened 2016-08-01
Sleeper: pause for some time every day

buluba89:
Short Description: 
The bot is playing 24/7, this is not a normal user behavior. This worker pauses the bot every day for some time to simulate the user having some rest. The time and the duration is configurable from the config file, also a random offset is added for both values.  Example of config entry: 
`

    {
         "type": "Sleeper",
         "config": {
           "time": "12:00",
           "duration":"5:30",
           "time_random_offset": "00:30",
           "duration_random_offset": "00:30"
         }
     }
     time: (HH:MM) local time that the bot should sleep
     duration: (HH:MM) the duration of sleep
     time_random_offset: (HH:MM) random offset of time that the sleep will start
                         for this example the possible start time is 11:30-12:30
     duration_random_offset: (HH:MM) random offset of duration of sleep
                         for this example the possible duration is 5:00-6:00

`
I made this to prevent detecting the bot and avoid ban, i hope you find it useful !

### DIFF
--- a/pokemongo_bot/cell_workers/__init__.py
+++ b/pokemongo_bot/cell_workers/__init__.py
@@ -15,4 +15,4 @@ from follow_spiral import FollowSpiral
 from collect_level_up_reward import CollectLevelUpReward
 from base_task import BaseTask
 from follow_cluster import FollowCluster
-from sleeper import Sleeper
+from sleeper import SleepSchedule
--- a/pokemongo_bot/cell_workers/__init__.py
+++ b/pokemongo_bot/cell_workers/__init__.py
@@ -15,4 +15,4 @@ from follow_spiral import FollowSpiral
 from collect_level_up_reward import CollectLevelUpReward
 from base_task import BaseTask
 from follow_cluster import FollowCluster
-from sleeper import SleepSchedule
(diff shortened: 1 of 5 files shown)


solderzzc:
lol, this is really funny :) @douglascamata @TheSavior

TheSavior:
For a first PR, this is great. Thanks for the quality, and I'm also very impressed that you wrote tests. I think this is the first worker that has tests (other than some helper logic for one of the navigators). Since this doesn't modify any other workers and just adds one, the barrier to merge is pretty low here. 

Can we name it something more clear than just Sleeper? I'm concerned users won't be able to guess what this does just by seeing the name. Perhaps SleepSchedule?

buluba89:
Thanks a lot !
Yes maybe Sleeper is not clear enought... i will rename it asap !

buluba89:
@TheSavior  Its ready !

TheSavior:
:+1:

cwild:
Great work, I had been thinking about something similar and this looks like it will do just the trick!

MZorzy:
hat to change
"type": "Sleeper",
to 
 "type": "SleepSchedule",
in config, but now in test

PR #2194, opened 2016-08-01
Modify SpiralTask to use 70m as stepsize and diameter as step_count

tstumm:
Short Description: 
SpiralNavigator used 200m as `step_size`, which causes kinda big blind spots since maximum radius for detecting pokemon has ben lowered to 70m.
Using 140m wouldnt be sufficient since we would still get small blind spots of over 1000m^2.
70m would reduce those blind spots. There is some big overlapping though, but this way respawned pokemons can be catched too.

I also changed the configuration from a step count to a diameter. The spiral is always a square, so the diameter describes the width/height of the spiral (`Height = diameter * 70m`)


solderzzc:
👍

douglascamata:
@tstumm make step size a parameter too ;)

douglascamata:
@tstumm also please note that step size is the Walker's responsibility. `FollowSpiral` only gives the spiral points and the walker changes position in small steps.

tstumm:
@douglascamata The defined step size in this Spiral Navigator has NOTHING to do with the walkers. Its just the size of the steps between the waypoints. If you really think that this PR just generates points in distance smaller than the walkers steps then you should really take a look into the code before claiming this.

tstumm:
The size of the steps the walkers are doing is defined by the config-param "walk" and it was always like this.

tstumm:
Please look at the changes to clear up this MAJOR misunderstanding.

douglascamata:
You're missing the point. If step size is defined by the `config.walk`, why you're enforcing a 70m step size in `FollowSpiral`?

tstumm:
Because the step with of the walker was to do more human-like steps before the bot was flodded with features. The step size in `FollowSpiral` is just do give waypoints so we can cover almost each point in our area. The stepsize of `FollowSpiral` was never related to walking at all, so you are misunderstanding the code.

tstumm:
The distance between the waypoints has been enforced since we have the Spiral-code

douglascamata:
@tstumm solve merge conflicts, please.

solderzzc:
👍

TheSavior:
LGTM

:+1:

PR #2201, opened 2016-08-01
Output hatched egg info to file

Shoh:
Short Description: Output hatched egg info to file

Fixes:
- Write hatched egg info to file to easily see what hatched from eggs while AFK
- Fixed follow_path worker to use 'with' statement when reading file in so it's auto-closed
- Fixed typo


douglascamata:
@Shoh this PR cannot be accept, because I'm moving all the logs to events (events are logged by a LoggingHandler of the system). But in the future, when PR gets merged and I add config, you will be able to redirect any given worker (or all workers) output to a file.

Shoh:
Got it, np

PR #2203, opened 2016-08-01
Fix path config example.

rikkert:
FollowPath will not work properly while also using MoveToFort.


TheSavior:
I believe these things are working properly, just maybe not what people would expect.

The FollowPath task essentially never gets run because there is always another fort for MoveToFort to go to. I'm happy to remove this as it is an example file specifically for FollowPath.

PR #2205, opened 2016-08-01
Insert random speed to make bot more human like

BoGnY:
Short Description: 
Insert a random speed in interval 0/+2 km/h in step_walker.py to make bot more human like

### DIFF
--- a/pokemongo_bot/step_walker.py
+++ b/pokemongo_bot/step_walker.py
@@ -1,8 +1,8 @@
+import random
 from math import sqrt
 
 from cell_workers.utils import distance
 from human_behaviour import random_lat_long_delta, sleep
-
 
 class StepWalker(object):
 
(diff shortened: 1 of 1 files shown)


douglascamata:
This doesn't make a lot of difference, anyway. And speed isn't in any unit, it's just an empirical number. Docs are outdated with regards to this.

tstumm:
km/h would be the wrong unit too.

TheSavior:
I agree with douglas. This is essentially random because of the random delay between each step we take, and we modify the coordinates we go to with a random amount of jitter.

BoGnY:
So what is the utility of speed on config.json???

tstumm:
It's the distance you "walk" after each tick if there is no other task blocking it.

Nik0lashka:
@tstumm , distance in km or miles or smth else? And what is tick and what is the frequency of it?

tstumm:
Meters. A tick is just a term for the whole task chain running once. There is no exact frequency, since the runtime of one tick is not deterministic.

Nik0lashka:
@tstumm, i suppose that tick is like processor time and it doesn't directly correlate with real time? Is it about millisecond or so?

PR #2208, opened 2016-08-01
Delay after printing the exchanged Pokemon but before exchanging, so …

skorokithakis:
Short Description: 
The bot currently prints an exchange message, then immediately exchanges, then waits for a few seconds. I changed it so it waits *before* releasing, instead of after, so someone has time to kill the bot if it's about to exchange a good pokemon.


pokecunt:
download

Nihisil:
This PR is not actual with latest dev version.

But you are welcome to make a new PR with same changes.

skorokithakis:
I'd rather not, if it takes ten days to merge a one-line PR. It's just going to go stale too anyway.

PR #2216, opened 2016-08-01
release logic fix

reddivision:
Short Description: 
Release logic was changed in `keep_best` branch to also process `release_under` config rules. However, many users using `keep_best` logic forgo these rules, so it was keeping all pokemon instead of releasing as it should have. This fix addresses that issue by allowing users _with_ `release_under` rules to have that logic processed, while other users can continue to have existing behavior for `keep_best` rules.
Fixes:
- Pokemon not releasing when only `keep_best` logic in release section


reddivision:
saw someone else has addressed this

PR #1992, opened 2016-07-31
PokemonGo-Map Synergy

mhdasding:
**Description:**
This worker will fetch current pokemon spawns from /raw_data of an PokemonGo-Map instance. It will also update the location of PokemonGo-Map when the bot is moving away.
There are two modes available:

**distance:** will move to the nearest pokemon
**priority:** will move to the pokemon with the highest priority assigned (tie breaking by distance)

**other configs:**
**address**
address of the webserver of PokemonGo-Map
**prioritize vips**
will prioritize vips in distance and priority mode above normal pokemon
**min_time:**
minimum time the pokemon has to be available before despawn
**max_distance:**
maximum distance the pokemon is allowed to be when walking, ignored when sniping
**snipe**
if enabled will teleport to target pokemon, encounter it, teleport back then catch it
if disabled will walk
**update_map**
disable/enable if the map location should be automatically updated
**catch**
a dictionary of pokemon to catch with an assigned priority (higher => better)
_vip pokemon are automatically catched without an entry here_

PR #2217, opened 2016-08-01
pokemon catcher interface for https://github.com/AHAAAAAAA/PokemonGo-Map

GNSPS:
**Short Description:**

This PR has a whole new worker to interface with https://github.com/AHAAAAAAA/PokemonGo-Map and go after rare pokemons (specified in the configs by id) found in the map.

When it is working it stops the move_to_fort worker because the pokemons have a limited lifetime and the bot cannot run the other worker simultaneously.

_Fixes_:
- No fixes, just a new  feature


GNSPS:
Fixing the conflicts which I didn't notice but broke it all!

DayBr3ak:
Don't merge, rebase

Anakin5:
max_travel_time  seems to be seconds. So the needed_speed is in distance_unit / seconds.
So I guess 30 cannot be the default value.

Anakin5:
Also, I think such mechanism should check pokemon cp / iv to make sure we don't run for a pokemon that we are going to release.

GNSPS:
Yes, you're right!

30 m/s was what I was using in the tests which is 108 km/h it is a bit fast but on the other hand really rare pokemons tend to spawn far away if you let the task define the needed speed you take ages to reach the desired pokemon and cannot move to forts at the same time.

So it's kind of a difficult choice!

GNSPS:
Nice point! Going to add that when I can.

For now let's assume you'll only use this for VIP pokemons in your config!

Anakin5:
We could configure the minimum and maximum speed.
Bot would try to reach the pokemon at the necessary speed to catch it (with a buffer for capture). If that speed is less than the minimum in config, use the minimum. If that speed is greater than the maximum, consider you cannot reach that pokemon in time.

GNSPS:
That is exactly what is happening the speed in the config is the minimum speed. The maximum is the 120 km/h which is tested by the pokemongo-webspoof guys!

But no harm in making the max configurable too! 😄

Anakin5:
On detail is that we do not check the speed in get_nearest_pokemon. So we  might return an encounter that is not reachable. That would be good to filter out all encounter not reachable at max speed.

GNSPS:
Right! important point!
Better than filter out one for each run of the worker

douglascamata:
@GNSPS please, also use a task config, like the `IncubateEgg` task does. We do not add config like that anymore, it's being deprecated.

nikhil-pandey:
Is this ~same to #1992 ?

Anakin5:
There is other pull requests to take into account:
https://github.com/PokemonGoF/PokemonGo-Bot/pull/1992
https://github.com/PokemonGoF/PokemonGo-Bot/pull/1887

GNSPS:
I thought about #1992 but that way you can't connect to an external server and this way you can

douglascamata:
@GNSPS you redid the whole work just to be able to connect to an external server? This is easy to implement there.

douglascamata:
@GNSPS also, tasks don't store state in `self.bot`, they store state in themselves now.

GNSPS:
No! I didn't see #1992, I was just saying I thought of that method and it wasn't as effective!

Gonçalo

----------
Sent from my mobile,
please excuse brevity & typos

mhdasding:
@GNSPS it's the same, i forgot to update the description to reflect the change that it now parses the spawns from /raw_data of PokemonGo-Map ;)

GNSPS:
Ok! ;)

Closing this now!

Issue #2218, opened 2016-08-01
Issue of -1 Pokeballs


db-coder:
### Expected Behavior
Should switch to farming mode.

### Actual Behavior
[17:43:57]  [x] Next egg incubates in 1.86 km 
[17:43:58] Lured pokemon at fort 3ebc9f6926254d32ab41cdebdb98e757.11
[17:43:58]  A Wild Rattata appeared! [CP 227] [Potential 0.29] 
[17:43:58] IV [Attack/Defense/Stamina] = [10/0/3]
[17:44:07] Move to destiny. 2 lured pokestops will be in range of 50m. Arrive in 163.051117374m.
[17:44:13] Lured pokemon at fort 3ebc9f6926254d32ab41cdebdb98e757.11
[17:44:14]  A Wild Rattata appeared! [CP 227] [Potential 0.29] 
[17:44:14] IV [Attack/Defense/Stamina] = [10/0/3]
[17:44:22] Move to destiny. 2 lured pokestops will be in range of 50m. Arrive in 157.922716904m.
[17:44:28] Lured pokemon at fort d0d0b295bd8746e29d680d9c8c204d60.12
[17:44:29]  A Wild Pidgey appeared! [CP 134] [Potential 0.62] 
[17:44:29] IV [Attack/Defense/Stamina] = [6/11/11]
[17:44:31] Using Pokeball (chance: 53.30%)... (-1 left!)
[17:44:37] Move to destiny. 2 lured pokestops will be in range of 50m. Arrive in 12.1994839674m.
[17:44:44] Lured pokemon at fort d0d0b295bd8746e29d680d9c8c204d60.12
[17:44:44]  A Wild Pidgey appeared! [CP 134] [Potential 0.62] 
[17:44:44] IV [Attack/Defense/Stamina] = [6/11/11]
[17:44:47] Using Pokeball (chance: 53.30%)... (-1 left!)
[17:44:53] Move to destiny. 2 lured pokestops will be in range of 50m. Arrive in 4.01179994087m.
[17:44:55] Lured pokemon at fort d0d0b295bd8746e29d680d9c8c204d60.12
[17:44:55]  A Wild Pidgey appeared! [CP 134] [Potential 0.62] 
[17:44:55] IV [Attack/Defense/Stamina] = [6/11/11]
[17:44:59] Using Pokeball (chance: 53.30%)... (-1 left!)
[...]

PR #2222, opened 2016-08-01
Fix -1 pokeballs

DayBr3ak:
__Short Description__

If balls come down to 0 during an encounter, the bot continue to try to catch even though there is no balls available.
Now it correctly exits the worker

__Fixes:__
- #2218


DayBr3ak:
Neverming, still broken

Issue #2184, opened 2016-08-01
BUG cluster


Marco208:
### Expected Behavior
stay connected

### Actual Behavior
disconnects

### Steps to Reproduce
use cluster

[07:54:32] No lured pokestops in vicinity. Search for normal ones instead. Move to destiny. 3 lured pokestops will be in range of 50m. Arrive in 16.7865123168m.
[07:54:36] No lured pokestops in vicinity. Search for normal ones instead. Move to destiny. 3 lured pokestops will be in range of 50m. Arrive in 15.6602194739m.
[07:54:41] No lured pokestops in vicinity. Search for normal ones instead. Move to destiny. 3 lured pokestops will be in range of 50m. Arrive in 9.98004002201m.
[07:54:45] No lured pokestops in vicinity. Search for normal ones instead. Move to destiny. 3 lured pokestops will be in range of 50m. Arrive in 5.99300669829m.
[07:54:48] Arrived at destiny. 3 pokestops are in range of 50m.
[07:54:50]  Server seems to be busy or offline - try again - 1/5 
[07:55:00]  Server seems to be busy or offline - try again - 1/5 
[07:55:08]  Server seems to be busy or offline - try again - 1/5 
^C[07:55:09]  Exiting PokemonGo Bot 
[07:55:09] 
[07:55:09]  Ran for 0:03:39 
[07:55:09]  Total XP Earned: 50  Average: 820.58/h 
[07:55:09]  Travelled 0.31km 
[07:55:09]  Visited 1 stops 
[07:55:09]  Encountered 0 pokemon, 0 caught, 0 released, 0 evolved, 0 never seen before 
[07:55:09]  Threw 0 pokeballs 
[07:55:09]  Earned 0 Stardust 
[07:55:09] 
[07:55:09]  Highest CP Pokemon:  
[07:55:09]  Most Perfect Pokemon:

leovarmak:
@Marco208 you mean that the bot closes after sometime ?

Agentscreech:
I think what he's saying is that it'll run for a bit, then it'll start giving you the "Server seems busy" error.  I've seen this too just this evening.  It'll go fine for a few min, then it'll start getting that message.  If I let it sit for a while, it'll go longer, but still hit this wall eventually.

MFizz:
Could that happen because of  niantic throttling (look at #2171)? Because I can't recreate.

xorinzor:
[...]

PR #2241, opened 2016-08-01
Dev - Fixed the loss of fort data

Abraxas000:
Short Description:  This fixes loss of fort data when the server returns no fort data. 

It seems like the server is saving cycles/bandwidth by only returning fort data when there has been a change to forts in your cells. When it returns no forts, we currently replace our existing data with the empty data.  This fixes that so we keep our fort data if the server returned no fort data.  Wild/lured pokemon are still updated in any case.

Fixes:
- If the server returns no fort data in the cell data, keep our existing fort data. (Web location data is still going to lose fort data, but the bot itself won't)
- I removed duplicate code
- Edited cluster log function to round to nearest meter

Known related Issues:
-https://github.com/PokemonGoF/PokemonGo-Bot/issues/2184

### DIFF
--- a/pokemongo_bot/__init__.py
+++ b/pokemongo_bot/__init__.py
@@ -94,19 +94,30 @@ class PokemonGoBot(object):
         forts = []
         wild_pokemons = []
         catchable_pokemons = []
+        fortspresent = False
         for cell in cells:
             if "forts" in cell and len(cell["forts"]):
+                fortspresent = True
                 forts += cell["forts"]
             if "wild_pokemons" in cell and len(cell["wild_pokemons"]):
                 wild_pokemons += cell["wild_pokemons"]
             if "catchable_pokemons" in cell and len(cell["catchable_pokemons"]):
                 catchable_pokemons += cell["catchable_pokemons"]
 
-        return {
-            "forts": forts,
-            "wild_pokemons": wild_pokemons,
-            "catchable_pokemons": catchable_pokemons
-        }
+        # If there are forts present in the cells sent from the server or we don't yet have any cell data, return all data retrieved
+        if fortspresent or not self.cell:
+            return {
+                "forts": forts,
+                "wild_pokemons": wild_pokemons,
+                "catchable_pokemons": catchable_pokemons
+            }
+        # If there are no forts present in the data from the server, keep our existing fort data and only update the pokemon cells.
+        else:
+            return {
+                "forts": self.cell["forts"],
+                "wild_pokemons": wild_pokemons,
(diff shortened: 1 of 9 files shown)


pokemongo_bot/__init__.py
```diff
@@ -94,19 +94,30 @@ def get_meta_cell(self):
         forts = []
         wild_pokemons = []
         catchable_pokemons = []
+        fortspresent = False
```
TheSavior: Do you really need this flag? Could we just check for `len(forts)`?

pokemongo_bot/cell_workers/follow_cluster.py
```diff
@@ -42,7 +42,7 @@ def work(self):
 
                 log_str = log_lure_avail_str + 'Move to destiny. ' + str(cnt) + ' ' + log_lured_str + \
                           'pokestops will be in range of ' + str(self.radius) + 'm. Arrive in ' \
-                          + str(distance(self.bot.position[0], self.bot.position[1], lat, lng)) + 'm.'
+                          + str(round(distance(self.bot.position[0], self.bot.position[1], lat, lng))) + 'm.'
```
TheSavior: This change doesn't seem relevant to this PR

douglascamata:
Also fix conflicts with dev branch, please.

pokemongo_bot/cell_workers/follow_cluster.py
Abraxas000: Agreed, I'm not sure how to separate the commits into different PRs yet. Feel free to drop this commit from the PR.

TheSavior: The commits would need to be on different branches. It is a lot of work for us to split up your PR. Since this is such a small change it would be easiest for you to just add a new commit undoing this change, and then on a new branch adding it again.

matheussampaio:
@Abraxas000 One way to separate commits is (or remove commits from that branch) is using `git rebase -i dev` and change `pick` to `drop`.

Abraxas000:
Thanks all for the assistance! I'm used to using TFS so I'm still bumbling my way through git processes.

@TheSavior The cluster change has been reverted and the conflict from the previous merge is resolved.

Abraxas000:
This PR got very messy. I'm starting with a fresh fork and branching my edits so I won't have to deal with this again later. I'm sure there's a more elegant way to clean up my forked repo, but I've already spent too long trying to fix this xD

Abraxas000:
New PR: https://github.com/PokemonGoF/PokemonGo-Bot/pull/2269

Issue #1719, opened 2016-07-29
Refactor dev branch Evolve Logic


nburn42:
As per discussion on #1649.

Current dev branch evolve worker should be refactored.

accident12123:
Is this needed?   isn't https://github.com/klingan/PokemonGo-Bot/commit/ac3bd71d7f08b05c3a0914f3ebc5553b17b0f9c5 the fix for why it's walking slow?   I manually put in the change and it's going great.

Issue #16, opened 2016-07-21
[Login] Support Google 2-Factor


tstumm:
Some Google accounts use 2-factor authorization. Login for 2-factor currently fails.

tstumm:
Workaround should be app passwords:
https://security.google.com/settings/security/apppasswords

Saturate:
Seeing as most accounts may get banned, it's not really good to use ones real google account with 2-Factor.

tstumm:
Well using better techniques than teleporting should avoid getting banned

s1x:
Till that you can use an app password: https://support.google.com/accounts/answer/185833?hl=en

tstumm:
I already proposed that as a fallback solution

PR #2244, opened 2016-08-01
Refactor evolve_all worker

achretien:
Short Description: 

Refactor of the evolve_all task

Fixes:

- Remove transfer of evolved pokemon (should be handle by transfer task)
- Add order_by config flag to choose to evolve by iv or cp (default: cp)
- Add evolve_iv_min as threshold for evolve by iv (order by cp under threshold)
- Fix _validate_config not called before
- Get candy list to test if enough candy in the bag
- Filter out pokemon which can't be evolved


douglascamata:
Here, sir... have my +1:

:+1:

douglascamata:
This will solve #1719

maestroi:
It evolved 
`[01:27:51]  Starting PokemonGo Bot....
[01:27:52]  [x] Next egg incubates in 0.17 km
[01:27:53] Successfully evolved Pidgeotto with 150 CP and 0.56 IV!
[01:27:56] Successfully evolved Pidgey with 264 CP and 0.44 IV!
`
with these settings
`    {
      "type": "EvolveAll",
      "config": {
        "evolve_all": "all",
        "order_by": "iv",
        "evolve_iv_min": 0.8,
        "use_lucky_egg": false
      }`

not working and please add example config

douglascamata:
@TheSavior hard review and test here, pls.

achretien:
We change the https://github.com/PokemonGoF/PokemonGo-Bot/wiki/Configuration-files#evolve-all-configuration behavior by not evolving pokemon under the threshold (cp or iv).

It should be more clear to the user reading the config file

douglascamata:
@achretien did you fix the bug that was found?

pokecunt:
download

achretien:
I changed the variable names to be more clear. And fixed the behavior of evolving by the expecting one.

maestroi:
`[17:13:41]  --- Chuckb0y1 ---
[17:13:41]  Level: 21 (Next Level: 14655 XP) (Total: 320345 XP)
[17:13:41]  Pokemon Captured: 1270 | Pokestops Visited: 1410
[17:13:41]  Pokemon Bag: 240/250
[17:13:41]  Items: 345/350
[17:13:41]  Stardust: 155104 | Pokecoins: 0
[17:13:41]  PokeBalls: 85 | GreatBalls: 47 | UltraBalls: 54
[17:13:41]  RazzBerries: 72 | BlukBerries: 0 | NanabBerries: 0
[17:13:41]  LuckyEgg: 1 | Incubator: 0 | TroyDisk: 5
[17:13:41]  Potion: 10 | SuperPotion: 20 | HyperPotion: 11
[17:13:41]  Incense: 8 | IncenseSpicy: 0 | IncenseCool: 0
[17:13:41]  Revive: 30 | MaxRevive: 0
[17:13:41]
[17:13:41]
[17:13:45]  Starting PokemonGo Bot....
[17:13:45]  [x] Current egg hatches in 3.62 km
[17:13:46] Successfully evolved Magikarp with 120 CP and 0.8 IV!
[17:13:50] Successfully evolved Staryu with 521 CP and 0.87 IV!
[17:13:54] Successfully evolved Staryu with 506 CP and 0.78 IV!
[17:13:58] Successfully evolved Poliwhirl with 673 CP and 0.49 IV!
[17:14:03] Successfully evolved Psyduck with 563 CP and 0.36 IV!
^C[17:14:05]  Exiting PokemonGo Bot


      {
      "type": "EvolveAll",
      "config": {
        "evolve_all": "all",
        "order_by": "iv",
        "evolve_iv_min": 0.8,
        "use_lucky_egg": false
        }`

TheSavior:
If we are renaming this task, we are going to be breaking everyones config. We should give people better error messages so they know what changed.

Can you change this https://github.com/PokemonGoF/PokemonGo-Bot/blob/dev/pokemongo_bot/tree_config_builder.py#L24 to:

```
if task_type is None:
                raise ConfigException('No type found for given task {}'.format(task))
            elif task_type is 'EvolveAll'
                raise ConfigException('The EvolveAll task has been renamed to EvolvePokemon')
```

And add a test for this case here: https://github.com/PokemonGoF/PokemonGo-Bot/blob/dev/tests/tree_config_builder_test.py

douglascamata:
And please solve merge conflicts, thanks!

achretien:
Updated with the new api wrapper. 

@douglascamata is it ok for merge?

TheSavior:
:+1:

douglascamata:
👍

coderplay:
1. The refactoring try to sort by pokemon index desc which is not
good. For example, why Venusaur which is #3 should be with lower
priority than Pidgey(#16)?

2. The refactoring always use dict.get(key, {}).get(key, {}) which will
eventually return a {} which will cause side effect. It's better
fail fast when the dict has no attribute which is unexpected by the
developer.

3. The refctoring try to use cache when the pokemon is not
evovlable. That developer doesn't know previously we didn't caculate the
candy before evovling, thus caused a lot of failure when evovling. After
caculating candy requirements by the refactoring and this patch, the
failure rarely happens. Thus cache is not necessary when the pokemon is
not evovlable.

4. The refactoring actually change the behavior after evolution. Previously,
 after evolution, the bot will release some pokemon which is not match 
user's rule. But this refactoring removed that logic. It's not a real 
refactoring which by definition doesn't change any behavior.

douglascamata:
1. It sorts pokemon by the status DESC, which means more powerful first. Please show the code that does as you say. Your Venusaur com be lower priority is you use IV as criteria and Pidgey has better IV.

2. `dict.get` is always better than try catch. The problem, if there are any, can be easily avoided by a two-liner:

```
if not my_dict:
    return
```

3. Show the diff.

4. EvolvePokemon task should not transfer pokemon. TransferPokemon task is the ONLY task supposed to transfer pokemon.

pokemongo_bot/cell_workers/evolve_pokemon.py
```diff
@@ -0,0 +1,152 @@
...
+                'cp': pokemon.get('cp', 0),
+                'iv': self._compute_iv(pokemon),
+                'candies_family': next_evol.get('Name', ""),
+                'candies_amount': next_evol.get('Amount', 0)
+            }
+            if pokemon["id"] > 0 and pokemon["candies_amount"] > 0 and (logic_to_function[self.cp_iv_logic](pokemon)):
+                pokemons.append(pokemon)
+
+        if self.first_evolve_by == "cp":
+            pokemons.sort(key=lambda x: (x['num'], x["cp"], x["iv"]), reverse=True)
```
coderplay: @douglascamata This line sort by pokemon index first.

pokemongo_bot/cell_workers/evolve_pokemon.py
```diff
@@ -0,0 +1,152 @@
...
+        else:
+            pokemons.sort(key=lambda x: (x['num'], x["iv"], x["cp"]), reverse=True)
+
+        return pokemons
+
+    def _can_evolve(self, pokemon, candy_list, cache):
+
+        if pokemon["name"] in cache:
```
coderplay: This line try to use cache which is unnecessary

pokemongo_bot/cell_workers/evolve_pokemon.py
```diff
@@ -0,0 +1,152 @@
...
+    def _validate_config(self):
+        if isinstance(self.evolve_all, basestring):
+            self.evolve_all = [str(pokemon_name) for pokemon_name in self.evolve_all.split(',')]
+
+    def work(self):
+        if not self._should_run():
+            return
+
+        response_dict = self.api.get_inventory()
+        inventory_items = response_dict.get('responses', {}).get('GET_INVENTORY', {}).get('inventory_delta', {}).get(
```
coderplay: @douglascamata 
If any of those key is not exist,  inventory_items will get an valid empty dict. Then it will be passed to the subsequence methods.

pokemongo_bot/cell_workers/evolve_pokemon.py
douglascamata: Looks like it works correctly. 

image

douglascamata: It sorts pokemon by NUMBER (from 1 to 151) and uses the CP as tiebreaker. All pokemon of some species will tie and be ordered by CP from bigger to smaller.

coderplay: @douglascamata 
The pokemon index I mentioned is the NUMBER you said.  We are actually the same understanding about the logic of those lines.  Let me describe more why I think sorting by pokemon number is not good.  As I mentioned in another thread. I have a patch to automatically evolve mid-tier pokemons like Ivysaur, Metapod or Pidgeotto which have further evolution. I tried to reserve candies for mid-tier pokemons with better potential (a.k.a better iv/cp) otherwise because the basic tier pokemon like Bulbasaur, Caterpie or Pidgey will be eariler to reach the candy requirement than their upper levels. In order to avoid starvation of the upper evolution, I rely on the sort, otherwise I might need to resort it again among those pokemons in the same family. If we sort by pokemon number, the higher level (which is mid-tier) of a family will get the chance to evolve. It's like a stronger first algorithm, not a better potential first algorithm. While if only sort by cp & iv, user have 2 options on choosing either stronger first or better potential first.   Furthermore, from the literal meaning of those config options of 'first_evolve_by', there is no pokemon index, right? I just think keeping the literal meaning with no surprise to the user is better.

douglascamata: Evolving logic has lots of personal preference into it. It will be close to, or, impossible to supports everyone's desires. The bot will never be something that you just open, never close and your account will be perfectly managed like everyone wants. So...

But anyway, I'm pretty sure you can use this sorted list somehow to work on and get your logic working. The more small functions we have, more easier is to reuse code and implement everyone's idea.

PR #2246, opened 2016-08-01
[FIX] Improper use of exception 

MFizz:
Short Description: 
Exception was used to filter response and additionally catch everything.

### DIFF
--- a/pokemongo_bot/__init__.py
+++ b/pokemongo_bot/__init__.py
@@ -413,15 +413,11 @@ class PokemonGoBot(object):
         items_stock = {x.value: 0 for x in list(Item)}
 
         for item in inventory_dict:
-            try:
-                # print(item['inventory_item_data']['item'])
-                item_id = item['inventory_item_data']['item']['item_id']
-                item_count = item['inventory_item_data']['item']['count']
-
+            item_id = item.get('inventory_item_data', {}).get('item', {}).get('item_id', {})
+            item_count = item.get('inventory_item_data', {}).get('item', {}).get('count', {})
+            if item_count != {} and item_id != {}:
                 if item_id in items_stock:
                     items_stock[item_id] = item_count
-            except Exception:
-                continue
         return items_stock
 
     def item_inventory_count(self, id):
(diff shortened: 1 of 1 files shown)


cwild:
Looks messy to me. Can't you just catch ```KeyError``` instead?

douglascamata:
👍 

@cwild much better use `dict.get` than catching `KeyError`, imho.

@TheSavior review please.

pokemongo_bot/__init__.py
```diff
@@ -413,15 +413,11 @@ def current_inventory(self):
         items_stock = {x.value: 0 for x in list(Item)}
 
         for item in inventory_dict:
-            try:
-                # print(item['inventory_item_data']['item'])
-                item_id = item['inventory_item_data']['item']['item_id']
-                item_count = item['inventory_item_data']['item']['count']
-
+            item_id = item.get('inventory_item_data', {}).get('item', {}).get('item_id', {})
+            item_count = item.get('inventory_item_data', {}).get('item', {}).get('count', {})
```
douglascamata: better do:

```
item_dict = item.get('inventory_item_data', {}).get('item', {})
item_count = item_dict.get('count')
item_id = item_dict.get('item_id')

if not item_count and not item_id
```

MFizz:
Thx, changed accordingly.

TheSavior:
:+1:

pokemongo_bot/__init__.py
```diff
@@ -413,15 +413,13 @@ def current_inventory(self):
...
         for item in inventory_dict:
-            try:
-                # print(item['inventory_item_data']['item'])
-                item_id = item['inventory_item_data']['item']['item_id']
-                item_count = item['inventory_item_data']['item']['count']
+            item_dict = item.get('inventory_item_data', {}).get('item', {})
+            item_count = item_dict.get('count')
+            item_id = item_dict.get('item_id')
 
+            if not item_count and not item_id:
```
douglascamata: I misunderstood the if condition. Previous one was `!= {}`, so the correct is `if item_count and item_id`, my bad.

MFizz: I'm sorry, I should have checked too.

PR #2249, opened 2016-08-01
Change FortID to FortName in lure

leadboots5:
Short Description: 
Making it more human readable. I did not test this change, just used the same field from movetofort line 41 and spinfort line 33.


solderzzc:
Thanks, it safe to merge.
👍

PR #2252, opened 2016-08-01
Display stats in the terminal title

Moonlight-Angel:
**Short Description**: A worker that periodically updates the terminal title to display some statistics about what's going on.

**Long description**: There's nothing really new here, I reused what was already done in the `Metrics` class.
Instead of having to stop the bot to access the summary, the stats will now also appear in the window title.
There are some configuration options too. You can totally customize what's shown. A window title isn't that big, just show what you really want to monitor. :eyes: 

```
Example config :
{
    "type": "UpdateTitleStats",
    "config": {
        "min_interval": 10,
        "stats": ["uptime", "km_walked", "level_stats", "xp_earned", "xp_per_hour"]
    }
}

Available stats :
- uptime : The bot uptime.
- km_walked : The kilometers walked since the bot started.
- level : The current character's level.
- level_completion : The current level experience, the next level experience and
                     the completion percentage.
- level_stats : Puts together the current character's level and its completion.
- xp_per_hour : The estimated gain of experience per hour.
- xp_earned : The experience earned since the bot started.
- stops_visited : The number of visited stops.
- pokemon_encountered : The number of encountered pokemon.
- pokemon_caught : The number of caught pokemon.
- pokemon_released : The number of released pokemon.
- pokemon_evolved : The number of evolved pokemon.
- pokemon_unseen : The number of pokemon never seen before.
- pokemon_stats : Puts together the pokemon encountered, caught, released,
                  evolved and unseen.
- pokeballs_thrown : The number of thrown pokeballs.
- stardust_earned : The number of earned stardust since the bot started.
- highest_cp_pokemon : The caught pokemon with the highest CP since the bot started.
- most_perfect_pokemon : The most perfect caught pokemon since the bot started.

min_interval : The minimum interval at which the title is updated,
               in seconds (defaults to 10 seconds).
               The update interval cannot be accurate as workers run synchronously.
stats : An array of stats to display and their display order (implicitly),
        see available stats above.
```

Successfully tested on Linux, Windows and OS X. Also compatible with `cygwin` on Windows.

**NOTE**: Fetching some stats requires making API calls. If you're concerned about the amount of calls  your bot is making, don't enable this worker or increase the minimum interval.

***

**Opinions/help wanted**:
- Searching a name for this worker was really hard. `UpdateTitleStats` still sounds odd. Any ideas ?
- I implemented this as a worker cause it seemed obvious to me, but there's a minor issue as mentioned in the above note. Also, the interval cannot be accurate as the workers execute synchronously. What's your opinion on this ? Should it stay as a worker or become something closer the the *core* ?
- I wasn't able to make a functional unit test for the title update on Windows because of some mocking issues on a particular method, see the skipped test for more information. If anyone has a solution, feel free to contribute.

### DIFF
--- a/pokemongo_bot/cell_workers/__init__.py
+++ b/pokemongo_bot/cell_workers/__init__.py
@@ -17,3 +17,4 @@ from collect_level_up_reward import CollectLevelUpReward
 from base_task import BaseTask
 from follow_cluster import FollowCluster
 from sleep_schedule import SleepSchedule
+from update_title_stats import UpdateTitleStats
--- a/pokemongo_bot/cell_workers/update_title_stats.py
+++ b/pokemongo_bot/cell_workers/update_title_stats.py
@@ -1,0 +1,222 @@
+from ctypes import *
+from sys import stdout, platform as _platform
+from datetime import datetime, timedelta
+
+from pokemongo_bot.cell_workers.base_task import BaseTask
+from pokemongo_bot.worker_result import WorkerResult
+from pokemongo_bot.tree_config_builder import ConfigException
+
+
+class UpdateTitleStats(BaseTask):
+    """
+    Periodically updates the terminal title to display stats about the bot.
+
+    Fetching some stats requires making API calls. If you're concerned about the amount of calls
+    your bot is making, don't enable this worker.
+
+    Example config :
+    {
+        "type": "UpdateTitleStats",
+        "config": {
+            "min_interval": 10,
+            "stats": ["uptime", "km_walked", "level_stats", "xp_earned", "xp_per_hour"]
+        }
+    }
+
+    Available stats :
+    - uptime : The bot uptime.
+    - km_walked : The kilometers walked since the bot started.
+    - level : The current character's level.
+    - level_completion : The current level experience, the next level experience and the completion
+                         percentage.
+    - level_stats : Puts together the current character's level and its completion.
+    - xp_per_hour : The estimated gain of experience per hour.
+    - xp_earned : The experience earned since the bot started.
+    - stops_visited : The number of visited stops.
+    - pokemon_encountered : The number of encountered pokemon.
+    - pokemon_caught : The number of caught pokemon.
+    - pokemon_released : The number of released pokemon.
+    - pokemon_evolved : The number of evolved pokemon.
+    - pokemon_unseen : The number of pokemon never seen before.
+    - pokemon_stats : Puts together the pokemon encountered, caught, released, evolved and unseen.
+    - pokeballs_thrown : The number of thrown pokeballs.
+    - stardust_earned : The number of earned stardust since the bot started.
+    - highest_cp_pokemon : The caught pokemon with the highest CP since the bot started.
+    - most_perfect_pokemon : The most perfect caught pokemon since the bot started.
+
+    min_interval : The minimum interval at which the title is updated,
+                   in seconds (defaults to 10 seconds).
+                   The update interval cannot be accurate as workers run synchronously.
+    stats : An array of stats to display and their display order (implicitly),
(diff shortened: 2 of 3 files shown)


MrCarlosJr:
This is great idea

douglascamata:
Very nice job! BTW, you have merge conflicts with remote.

@TheSavior, look this!

douglascamata:
Maybe we can transform this into a handler of the event system. 💭

Moonlight-Angel:
@douglascamata Noticed there was conflicts but didn't find where to get the details.
I bet there's one with `cell_workers/__init__.py`.

Moonlight-Angel:
Conflict fixed. :+1:

Moonlight-Angel:
Also, I noticed an issue with the experience thresholds.
It seems that the values given by the server corresponds to the previous and the next level experience thresholds. There is no value giving the experience for the current level.
As a result, when the character gains a level, the level completion jumps directly to ~50%.
Can we do something to get the experience for the current level and fix that calculation ?

TheSavior:
@Moonlight-Angel This looks like an awesome feature and your PR looks great. I'm a little hesitant to merge it because it doesn't seem like this is really something that should be done in a worker, it seems like this is more a configuration for the bot itself. Probably tied into the event system that @douglascamata mentioned. I'm not really sure how that code would look though.

In the spirit of getting this functionality added as soon as possible while we figure out how it is actually used, how would you feel about making this just a generic class that could be called to update the title bar. Then the unit tests could just test that functionality directly without it needing to be a worker. In this PR let's not actually hook up that functionality to the bot, but have the class and logic to do it.

Then we can figure out whether this should be a worker like you have it here, or a core part of the bot. Either approach should be pretty straight forward to implement as it would essentially just be telling your standalone class to run that you have already written.

@douglascamata Do you have any more thoughts on this?

TheSavior:
Bah, Screw it. Your PR is exactly what I want contributions to this repo to be. If the team decides we can come up with a better way to set this up, we will refactor it. I'm good with this as it is. Thanks @Moonlight-Angel!

:+1:

Moonlight-Angel:
@TheSavior I saw some commit about that event system a couple of days ago but can't find it anymore. Didn't take the time to read it carefully so I'm not sure why using the event system would be better for a feature like this one.
Implementing this as a worker seemed cleaner at first glance. Existing classes (mainly the entry ones like `pokecli` and `pokemongo_bot.__init__` were already pretty big, didn't have the courage to dig into them :joy: ).
Also, adding this into `metrics.py` would have lengthen it too.
The easy way would have been to create another non-worker class with some timer on it to periodically update the title bar.

Anyway, if you think this is good enough to marry the `dev` branch, let's do this. ( ͡° ͜ʖ ͡°)

BTW, there is still some testing that needs to be done on Windows and OS X. I'll try to fire up some VMs to test it myself. I'll keep you updated.

Moonlight-Angel:
Fully working under Windows.
I also added `cygwin` support for the people who understood life and are using cywgin on Windows. ( ͡° ͜ʖ ͡°)

Moonlight-Angel:
And fully working on OS X too.
I'm all set on my side. Waiting for your approval. :+1:

TheSavior:
:+1:

TheSavior:
@douglascamata Please review this

pokecunt:
download

douglascamata:
👍

abusalman:
Thanks guys, it works!
could we add username or botname in this title array?
in case of running several window of bots, we could use this indicator to distinguish which window belong to which bot.

Moonlight-Angel:
@abusalman Please check out PR https://github.com/PokemonGoF/PokemonGo-Bot/pull/2494.

Issue #2247, opened 2016-08-01
[Suggestion] Variable soft ban pokestop flips. 


Kassius509:
Currently we're just using a fixed number of 50.  Variation in the # of softban stop flips will help prevent any flagging of anomalous play. 

Thank

PR #2253, opened 2016-08-01
Randomize spins for softban #2247

scalp42:


Kassius509:
Thx @scalp42!

maestroi:
but why ? :)
we can set it to 60

pokecunt:
download

Kassius509:
@maestroi When we do the EXACT same number of spins, every single time, it creates a distinct and easily recognizable pattern that Niantic can use to flag accounts as run by a bot.

TheSavior:
:+1:

Seems like a low risk change.

Issue #1973, opened 2016-07-31
Keep # Pokemon based on Candy on hand


k4n30:
#1395 brought the functionality of keeping Pokemon based on CP or IV
#1472 brought the ability to release based on BOTH

New feature requested:

Add an additional check based on candy amount (global on/off - configurable in config) that allows you to keep Pokemon based on the number of candy.

eg. 
1. Pidgey requires 12 candy to evolve and you currently have 118 pidgey candy, and 9 pidgey's in your Pokemon bag
2. You capture a new pidgey (taking candy to 121, and pidgeys in your bag to 10)
3. The release config has something like 
`"Pidgey": {"keep_best_cp": 2, "keep_best_iv": 3, "keep_candy_evolution_ratio" : true}`
4. The config would also have a "keep Pokemon for lucky egg mass evolve spree" true/false field set to true
5.  The new pidgey is NOT released due to keeping the candy to evolution ratio

This would be useful for when it came time to drop a lucky egg and do the mass evolving, but if you didn't actually have the right ratio of Pokemon to candy on hand, the lucky egg would be wasted (or run out before finding enough to evolve)

k4n30:
#1972 - Seems we both had the same idea at the same time :p

HecticSeptic:
Indeed, uncanny. It's the same suggestion.

slatspie:
This would be a handy feature. I mentioned it in #1528 a couple days ago too.

TheSavior:
I think this is a great feature. New workers are appreciated

nanogirl21:
I would love this

k4n30:
@MFizz - any chance?

k4n30:
#394 another issue with the same request.

All the issues need to be cleaned up and consolidated

TheSavior:
@k4n30 Thanks for pointing that out. Closing this as a duplicate of #394.

Issue #1972, opened 2016-07-31
[Feature Request] Prevent release of bulk-evolvable pokemon (config)


HecticSeptic:
### Desired Behavior
Keep only just enough (e.g.) Pidgeys for a batch evolve:

If the total candies needed to evolve all pokemon 'x' is less than held x candies, then do not release any pokemon x - because we want more of x for the next [manual] xp egg pop and batch evolve
If we have more Pidgeys than we have (Pidgey candies /12), we can then use normal release logic

pseudocode example:
candiesPerEvolve = 12 ## in the most important cases
...
if (keepEvolvable)
    if (candiesPerEvolve * count(bot.getPokemon("Pidgey")) < bot.getCandies("Pidgey"))
        release = False
...
### Actual Behavior
No apparent way to keep a good number of pidgeys, caterpies and weedles for batch evolve. Setting keep_best_cp (and or IV) too high just results in a full box/inventory (and error).

### Steps to Reproduce
Add keepEvolvable flag to release config in json file.
Handle unset or mistyped flags.
Change transfer_pokemon.py to cancel any release which would reduce batch evolves if flag is true

Future addition: if evolve is set in config, only do so once pokemon inventory fills up, that way the max batch evolve output can be reached with the xp egg
Potential max: if using 20s evolves, only 30 are needed, could have 
"Pidgey,Weedle,Caterpie": {"keep_best_cp": 30, "keepEvolvable": True}
// (i.e. all three together total 30 and can all evolve) 
### Other Information
OS: Should be independent 
Git Commit: (run 'git log -n 1 --pretty=format:"%H"' and paste it here) N/A

HecticSeptic:
#1973  suggests the same feature, within seconds of this one.

Issue #1528, opened 2016-07-29
[Feature Request] Keep X amount of Z,Y,A pokemon. When X amount is reached use lucky egg  and evolve Z, Y, A.


Henji96:
This would be very useful for power leveling.

slatspie:
I like this. 
Even just the first part would help. Keep x amount pokemon based on the candies needed for evolution would keep enough of each specific pokemon to use all candies for leveling in batch .. eg. have 9 Pidgey candies: keep 1 Pidgey, have 15 candies: keep 2, have 110 candies, keep 10 (candies-on-hand div candies-to-evolve +1). As you said, It would be for power leveling so keeping specific pokemon based on iv or cp would not really matter

RedViper9:
You might be able to do part of this now by setting the keep_best_iv or keep_best_cp parameter. eg:
```
"release": {
    "Pidgey": { "release_below_cp": 237, "release_below_iv": 0.9, "logic": "and", "keep_best_iv": 70 },
```
That should let you keep your top 70 Pidgeys.

You could then modify config for Egg/Evolve cycle and reload the bot.

slatspie:
I did have it set up like that. I finished doing a whack of evolves just now. I had more pidgey candies than I thought so didn't save enough pidgeys (my map shows undefined for candy counts and the fixes here didn't work for me so I lost count). Anyway, it was all good. I'll set it higher as you suggest.

Henji96:
Slatspie, how do you go about evolving? I can't seem to get it to work on master or dev.

slatspie:
@Henji96 sorry for the delay, was out of town. I evolved manually at that time. I think the evolve code is working now but I have not fully caught up on everything

PR #2255, opened 2016-08-01
[Feature] added keep pokemon for batch evolution

MFizz:
Short Description: 
After all keep/release configs are processed, the remaining pokemon are released on how many evolutions a currently possible. E.g, if you have additional 12 squirtles (to the ones you would keep normally) and 72 squirtle candies, you release 9 and keep 3, because you can evolve 3 (25 candies required). 

Fixes:
#1973
#1972
#1528

Please help me with testing this, since there are many moving parts and release is crucial.

### DIFF
--- a/pokemongo_bot/cell_workers/transfer_pokemon.py
+++ b/pokemongo_bot/cell_workers/transfer_pokemon.py
@@ -9,6 +9,7 @@ class TransferPokemon(BaseTask):
     def work(self):
         pokemon_groups = self._release_pokemon_get_groups()
         candies = self._get_candies()
+        evolvable = 0
         for pokemon_id in pokemon_groups:
             group = pokemon_groups[pokemon_id]
 
@@ -82,6 +83,7 @@ class TransferPokemon(BaseTask):
                             evo_pokemon = group
                             group = []
 
+                        evolvable += len(evo_pokemon)
                         if len(evo_pokemon) > 0:
                             logger.log("Keep {} {}, for evolution - {} candies".format(len(evo_pokemon),
                                                                              pokemon_name, candy), "green")
@@ -94,6 +96,9 @@ class TransferPokemon(BaseTask):
 
                 for pokemon in group:
                     self.release_pokemon(pokemon_name, pokemon['cp'], pokemon['iv'], pokemon['pokemon_data']['id'])
+
+        logger.log("{} pokemon transferred total. {} evolutions ready (based on pokemons additional to the ones kept"
+                   " with cp/iv criteria)".format(len(group), evolvable), "green")
 
     def _release_pokemon_get_groups(self):
         pokemon_groups = {}
--- a/pokemongo_bot/cell_workers/transfer_pokemon.py
+++ b/pokemongo_bot/cell_workers/transfer_pokemon.py
@@ -70,24 +70,25 @@ class TransferPokemon(BaseTask):
                         logger.log("Api call for candies failed, try again")
                         return
                     candy = candies[pokemon_id]
-                    req_candy = self.bot.pokemon_list[pokemon_id - 1]['Next Evolution Requirements']['Amount']
-                    num_keep = (len(group) + candy) / (req_candy + 1)
+                    if 'Next Evolution Requirements' in self.bot.pokemon_list[pokemon_id - 1]:
+                        req_candy = self.bot.pokemon_list[pokemon_id - 1]['Next Evolution Requirements']['Amount']
+                        num_keep = (len(group) + candy) / (req_candy + 1)
 
-                    if len(group) > num_keep:
-                        group.sort(key=lambda x: x['iv'], reverse=True)
-                        evo_pokemon = group[:num_keep]
-                        group = group[num_keep:]
-                    else:
-                        evo_pokemon = group
-                        group = []
+                        if len(group) > num_keep:
+                            group.sort(key=lambda x: x['iv'], reverse=True)
+                            evo_pokemon = group[:num_keep]
+                            group = group[num_keep:]
+                        else:
+                            evo_pokemon = group
+                            group = []
 
-                    if len(evo_pokemon) > 0:
-                        logger.log("Keep {} {}, for evolution - {} candies".format(len(evo_pokemon),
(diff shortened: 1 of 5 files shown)


MFizz:
Btw. are there PRs anywhere to write tests for release logic? Would be a good idea, but I don't have time atm.

achretien:
Maybe we could move candy check in an other part, I am also using it to test if we should evolve in #2244

MFizz:
Sure, maybe into cell_workers/utils?

pokemongo_bot/cell_workers/transfer_pokemon.py
```diff
@@ -33,41 +37,69 @@ def work(self):
...
+                        logger.log("{} [CP {}] [Potential {}]".format(pokemon_name,
+                                                                      high_pokemon['cp'],
+                                                                      high_pokemon['iv']), 'green')
+
+                if keep_for_evo and len(group) > 0:
+
+                    if candies == {}:
+                        logger.log("Api call for candies failed, try again")
+                        return
+                    candy = candies[pokemon_id]
```
achretien: Pokemon with 2 evolutions always use candy of the first kind of pokemon.
We have two solutions use the ['Next Evolution Requirements']['Name'] to order candies array (as my PR #2244 ). Or add the Evolution candy family in the JSON used for pokemon_list variable

MFizz: Ah yes I didn't think of that. In my mind the best solution would be only to include pokemon with 2 evolutionsteps possible the normal use case would be mass evolution with lucky egg for exp. And in this case 1st stage evolutions would be preferable because they use less candy. 

What do you think?

achretien: Maybe we should ask bot masters :).
In this case I will add an option in my PR, I currently evolve pokemon best evolution first.
In this case I will burn all the candies.

achretien: ping @pokemongo-dev

MFizz:
As @achretien pointed out 2nd stage pokemon use the same candy as 1st stage. The normal use case would be mass evolution with lucky egg for exp. And in this case 1st stage evolutions would be preferable because they use less candy.

So i excluded 2nd stage pokemon. Please tell me if 2nd stage pokemon should be included and why.

solderzzc:
Do you have test result here? And the config example should be updated.

TheSavior:
I'd have no problem with us adding a JSON file that lists the candy per pokemon per evolution. 

It sounds like we essentially want the calculations that pidgeycalc.com do. If so, can you move all of that general candy calculation logic out of transfer_pokemon and into a new file / new class called something like CandyEvolveCalculator? That way it will be easier to test and the transfer_pokemon task can just call a method on the CandyEvolveCalculator class to figure out what it should do for each of the pokemon.

pokecunt:
download

MFizz:
@solderzzc 
> Do you have test result here? And the config example should be updated.
Added configs and and example output would be
```
...
[11:41:45]  Keep 2 best Staryu, based on cp and iv 
[11:41:45]  Staryu [CP 94] [Potential 0.84] 
[11:41:45]  Staryu [CP 498] [Potential 0.62] 
[11:41:45]  Keep 1 Staryu, based on cp/iv criteria 
[11:41:45]  Staryu [CP 463] [Potential 0.13] 
[11:41:45]  Keep 1 Staryu, for evolution - 55 candies 
[11:41:45]  Staryu [CP 283] [Potential 0.8] 
[11:41:45]  0 pokemon transferred total. 14 evolutions ready (based on pokemons additional to the ones kept with cp/iv criteria) 
```

MFizz:
@TheSavior 
> I'd have no problem with us adding a JSON file

We already have that, take a look at data/pokemon.json. 

I would not like to make an extern task for it, since it is essentially transferring pokemon and it needs to know which pokemon are off limits based on config. Additionally transfer_pokemon would need to call this extern task in order to know which pokemon may or not may be released. So I think encapsulating it would be somewhat awkward, because it is in itself release/transfer logic.

achretien:
Can we unminified data/pokemon.json to have clearer commit for this file and add family candy id to the "Next Evolution Requirements" node ?

MFizz:
> Can we unminified data/pokemon.json to have clearer commit for this file and add family candy id to the "Next Evolution Requirements" node ?

Why not make an extra PR for that so it can be reverted more easily if it crashes something

douglascamata:
@MFizz what if I want to keep 1 Squirtle to get a Blastoise and power up the blasto with the remaining candies? You can emulate this configuration by using `keep_best_*` release filter, no?

MFizz:
@douglascamata this worker doesn't evolve pokemon. All it does is keeping enough squirtles to evolve with you current candy. If you power up your squirtle it just registers less candy and will release more Squirtles next tick.

For example say you have your config set to 
`'Squirtle': {'keep_best_iv': 2, 'release_below_cp': 900, 'keep_for_evo': true}`
and you have 72 candy and 20 Squirtles, of which 6 have cp over 900 (set 1)and 2 have the best iv (set 2) obv. Assuming set 1 and 2 are disjoint the current algorithm would release the 12(set3) remaining pokemon which do not match the criteria.

This feature then would now compute how many of these remaining 12 squirtles you could evolve, if you wanted to in this instant, which would be 3 (72 candy + 9 candy for transfer = 81 candy / 25 candy required = 3 evolutions). So it would release 9 pokemon of set 3.

Now if you wanted to you could start now a mass evolution/ lucky egg because you kept enough pokemon for your candy. Or you could do other stuff, like evolving your squirtle and powering it up. In this case, lets say that costed 15 candy and you used a pokemon of set 2:

set1 would be unchanged

set2 would have one less because of the evolution

set3 has 3 pokemon because of the tranfer

there are 57 candies left

now the algorithm would again keep set 1 and 2 because if the other criteria and compute maximum evolutions for set 3: 57/25 req = 2 evolutions -> release 1 squirtle. 

I hope this clears it up.

TheSavior:
:+1:

douglascamata:
👍

PR #2258, opened 2016-08-01
Change egg hatching text

Jasperrr91:
Short Description: Clearify the egg hatching text

Fixes:
- changed egg hatching text


maestroi:
@Jasperrr91 i agree! it's just weird :+1:

PR #2264, opened 2016-08-01
[Dev] Printing Catched pokemon list before exit

iruy:
Short Description: 
By adding `"print_catches": true` to config.json, a table listing all the catched pokemon during session will be printed after bot closes
sample iamge

Fixes:
- 
- 
-


mzygmunt:
Maybe it makes sense to add something like this, but showing all of own Pokemons on bot's startup?

vanrzuf:
Is it possible to show this and other infos on demand, for example by pressing i on keyboard?

TheSavior:
@vanrzuf That functionality is more likely to come from the web app and less likely for us to have a listener to keyboard events as the bot is running.

pokecunt:
download

Nihisil:
It can't be easily merged, because of all latest changes in dev version.

But you are welcome to send a new PR with same changes, but you need to rely on the new logger instead of old colored logger.

PR #2269, opened 2016-08-01
Dev - Fixed the loss of fort data (updated)

Abraxas000:
Short Description: This fixes loss of fort data when the server returns no fort data.

It seems like the server is saving cycles/bandwidth by only returning fort data when there has been a change to forts in your cells. When it returns no forts, we currently replace our existing data with the empty data. This fixes that so we keep our fort data if the server returned no fort data. Wild/lured pokemon are still updated in any case.

Fixes:
- If the server returns no fort data in the cell data, keep our existing fort data. (Web location data is still going to lose fort data, but the bot itself won't)
- I removed duplicate code

Known related Issues:
-https://github.com/PokemonGoF/PokemonGo-Bot/issues/2184

### DIFF
--- a/CONTRIBUTORS.md
+++ b/CONTRIBUTORS.md
@@ -47,3 +47,4 @@
  * MFizz
  * NamPNQ
  * matheussampaio
+ * Abraxas000
--- a/pokemongo_bot/__init__.py
+++ b/pokemongo_bot/__init__.py
@@ -105,7 +105,7 @@ class PokemonGoBot(object):
                 catchable_pokemons += cell["catchable_pokemons"]
         
         # If there are forts present in the cells sent from the server or we don't yet have any cell data, return all data retrieved
-        if len(forts) or not self.cell:
+        if len(forts) > 1 or not self.cell:
             return {
                 "forts": forts,
                 "wild_pokemons": wild_pokemons,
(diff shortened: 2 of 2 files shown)


Abraxas000:
@douglascamata 
@TheSavior I was using the wrong reference the first time I used "len". Go it working with the most recent revision in this PR.

Let me know if I need to fix anything else. Thanks for the advice!

douglascamata:
@Abraxas000 close the other PR please. Did you test? For how long?

Abraxas000:
Added a function to print the length of the forts collection on each update. It looks like the throttling from @matheussampaio has significantly reduced the frequency that I'm seeing the empty forts from Niantic.  Every time so far (50+ updates) there have been 78 or 86 (it increased about halfway through).

Abraxas000:
I had my original implementation of this in-place all day and I've been running the latest revision for about an hour (changing from the bool variable to using len(forts)). My bot's path has been significantly more stable and logical now that it hasn't been losing the fort data.

Before when the fort data would get lost, it would start making very repetitive movements and would stop spinning forts altogether in the worst cases.

matheussampaio:
What if you keep every fort in cache and not the last "good" response? Every response, you check/add new forts?

douglascamata:
@matheussampaio good suggestion!

Abraxas000:
@matheussampaio That's a great idea and would certainly be safer, data-wise, but would require more cycles and more time coding a system to do so. This is generally meant as a quick, "make sure we stop shooting ourselves in the foot" measure :-)

Abraxas000:
Been going strong for a solid hour with no issues.

PR #2274, opened 2016-08-01
Update web to latest master commit

bstpierr:
The current dev commit has an issue where it doesn’t show the number of candies due to the change in the API several days ago.

This moves the the web submodule forward by 27 commits.


BoGnY:
This isn't latest commit.. there are more than 25 commits for get latest

PR #2279, opened 2016-08-01
Replaced fort with pokestop in logger

Jasperrr91:
Short Description: 
Replaced fort with pokestop in logger

Fixes:
- Replaced fort with pokestop in logger


maestroi:
:+1: @douglascamata  push :horse_racing:

Evanito:
I don't understand why it is forts in the first place, explanation?

douglascamata:
Name in the API responses from servers.

Jasperrr91:
@Evanito in the API, both Gyms and Pokestops are named Forts. But we only use the Pokestops so that name is a better fit.

douglascamata:
Closed because events.

Issue #2250, opened 2016-08-01
Never seem to get 100exp


metheos:
### Expected Behavior
Should sometimes get 100exp when catching pokemon.

### Actual Behavior
Lowest value I've seen is 110exp.

### Steps to Reproduce
Set the spin and reticle factors to 0. 
    "catch_randomize_reticle_factor": 0.0,
    "catch_randomize_spin_factor": 0.0,
Wait while pokemon are caught and monitor the EXP earned.

### Other Information
OS: Win64
Git Commit: 494bc055543c3cb674cca781ec545311d3c9bad1
Python Version: 2.7

Nathanian:
Since i updated to the newest version i even get 210 exp with every Pokemon i catch

Likael:
Hello metheos,

It is exact that in the current state you never get to do just a normal it. I am working on the dev that will allow to have a better control over the type of throw you do, and normal throw without bonus will be reintroduced,

metheos:
@Likael Great! Thanks!

yoot55:
@Likael Isn't this dangerous as currently its not possible to get the aim (Nice/Great/Exceellent) bonus on the platform. Not sure if the change was intentional from Niantic's side. But if it was, it would easy to flag botters.

Likael:
Well, that's one more reason to be able to customize the type of throw you want. Also, as far as I know they don't monitor the app version we are using, so even if it was removed, we could still be having an old version that trigger the nice/great/... throws.

PR #2286, opened 2016-08-02
Allow to customize quality of pokeball throws

Likael:
**_WARNING:_**
The  code below is safe to use.
However, I tried to test further the possibilities of the field _normalized_hit_position_, and, coincidence or not, couldn't log again with my account any more for 1 hour immediately after sending a weird value. If you want to proceed and modify _normalized_hit_position_ to test randomization, be careful.

**Short Description:** 
The code allows to modify the frequency of the various throw possibilities.
For example, with the following setting : 
"catch_throw_parameters": {
	  "excellent_rate": 4,
	  "great_rate": 10,
	  "nice_rate": 4,
	  "normal_rate": 2,
	  "spin_success_rate" : 0.6,
	}
The bot would have 4 chances out of  20 (4+10+4+2) to land an _Excellent_ throw.
In the sample config, those value add up to 1 to look like percentage, but the user is free to set the values to what speaks most to him/her.
The _spin_success_rate_ is a percentage of landing a successful spin. In the example above, the bot has 60% chance of landing a spin that will generate the bonus exp. 

**More information about the _normalized_hit_position_ :** 
The code below should work safely as it doesn't contain the further tests i have tried, however my paranoid nature would ask for people to test with account they don't care about for 1 day before merging it. I will do so myself tomorrow too. Since the (unban?) I have been testing the code below, and got no further problem.
The value of 0 for _normalized_hit_position_  is safe, as it was what we were sending before I correct the API call.
The value of 1.0 is also safe, as we used it forever.
I didn't test any values between 0 et 1.
Values above 1 seems to be what could have caused the potential temporary ban.
The last case don't happen with this code, as the value is set either to 0 or 1.

**Fixes:**
- #2250

### DIFF
--- a/configs/config.json.cluster.example
+++ b/configs/config.json.cluster.example
@@ -94,11 +94,6 @@
       "// any": {"keep_best_iv": 3},
       "// Example of keeping the 2 strongest (based on CP) and 3 best (based on IV) Zubat:": {},
       "// Zubat": {"keep_best_cp": 2, "keep_best_iv": 3}
-      "// Example of keeping as many Zubat as you have candy to evolve,
-      "// Zubat": {"keep_for_evo": true}
-      "// Example of keeping the 2 strongest (based on CP) and 3 best (based on IV) Zubat and on top as many Zubat
-      as you have candy to evolve:": {},
-      "// Zubat": {"keep_best_cp": 2, "keep_best_iv": 3, "keep_for_evo": true}
     },
     "vips" : {
          "Any pokemon put here directly force to use Berry & Best Ball to capture, to secure the capture rate!": {},
--- a/configs/config.json.cluster.example
+++ b/configs/config.json.cluster.example
@@ -94,6 +94,11 @@
       "// any": {"keep_best_iv": 3},
       "// Example of keeping the 2 strongest (based on CP) and 3 best (based on IV) Zubat:": {},
       "// Zubat": {"keep_best_cp": 2, "keep_best_iv": 3}
+      "// Example of keeping as many Zubat as you have candy to evolve,
+      "// Zubat": {"keep_for_evo": true}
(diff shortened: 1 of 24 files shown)


pokecunt:
download

douglascamata:
@Likael please check how to put all the settings you add properly inside another key in the setup. The websocket settings are a good example.

Likael:
@douglascamata I need your help here, I must be missing something. I can't find the websocket settings (other than the key _websocket_server_).
I packed all the settings under the key _catch_throw_parameters_. The key itself is at the base of the config, mainly because : 
- I couldn't put it in a task, as this config is used by an undeclared worker, instanced by both _CatchVisiblePokemon_ and _CatchLuredPokemon_.
- I did not wanted to put it in the _Catch_ key, as it is used a list of pokemons with a default _any_ and additional keys there could be confusing.

I will be happy look at the websocket settings if you can guide me to it, and revise the settings to make it look more like it.

douglascamata:
@Likael this config:

```
add_config(
        parser,
        load,
        short_flag="-ws",
        long_flag="--websocket.server_url",
        help="Connect to websocket server at given url",
        default=False
    )
```

Automatically translates to a variable named `websocket_server_url`, which in the JSON configuration will be used like this:

```
"websocket": {
    "server_url": "my websocekt url"
}
```

This way you don't need any `.get` call or to mess with dicts. (:

RedViper9:
This is great! Can't wait.

Would it be possible to randomize these parameters such that Excellent throws are made 15-20% of the time, for example?

Likael:
@douglascamata Excellent! I must have missed it. I am on it.
@RedViper9 That's exactly what it will used too :) Defining your own odds for throw quality to be used.

Likael:
OK rebase went wrong. Will fix that tomorrow.

Likael:
Closing pull request to reopen a new clean one

PR #2288, opened 2016-08-02
Dev

tadejv:
Short Description: 
updated logging and other minor fixes

Fixes:
- set unnecessary log entries to debug
- fixed extra spaces when using colors
- changed some log colors to more apropriate color
- move to fort moves to next fort even if inventory full (might catch pokemon on the way?)
- updated how catching and transferring pokemons is logged
- print fort name even if not spinning
- removed duplicate entries

### DIFF
--- a/pokecli.py
+++ b/pokecli.py
@@ -53,7 +53,8 @@ def main():
     config = init_config()
     if not config:
         return
-    logger.log('Configuration initialized', 'yellow')
+    if config.debug:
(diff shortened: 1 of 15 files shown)


solderzzc:
There's new event system in progress, it will change the behavior as now, so we need wait for the new merging coming.

TheSavior:
Closing due to soon incoming log changes

pokecunt:
download

PR #2299, opened 2016-08-02
Replace fort with Pokestop in log output

gweedo767:
We shouldn't call them forts anywhere really.


Jasperrr91:
https://github.com/PokemonGoF/PokemonGo-Bot/pull/2279

pokecunt:
download

PR #2306, opened 2016-08-02
Follow Path with googlemaps directions api integration (follows roads and paths)

LasseSkogland:
Short Description: 
Added a new cell_worker based on `pokemongo_bot/cell_workers/follow_path.py`
It utilizes the Google Maps Directions API to follow roads and paths.
This is a "proof of concept" and could probably use some more work.
I will run some test to see how stable the code is. (it should be stable)

Fixes:
- Added  Class FollowHumanPath

### DIFF
--- a/configs/config.json.humanpath.example
+++ b/configs/config.json.humanpath.example
@@ -1,0 +1,99 @@
+{
+    "auth_service": "google",
+    "username": "YOUR_USERNAME",
+    "password": "YOUR_PASSWORD",
+    "location": "SOME_LOCATION",
+    "gmapkey": "GOOGLE_MAPS_API_KEY",
+    "tasks": [
+      {
+        "type": "HandleSoftBan"
+      },
+      {
+        "type": "CollectLevelUpReward"
+      },
+      {
+        "type": "IncubateEggs",
+        "config": {
+          "longer_eggs_first": true
+        }
+      },
+      {
+        "type": "TransferPokemon"
+      },
+      {
+        "type": "EvolveAll",
+        "config": {
+          "evolve_all": "NONE",
+          "evolve_cp_min": 300,
+          "evolve_speed": 20,
+          "use_lucky_egg": false
+        }
+      },
+      {
+        "type": "RecycleItems",
+        "config": {
+          "item_filter": {
+            "Pokeball":       { "keep" : 100 },
+            "Potion":         { "keep" : 10 },
+            "Super Potion":   { "keep" : 20 },
+            "Hyper Potion":   { "keep" : 30 },
+            "Revive":         { "keep" : 30 },
+            "Razz Berry":     { "keep" : 100 }
+          }
+        }
+      },
+      {
+        "type": "CatchVisiblePokemon"
+      },
+      {
+        "type": "CatchLuredPokemon"
+      },
+      {
+        "type": "SpinFort"
+      },
+      {
+        "type": "FollowHumanPath",
+        "config": {
+          "path_mode": "linear",
+          "path_file": "configs/path.example.json"
+        }
+      }
+    ],
+    "map_object_cache_time": 5,
+    "max_steps": 5,
+    "forts": {
+      "avoid_circles": true,
+      "max_circle_size": 50
+    },
+    "websocket_server": false,
+    "walk": 4.16,
+    "action_wait_min": 1,
+    "action_wait_max": 4,
+    "debug": false,
+    "test": false,
+    "health_record": true,
+    "location_cache": true,
+    "distance_unit": "km",
+    "reconnecting_timeout": 15,
+    "evolve_captured": "NONE",
+    "catch_randomize_reticle_factor": 1.0,
+    "catch_randomize_spin_factor": 1.0,
+    "catch": {
+      "any": {"catch_above_cp": 0, "catch_above_iv": 0, "logic": "or"},
+      "// Example of always catching Rattata:": {},
+      "// Rattata": { "always_catch" : true }
+    },
+    "release": {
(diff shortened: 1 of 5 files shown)


solderzzc:
This is nice, in my option, the follow path and follow GWalking should be separated, how do you think about?

LasseSkogland:
They are seperated.
FollowPath in `pokemongo_bot/cell_workers/follow_path.py`
FollowHumanPath in `pokemongo_bot/cell_workers/follow_path.py`

solderzzc:
@LasseSkogland  Yes, I just figured it out.
Could you please have the config example updated?

solderzzc:
Which kind of key should be used, the JS key?

pokecli.py
```diff
@@ -33,6 +33,7 @@
 import ssl
 import sys
 import time
+import googlemaps
```
solderzzc: We don't need import the googlemaps here, right?

pokemongo_bot/__init__.py
```diff
@@ -34,6 +35,10 @@ def position(self):
 
     def __init__(self, config):
         self.config = config
+        try:
+            self.gmap_client = googlemaps.Client(config.gmapkey)
```
solderzzc: The initial should better be inside Task.

LasseSkogland:
JS Key? Do you mean userdata,js.
I get the api key from config.json

solderzzc:
I mean the google map key type, it should be js key to be used inside the googlemaps package

configs/config.json.humanpath.example
```diff
@@ -0,0 +1,99 @@
...
+        "type": "CatchLuredPokemon"
+      },
+      {
+        "type": "SpinFort"
+      },
+      {
+        "type": "FollowHumanPath",
+        "config": {
+          "path_mode": "linear",
+          "path_file": "configs/path.example.json"
```
solderzzc: The google path waypoint has 21 points limitation, is that a problem?

MZorzy:
There are too much player for niantic to check if bot is not on road, and it can be on park.
but maybe will be a check if player cross a big building (minus gps lag/errors) so follow map can be good.

solderzzc:
@MZorzy  I think the GPX configuration is not user friendly, should we just calc the best route in a specific area instead of the GPX input? The bot will be much more AI driving.

LasseSkogland:
The way I do it is getting the overview_polyline points, but it isn't the most accurate.
I'll work on using the more accurate polyline points under ['routes']['legs']['steps']

solderzzc:
I have code since the project started, I intended to use GWalk for the walk. 
We can draw a route connected 'Spawn Point'.'Fort' one by one, and loop in it or jump to another cell when done according the configuration.
```
def pokemon_working():
    gmaps = googlemaps.Client(key=GOOGLEMAPS_KEY)
    directions_result = gmaps.directions((49.004, 8.456),
                                         (49.004, 8.469),
                                         mode="walking")
    if directions_result and len(directions_result) > 0:
        """
        print directions_result[0]
        for ka, va in directions_result[0].iteritems():
            print ka+'\n'
            print va
        exit(0)
        """
        steps = directions_result[0]['legs'][0]['steps']
        print len(steps)
        for index, item in enumerate(steps):
            print index,item['start_location'],'-->',item['end_location'],'duration: ',item['duration']['value'],'sec'
```

pokemongo_bot/__init__.py
```diff
@@ -103,7 +103,7 @@ def get_meta_cell(self):
                 wild_pokemons += cell["wild_pokemons"]
             if "catchable_pokemons" in cell and len(cell["catchable_pokemons"]):
                 catchable_pokemons += cell["catchable_pokemons"]
-        
+
```
solderzzc: This code should be clean up.

LasseSkogland:
Preview: (marker is original position, lines are GMaps Directions API logic)
image

pokemongo_bot/cell_workers/follow_human_path.py
```diff
@@ -0,0 +1,126 @@
...
+    def load_path(self):
+        if self.path_file is None:
+            raise RuntimeError('You need to specify a path file (json or gpx)')
+
+        path = []
+        if self.path_file.endswith('.json'):
+            path = self.load_json()
+        elif self.path_file.endswith('.gpx'):
+            path = self.load_gpx()
+        if not self.gmap_client == None:
```
solderzzc: If there's no path pass in, we should better calc the path from the cell list.

TheSavior:
I thought this is what the polyline walker is supposed to do. https://github.com/PokemonGoF/PokemonGo-Bot/blob/master/pokemongo_bot/polyline_walker/polyline_walker.py#L11

We already have this. We need our navigators to use the polyline walker instead of the stepwalker and we are golden. It probably just needs a config flag. @th3w4y Can you provide some more context on the polyline walker and its usage please?

pokemongo_bot/cell_workers/follow_human_path.py
```diff
@@ -0,0 +1,126 @@
...
+from pokemongo_bot.cell_workers.base_task import BaseTask
+from pokemongo_bot.cell_workers.utils import distance, i2f, format_dist
+from pokemongo_bot.human_behaviour import sleep
+from pokemongo_bot.step_walker import StepWalker
+from pgoapi.utilities import f2i
+
+
+class FollowHumanPath(BaseTask):
+    def get_gmap_directions(self, origin, destination):
+        directions_json = googlemaps.directions.directions(self.gmap_client, origin, destination, mode="walking", optimize_waypoints=True)
```
th3w4y: optimize_waypoints is not needed since you don't give waypoints

PR #2309, opened 2016-08-02
Add readme GNU License

maestroi:
Short Description: 
Add the gnu license to readme

Fixes:
- 
- 
-

### DIFF
--- a/README.md
+++ b/README.md
@@ -10,6 +10,7 @@
 
 # PokemonGo-Bot
 The Pokemon Go Bot, baking with community.
+"This project is licensed under the terms of the GNU General Public License."
 ## Help Needed on [Desktop Version](https://github.com/PokemonGoF/PokemonGo-Bot-Desktop)
 ## Help with the project [Dev Bot](https://github.com/PokemonGoF/PokemonGo-Bot/wiki/Develop-PokemonGo-Bot)
 ## Project Chat
--- a/README.md
+++ b/README.md
@@ -164,6 +164,7 @@ Research Website [Nice Tool](https://thesilphroad.com/research)
  * th3w4y
  * Leaklessgfy
  * GregTampa
+ * Maestro
 
 -------
 ## Credits
--- a/README.md
+++ b/README.md
@@ -9,7 +9,7 @@
 </p>
 
 # PokemonGo-Bot
-The Pokemon Go Bot, baking with community.
+The Pokemon Go Bot, baking with community.  
 "This project is licensed under the terms of the GNU General Public License Version 3."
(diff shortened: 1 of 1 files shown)


ironbishop:
which version?

tdurieux:
According to the file [LICENCE](https://github.com/PokemonGoF/PokemonGo-Bot/blob/dev/LICENSE) in dev branch the licence is MIT. It is not a good idea to have two different licence for the same project

ironbishop:
It is possible to multi-license a product. and many projects do that. It really depends on the original developer's choice.

tdurieux:
In this case, it's a licence change (no licence -> GNU licence), theoretically you must have the agreement of all the contributors.

ironbishop:
That's already been done in dev, I assume you agreed on that.

maestroi:
@tdurieux pokemongof wanted to know opinion of others but he liked the idea.

maestroi:
The problem is people selling this bot with close source.

tdurieux:
My point of view does not count I'm not a contributor... 
My comment is just to point out that the license change is not a trivial matter.

> The problem is people selling this bot with close source.

I was not aware of this situation.

ironbishop:
that's funny :) selling something useless (well, the only use is earning fake points in a fake world), and being able to find somone who buys it anyway with real name

ironbishop:
By the way, this is not going to stop people from selling the bot: in fact, using a GNU license means encuraging them to sell it! :) 

Without a license, every software is closed source, so it's illegal to copy and sell it. With a MIT or GNU license, copying and selling gets legal.

maestroi:
@ironbishop yes but it needs to be open source not closed source,
```The GNU GPLv3 is a copyleft license that requires anyone who distributes your code or a derivative work to make the source available under the same terms, and also provides an express grant of patent rights from contributors to users.```

pokecunt:
download

th3w4y:
@maestroi yes so they will be obliged to publish their changes... how will one keep track of what those changes are? and were are the modified sources available?
under GPL one is obliged to provide the code only on request even against a fee for delivering the source code on a DVD or something..

practically i am not sure that relicensing is of any use case, it only adds more obligations to the contributors at this stage.

if i want to reuse parts that i wrote for this project in a different project for example, 
then i would have to license under GPL my full new product...

maestroi:
@th3w4y there is no reference to any license atm, we could make it MIT or GPL it was just a suggestion have a license

solderzzc:
let me close, thanks for all the discuss.

PR #2334, opened 2016-08-02
add NicknamePokemon

yqrashawn:
Short Description: 

Fixes:
- 
- 
-

### DIFF
--- a/configs/config.json.pokemon.example
+++ b/configs/config.json.pokemon.example
@@ -18,6 +18,12 @@
         }
       },
       {
+        "type":"NicknamePokemon",
+        "config":{
+        "nickname_template":"{name}"
+        }
+      },
+      {
         "type": "TransferPokemon"
       },
       {


maestroi:
this is more confusing than usefull

configs/config.json.pokemon.example
```diff
@@ -18,6 +18,12 @@
         }
       },
       {
+        "type":"NicknamePokemon",
+        "config":{
+        "nickname_template":"{name}"
```
solderzzc: How about this :
"nickname_template":"{name}_{iv_pct}"

reddivision:
there are a few other example files that would need updating if we're going to do this. Also, we had a wave of reaction by our users that just "plug and play" with having all their pokemon renamed without them wanting that, so the default template should actually be `"nickname_template": ""` if anything.

TheSavior:
This was removed from the json because people copy and paste the example config and didn't want to have everything duplicated. We need a better (different) way to show people all of the different options available to them.

reddivision:
@TheSavior we've got some people that, such as brantje in slack, that have implemented config generators. maybe we should consider maintaining one with the bot as a submodule, which might ease the whole config generation process.

Issue #2332, opened 2016-08-02
KeyError: 'GET_INVENTORY'


kinvaris:
### Expected Behavior

The application to work

### Actual Behavior

```
(PokemonGo-Bot) jonas@moon:~/PokemonGo-Bot⟫ ./pokecli.py --config config.json 
[2016-08-02 06:50:01] [x] PokemonGO Bot v1.0
[2016-08-02 06:50:01] [x] Configuration initialized
[2016-08-02 06:50:01] 
[2016-08-02 06:50:01] [x] Address found: Belgium, Kortrijk
[2016-08-02 06:50:01] [x] Position in-game set as: (50.8194776, 3.2577263, 0.0)
[2016-08-02 06:50:01] 
Traceback (most recent call last):
  File "./pokecli.py", line 220, in <module>
    main()
  File "./pokecli.py", line 206, in main
    bot.start()
  File "/home/jonas/PokemonGo-Bot/pokemongo_bot/__init__.py", line 31, in start
    self._setup_api()
  File "/home/jonas/PokemonGo-Bot/pokemongo_bot/__init__.py", line 159, in _setup_api
    balls_stock = self.pokeball_inventory()
  File "/home/jonas/PokemonGo-Bot/pokemongo_bot/__init__.py", line 234, in pokeball_inventory
    inventory_dict = inventory_req['responses']['GET_INVENTORY'][
KeyError: 'GET_INVENTORY'
```

### Steps to Reproduce

`./pokecli.py --config config.json`

### Other Information
OS: `Linux moon 3.13.0-85-generic #129-Ubuntu SMP Thu Mar 17 20:50:15 UTC 2016 x86_64 x86_64 x86_64 GNU/Linux`
Git Commit: `4182e93f9bfbe471d07b74e95c55cdfbacda127b`
Python Version: `Python 2.7.6`

Pawamoy:
Happened to me in master branch few days ago. I just launched the bot again and it worked.

Lunik:
Issue: #1491

maestroi:
works in dev.

INGCRENGIFO:
Hello, I read in several posts that say, in dev works. But I can download dev ??? that command must write ??

maestroi:
@INGCRENGIFO read the wiki..

maestroi:
please close @MFizz is right

INGCRENGIFO:
[...]

PR #2344, opened 2016-08-02
[Fix]inventory key error

MFizz:
Short Description: 
Added call loop if response is empty, also checking for keys before using them

Fixes:
- #2332
- #1491


TheSavior:
Looks like you have some merge conflicts @MFizz

PR #2350, opened 2016-08-02
fix(docker): correct web config file path

geminiyellow:
Short Description: 

Fixes:
- `docker-compose` run web will cause `userdata.js` cannot found error


Kassius509:
This is in correct according to the current documentation. The config folder only holds the userdata.js.example file. The active userdata.js file should be stored in the /web folder.

TheSavior:
👍

PR #2354, opened 2016-08-02
Updated api wrapper to handle pgoapi updates

reddivision:
Short Description: 
Title says it all. I added some checks to make the wrapper also work with earlier pgoapi versions, just in case.


DayBr3ak:
You beat me to it haha #2357

PR #2361, opened 2016-08-02
Recycle lower level pokeballs to retain the best balls.

Jasperrr91:
Short Description: 
At the moment, you need to specify how many PokeBalls, GreatBalls, UltraBalls and MasterBalls you want to keep. Preferably you only want MasterBalls, but if you set Pokeballs to 0; you'll need to spin tons of pokestops before finally getting MasterBalls to catch a pokemon.

A better system would be to just set the total amount of Balls you want. This script allows you to do that. Simply update your config to:
`{
        "type": "RecycleItems",
        "config": {
          "item_filter": {
            "101": { "keep" : 0 },
            "102": { "keep" : 0 },
            "103": { "keep" : 25 },
            "104": { "keep" : 25 },
            "201": { "keep" : 25 },
            "701": { "keep" : 50 }
          },
          "balls": 200
        }`

Then when you have 10 normal PokeBalls and 190 GreatBalls and a pokestop gives you 3 GreatBalls; the bot will discard 3 normal Pokeballs.

So to summarize: this PR makes sure you'll have the best PokeBalls available.

Todo:
- Update config examples

### DIFF
--- a/pokemongo_bot/cell_workers/recycle_items.py
+++ b/pokemongo_bot/cell_workers/recycle_items.py
@@ -8,6 +8,7 @@ class RecycleItems(BaseTask):
     def initialize(self):
         self.item_filter = self.config.get('item_filter', {})
         self._validate_item_filter()
+        self.balls = self.config.get('balls', {})
 
     def _validate_item_filter(self):
         item_list = json.load(open(os.path.join('data', 'items.json')))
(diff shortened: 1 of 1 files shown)


douglascamata:
@Jasperrr91 this needs to come with better pokeball choice logics, otherwise ppl will just get a full inventory and never use better balls.

Jasperrr91:
@douglascamata true, it's not giving you the full benefit because any normal pokeballs will be used first. But it's giving you a small benefit nonetheless. So I think it's best to open a new PR to maximise the benefits of this hack.

Kassius509:
Could we make it so the Order of the items in RecycleItems config item_filter would determine which items to keep (prioritizing).

douglascamata:
@Kassius509: dicts have no order.

Jasperrr91:
@Kassius509 the script compares each item's inventory count to their respective filter setting.

adrianoresende:
@bstpierr @douglascamata please aprove this feature ;)

k4n30:
@Jasperrr91 this merge is severly outdated. How do you want to proceed?

PR #2370, opened 2016-08-02
Improves bot performance

mako-taco:
- removes sleep from step-walker in favor of minimum-tick-time in main loop
- prevents evolve-all-worker from running every tick
- transferring pokemon worker runs every nth tick (configurable)
- adds performance time logging with --log_timing_info (defaults to false)

image

image

```
    "location": "Old Town Alexandria",
    "walk": 40.0,
    "minimum_tick_time": 2.0,
    "action_wait_min": 1,
    "action_wait_max": 4,
```

### DIFF
--- a/pokemongo_bot/cell_workers/transfer_pokemon.py
+++ b/pokemongo_bot/cell_workers/transfer_pokemon.py
@@ -5,7 +5,13 @@ from pokemongo_bot.human_behaviour import action_delay
 from pokemongo_bot.cell_workers.base_task import BaseTask
 
 class TransferPokemon(BaseTask):
+    def initialize(self):
+        self.every_nth_tick = self.config.get('every_nth_tick', 5)
+
     def work(self):
+        if not self._should_run():
+            return
+
         pokemon_groups = self._release_pokemon_get_groups()
         for pokemon_id in pokemon_groups:
             group = pokemon_groups[pokemon_id]
@@ -68,6 +74,9 @@ class TransferPokemon(BaseTask):
 
                         if self.should_release_pokemon(pokemon_name, pokemon_cp, pokemon_potential):
                             self.release_pokemon(pokemon_name, item['cp'], item['iv'], item['pokemon_data']['id'])
+
+    def _should_run(self):
+        return self.bot.tick_count % self.every_nth_tick is 0
 
     def _release_pokemon_get_groups(self):
         pokemon_groups = {}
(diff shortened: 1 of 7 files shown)


pokemongo_bot/tree_config_builder.py
```diff
@@ -27,7 +27,7 @@ def build(self):
             task_config = task.get('config', {})
 
             worker = self._get_worker_by_name(task_type)
-            instance = worker(self.bot, task_config)
+            instance = worker(self.bot, task_config, task_type)
```
TheSavior: Why do we want to pass in the name of the task to the task itself? It already knows. It's its class name

pokemongo_bot/cell_workers/transfer_pokemon.py
```diff
@@ -5,7 +5,13 @@
 from pokemongo_bot.cell_workers.base_task import BaseTask
 
 class TransferPokemon(BaseTask):
+    def initialize(self):
```
TheSavior: I think we need a smarter way to speed up this task, I'm definitely a :-1: on checking the tick. I don't even like that we have a flag for the first tick.

Another approach here would be for the task to have a list of the pokemon it saw on the last tick, and then on the next tick if the pokemon have changed, run the transfer check on only the relevant ones.

I think this should be undone in this PR and done properly in a standalone PR.

pokemongo_bot/tree_config_builder.py
mako-taco: I don't know python

TheSavior:
@mako-taco I think a lot of this functionality is valuable, but this PR is changing too many things. It makes it too hard for us to review, and revert individual pieces if necessary. It also keeps us from being able to make intelligent merge decisions to reduce the number of merge conflicts people will have.

Can you break this into individual PRs that can be reviewed separately? Logging, EvolveAll, TransferPokemon, StepWalker? Then we can review the individual merits and implementation of each thing. Doing this also means they will also get merged faster.

pokemongo_bot/cell_workers/transfer_pokemon.py
mako-taco: I think it should only run when the pokemon bag is full or on startup, but I don't think we have a way to check number of pokemon w/o hitting the server, and this was a much faster fix.  Do you have any ideas?

TCB13:
I wish this was added!

PR #2358, opened 2016-08-02
Dev Proxy support

alecuba16:
Short Description:  Proxy support implemented with exports.

Fixes:
-  Cannot log-in on banned cloud services by using a proxy.
- 
-

PR #2374, opened 2016-08-02
Revert "Dev Proxy support"

TheSavior:
Reverts PokemonGoF/PokemonGo-Bot#2358

The config values should not be required


TheSavior:
@alecuba16, can you fix those config values?

alecuba16:
@TheSavior Removed the required directive, now they are fully optional and none by default.

Issue #2372, opened 2016-08-02
[FEATURE] Only use lucky egg/evolve worker when at least x # of pokemon to evolve. 


Kassius509:
Would be nice to only use candies/evolve when there's a minimum number of pokemon that are eligible to evolve (i.e. Pkmn that are both on the list of evolution and meet the given CP reqs.)

PR #2390, opened 2016-08-02
Evolve only if evolvable candidates no less than a certain number.

coderplay:
Previously, when user configured use_lucky_egg as true, the bot will use
lucky egg even evolvale candidates is zero. That's kinda of
waste. Meanwhile, if user didn't configured use_lucky_egg, the bot will
try to evolve as soon as possible after a catch, but sometimes no good
 potential pokemon in that batch, it's better delay the evolution after a
while. By adding a config item evolve_num_min can cope above two
problems. The bot evolves only if evolvable candidates no less than a
certain number.

1. Evolve only if evolvable candidates no less than a certain number.
2. Dragoniar's configuration is abnormal in pokemon.json, which lack of
"Previous evolution(s)" attribute, while the other middle-tier pokemon
has.


coderplay:
This fixed #2372

douglascamata:
@coderplay you PR is completely messed up. You mind opening another one?

coderplay:
@douglascamata sure.

PR #2392, opened 2016-08-02
remove sleep from step walker, makes ticks take minimum time

mako-taco:
@TheSavior I have begun breaking changes out of #2370 
This removes `sleep(1)` from movement, and replaces it with a configurable minimum time for each tick to take.

### DIFF
--- a/pokemongo_bot/__init__.py
+++ b/pokemongo_bot/__init__.py
@@ -78,6 +78,8 @@ class PokemonGoBot(object):
         # self.event_manager.emit('location', 'level'='info', data={'lat': 1, 'lng':1}),
 
     def tick(self):
+        start_time = time.time()
+
         self.cell = self.get_meta_cell()
         self.tick_count += 1
 
@@ -86,7 +88,11 @@ class PokemonGoBot(object):
 
         for worker in self.workers:
             if worker.work() == WorkerResult.RUNNING:
-                return
+                break
+
+        elapsed_time = time.time() - start_time
+        sleep_time = max(0, self.config.minimum_tick_time - elapsed_time)
+        time.sleep(sleep_time)
 
     def get_meta_cell(self):
         location = self.position[0:2]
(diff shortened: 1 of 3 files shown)


yungfrancis:
  File "pokecli.py", line 450, in <module>
    main()
  File "pokecli.py", line 85, in main
    report_summary(bot)
  File "pokecli.py", line 93, in report_summary
    metrics.capture_stats()
  File "/Users/Fran/PokemonGo-Bot/pokemongo_bot/metrics.py", line 73, in capture_stats
    request = self.bot.api.create_request()
AttributeError: 'PokemonGoBot' object has no attribute 'api'

Maybe merged from an old version? :(

pokemongo_bot/__init__.py
```diff
@@ -78,15 +78,26 @@ def _setup_event_system(self):
...
+            worker_start = time.time()
+            result = worker.work()
+            worker_time = time.time() - worker_start
+
+            if result == WorkerResult.RUNNING:
+                break
+
+        elapsed_time = time.time() - start_time
+        sleep_time = max(0, self.config.minimum_tick_time - elapsed_time)
+        if sleep_time > 0:
```
douglascamata: `sleep(0)` works, don't need this `if` here.

mako-taco: yeah this, and some of the  `worker_start` and `worker_time` lines were actually part of a commit that involved performance logging.  I'll remove these.

Nihisil:
@mako-taco can you, please, merge latest dev to your PR?

TheSavior:
The only reason to specify a minimum tick time is because we want to rate limit our requests to the api. We are already artificially slowing the bot by making a max of 2 requests a second. We don't want to artificially slow it this way because it may not make sense if people only have one task in the task list, for example.

mako-taco:
I've stopped working on this project, feel free to commandeer the PR

Issue #2391, opened 2016-08-02
[dev] Bot using lucky eggs before reconnect after API exception


Vvkmnn:
### Expected Behavior
Bot should retry until API lets it through, not break on the `pgoapi` exception.

### Actual Behavior
```
[16:16:23] Server is throttling, let's slow down a bit
[16:16:25] Server is throttling, let's slow down a bit
[16:16:26] Server is throttling, let's slow down a bit
[16:16:27] Server is throttling, let's slow down a bit
[16:16:28] Server is throttling, let's slow down a bit
[16:16:29]
[16:16:29]  Ran for 0:00:12
[16:16:29]  Total XP Earned: 0  Average: 0.00/h
[16:16:29]  Travelled 0.00km
[16:16:29]  Visited 0 stops
[16:16:29]  Encountered 0 pokemon, 0 caught, 0 released, 0 evolved, 0 never seen before
[16:16:29]  Threw 0 pokeballs
[16:16:29]  Earned 0 Stardust
[16:16:29]
[16:16:29]  Highest CP Pokemon:
[16:16:29]  Most Perfect Pokemon:
Traceback (most recent call last):
  File "pokecli.py", line 442, in <module>
    main()
  File "pokecli.py", line 71, in main
    bot.tick()
  File ".../PokemonGo-Bot/pokemongo_bot/__init__.py", line 92, in tick
    if worker.work() == WorkerResult.RUNNING:
  File ".../PokemonGo-Bot/pokemongo_bot/cell_workers/evolve_all.py", line 19, in work
    if not self._should_run():
  File ".../PokemonGo-Bot/pokemongo_bot/cell_workers/evolve_all.py", line 73, in _should_run
    response_dict_lucky_egg = self.bot.use_lucky_egg()
  File ".../PokemonGo-Bot/pokemongo_bot/__init__.py", line 376, in use_lucky_egg
    return self.api.use_item_xp_boost(item_id=301)
  File ".../PokemonGo-Bot/venv/src/pgoapi/pgoapi/pgoapi.py", line 84, in function
    return request.call()
  File ".../PokemonGo-Bot/pokemongo_bot/api_wrapper.py", line 109, in call
    raise ServerSideRequestThrottlingException('Server throttled too many times')
pgoapi.exceptions.ServerSideRequestThrottlingException: Server throttled too many times
```

### Steps to Reproduce
Just switch locations and retry. 

### Other Information
OS: OSX 10.10.5
[...]

PR #2393, opened 2016-08-02
Followup on PGoApi -- delaying errors, and reducing noise

DayBr3ak:
__Short Description__: 

More retries, and a catch for the exception ServerSideThrottling, that try to reconnect the bot

__Fixes__:
- #2391

### DIFF
--- a/pokecli.py
+++ b/pokecli.py
@@ -35,7 +35,7 @@ import sys
 import time
 from datetime import timedelta
 from getpass import getpass
-from pgoapi.exceptions import NotLoggedInException
+from pgoapi.exceptions import NotLoggedInException, ServerSideRequestThrottlingException, ServerBusyOrOfflineException
 from geopy.exc import GeocoderQuotaExceeded
 
 from pokemongo_bot import PokemonGoBot, TreeConfigBuilder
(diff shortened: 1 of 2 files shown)


DayBr3ak:
@douglascamata could you review this when you have time?

JSchwerberg:
30 seconds when you get throttled?  How does this work in practice?  Seems like it would slow the bot down immensely, given how often the bot likes to throttle you to begin with.

DayBr3ak:
I mean I raised the try count to 15... if you get throttled 15+ times, it's really not the bot the problem. For more details we need a niantic networking expert

JSchwerberg:
Oh! I was reading this backwards.  I see now; if you get throttled 15 times in a row, it waits 30 seconds.  For some reason, I read it as waiting 30 seconds after each individual throttle response.  My bad. :)

pokecli.py
```diff
@@ -74,9 +74,12 @@ def main():
             logger.log('Exiting PokemonGo Bot', 'red')
             finished = True
             report_summary(bot)
-        except NotLoggedInException:
+        except (NotLoggedInException, ServerBusyOrOfflineException):
             logger.log('[x] Error while connecting to the server, please wait %s minutes' % config.reconnecting_timeout, 'red')
             time.sleep(config.reconnecting_timeout * 60)
+        except ServerSideRequestThrottlingException:
+            logger.log('Server is throttling, reconnecting in 30sec')
+            time.sleep(30)
```
douglascamata: 30 SECS?!

JSchwerberg: You're making the same mistake I did, @douglascamata -- it's 30 seconds if you get throttled 15 times in a row.

DayBr3ak: Yes it's a pause after 15 tries. This except is there to prevent the bot from crashing

douglascamata: then.... it's a trap!

PR #2396, opened 2016-08-02
evolve_all filter is not functional

coderplay:
During the refactoring 987f730, a _validate_config method was
introduced, which never be used anywhere, which make the config item
evolve_all is not functional as before.


JSchwerberg:
This has already been fixed, and merged into dev branch. 
550b2c3c357ce2f40aade8e63c3efda9beae6ab7

PR #2402, opened 2016-08-02
Allow for 3-7 decimal points for coordinates

L422Y:
Some exports only provide three decimal point accuracy, and sometimes 4.440000 turns into 4.44, need to adjust the regex.


TheSavior:
:+1:

PR #2415, opened 2016-08-03
Showing disance while walking

SidSyrus:
Short Description: 

Added text to show the distance while **Walking**. That would give a user an idea on when their egg might hatch.

### DIFF
--- a/pokemongo_bot/stepper.py
+++ b/pokemongo_bot/stepper.py
@@ -68,7 +68,7 @@ class Stepper(object):
         steps = (dist + 0.0) / (speed + 0.0)  # may be rational number
         intSteps = int(steps)
         residuum = steps - intSteps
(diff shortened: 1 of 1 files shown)


pokemongo_bot/stepper.py
```diff
@@ -68,7 +68,7 @@ def _walk_to(self, speed, lat, lng, alt):
         steps = (dist + 0.0) / (speed + 0.0)  # may be rational number
         intSteps = int(steps)
         residuum = steps - intSteps
-        logger.log('[#] Walking from ' + str((i2f(self.api._position_lat), i2f(
+        logger.log('[#] Walking ' + str(ceil(dist)) + 'M from ' + str((i2f(self.api._position_lat), i2f(
```
TheSavior: Can you use the format_dist function from utils? That way it will be in whatever unit the user has configured.

PR #2432, opened 2016-08-03
Automated a script that restarts pokecli.py if it crashes

hubchub:
not sure where to post this but if you create a script with the following it will restart pokecli.py if it crashes so you can leave it running all day. 

import subprocess

while True:
    try:
        print subprocess.check_output(['python', 'pokecli.py'])
    except KeyboardInterrupt:
        break
    except subprocess.CalledProcessError:
        continue


route12345:
`import subprocess

while True:
try:
print subprocess.check_output(['python', 'pokecli.py'])
except KeyboardInterrupt:
break
except subprocess.CalledProcessError:
continue`

Is not working:
$ python restart.py
  File "restart.py", line 4
    try:
      ^
IndentationError: expected an indented block

reddivision:
You're trying to merge a branch directly to master. Also, it doesn't look like you're actually merging any new code, just making a suggestion? If you have something to submit, please create a new branch and merge that to dev. Otherwise, open an Issue with your suggestion and it can probably be added to the wiki. Closing this regardless.

PR #2440, opened 2016-08-03
Fix to display stats on iTerm2 terminal

lucasfevi:
**Short Description:**
Displays stats on iTerm2 terminal.

**Long Description:**
The PR #2252 was working on the default OS X terminal but not on iTerm2, now with this fix it is working on both.

### DIFF
--- a/pokemongo_bot/cell_workers/update_title_stats.py
+++ b/pokemongo_bot/cell_workers/update_title_stats.py
@@ -6,7 +6,6 @@ from pokemongo_bot.cell_workers.base_task import BaseTask
 from pokemongo_bot.worker_result import WorkerResult
 from pokemongo_bot.tree_config_builder import ConfigException
 
-
 class UpdateTitleStats(BaseTask):
     """
     Periodically updates the terminal title to display stats about the bot.
@@ -108,8 +107,10 @@ class UpdateTitleStats(BaseTask):
         :raise: RuntimeError: When the given platform isn't supported.
         """
         if platform == "linux" or platform == "linux2"\
-                or platform == "darwin" or platform == "cygwin":
+                or platform == "cygwin":
             stdout.write("\x1b]2;{}\x07".format(title))
+        elif platform == "darwin":
+            stdout.write("\033]0;{}\007".format(title))
         elif platform == "win32":
             ctypes.windll.kernel32.SetConsoleTitleA(title)
         else:
--- a/CONTRIBUTORS.md
+++ b/CONTRIBUTORS.md
@@ -50,3 +50,4 @@
  * z4ppy.bbc
  * matheussampaio
  * Abraxas000
+ * lucasfevi


pokemongo_bot/cell_workers/update_title_stats.py
```diff
@@ -108,8 +107,10 @@ def _update_title(self, title, platform):
         :raise: RuntimeError: When the given platform isn't supported.
         """
         if platform == "linux" or platform == "linux2"\
-                or platform == "darwin" or platform == "cygwin":
+                or platform == "cygwin":
             stdout.write("\x1b]2;{}\x07".format(title))
+        elif platform == "darwin":
```
TheSavior: The intended behavior seems backwards to me. Darwin is mac, right? So that would mean that iterm2 is cygwin, right? In which case, shouldn't the new format option be for cygwin instead of darwin?

lucasfevi: on my tests on my mac iterm2 was darwin too

Moonlight-Angel:
That's weird indeed.
I tested this on iTerm inside an OS X (El Capitan) VM and it worked just fine.

Thanks for contributing. :+1:

TheSavior:
@Moonlight-Angel Can you confirm that this change has the intended behavior in mac terminal, iterm and iterm2? If you are cool with this change, then so am I. :)

nullpixel1:
I know I'm not really in this conversation, but I use iTerm 2, and using this fork did give me the intended behaviour @TheSavior

Moonlight-Angel:
Tested on Mac default terminal and iTerm2 3.0.5 (latest stable).
Behaves as intended.
:+1:

TheSavior:
👍

Nihisil:
👍

douglascamata:
👍

Moonlight-Angel:
@lucasfevi I just noticed something.
Maybe you should have also modified the unit test to test `stdout.write` gets called with the good markup under `darwin` platform.

I'm currently working on adding some more stats to display. If you don't mind, I'll take care of modifying the tests.

Moonlight-Angel:
@lucasfevi Whatever, had time to kill, included it in this PR. https://github.com/PokemonGoF/PokemonGo-Bot/pull/2494

lucasfevi:
@Moonlight-Angel haha perfect. Since you're still working on it, do you know if it is a normal behavior to not show the stats on the tab if I'm connected via SSH? My bot is running on google cloud and the tab stats is not working, which I think is an expected behavior. So I added a `logger.log` to show the same stats.

Also that may be a good feature: choose to either change the title or print the stats.

Moonlight-Angel:
@lucasfevi I'm not sure if the client terminal title is updated when some process updates the title while you're ssh'ing on it.

lucasfevi:
@Moonlight-Angel yeah you're right lol

PR #2453, opened 2016-08-03
Add simple nearby pokemon hunter cellworker (+ radar output feature)

z4ppy:
Generate simple search path and run it when pokemons are nearby bot (config : vips, count).
Can output the game radar (option)

TODO
- better search path
- an algorithm to approximate Pokemon vips
- handle vip or better place to hunt meanwhile started search (reinit)
- more options (start when no move, lures camping, bag full, ...)
- PokemonMap Synergie : vips position deduction (encounter_id)

### DIFF
--- a/configs/config.json.example
+++ b/configs/config.json.example
@@ -57,11 +57,11 @@
       {
         "type": "HuntNearbyPokemon",
         "config": {
-          "start_count_pokemon": 9,
+          "start_count_pokemon": 8,
           "start_vip": true,
           "start_min_anyball": 30,
           "stop_min_anyball": 10,
-          "stop_max_time": 60,
+          "stop_max_time": 40,
           "detect_radius": 40,
           "search_radius": 200,
           "show_radar": true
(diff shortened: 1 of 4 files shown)


yungfrancis:
Awesome work fella, testing it and I love it.

douglascamata:
There's already a task to snipe pokemons based on PokemonMap. 

This PRs has too much output, please only log as little as possible and only actions that are really necessary to be logged.

Nik0lashka:
@z4ppy, in #1992 script loads data from https://github.com/AHAAAAAAA/PokemonGo-Map and tries to catch pokemons by simple logic and some hacks. But if you will add your path ideas or may be some sort of "best route", that helps to visit maximum amount of visible pokemons before they vanish using "Traveling Salesman Problem" algorythms - that would be great.

z4ppy:
Fixed:  log  & snail speed

TheSavior:
I'm not sure why this task looks like it does so many things. Catching pokemon should be separate from movement. What does this do that the other tasks don't do?

k4n30:
@z4ppy this merge is severly outdated. How do you want to proceed?

solderzzc:

PR #2454, opened 2016-08-03
Added a configuration option "path_startmode"

JaapMoolenaar:
_Python is most definitely not my main programming language, however, I thought this was a nice addition to the path walker nonetheless._

Short Description: 
Updates the FollowPath walker to allow a "path_startmode" configuration option. Setting it to "first" mimics the current behavior: move to the first point in the list and walk through them from there. Setting it to the "new" value of "closest" will try to find the closest point to the bots position and first move there before continuing along the other point in the path.

I also added a log message to see which point the bot is walking to (so one can tell it's still walking)

### DIFF
--- a/configs/config.json.path.example
+++ b/configs/config.json.path.example
@@ -58,6 +58,7 @@
         "type": "FollowPath",
         "config": {
           "path_mode": "loop",
+          "path_startmode": "first",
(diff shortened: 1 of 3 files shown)


JaapMoolenaar:
I botched up the commits. I will create a new PR.

PR #2166, opened 2016-08-01
Revert "Adding an Anyball item to the release config"

douglascamata:
Reverts PokemonGoF/PokemonGo-Bot#2140, because it is causing error:

```
Traceback (most recent call last):
  File "./pokecli.py", line 465, in <module>
    main()
  File "./pokecli.py", line 70, in main
    bot.tick()
  File "/root/testen/dev-ptc2/pokemongo_bot/__init__.py", line 86, in tick
    if worker.work() == WorkerResult.RUNNING:
  File "/root/testen/dev-ptc2/pokemongo_bot/cell_workers/recycle_items.py", line 18, in work
    self.recycle_pokeballs()
  File "/root/testen/dev-ptc2/pokemongo_bot/cell_workers/recycle_items.py", line 72, in recycle_pokeballs
    recycle_amount = total_amount_of_balls - number_of_balls_to_keep
TypeError: unsupported operand type(s) for -: 'int' and 'NoneType'```

PR #2459, opened 2016-08-03
Added option to prioritize the best balls

aleskr:
Short Description: 
Added the option to use Anyball in the config of what to keep. If there is more pokeballs(of all types combined) than specified by the config to keep the worst balls will be recycled first.

Usage:
Remove all Pokeball filters and add Anyball and the number of balls you want to keep in total.

Fixes:
- #2166

### DIFF
--- a/data/items.json
+++ b/data/items.json
@@ -5,6 +5,7 @@
     "2": "Greatball",
     "3": "Ultraball",
     "4": "Masterball",
+    "5": "Anyball",
 
     "101": "Potion",
     "102": "Super Potion",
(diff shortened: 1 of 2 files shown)


amal:
Brilliant! Please do the same for potions.

aleskr:
I was considering rewriting it and doing the same for potions and revives, and just throwing away the rest. I really don't se any reason for keeping "Pokeball" "Greatball" and so on...

MrCarlosJr:
Necessary use any ball to for capture pokemon.

Evanito:
For some reason current code doesn't use ultraballs when out of poke and great balls

aleskr:
@Evanito @MrCarlosJr This has nothing to do with capturing, this is for recycling as stated in the description...

mjmadsen:
Love this. @aleskr any progress on this lately?

Issue #2443, opened 2016-08-03
Error on startup and with Map


alexanderrock:
### Expected Behavior
Bot starts

### Actual Behavior
Bots stopped with following error

`return log.logger('Map location data was not valid', 'red')
NameError: global name 'log' is not defined`

Could not reach PokemonGo-Map Server

### Steps to Reproduce
Config

> {
    "auth_service": "xxx",
    "username": "xxx",
    "password": "xxx",
    "location": "xxx",
    "gmapkey": "xxx",
    "tasks": [
      {
        "type": "HandleSoftBan"
      },
      {
        "type": "CollectLevelUpReward"
      },
      {
        "type": "IncubateEggs",
        "config": {
          "longer_eggs_first": true
        }
      },
      {
        "type": "TransferPokemon"
      },
      {
        "type": "EvolvePokemon",
        "config": {
          "evolve_all": "NONE",
          "evolve_cp_min": 300,
          "evolve_speed": 20,
          "use_lucky_egg": false
        }
      },
      {
        "type": "RecycleItems",
        "config": {
          "item_filter": {
            "Potion":         { "keep": 5 },
            "Super Potion":   { "keep": 5 },
            "Hyper Potion":   { "keep": 5 },
            "Max Potion":     { "keep": 5 },
            "Revive":         { "keep": 5 },
            "Razz Berry":     { "keep": 5 },
            "Pokeball":       { "keep": 200 }
          }
        }
      },
      {
        "type": "CatchVisiblePokemon"
      },
      {
        "type": "CatchLuredPokemon"
      },
      {
        "type": "SpinFort"
      },
      {
        "type": "MoveToMapPokemon",
        "config": {
          "address": "http://localhost:8000",
          "max_distance": 500,
          "min_time": 60,
          "prioritize_vips": true,
          "snipe": false,
          "update_map": true,
          "mode": "priority",
          "catch": {
            "==========Legendaries==========": 0,
            "Aerodactyl": 1000,
            "Snorlax": 1000,
            "Articuno": 1000,
            "Zapdos": 1000,
            "Moltres": 1000,
            "Dratini": 1000,
            "Dragonair": 1000,
            "Dragonite": 1000,
            "Mewtwo": 1000,
            "Mew": 1000,

            "==========Region Locked==========": 0,
            "Farfetch'd": 1000,
            "Kangaskhan": 1000,
            "Mr. Mime": 1000,
            "Tauros": 1000,

[...]

Issue #2450, opened 2016-08-03
[Feature] Snipe_min


Lunik:
### Expected Behavior
Only snipe pokemon with more than 100 in the `MoveToMapPokemon` config
adding `snipe_min: 100`

### Actual Behavior
Snipe every pokemon

### Steps to Reproduce
````
"config": {
          "address": "http://localhost:5050",
          "max_distance": 500,
          "min_time": 60,
          "prioritize_vips": true,
          "snipe": false,
          "update_map": true,
          "mode": "priority",
          "catch": {
            ...
          }
}
````

### Other Information
OS: Linux
Git Commit: 361c0fa11783085ea201a71a99f6c43859e8c239
Python Version: Python 2.9.7

mhdasding:
just remove the pokemon from the caught list if you don't want them to be sniped?

Lunik:
Hum yes xD That is a stupide question... 
Thanks :)

Lunik:
Yes why not :)
But my point is: If snip option is enable, the bot will snip every pokemon. Could we add some filter.

mhdasding:
it doesn't snipe every pokemon, only those defined in the caught list

Lunik:
Yes, but we could have some pokemon in the caught list, but not snip all of them

mhdasding:
why have them in the caught list if you don't want to snipe them?

Lunik:
To follow them if there is anything to catch

mhdasding:
created #2463

Lunik:
Thanks :)

PR #2463, opened 2016-08-03
[Feature] add config for minimum snipe priority #2450

mhdasding:
Short Description: 
if snipe is enabled only snipe pokemon with a minimum priority, still walk to pokemon below this priority.

### DIFF
--- a/configs/config.json.map.example
+++ b/configs/config.json.map.example
@@ -59,6 +59,7 @@
           "min_time": 60,
           "prioritize_vips": true,
           "snipe": false,
+          "snipe_min_prio": 100,
           "update_map": true,
           "mode": "priority",
           "catch": {
--- a/pokemongo_bot/cell_workers/move_to_map_pokemon.py
+++ b/pokemongo_bot/cell_workers/move_to_map_pokemon.py
@@ -46,7 +46,7 @@ class MoveToMapPokemon(BaseTask):
             try:
                 pokemon['encounter_id'] = long(base64.b64decode(pokemon['encounter_id']))
             except TypeError:
-                log.logger('base64 error: {}'.format(pokemon['encounter_id']), 'red')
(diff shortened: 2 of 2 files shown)


mhdasding:
also fixes #2443

chrismapes:
This sounds great! How would you feel about a PR that did:

- Attempt to "drive" to pokemon instead of snipe
   - Get lat lon of pokemon, send current coordinates and pokemon coordinates to Google Directions Server
   - Get Path
   - Follow Path at random speed between 30-45 mph
   - Ease out speed upon approach to pokemon to 5 mph.
   - Catch pokemon
   - Either drive to next pokemon on snipe queue, or return to walk area and resume walk.

- Have option to snipe only if pokemon is impossible to reach in time via "driving". Maybe also drive closer before sniping to minimize distance and potentially reduce future ban hammer likelihood.

If you don't think this is valuable I'll probably end up writing a PR for it in the next week, I'm just up to my ears with real "work" dev work and don't have time ATM. My only hesitation with the snipe feature is that if Niantic is keeping a history of pokemon spawned and the nest they were spawned at (thereby knowing lat long), and user location history, it's going to be really easy for them to write a filter to find pokemon catches that weren't possible due to unrealistic player distances from the nest and do a mass ban. I'd hate to build up my bot player and then suddenly out of the blue get banned.

mhdasding:
@PlanetoftheMapes that would be totally valuable :D i suggest putting the driving logic tho in a new `DriveWalker` so it could be interchanged for `StepWalker`.

PlanetoftheMapes:
@mhdasding Good call, I'll dig into it next week unless someone else ends up championing it.

maradlgl:
please fix line 107
error

mhdasding:
will close this for now, will reopen a PR after some major PRs got merged

PR #2465, opened 2016-08-03
[Feature] Recycle Threshold

nikhil-pandey:
**Short Description:** 
Do not recycle items in bag until a threshold is reached.

### DIFF
--- a/configs/config.json.cluster.example
+++ b/configs/config.json.cluster.example
@@ -36,7 +36,7 @@
       {
         "type": "RecycleItems",
         "config": {
-          "threshold": 300,
+          "min_inventory_items_to_run": 300,
           "item_filter": {
             "Pokeball":       { "keep" : 100 },
             "Potion":         { "keep" : 10 },
--- a/configs/config.json.cluster.example
+++ b/configs/config.json.cluster.example
@@ -36,7 +36,7 @@
       {
         "type": "RecycleItems",
         "config": {
-          "min_inventory_items_to_run": 300,
+          "min_empty_space": 15,
           "item_filter": {
             "Pokeball":       { "keep" : 100 },
             "Potion":         { "keep" : 10 },
--- a/configs/config.json.cluster.example
+++ b/configs/config.json.cluster.example
@@ -36,6 +36,7 @@
       {
         "type": "RecycleItems",
         "config": {
+          "threshold": 300,
           "item_filter": {
             "Pokeball":       { "keep" : 100 },
             "Potion":         { "keep" : 10 },
--- a/pokemongo_bot/cell_workers/recycle_items.py
+++ b/pokemongo_bot/cell_workers/recycle_items.py
@@ -9,7 +9,7 @@ class RecycleItems(BaseTask):
 
     def initialize(self):
         self.min_inventory_items_to_run = self.config.get('min_inventory_items_to_run', None)
-        self.run_when_storage_less_than = self.config.get('min_inventory_items_to_run', None)
+        self.run_when_storage_less_than = self.config.get('run_when_storage_less_than', None)
         self.item_filter = self.config.get('item_filter', {})
         self._validate_item_filter()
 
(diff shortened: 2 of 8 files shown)


Lunik:
+1

nikhil-pandey:
@douglascamata @TheSavior Could you please review this

douglascamata:
@nikhil-pandey I suggest that you rename the config to `"min_inventory_items_to_run"`. It's verbose that it's more intuitive.

nikhil-pandey:
@douglascamata Done.

TheSavior:
:+1:

I believe we will be waiting to merge this until after logging gets converted to events.

douglascamata:

pokemongo_bot/cell_workers/recycle_items.py
```diff
@@ -3,10 +3,12 @@
 from pokemongo_bot.base_task import BaseTask
 from pokemongo_bot.tree_config_builder import ConfigException
 
+
 class RecycleItems(BaseTask):
     SUPPORTED_TASK_API_VERSION = 1
 
     def initialize(self):
+        self.min_inventory_items_to_run = self.config.get('min_inventory_items_to_run', 0)
```
TheSavior: Instead of the minimum number of items we have to have, maybe it should be the number of slots we have left in the bag. That way if people purchase bag upgrades, they don't need to change their config.

TheSavior: I think it would still be great to make this change.

TheSavior:
I think it would still be great if instead of the minimum number of items we have to have, maybe it should be the number of slots we have left in the bag. That way if people purchase bag upgrades, they don't need to change their config.

nikhil-pandey:
@TheSavior I have added both options.
Working on fixing errors.

nikhil-pandey:
Working on improving names and messages.

nikhil-pandey:
Any suggestions on name of the configs? Or on any other thing?

configs/config.json.cluster.example
```diff
@@ -36,6 +36,7 @@
       {
         "type": "RecycleItems",
         "config": {
+          "min_inventory_items_to_run": 300,
```
TheSavior: I think these need to be updated

pokemongo_bot/cell_workers/recycle_items.py
```diff
@@ -3,10 +3,13 @@
 from pokemongo_bot.base_task import BaseTask
 from pokemongo_bot.tree_config_builder import ConfigException
 
+
 class RecycleItems(BaseTask):
     SUPPORTED_TASK_API_VERSION = 1
 
     def initialize(self):
+        self.min_inventory_items_to_run = self.config.get('min_inventory_items_to_run', None)
+        self.run_when_storage_less_than = self.config.get('run_when_storage_less_than', None)
```
TheSavior: perhaps `min_empty_space`?

TheSavior: I don't think you need both config options. I'd be fine with just `min_empty_space`. If you feel like you really want to support both, we should probably throw if both are set.

nikhil-pandey:
I think its ready for merge.

TheSavior:
:+1:

jimmywarting:
Don't think you really need a config for it. I say #3135 solves this more elegant

TheSavior:
@jimmywarting I don't think these are mutually exclusive. I think #3135 looks good, and improves what items to recycle. This PR helped skip recycling unless we had too many items

Nik0lashka:
@nikhil-pandey  got a loop issue with that.

photo843403184155499

nikhil-pandey:
@Nik0lashka What does your configuration look like

Nik0lashka:
@nikhil-pandey here it is:
[config.txt](https://github.com/PokemonGoF/PokemonGo-Bot/files/408864/config.txt)

douglascamata:
@nikhil-pandey it looks to me that this log is very spammy. It fires the event every single time no items are recycled? Isn't it better to inform users that the item limit threshold was reached? This option sounds less spammy.

PR #2471, opened 2016-08-03
made file paths absolute

megamorphf:
Short Description: 
This fix will enable users to  call script anywhere from their home directory. Also this change will let you to run this script as a daemon on linux machine.  ! You will probably get an **IOError** on linux, therefore you need to change project folder permissions

> chmod -R 777 project_folder

Fixes:
- Made file paths absolute and relative to  the project folder at pokemon_bot/__init__.py file
- Made file paths absolute relative to pokecli.py

### DIFF
--- a/pokecli.py
+++ b/pokecli.py
@@ -46,9 +46,11 @@ if sys.version_info >= (2, 7, 9):
 
 def init_config():
     parser = argparse.ArgumentParser()
-    config_file = "config.json"
-    release_config_json = "release_config.json"
+
     web_dir = "web"
+    _root_dir = os.path.dirname(os.path.realpath(__file__))
+    _config_dir = os.path.join(_root_dir, "config.json")
+    _release_config_dir = os.path.join(_root_dir, "release_config.json")
 
     # If config file exists, load variables from json
     load = {}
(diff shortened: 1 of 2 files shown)


pokecli.py
```diff
@@ -46,21 +46,24 @@
 
 def init_config():
     parser = argparse.ArgumentParser()
-    config_file = "config.json"
-    release_config_json = "release_config.json"
+
     web_dir = "web"
+    _root_dir = os.path.dirname(os.path.realpath(__file__))
+    _config_dir = os.path.join(_root_dir, "config.json")
+    _release_config_dir = os.path.join(_root_dir, "release_config.json")
```
mmnsgo: dev branch doesn't use release_config.json anymore and config files are in configs/, not root

pokemongo_bot/__init__.py
```diff
@@ -21,10 +22,16 @@
 
 
 class PokemonGoBot(object):
+    _script_dir = os.path.dirname(os.path.realpath(__file__))
+    _root_dir = os.path.dirname(_script_dir)
+    _release_conf_dir = os.path.join(_root_dir, 'release_config.json')
```
mmnsgo: not used anymore

TheSavior:
I believe we made these changes as part of another PR. Thanks for contributing!

Issue #2359, opened 2016-08-02
(Request dev) - transfer and recycle threshold 


Philipp59:
now release and recycle is done by the task order, which is ok.
But it would be better to just modify the tasks, so that the release and recycle task are executed when the there is no space left 

for e.g. 
recycle task is done with config setting (keep X,Keep X, etc) only when bag is full or saver when inventory count reaches 349

same with pokemon space, release after release settings but only when bag reaches full or saver 249 

it might not sound necessary at all but in a long run it will save so much time. Because the bot should just farm and clear inventory only when its needed .

for e.g. one loop (pokestop->capture->loot->recycle->release)
```
[16:47:04] Moving towards fort 伊勢丹 Plate, 0.08km left (attraction of lure 0.16km)
[16:47:04] Arrived at pokestop.
[16:47:04]  [x] Current egg hatches in 0.22 km 
[16:47:05] Something rustles nearby!
[16:47:06]  A Wild Spearow appeared! [CP 184] [Potential 0.78] 
[16:47:06] IV [Attack/Defense/Stamina] = [13/13/9]
[16:47:09] Using Pokeball (chance: 47.34%)... (126 left!)
[16:47:10]  Captured Spearow! [CP 184] [Potential 0.78] [13/13/9] [+210 exp] 
[16:47:11] Failed to evolve Sparrow!
[16:47:16]  Now at Pokestop: 伊勢丹 Plate 
[16:47:16]  Spinning ... 
[16:47:17]  Loot:  
[16:47:17]  100 xp 
[16:47:17]  - 5x Pokeball (Total: 129) 
[16:47:17]  - 2x Potion (Total: 2) 
[16:47:17]  - 1x Revive (Total: 1) 
[16:47:17] PokeStop on cooldown. Time left: 4 minutes, 49 seconds
[16:47:20]  -- Discarded 4x Pokeball (keeps only 125 maximum)  
[16:47:21]  -- Discarded 1x Revive (keeps only 0 maximum)  
[16:47:21]  -- Discarded 2x Potion (keeps only 0 maximum)  
[16:47:22]  Releasing Spearow with CP 184 and IV 0.78. Matching release rule: CP < 9 OR IV < 0.97.  
[16:47:22]  Exchanging Spearow [CP 184] [Potential 0.78] for candy! 
```

whole loop takes about 18sec from that 18 sec about 3sec are wasted for recycle/release which is nearly a 16% speed down.
[...]

PR #2482, opened 2016-08-03
Improved item recycling

BriceSD:
Short Description: 
- Now recycling only if less than 5 space left in inventory
- Now trying to recycle before moving to/spinning fort if less than 5 space left in inventory
- Refactored recycle_items

#2359


douglascamata:
@BriceSD SpinFort task should NOT call RecycleItem task. It's the user's responsibility to put the RecycleItem task BEFORE SpinFort in their configuration. We want each task to be as independent as possible and not call other tasks.

douglascamata:
@BriceSD your refactoring to RecycleItems task is very welcome.

BriceSD:
@douglascamata You're complety right, tasks must be as independant as possible.
Items recycling optimisation issue is a subpart of tasks management issues. Bypassing the problem won't resolve it.

web
```diff
@@ -1 +1 @@
-Subproject commit 6ba5609c6151507b5b832a74e471b6b7b1a182c9
+Subproject commit dc742c598a2636337bd358dae8a558ef02159e8e
```
douglascamata: this change shouldn't be here

douglascamata:
@BriceSD please fix my comment, I want to merge this :P

BriceSD:
@douglascamata We should be good to go

douglascamata:
@BriceSD can you find a better place for the Item class? It should not be embedded in the RecycleItem task.

BriceSD:
@douglascamata I don't think the inner class Item should be in an other file.
The class Item shouldn't be use outside RecycleItem, it wouldn't be wise to allow an other task to recycle an item. In java i'd have make it private.

The purpose of this class it to make the code more readable/testable. It allows me to split the code in smaller methods and expose the logic without having 5 parameters everytime.

douglascamata:
@BriceSD should not be private because other tasks deal with items and they could use that class.

douglascamata:
@TheSavior please, review when you can.

BriceSD:
@douglascamata I think the name Item is a poor choice. Naming instances of that class item was my real goal. But i haven't found a better name so far.

The class contains all the low level details of the recycling process and doesn't really represent an item.

Instead of item.recycle() it should rather be something like Item_Recycler.recycle(item) if we would like to share the class Item.
Maybe I should make a class Item and a class Item_Recycler, but it sounds a little too overkill.

douglascamata:
@BriceSD hummm, ok ok... the it still bothers me that you're defining a class inside a class. Anyway, I wait for @TheSavior thoughts.

pokemongo_bot/cell_workers/recycle_items.py
```diff
@@ -1,65 +1,162 @@
...
+            for item_id, item_count_in_bag in item_counts_in_bag_dict.iteritems():
+                item = RecycleItems._Item(item_id, self.items_filter, self)
+
+                if item.should_be_recycled():
+                    item.request_recycle()
+                    item.emit_recycle_result()
+
+        return WorkerResult.SUCCESS
+
+    class _Item:
```
TheSavior: Excuse my python noobishness, but what is the difference between `class _Item:` and `class _Item(object):`?

TheSavior:
I'm not really sure what the need is for that inner class vs just some more functions, but *shrug*, whatever. I don't really have a problem with it. 

:+1:

pokemongo_bot/cell_workers/recycle_items.py
douglascamata: When you inherit from `object` you have new-style class. 

New style objects have a different object model to classic objects, and some things won't work properly with old style objects (those that doesn't inherit from `object`), for instance, `super()`, `@property` and `descriptors`.

BriceSD:
@douglascamata 
We should merge #2528 first. ItemRecycle should use the new inventory manager.

douglascamata:
@BriceSD definitely! You can already merge that PRs work into ours if you wish and know how to.

douglascamata:
@BriceSD yet once again, merge conflicts. Sorry :/

BriceSD:
@douglascamata Yes is was issues from the new inventory management that I merged with. Working on impementing it.

BriceSD:
@douglascamata once #3256 got fixed i can commit again. It's working and using the new inventory manager.

pokemongo_bot/inventory.py
```diff
@@ -100,7 +100,27 @@ class Items(_BaseInventoryComponent):
...
+
+    def get_space_used(self):
+        """
+        Counts the space used in item inventory.
+        :return: The space used in item inventory.
+        :rtype: int
+        """
+        itemcount = 1
+        for item in self._data:
+                itemcount += self.count_for(item)
```
douglascamata: if `self._data[item_id].get('count', False)` fails and return `False`, then this line fails. It should only add if `self.count_for(item)` is a truthy value.

pokemongo_bot/cell_workers/recycle_items.py
```diff
@@ -1,85 +1,185 @@
...
-            id_filter = self.item_filter.get(item_name, 0)
-            if id_filter is not 0:
-                id_filter_keep = id_filter.get('keep', 20)
+        """
+        Discard items if necessary.
+        :return: Always returns WorkerResult.SUCCESS.
+        :rtype: WorkerResult
+        """
+        # Updating inventory
+        inventory.init_inventory(self.bot)
```
douglascamata: Inventory is already initialized when the API is set up, you don't need this here. This will ready all JSON cache files again and can slow down the tick a lot.

pokemongo_bot/cell_workers/recycle_items.py
```diff
@@ -1,85 +1,185 @@
...
+        if self.should_run():
+            # For each user's item in inventory recycle it if needed
+            for item_in_inventory in inventory.items().all():
+                item = RecycleItems._Item(item_in_inventory['item_id'], self.items_filter, self)
+
+                if item.should_be_recycled():
+                    item.request_recycle()
+                    item.emit_recycle_result()
+                    self.update_inventory(item)
```
douglascamata: Inventory update should be automatically done in `item.request_recycle` if the result is successful. We cannot trust other devs to always remember to call `self.update_inventory(item)`, it's better to embed it in the recycle request.

pokemongo_bot/cell_workers/recycle_items.py
BriceSD: Once every task track inventory changes I’ll hapily remove the api call.

douglascamata: If everyone thinks like that, we will never reach full integration. You can use the cached inventory here, we update other tasks as necessary.

douglascamata:
@BriceSD solve merge conflicts please, this is ready to merge, imho.

BriceSD:
@douglascamata Pretty big changes. Please review them while I resolve conflicts

douglascamata:
@BriceSD what you think about moving the class that recycles the items to a service/lib folder? I think we should start to extract this type of logic like this now.

BriceSD:
@douglascamata Sounds good to me.

BriceSD:
@douglascamata Seems like ready to merge to me.

douglascamata:
👍 

Thanks a lot and sorry for tons of merge conflicts you solved.

PR #2487, opened 2016-08-03
Moved Pokemon Catch Worker

nikhil-pandey:


douglascamata:
@nikhil-pandey you moved a lot of code that belonged in the task itself. I don't believe PokemonCatch task should be removed, because it has its own configuration and we don't want this configuration to take effect in every single place that we reuse the code to catch pokemon. But indeed we need to extract that ONLY captures pokemon. Those verifications of VIPs and etc should still be there.

PR #2489, opened 2016-08-03
Added a configuration option "path_startmode"

JaapMoolenaar:
_Python is most definitely not my main programming language, however, I thought this was a nice addition to the path walker nonetheless._

Short Description: 
Updates the FollowPath walker to allow a "path_startmode" configuration option. Setting it to "first" mimics the current behavior: move to the first point in the list and walk through them from there. Setting it to the "new" value of "closest" will try to find the closest point to the bots position and first move there before continuing along the other point in the path.

I also added a log message to see which point the bot is walking to (so one can tell it's still walking)

### DIFF
--- a/CONTRIBUTORS.md
+++ b/CONTRIBUTORS.md
@@ -51,3 +51,4 @@
  * matheussampaio
  * Abraxas000
  * lucasfevi
+ * JaapMoolenaar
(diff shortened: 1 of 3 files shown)


TheSavior:
@JaapMoolenaar Wow, this is great. Sorry it slipped through our radar for so long. Would you be willing to fix up your merge conflicts? We'd love to get this in.

JaapMoolenaar:
Please see #3270

_(I couldn't seem to merge the upstream somehow, which i coincidentally haven't done much before either)_

PR #2492, opened 2016-08-03
Added better rng, its model file, and GPS noise simulation

kanemasa1987:
**"Updated"**
Short Description: 

Added somewhat better rng to replicate a human player.
**Added Config class so that the behavior is customizable by user.**

#1980, merging branch was wrong.

New Features
- GPS noise replication
- Ball throw fail, wait, precisions can be configured now

Fixes:
- Added better rngs (pokemongo_bot/human_behavior.py)
- **Added singleton Config class and get_config utility (pokemongo_bot/config.py)**
- **Changed api.set_position to replicate gps errors (pokemongo_bot/api_wrapper.py, __init__.py).**
-  **api.get_position and bot.position returns the actual position the bot is meant to be.**


douglascamata:
@kanemasa1987 we need to be able to support not using these randomisations at all.... for example I like to use only sleep for 1 or 2 seconds max whenever a sleep is needed.

kanemasa1987:
It seems to be working fine as far as I tested.
Please review the code.

pokemongo_bot/human_behaviour.py
```diff
@@ -1,46 +1,358 @@
...
+        maximum = 1.0
+        return uniform(
+            minimum + (maximum - minimum) * factor,
+            maximum)
+    elif 'human' == mode:
+        return aim_rng(factor)
+    return 1.0
+
+"""
+def _visualize():
```
Nihisil: Why there is a big piece of commented code?

kanemasa1987: _visualize was used to draw graphs to check the distribution of rng (which I showed in #1980,) to verify that there are no anomalies at the bounds etc.
Suppose you use a gaussian of mu=0.5, with big std. If you clip the min and max value between [0, 1], then you will see pillars at 0 and 1, which is not nice as a rng to replicate human behaviour.
It might be useful for future rng development, however, the CI failed due to "import seaborn" line in the function (which is never used or called in the actual app, but detected by CI,) so I decided to comment them out rather than erasing.

The other function (_precalc_lognormal_ping_param) was used to calculate parameter for the log-normal. It uses theano, and since it is somewhat heavy task, I precalculated some of them and commented out for same reason, just incase we want to calculate parameter for our desired rng.

peter-bonanni:
I've been testing this locally merged with the latest dev @ 138b66413e45efcef1ae015388b80a2725e7a7d8. Everything works with a simple manual merge of the new config options.

I'm seeing a lot of drift and reflex with both replicate_reflex and replicate_gps_noise set to true. It makes it very difficult for the path worker to follow a path when they're rubber-banding. I also noticed that with the default values and the above two flags set to true I received a soft-ban pretty regularly. I was able to fix this using the following config options:

`    "replicator": {
        "max_sleep" : 60,
        "replicate_reticle_fail_delay": true,
        "reticle_fail_chance" : 0.05,
        "reticle_fail_max_trial" : 10,
        "reticle_fail_max_sleep" : 10,
        "jitter_ping" : 80,
        "jitter_range" : 15,
        "jitter_max_sleep" : 1.0,
        "replicate_reflex" : false,
        "reflex_time" : 270,
        "reflex_range" : 100,
        "reflex_max_sleep" : 1.0,
        "replicate_gps_noise" : false,
        "ball_throw_hit_success_chance": 0.80,
        "gps_noise_radius" : 0.00032,
        "replicate_ball_throw_reticle" : "human",
        "replicate_ball_throw_spin" : "human"
    },`

peter-bonanni:
The changes related to throwing pokeballs seem to working well. I'm missing a decent amount of throws and my stats are starting to look more 'normal' however I'd really like it if there was a VIP flag that would let me do a perfect toss for pokemon in that group. 

I just missed a snorlax with 3 ultraballs and then he fled.

douglascamata:
Just saying that if you think this will protect you from bans, it won't. There are much easier ways for Niantic to detect and ban all bots user base (not just ours).

peter-bonanni:
@douglascamata yeah the stuff getting sent in the signature is pretty telling.

kanemasa1987:
@douglascamata 
That's true. What I'm trying here is at least not get banned statistically from the user's action log.
We don't know from what features they ban users in the future, but at least people with 100% hit chance are easier to ban than those does not.

@peter-bonanni 
Thankyou for the test! The human reflex is not currently used, so it might be the gps that triggered softban.  The value was set to 0.00075 according to the gps error of smartphone (somewhere I read,) but it might have been too high. I'm currently setting it at 0.00025 with PolylineWalker, seemingly not having softban issue (and I believe we need a factory class for walkers to use PolylineWalker atm..., I may work on this later.)
For the vips, there can be a change in the future.

kanemasa1987:
At this point, the issue is whether to centralize the human_behavior config (this PR,) or split each config to each task's config (current dev.) ...
Any ideas?

douglascamata:
@kanemasa1987 I think split in each task's config is better, imho.

kanemasa1987:
I will close this PR and resend for GPS part later.
The human_behavior part has overextended.

PR #2496, opened 2016-08-03
New Option: "dont_nickname_favorite"

aceradryd:
This change (line 19) adds the option, that the user can choose, whether their favorite pokemons should also get a new nickname or not.

If a user want this, then he or she has to add the line ("dont_nickname_favorite" = true) after ("nickname_template": " ... ",).


douglascamata:
👍

douglascamata:
@aceradryd please, solve merge conflicts.

@TheSavior please, review this.

pokemongo_bot/cell_workers/nickname_pokemon.py
```diff
@@ -16,7 +16,8 @@ def work(self):
         else:
             pokemon_data = self._get_inventory_pokemon(inventory)
             for pokemon in pokemon_data:
-                self._nickname_pokemon(pokemon)
+                if not(pokemon.get('favorite', 0) is 1 and self.config.get('dont_nickname_favorite','')):
```
reddivision: since `dont_nickname_favorite` is a boolean, the `get` call should default to a boolean as well (`False`)

aceradryd: Thank you, I changed it.

TheSavior:
:+1:

aceradryd:
Would it be cleaner, if I add the request, whether the user don't want to nickname his/her favourite pokemon, to "if not pokemon.get('is_egg',False):"?

douglascamata:
@aceradryd I recommend using that `False` value in the `.get` indeed.

aceradryd:
@douglascamata I changed it some hours ago, from ".get('dont_nickname_favorite',''))" to ".get('dont_nickname_favorite',False))" or do you mean something else?

douglascamata:
`if not(pokemon.get('favorite', 0) is 1 and self.config.get('dont_nickname_favorite',False)):`

This code line still has problems: unnecessary parenthesis, and a comparison between ints should use `==` and not `is`.

aceradryd:
I know, that the comparison between ints should use "==", but this part is copied from transfer_pokemon.py. This is the reason, why it's an "is" instead of a "==".

Which parenthesis shall be unnecessary? The only parenthesis, which you could mean, are necessary, because "not (A and B)" is not equal to "not A and B", but to "not A or not B". But if i code the last one, i have to use parenthesis before and after the comparison by reason of the AND-assignment.

douglascamata:
@aceradryd if you need to apply a not to the every single condition in the `if`, surely you can switch the code's logic to avoid those nots.

TheSavior:
:+1:

aceradryd:
@douglascamata yes of course, but which unnecessary parenthesis do you mean?
Is everything else with the code okay now?

douglascamata:
👍 

@aceradryd yes, but I'll fix it later. Merging for now.

Issue #2498, opened 2016-08-03
NameError: global name 'log' is not defined


sffetlio:
Sometimes I have:
[18:26:51]  Map data was not valid
[18:26:51] Moving towards fort Waterfall, 0.06km left (attraction of lure 0.35km)
[18:26:59]  [x] Current egg hatches in 3.41 km
[18:27:00]  Map data was not valid
[18:27:00] Moving towards fort Waterfall, 0.05km left (attraction of lure 0.34km)
...

and it crashed in the end:
[18:28:10]  Ran for 0:23:23
...
Traceback (most recent call last):
  File "pokecli.py", line 446, in <module>
    main()
  File "pokecli.py", line 71, in main
    bot.tick()
  File "/PokemonGo-Bot/pokemongo_bot/__init__.py", line 92, in tick
    if worker.work() == WorkerResult.RUNNING:
  File "/PokemonGo-Bot/pokemongo_bot/cell_workers/move_to_map_pokemon.py", line 161, in work
    self.update_map_location()
  File "/PokemonGo-Bot/pokemongo_bot/cell_workers/move_to_map_pokemon.py", line 107, in update_map_location
    return log.logger('Map location data was not valid', 'red')
NameError: global name 'log' is not defined

### Other Information
OS: Ubuntu
Git Commit: 2f91fd0410eeda430aeb9cb956885eda8bf04e22
Python Version: Python 2.7.6
Pokemon-Map: f632c023e905a70c775cdc3fc3e7d0b8ce95b5c1

PR #2503, opened 2016-08-03
Fixed typo logger.log

sffetlio:
Fixes: #2498


mhdasding:
Duplicate, got also fixed in #2463

PR #2513, opened 2016-08-03
Call heartbeat on step_walker even if speed is higher than distance

lucasfevi:
**Short Description:** Bot wasn't calling `bot.hearbeat()` when speed was higher than distance.

**Long Description:** I found this bug while using walk speed of 80, which causes the speed to be higher than the distance to the next fort most of the times (in places with loads of forts). Basically the position was being updated but no call to `bot.hearbeat()`, therefore not updating the location on the web, which made me look for what was happening.

Fixes:
- call `bot.hearbeat()` even if speed is higher than distance


Kassadd:
Please, merge this PR. I'm trying #2889 putting walk speed to 80, and the web doesn't track the bot at all before apply this fix.

TheSavior:
Thanks for contributing, sorry it took so long to see this!

PR #2527, opened 2016-08-03
Refactor catch worker

pokepal:
This is a refactor of the catch worker to eliminate duplicate berry throwing and ball choosing code and make it more readable/extensible (less nesting, etc)

This should fix a few things like defaulting to first available ball instead of Pokeball (I had issues with throwing -1 pokeballs) and I believe it fixes some issues like:
https://github.com/PokemonGoF/PokemonGo-Bot/issues/2505
https://github.com/PokemonGoF/PokemonGo-Bot/issues/3036
https://github.com/PokemonGoF/PokemonGo-Bot/issues/3092
https://github.com/PokemonGoF/PokemonGo-Bot/issues/3107

PR #2528, opened 2016-08-03
[Inventory Management] Add a central class for caching/parsing inventory & static data

aeckert:
Short Description: One of the biggest pains adding more complicated features is that the inventory parsing logic is spread out all over the place. This means that individual cell_workers must often parse the giant response dict themselves, which is very error prone and leads to a bunch of duplicated code. This PR adds a central class that parses every item we get back in the GET_INVENTORY request in to more usable classes so that it easy to retrieve inventory data from anywhere. This is cached at the module level and we should try to only initialize this once per tick.

Layered on top of this is the ability to load a static json file to get more data that is relevant to a certain type of object. For example, parsing pokemon.json to get next evolution, cost, etc.

The usage for this is very simple

```python
from pokemongo_bot import inventory

# Access various inventories
inventory.candies()
inventory.pokemons()
inventory.pokedex()
inventory.items()

# Some example uses
if inventory.pokedex().captured(pokemon_id):
  print('Have captuered {}'.format(inventory.Pokemons.name_for(pokemon_id))

# Access static data
inventory.Pokemons.name_for(pokemon_id)
inventory.Pokemons.first_evolution_id_for(pokemon_id)
```

I also added two small examples of how this can be used.

- Output # of candy after capturing a pokemon
- Replaced all logic for fetching/tracking Candy in EvolvePokemon

I have been running with this for a while. I also implemented a few more complex features using this class (optimizing evolutions for xp/unseen pokemon), keeping the number of pokemon you need to perform the maximum # of evolutions, but I will send them out as separate PRs since they require a little clean-up.

In the long run we should probably move ALL inventory management through this mechanism, but I didn't want to track down every place we parse the response_dict directly so I added this in a way that should be easy to move towards incrementally.

### DIFF
--- a/pokemongo_bot/cell_workers/pokemon_catch_worker.py
+++ b/pokemongo_bot/cell_workers/pokemon_catch_worker.py
@@ -1,6 +1,7 @@
 # -*- coding: utf-8 -*-
 
 import time
+from pokemongo_bot import inventory
 from pokemongo_bot.base_task import BaseTask
 from pokemongo_bot.human_behaviour import normalized_reticle_size, sleep, spin_modifier
 
@@ -27,8 +28,8 @@ LOGIC_TO_FUNCTION = {
 class Pokemon(object):
 
     def __init__(self, pokemon_list, pokemon_data):
-        self.num = int(pokemon_data['pokemon_id']) - 1
-        self.name = pokemon_list[int(self.num)]['Name']
+        self.num = int(pokemon_data['pokemon_id'])
+        self.name = pokemon_list[int(self.num) - 1]['Name']
         self.cp = pokemon_data['cp']
         self.attack = pokemon_data.get('individual_attack', 0)
         self.defense = pokemon_data.get('individual_defense', 0)
@@ -386,6 +387,19 @@ class PokemonCatchWorker(BaseTask):
                         'exp': sum(response_dict['responses']['CATCH_POKEMON']['capture_award']['xp'])
                     }
                 )
+
+                # We could refresh here too, but adding 3 saves a inventory request
+                candy = inventory.candies().get(pokemon.num)
+                candy.add(3)
+                self.emit_event(
+                    'gained_candy',
+                    formatted='You now have {quantity} {type} candy!',
+                    data = {
+                        'quantity': candy.quantity,
+                        'type': candy.type,
+                    },
+                )
+
                 self.bot.softban = False
 
                 # evolve pokemon if necessary
--- a/pokemongo_bot/inventory.py
+++ b/pokemongo_bot/inventory.py
@@ -54,6 +54,11 @@ class Candy(object):
         self.quantity = quantity
 
 
+    def add(self, amount):
+        if amount < 0:
+            raise Exception('Must add positive amount of candy')
+        self.quantity += amount
+
 class Candies(_BaseInventoryComponent):
     TYPE = 'candy'
     ID_FIELD = 'family_id'
@@ -63,10 +68,8 @@ class Candies(_BaseInventoryComponent):
         return Pokemons.first_evolution_id_for(pokemon_id)
 
     def get(self, pokemon_id):
-        try:
-            return self._data[self.family_id_for(pokemon_id)]
-        except KeyError:
-            return Candy(family_id, 0)
+        family_id = self.family_id_for(pokemon_id)
+        return self._data.setdefault(family_id, Candy(family_id, 0))
 
     def parse(self, item):
         candy = item['candy'] if 'candy' in item else 0
--- a/pokemongo_bot/inventory.py
+++ b/pokemongo_bot/inventory.py
@@ -1,0 +1,240 @@
+import json
+import os
+
+'''
+Helper class for updating/retrieving Inventory data
+'''
+
+class _BaseInventoryComponent(object):
+    TYPE = None  # base key name for items of this type
+    ID_FIELD = None  # identifier field for items of this type
+    STATIC_DATA_FILE = None  # optionally load static data from file,
+                             # dropping the data in a static variable named STATIC_DATA
+
+    def __init__(self):
+        self._data = {}
+        if self.STATIC_DATA_FILE is not None:
+            self.init_static_data()
+
+    @classmethod
+    def init_static_data(cls):
+        if not hasattr(cls, 'STATIC_DATA') or cls.STATIC_DATA is None:
+            cls.STATIC_DATA = json.load(open(cls.STATIC_DATA_FILE))
+
+    def parse(self, item):
+        # optional hook for parsing the dict for this item
+        # default is to use the dict directly
+        return item
+
+    def retrieve_data(self, inventory):
+        assert self.TYPE is not None
+        assert self.ID_FIELD is not None
+        ret = {}
+        for item in inventory:
+            data = item['inventory_item_data']
+            if self.TYPE in data:
+                item = data[self.TYPE]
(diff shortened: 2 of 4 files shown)


cprior:
Very nice, just this evening I also started to look at "software inventory".

I came here to look if I could find a place to dump a minimal example and this PR might just do the trick \*g\*
```
#!/usr/bin/env python
# -*- coding: utf-8 -*-
 
import json
from pgoapi import PGoApi

def main():
    api = PGoApi()
    api.set_position(LAT, LON, 0.0)
    api.login('ptc', 'USER', 'PASS')

    req = api.create_request()
    req.get_player()
    req.get_inventory()

    import json
    with open('data.json', 'w') as fp:
        json.dump(response, fp)

if __name__ == '__main__':
    main()
```

The resulting data.json can be read with
```jq '.responses.GET_INVENTORY.inventory_delta.inventory_items[] | .inventory_item_data | [ {pokemon: [ .pokemon_data.pokemon_id , .pokemon_data.cp]}, { pokedex: [.pokedex_entry.pokemon_id, .pokedex_entry.times_captured]}, {item: [.item.item_id, .item.count]} ]' -c data.json```

This PR is much nicer, of course! (y)
But maybe us "inventory-minded" can find a place for an inventory discussion somewhere? I am excited to see this PR because although I never do computer games I discovered this one last week explicitly because it generates "data with a story" as a basis for data analysis. So my ~50 bots are meant to produce "software inventory over time". (I do have an automated PTC account 'sniper', btw. The CSRF-token is easy to get with curl and their Django checks for the referer which took me a day to realize. The Gmail API then gives access to the confirmation link. Starting all bots with generated systemd services "pumps data" \*yay\* )

With that in place inventory processing becomes the next important item.


**potential usecases for better access to inventory data**

- visualization of the data, possibly over time
- comparison between bots
- identification of rare Pokemons in the "bot army"
- periodically dumping a bot's inventory

**data enrichment**

Some data is not present in the GET_INVENTORY structure and could be nice to persist otherwise:

- location where it was captured
- any kind of geo-data analysis is then possible

**Comment by aeckert** [comment]
@cprior  I definitely was mostly thinking about code structure when I wrote this, but I like a lot of your ideas. My measly 3 bot army does not have a need for super advanced analytics I guess lol :p. 

I think the main advantage of this model is that we should be able to extend the classes as much as we want, since they are organized more by conceptual objects than how the response dict is set up. This then allows use to add logic specific to different types in a central, re-usable way. Layering static data on top of the classes already makes them not strictly dealing with the inventory so it could be reasonable to add in some more logic (such as rarity, comparison etc). This could also be a separate class that just uses the Inventory component for the analytices

As for the accessing, I think there are two things we can do:

1) I could add a getter for the entire inventory (had that initially)
2) could also require every InventoryComponent to implement some kind of "serialize()" function that would give a json representation of the component. Then we could dump to a file for history etc.

**Comment by douglascamata** [comment]
DUDE I LOV YOU

**Comment by douglascamata** [comment]
@aeckert come to slack and let's have a chat. I have plans too add superpowers to your PR.

**Comment by cprior** [comment]
@aeckert : Yes, an equivalent of .toString() like .toJson() or similar is what I had in mind 4 hours ago. In Python this could be just a json.dump() away.

Your "cached at the module level " deserves highlighing, as I managed to get a 'ServerSideRequestThrottlingException("Request throttled by server... slow down man")' right away :)

The full inventory is already placed in the web directory so my above example is probably not the best. But this is written by  cell_workers/transfer_pokemon.py

**Comment by cprior** [comment]
Ah, and talking about "inventory data": As far as I understand there might well be additional Pokemons in the (near?) future. Just something to keep in mind -- this is all rather "NoSQL" than RDMBS, so to say.

**Comment by douglascamata** [comment]

**Comment by aeckert** [comment]
Sure fixing now.

File: pokemongo_bot/cell_workers/pokemon_catch_worker.py [hunk 1]
```diff
@@ -352,6 +354,19 @@ def work(self, response_dict=None):
...
+
+                                    # We could refresh here too, but adding 3 saves a inventory request
+                                    candy = inventory.candies().get(pokemon_id)
+                                    candy.add(3)
+                                    self.emit_event(
+                                        'gained_candy',
+                                        formatted='You now have {quantity} {type} candy!',
+                                        data = {
+                                            'quantity': candy.quantity,
+                                            'type': candy.type,
```
douglascamata: `quantity` and `type` should be added as `gained_candy` event parameters.

pokemongo_bot/inventory.py
```diff
@@ -0,0 +1,251 @@
...
+class Items(_BaseInventoryComponent):
+    TYPE = 'item'
+    ID_FIELD = 'item_id'
+    STATIC_DATA_FILE = os.path.join('data', 'items.json')
+
+    def count_for(self, item_id):
+        return self._data[item_id]['count']
+
+
+class Pokemons(_BaseInventoryComponent):
```
TheSavior: I think Pokemon is singular and plural

douglascamata:
@aeckert sorry to bother, but merge conflicts again :/

TheSavior:
Seems like the Pokemon class that has some display logic that was added here: https://github.com/PokemonGoF/PokemonGo-Bot/pull/2527/files#diff-d81926cdea81e55e6b994231eda7f95dR27

Could be consolidate with the pokemon class added in this PR.

aeckert:
@TheSavior Yea we probably should consolidate those. Will try to do that in a separate PR though as I have flaky internet the next few days and would like to get this in.

My plan was to slowly go through and use this class to clean up code in various cell workers in a separate PR.

aeckert:
@douglascamata Fixed conflicts :)

pokemongo_bot/cell_workers/pokemon_catch_worker.py
```diff
@@ -386,6 +387,19 @@ def _do_catch(self, pokemon, encounter_id, catch_rate_by_ball, is_vip=False):
                         'exp': sum(response_dict['responses']['CATCH_POKEMON']['capture_award']['xp'])
                     }
                 )
+
+                # We could refresh here too, but adding 3 saves a inventory request
+                candy = inventory.candies().get(pokemon.num)
```
TheSavior: Does this mean that there would be a separate count for pidgey candy vs pidgeotto candy (as an example)?

aeckert:
@TheSavior no internally it will remap Pidgeotto to Pidgey and maintain one count. This is also what is received in the response to a GET_INVENTORY request. The logic is to get the first evolution for the passed in pokemon number and then retrieve the number of candies for that. That logic is in the inventory.py file.

pokemongo_bot/cell_workers/pokemon_catch_worker.py
aeckert: @TheSavior no internally it will remap Pidgeotto to Pidgey and maintain one count. This is also what is received in the response to a GET_INVENTORY request. The logic is to get the first evolution for the passed in pokemon number and then retrieve the number of candies for that. That logic is in the inventory.py file.

TheSavior:
:+1:

TheSavior:
@aeckert This is great. Doing a better job of caching the inventory will help make everything better and reduce the number of unnecessary requests we need to make to the api, speeding everything up and improving the amount of xp/hr everyone can get.

Now that this is in, it should hopefully be relatively easy to add PRs and get them merged that use this in more places.

BriceSD:
@aeckert See #3256

PR #2534, opened 2016-08-03
Allow to set throw quality and spin odds

Likael:
This is a new PR that replaces #2286

**_WARNING:_**
The  code below is safe to use.
However, I tried to test further the possibilities of the field _normalized_hit_position_, and, coincidence or not, couldn't log again with my account any more for 1 hour immediately after sending a weird value. If you want to proceed and modify _normalized_hit_position_ to test randomization, be careful.

**Short Description:** 
The code allows to modify the frequency of the various throw possibilities.
For example, with the following setting : 
"catch_throw_parameters": {
	  "excellent_rate": 4,
	  "great_rate": 10,
	  "nice_rate": 4,
	  "normal_rate": 2,
	  "spin_success_rate" : 0.6,
	}
The bot would have 4 chances out of  20 (4+10+4+2) to land an _Excellent_ throw.
In the sample config, those value add up to 1 to look like percentage, but the user is free to set the values to what speaks most to him/her.
The _spin_success_rate_ is a percentage of landing a successful spin. In the example above, the bot has 60% chance of landing a spin that will generate the bonus exp. 

**More information about the _normalized_hit_position_ :** 
The code below should work safely as it doesn't contain the further tests i have tried, however my paranoid nature would ask for people to test with account they don't care about for 1 day before merging it. I will do so myself tomorrow too. Since the (unban?) I have been testing the code below, and got no further problem.
The value of 0 for _normalized_hit_position_  is safe, as it was what we were sending before I correct the API call.
The value of 1.0 is also safe, as we used it forever.
I didn't test any values between 0 et 1.
Values above 1 seems to be what could have caused the potential temporary ban.
The last case don't happen with this code, as the value is set either to 0 or 1.

**Fixes:**
- #2250

### DIFF
--- a/pokemongo_bot/cell_workers/pokemon_catch_worker.py
+++ b/pokemongo_bot/cell_workers/pokemon_catch_worker.py
@@ -1,9 +1,10 @@
 # -*- coding: utf-8 -*-
 
 import time
+from random import random
 from pokemongo_bot import inventory
 from pokemongo_bot.base_task import BaseTask
-from pokemongo_bot.human_behaviour import normalized_reticle_size, sleep, spin_modifier
+from pokemongo_bot.human_behaviour import sleep
 from pokemongo_bot.worker_result import WorkerResult
 
 CATCH_STATUS_SUCCESS = 1
@@ -69,6 +70,7 @@ class PokemonCatchWorker(BaseTask):
         # validate response
         if not response_dict:
             return WorkerResult.ERROR
+
         try:
             responses = response_dict['responses']
             response = responses[self.response_key]
@@ -301,8 +303,18 @@ class PokemonCatchWorker(BaseTask):
             if catch_rate_by_ball[current_ball] < ideal_catch_rate_before_throw and berry_count > 0 and not used_berry:
                 catch_rate_by_ball = self._use_berry(berry_id, berry_count, encounter_id, catch_rate_by_ball, current_ball)
                 berry_count -= 1
+            
+            # Randomize the quality of the throw
+            # Default structure
+            throw_parameters = {'normalized_reticle_size': 1.950,
+                                'spin_modifier': 1.0,
+                                'normalized_hit_position': 1.0,
+                                'throw_type_label': 'Excellent'}
+            self.generate_spin_parameter(throw_parameters)
+            self.generate_throw_quality_parameters(throw_parameters)
 
             # try to catch pokemon!
+            # TODO : Log which type of throw we selected
             items_stock[current_ball] -= 1
             self.emit_event(
                 'threw_pokeball',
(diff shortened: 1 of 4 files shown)


maryokhin:
According to this [Reddit post](https://www.reddit.com/r/pokemongodev/comments/4w1cvr/pokemongo_current_api_status/), something like this will potentially flag all accounts using the bot.
> That said we would like to confirm that Niantic can detect any MITM apps, these are apps that somehow modify data sent to the server. For example an app that ensures a perfect pokeballthrow. If you used an app like that Niantic could know.

maryokhin:
ah, ok, I thought it was the other way around. IMO the bot should just have a default randomness on how it throws the ball and have a setting to "enable perfect throws". After all, one of the goals of a bot is to emulate a human.

Likael:
Well, the problem with default randomness is that it is only pseudo-random. Without inputs, you take the risk to see all bots behaving the same on the long term. By allowing people to modify inputs, you generate a greater variety of behaviours, which makes each bot more unique. Of course some combinations could and/or would trigger flag (i.e. 100% perfect throw, or any behaviours that move too far away from the standard deviation), but it is not ours to judge. The exercise here is to challenge ourselves to design a bot that is able to get as close as possible of a human behaviour.

douglascamata:
Need to solve merge conflicts.

RedViper9:
@Likael Would it be possible to also add a "miss" option for the sake of more human behavior?

Likael:
@douglascamata Done.
@RedViper9 Sure, but I first need to have a check at the missed throw requests to ensure the data we will be sending are coherent with the app one.

pokemongo_bot/cell_workers/pokemon_catch_worker.py
```diff
@@ -432,3 +441,50 @@ def _do_evolve(self, pokemon, new_pokemon_id):
                 formatted='Failed to evolve {pokemon}!',
                 data={'pokemon': pokemon.name}
             )
+
+    def generate_random_throw_parameters_if_necessary(self, throw_parameters):
```
douglascamata: can you please break down this method into smaller ones?

douglascamata:
@Likael see my comment and fix merge conflicts again, please? :D

Likael:
@douglascamata Partly done. I merged, but couldn't really break down the code. I have extracted the spin function. The rest is as it is, just the randomization part. I unfolded it for clarity and easy debug, but it's possible de condensate it using a for loop and creating the adequate data structure in the class. However it will lose the clarity. I don't really have time to do that today, but if that's what you want, I can have a look this weekend.

douglascamata:
@Likael add an example configuration to teach us how to configure these parameters, please.

Likael:
@douglascamata Sorry the example has been trimmed away in the previous commit. Added it again.

douglascamata:
:+1: 

[Approved with PullApprove](https://pullapprove.com/PokemonGoF/PokemonGo-Bot/pull-request/2534/?utm_source=github-pr&utm_medium=comment-badge&utm_campaign=PokemonGoF/PokemonGo-Bot)

patrickbrouwers:
The example config is invalid: Strict JSON does not allow a final comma in an object (dictionary) literal (https://github.com/PokemonGoF/PokemonGo-Bot/pull/2534/commits/ec82e9475b6f4259abcdc6cd33bb7764c7788108#diff-12669c2a3cbb1440601e0581913cff68R109)

douglascamata:
@patrickbrouwers thx, already pushed a fix for this a few minutes ago.

PR #2561, opened 2016-08-04
Dev

codytrey:
Short Description: Added a player journal file feature that writes to a file similar to the "journal" in the Pokemon GO app


TheSavior:
This PR looks like it has a bad merge. Too many changed files.

PR #2600, opened 2016-08-04
Fixed GeocoderQuotaExceeded exception handling.

mjmadsen:
Changed how we handle the GeocoderQuotaExceeded exception. 

Are we intending remove all logger.log calls?


mjmadsen:
#2599 is better!

PR #2601, opened 2016-08-04
typo: update bot_event.py

gokaygurcan:
Fix a typo. 

I assume that it was "information" initially, but became "logrmation" when someone used replace all functionality to replace all infos with logs. But I might be totally wrong at this point, idk. Just didn't like the word and wanted to fix that typo.


TheSavior:
Thanks!

gokaygurcan:
No worries. Thanks for accepting it.

PR #2605, opened 2016-08-04
Added logic to allow users to release pokemon if they do not have the…

GreenLink5:
Short Description: 

I have created logic in transfer_pokemon.py that will allow users to release pokemon based on movesets and also added moves.json to data\

Fixes:
- 
- 
- 

… desired movesets


sffetlio:
Can we have some sort of moves priority so it can keep the pokemon with better moveset and not only the one with the best/perfect moves?

Nihisil:
Can you fix these things:

1. Merge latest dev to your PR and fix all conflicts
2. Use new Pokemon object, it contains all information about moves. So, you don't need to include it to the project.
3. Add ability to config it with logic like "don't release pokemons with move_1: 111,123" or so.
4. Add examples into cofig.example

PR #2656, opened 2016-08-07
Corrected "Spun" to "Spinning"

raulgbcr:
Short Description: 

Fixes:
- Corrected "Spun" to "Spinning" syntax.


solderzzc:
Thanks for the PR.

TheSavior:
Why is this valid? We already have spun it. So it should be past tense

PR #2662, opened 2016-08-07
fix for softban worker

reddivision:
Short Description: 
Softban worker fix
Fixes:
- softban worker failing


solderzzc:
Merge for testing ...

PR #2679, opened 2016-08-07
Adding plugin support

TheSavior:
This will support a plugins configuration item that contains paths to directories containing "plugins". A plugin is a python module that contains an `__init__.py`. Then you can specify `package_name.MyTask` in this format in the tasks object:

```
{
    // ...
    "tasks": [
      {
        "type": "package_name.MyTask"
      }
    ],
    // ...
}
```


TheSavior:
It won't actually be possible to create plugins until we move BaseTask out of the main repo, or come up with some other way to handle extending a core piece.

solderzzc:
👍

TheSavior:
@douglascamata Please review this even though it has been merged.

PR #2692, opened 2016-08-07
Throw the exact error lines

pmquan:
Short Description: 
The pokecli.py will throw the error lines information.


TheSavior:
I believe this was fixed in a different PR.

PR #2744, opened 2016-08-07
Use current working directory

brantje:
Short Description: 
Even with the encrypt.so is placed in the base directory, it gives the following error:
`'OSError: encrypt.so: cannot open shared object file: No such file or directory`
This pr gets the current working directory and uses that with the encrypt.so

### DIFF
--- a/pokemongo_bot/__init__.py
+++ b/pokemongo_bot/__init__.py
@@ -602,8 +602,10 @@ class PokemonGoBot(object):
         # chain subrequests (methods) into one RPC call
 
         self._print_character_info()
+        # self.api.activate_signature("/home/pi/PokemonGo-Bot/encrypt.so")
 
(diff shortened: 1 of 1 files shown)


brantje:
Closing, will improve this to load the dll on windows

pokemongo_bot/__init__.py
```diff
@@ -602,8 +602,10 @@ def _setup_api(self):
         # chain subrequests (methods) into one RPC call
 
         self._print_character_info()
+        # self.api.activate_signature("/home/pi/PokemonGo-Bot/encrypt.so")
 
-        self.api.activate_signature("encrypt.so")
+        cwd = os.getcwd()
```
TheSavior: I don't think you want cwd, I think you want `os.path.join` with `os.path.dirname(__file__)`

Issue #2698, opened 2016-08-07
Possibly Searching Too Often - keeps changing fort it walks to without catching Pokemon


jryd:
### Expected Behavior

Bot locates a fort, walks to it and collects loot along with any nearby pokemon.

### Actual Behavior

Bot begins walking to fort and then emits `Possibly searching too often, take a rest.` and then picks a new fort to walk to.

It then does this over and over again, see dump below:

```
2016-08-07 07:34:16,016 [  SpinFort] [INFO] [pokestop_searching_too_often] Possibly searching too often, take a rest.
2016-08-07 07:34:16,244 [MoveToFort] [INFO] [moving_to_lured_fort] Moving towards pokestop Per Capita  - 0.12km (attraction of lure 0.12km)
2016-08-07 07:34:18,205 [MoveToFort] [INFO] [moving_to_lured_fort] Moving towards pokestop Per Capita  - 0.10km (attraction of lure 0.11km)
2016-08-07 07:34:20,237 [MoveToFort] [INFO] [moving_to_lured_fort] Moving towards pokestop Per Capita  - 0.09km (attraction of lure 0.11km)
2016-08-07 07:34:22,274 [MoveToFort] [INFO] [moving_to_lured_fort] Moving towards pokestop Per Capita  - 0.08km (attraction of lure 0.10km)
2016-08-07 07:34:25,080 [MoveToFort] [INFO] [moving_to_lured_fort] Moving towards pokestop Running Pinhead - 0.10km (attraction of lure 0.10km)
2016-08-07 07:34:27,285 [MoveToFort] [INFO] [moving_to_lured_fort] Moving towards pokestop Running Pinhead - 0.08km (attraction of lure 0.08km)
2016-08-07 07:34:29,634 [MoveToFort] [INFO] [moving_to_lured_fort] Moving towards pokestop Running Pinhead - 0.07km (attraction of lure 0.07km)
2016-08-07 07:34:32,235 [MoveToFort] [INFO] [moving_to_lured_fort] Moving towards pokestop Running Pinhead - 0.06km (attraction of lure 0.06km)
2016-08-07 07:34:34,477 [CatchLuredPokemon] [INFO] [lured_pokemon_found] Lured pokemon at fort Running Pinhead (3b716d7483d14ccf8eab6757d0b04a83.16)
[...]

PR #2760, opened 2016-08-07
Fixes "Possibly Searching Too Often" after stale session re-login

geek-man:
Short Description: 

* Added call to `api.activate_signature` to prevent this issue after a re-login.
* Also fixed broken deprecated logger call and replaced it with an event_manager emit.

Fixes:
- #2698 
- 
-


brantje:
We don't accept pull requests to master. If you could point this to dev we'd be happy to review it.

PR #2764, opened 2016-08-07
When the google analytics domain is blocked the bot crashed.

brantje:
Short Description: 
When the google analytics domain is blocked the bot crashed. 
With a simple try / except this can be solved.


BobbyWibowo:
To begin with, does the google analytics necessary for it to run?

douglascamata:
@BobbyWibowo it's optional and allow us to count how many active bot users we got. No personal information is collected.

BobbyWibowo:
@douglascamata I see. Sorry, I didn't noticed the bit about Analytics in the README file.

PR #2768, opened 2016-08-07
OS Detection for encrypt lib

brantje:
Short Description: 
Including encrypt.so will not work on windows. 
This PR looks up the user OS and loads the according dll / so file.

### DIFF
--- a/pokemongo_bot/__init__.py
+++ b/pokemongo_bot/__init__.py
@@ -28,8 +28,8 @@ from pokemongo_bot.socketio_server.runner import SocketIoRunner
 from pokemongo_bot.websocket_remote_control import WebsocketRemoteControl
 from worker_result import WorkerResult
 from tree_config_builder import ConfigException, MismatchTaskApiVersion, TreeConfigBuilder
-
-
+from sys import platform as _platform
+import struct
 class PokemonGoBot(object):
     @property
     def position(self):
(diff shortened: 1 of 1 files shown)


pokemongo_bot/__init__.py
```diff
@@ -603,7 +603,25 @@ def _setup_api(self):
...
+        print os.path.dirname(__file__)
+
+
+        if _platform == "linux" or _platform == "linux2":
+            file_name = 'encrypt.so'
+        elif _platform == "darwin":
+            file_name = 'encrypt.so'
+        elif _platform == "win32":
+            # Check if we are on 32 or 64 bit
+            if(  8 * struct.calcsize("P") == 32):
```
TheSavior: It seems like the internet likes this check more these days: 

```
sys.maxsize > 2**32
```

pokemongo_bot/__init__.py
```diff
@@ -603,7 +603,25 @@ def _setup_api(self):
...
 
-        self.api.activate_signature("encrypt.so")
+
+
+        print os.path.dirname(__file__)
+
+
+        if _platform == "linux" or _platform == "linux2":
+            file_name = 'encrypt.so'
+        elif _platform == "darwin":
```
MikeDX: since darwin and linux both use the same file_name, is it necessary to have an additional check? Wouldn't it be better to default to enscrypt.so and then only change file_name if windows were detected.

brantje:
Can someone test this for windows 32 / 64 bit?

douglascamata:
Can you use `os.path.abspath` for absolute path instead of relatives?

mvolli:
working on win7 x64 with python x64

BobbyWibowo:
If I use python 32 bit on a win 64 machine with 32 bit encrypt.dll since I use python 32 bit, it's failing

BobbyWibowo:
@brantje How does the platform checking work? I'm on Windows 10 64 bit with 64 bit version of Python (I re-installed a 64-bit Python, if you've seen my previous comment), but the message in the bot said this:
`[PokemonGoBot] [INFO] Found encrypt_64.dll! Platform: win32 Bot root directory: fakepath\PokemonGo-Bot`

PR #2819, opened 2016-08-07
New default encoding + encrypt.so call

stylersnico:
Short Description: 

Fixes:
- sys module default encoding to UTF8
- Changed the way we call encrypt.so


brantje:
I think #2768 covered this too.

PR #2604, opened 2016-08-04
Colorized logging

gurupras:
Short Description: Updates to the (updated) logging system

Fixes:
- Adds lost colour to the event system
- Adds new methods to loggers

Some of the colours may not be the same as they were in the earlier `logger.log` system. I can change these if you're very particular. Some of the LCD code has also crept in. I can get rid of these if they're no longer required.

The commit in itself may have some redundant code in terms of functions that can be used like:
`self.logger.red('message in red')`, `...`
I thought these might be useful for cases where a message doesn't necessarily fit into the usual logging levels like `CRITICAL`, `ERROR`, `WARNING`, `INFO`, `DEBUG`

PR #2914, opened 2016-08-07
Add colored logs

fyooree:
Adds colored logs back.
Can be reverted when a "real" solution is found.

+ Colored logs
+ Colors configurable through config

Supported colors:
black, red, green, yellow, blue, magenta, cyan, white

### DIFF
--- a/configs/config.json.example
+++ b/configs/config.json.example
@@ -127,5 +127,12 @@
         "Weezing": {},
         "Flareon": {}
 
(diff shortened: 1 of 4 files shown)


douglascamata:
@fyooree I'm sorry, but there's already a PR for this at https://github.com/PokemonGoF/PokemonGo-Bot/pull/2604.

PR #2944, opened 2016-08-07
correctly re-raise exception to keep backtrace

rbignon:
When it crashes, we lose the original backtrace on stderr.


brantje:
:+1:  looks good.

PR #2948, opened 2016-08-07
Add cygwin to supported platform and improved log readability

villish:
Short Description: 
added cygwin as a supported platform (win32) and increased readability of log output
Fixes:
cygwin support
correct error message when encrypt.dll not found while using cygwin


solderzzc:
👍

PR #2956, opened 2016-08-08
First basic features of the pokemon optimizer

Anakin5:
Short Description: Pokemon optimizer, transfer and evolve
- Transfer and evolve pokemons based on its rank in his family according to configurable sorting criteria.
- Transfer and evolve to maximize xp = Keep enough basic pokemon to use your candies for evolution

TODO:
- Allow deactivation of xp optimization for max level players
- Support dps and moves in sorting criteria


abusalman:
+1 for this.

pokemongo_bot/__init__.py
```diff
@@ -265,7 +265,7 @@ def _register_events(self):
         )
         self.event_manager.register_event(
             'pokemon_evolved',
-            parameters=('pokemon', 'iv', 'cp')
+            parameters=('pokemon', 'iv', 'cp', 'xp')
```
douglascamata: @Anakin5 you added a required parameter (no optional parameter logic available yet) to `pokemon_evolved` event here, but the EvolvePokemon task doesn't send it, so an error will happen. You need to make EvolvePokemon send it.

douglascamata:
Quick question, why not improve the existing ReleasePokemon and EvolvePokemon tasks? This is awesome features, but we would have now a duplicate worker with two functions that already exist separately: transfer and evolve pokemon.

Anakin5:
If I separate evolution and transfer, I will have to duplicate the logic since the selection of who must be evolved and who must be transfered are linked together.

I will change the EvolvePokemon task to still be compatible.

I also need to fix a few things on the current code. So please, use dry run and let me know if you see something strange in the choice of Pokemon to transfer/evolve.
- For evolution, we must only keep Pokemon's with the lowest candy required for evolve.
- Add one config parameter to keep at least the beat iv Pokémon of each name instead of only the best iv of the family.
- Evolve only when we have lucky egg and enough evolution to cover the whole 5 minutes.

solderzzc:
👍 
@TheSavior  how do you think about this PR ? I'm starting to test locally.

solderzzc:
I think there need a config.example, right?

TheSavior:
I am not excited about having a new task that has duplicated logic from two of our other tasks. To follow on to @douglascamata's comment, I'm also not sure why we can't improve the tasks we have, or refactor the logic we currently have in our tasks to be able to reuse some of it in a task like this.

As this PR stands, I'd rather have this be as an external plugin than part of the core if it needs duplicated logic.

solderzzc:
This is good for XP. @Anakin5 Merge logic with core task will be good system.

If our module separated into three kinds
Event -> Decision  ->  Action 
The core will be much flexible.

screen shot 2016-08-07 at 10 04 44 pm

pokemongo_bot/cell_workers/__init__.py
```diff
@@ -18,3 +18,4 @@
 from follow_cluster import FollowCluster
 from sleep_schedule import SleepSchedule
 from update_title_stats import UpdateTitleStats
+from pokemongo_bot.cell_workers.pokemon_optimizer import PokemonOptimizer
```
solderzzc: -from pokemongo_bot.cell_workers.pokemon_optimizer import PokemonOptimizer
+from pokemon_optimizer import PokemonOptimizer

Anakin5:
There is no configuration parameters for now, but I am going to add a few.
Also I just realized I have to commit some fixes in the Pokemon.json data file.

BartKoppelmans:
I think this should me merged with the other workers, you could set configuration via a different setting, for example:
```
"optimizer" : {
    "setting 1" : True
}
```
Not everyone wants it on, but it could be that people want it partially on. To get this done, we will have to allow different settings for this. Still, the workers do the same thing and the logic should be kept in one place, imho.

There is another PR (#2482) which has a similar effect for the bag, perhaps these two could have merged settings to have them in one place as an optimizer?

pokemongo_bot/__init__.py
Anakin5: fixed

douglascamata:
> @Anakin5 said: If I separate evolution and transfer, I will have to duplicate the logic since the selection of who must be evolved and who must be transfered are linked together.

Ok ok, then we need to find a good way to split this logic. Maybe we can create a class for filtering pokemon lists, for example. Duplicating code because otherwise I would duplicate other code is not a good excuse :P

Anakin5:
I am fine with removing the EvolveAll and TransferPokemon tasks if you think all cannot live together. But there is no point in having tasks configurable if there is only one valid set.

Please explain why you want those 2 tasks to be split. The fact that it is like that now is not a good reason :P

TheSavior:
@Anakin5 One reason it should be split is that some people want one feature but not the other. Also, they might want to put them in different places in the task list. Merging them into one won't let them do that.

Anakin5:
Algo for Eevee family coming later. This one a bit tricky.

TheSavior:
@Anakin5 This PR is currently waiting for changes. We don't want to duplicate logic that is shared between this and the other workers we currently have. If you need shared logic, it might be best to pull that out into another class that can be reused from something like this. 

What might be valuable is to reduce the scope if this task. Perhaps it only replaces the existing TransferPokemon task to smartly transfer based on the number of necessary candy to evolve the ones it is keeping. Then people can use the existing Evolve task with a lucky egg.

TheSavior:
@Anakin5 How do you feel about my last comment? I want to make sure we help you move this in a direction that will actually allow it to get merged.

MicaelCarvalho:
How to use this feature ?
> Keep best iv / cp / relative cp / across whole family
Couldn't find the correct parameter in the committed changes, nor on the config file.

pokemongo_bot/cell_workers/pokemon_catch_worker.py
```diff
@@ -405,8 +405,7 @@ def _do_catch(self, pokemon, encounter_id, catch_rate_by_ball, is_vip=False):
                 )
 
                 # We could refresh here too, but adding 3 saves a inventory request
-                candy = inventory.candies().get(pokemon.num)
-                candy.add(3)
+                candy = inventory.candies(True).get(pokemon.num)
```
douglascamata: Refreshing is a bad behaviour. It would be great if we could make `PokemonCatchWorker` update the candy account using the inventory variable when pokemon is captured is avoid this refresh here.

pokemongo_bot/event_handlers/colored_logging_handler.py
```diff
@@ -1,172 +0,0 @@
-# -*- coding: utf-8 -*-
```
douglascamata: why was this file removed?

douglascamata:
Small tip: keep the methods ordered by call time. For example, if the first thing method `x` does is calling `y`, then `y` should be defined right below `x`. This makes a lot easier to read code in big files, otherwise you need a lot of scrolling all the time.

douglascamata:
And you got merge conflicts again, sorry :/

pokemongo_bot/cell_workers/pokemon_catch_worker.py
Anakin5: The refresh is not for the candy, but for the Pokemon. If this refresh is not done, there is nobody in the loop accounting for the -1 space in the Pokemon bag. It leads to inventory full.
I agree that we should be able to do that without a refresh but we have to be careful since it would mean desynchronizing the space left in the bag with the actually list of Pokemons return by inventory.pokemons.
I propose we create a counter in Pokemons inventory class, that we can increase/decrease just like candies.
But if someone call inventory.pokemons() and the size of that list is different from the counter, then we should immediately refresh.

pokemongo_bot/event_handlers/colored_logging_handler.py
Anakin5: Probably a mistake

pokemongo_bot/cell_workers/pokemon_catch_worker.py
douglascamata: So we need to modify the pokemon list itself then, not just a counter. The list is cached, so we remove an element from it.

Anakin5: It would be adding a Pokemon. Do we have all the info required to fake insert the new Pokemon ?

douglascamata: Not sure, but I think we do.

douglascamata:
Aaaand merge conflicts one more time. I'm so sorry that I have no time to merge your PR in time and you always get conflicts 😢

TheSavior:
I want to get this in so you stop having merge conflicts, and then we can improve it as need be separately. It seems pretty self contained. :+1:

nanogirl21:
`"sort": ["ncp"]`  what is ncp?

1cu:
the optimizer is only run at startup, never during run time:
2016-08-12 10:26:06,782 [PokemonCatchWorker] [INFO] [pokemon_inventory_full] Your Pokemon inventory is full! Could not catch!

After a restart it immediately evolves and transfers all pkmn.

Here is my task config:

    "tasks": [
      {
        "type": "HandleSoftBan"
      },
      {
        "type": "CollectLevelUpReward"
      },
      {
        "type": "IncubateEggs",
        "config": {
          "longer_eggs_first": true
        }
      },
      {
          "type": "PokemonOptimizer",
          "config": {
            "enable": true,
            "dry_run": false,
            "evolve_only_with_lucky_egg": false
          }
      },
      {
        "type": "RecycleItems",
        "config": {
          "min_empty_space": 15,
          "item_filter": {
            "Pokeball":       { "keep" : 15 },
            "Greatball":      { "keep" : 40 },
            "Ultraball":      { "keep" : 50 },
            "Potion":         { "keep" : 0 },
            "Super Potion":   { "keep" : 10 },
            "Hyper Potion":   { "keep" : 30 },
            "Max Potion":     { "keep" : 50 },
            "Revive":         { "keep" : 20 },
            "Razz Berry":     { "keep" : 50 }
          }
        }
      },
      {
        "type": "CatchVisiblePokemon"
      },
      {
        "type": "CatchLuredPokemon"
      },
      {
        "type": "SpinFort"
      },
      {
        "type": "UpdateTitleStats",
        "config": {
            "enable": true,
            "min_interval": 10,
            "terminal_log": false,
            "terminal_title": true,
            "// stats": ["uptime", "xp_per_hour", "stardust_earned", "captures_per_hour", "level_stats" ],
            "stats": ["uptime", "xp_per_hour", "stardust_earned", "level_stats" ]
        }
      },
      {
        "type": "MoveToFort",
        "config": {
          "lure_attraction": true,
          "lure_max_distance": 2000
        }
      },
      {
        "type": "FollowSpiral",
        "config": {
          "enable": false,
          "diameter": 4,
          "step_size": 70
        }
      }
    ],

balpoint:
I really like this idea, but what do the following things mean:

"top": 1 ?? 
ncp ?

i would use it like this if its possible:

>            "type": "PokemonOptimizer",
            "config": {
                "transfer": true,
                "evolve": true,
                "use_lucky_egg": true,
                "evolve_only_with_lucky_egg": true,
                "minimum_evolve_for_lucky_egg": 20,
                "keep": [
                    {
                        "weedle": 45,
                        "evolve": true,
                        "sort": ["iv"]
                    },
                    {
                        "pidgey": 45,
                        "evolve": true,
                        "sort": ["iv"]
                    },
                    {
>//below to keep every other pokemon than defined above based on IV CP, like the old release config did >//like this: "Slowbro": { "keep_best_cp": 1, "keep_best_iv": 1 },

>                       "top": 2,
                        "evolve": false,
                        "sort": ["ivcp"]
                    }
                ]
            }
        },

Anakin5:
ncp stands for normalized cp. It is called cp_percent in the code actually. You can use cp_percent as well if it is more meaningful to you.

If task not running during live, it means you catch pokemon task is not updating the Pokemon inventory.

Anakin5:
I will put some comment in the config to explain stuff a bit.

balpoint:
When can we expect it @Anakin5 really excited!

nanogirl21:
@Anakin5 Adding tips to explain in the config would help a lot. I have been waiting for a feature like this. 

Just a few questions. 

1. If the optimize code ONLY works at start up to evolve shouldn't there be some type of release config that runs all the time (how the code is now)? This will release lower CP or IV pokemon and not allow your pokebag to get complexly full with 200 pidgeys. If not, is there a way to automatically restart the bot when your pokebag get’s full to release and evolve to make room? I personally leave my bot running overnight and am not always there to manually restart when the bag is full. 

2. Is there any way to use candy to max out CP (or until you run out of candy) before evolving to next level?

Anakin5:
@nanogirl21 
I added comments in the config in the next PR: https://github.com/PokemonGoF/PokemonGo-Bot/pull/3743
For your issue 1., it is not a normal behavior. This should not happen with the little change I made to the pokemon catch worker. Make sure you have that change and let us know if you still have issues.
2. No, since there is no point to use candies before evolving rather than after. We could however add a pamater that allow you to use your candies to upgrade your best pokemons.

nanogirl21:
@Anakin5  Thank you for your help. I will try to run later when I am back home. I do see that you added parameter "use_candies_for_xp". Thank you.

PR #2964, opened 2016-08-08
Config/encrypt.so

heihachi:
Short Description: 
Set a folder where your encrypt.so/dll are located.

Also a sidenote this only points to the folder containing encrypt.so,encrypt.dll etc not to the actual file.

### DIFF
--- a/configs/config.json.example
+++ b/configs/config.json.example
@@ -4,6 +4,7 @@
     "password": "YOUR_PASSWORD",
     "location": "SOME_LOCATION",
     "gmapkey": "GOOGLE_MAPS_API_KEY",
+    "libencrypt_location": "",
     "tasks": [
       {
         "type": "HandleSoftBan"
--- a/pokecli.py
+++ b/pokecli.py
@@ -53,7 +53,7 @@ logger.setLevel(logging.INFO)
 
 def main():
     bot = False
-    
+
     try:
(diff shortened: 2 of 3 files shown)


configs/config.json.example
```diff
@@ -4,6 +4,7 @@
     "password": "YOUR_PASSWORD",
     "location": "SOME_LOCATION",
     "gmapkey": "GOOGLE_MAPS_API_KEY",
+    "libencrypt_location": "",
```
solderzzc: Is there a hard code location inside the PGOAPI package?

heihachi: Nope its set as encrypt="encrypt.so" or similar so having it not be set isnt a problem. If its "" it will check in the bots folder like usual

solderzzc:
👍

Meph88:
I'm a little confused....

I'm getting this output:
`[ERROR] encrypt.so is not found! Please place it in the bots root directory or set libencrypt_location in config.`
`[INFO] Platform: darwin Encrypt.so directory: /Users/USERNAME/Documents/POBOT/dev2/PokemonGo-Bot`

Encrypt.so is clearly a required file now but that's all I know and I can't find the answer.
I've tried creating an empty file named this in the root directory of the bot but that creates a huge amount of errors.
I don't know what to do here and help is hugely appreciated.

PR #2976, opened 2016-08-08
Terminal flush

mhdasding:
Short Description: 

some terminals buffer writes of stdout so UpdateTitleStats does not always update like it should

Fixes:
- flush stdout after stats update


mhdasding:
wrong commits

PR #3072, opened 2016-08-08
Fix for FileIO in Heartbeat. Threaded out the FileIO

ajurna:
Short Description: FileIO is causing delays on the heartbeat. this is fixed by threading this out.

Fixes:
Delay caused by writing files.


ajurna:
crap this is just wrong.

Issue #3117, opened 2016-08-08
[Confusing words] Fled vs Escape


DeviousFusion:
Pull #3118 

### Expected Behavior

[pokemon_escaped] Eevee escaped.

### Actual Behavior

[pokemon_fled] Eevee fled.

### Steps to Reproduce

Run the bot as usual. Whenever a pokemon escapes from pokeball, the word usage is confusing

### Other Information
OS: 
Git Commit: ff380cd0e07f225ad1ae2c739bea9f9f80e2f8ce
Python Version: Python 2.7.12

Marco208:
This is wrong. You can't just rename the event, please revert.

Marco208:
pokemongo_bot.event_manager.EventNotRegisteredException: Event pokemon_escaped not registered...

DeviousFusion:
@Marco208 : Fixing. I forgot to push __init__.py

Marco208:
Thanks for your contributions :)

RedViper9:
Thanks for noting this! I had thought the bot was just forcing the api to catch fled pokemon!

`2016-08-08 12:23:46,353 [PokemonCatchWorker] [INFO] [threw_pokeball] Used Greatball, with chance 46.80 (52 left)

2016-08-08 12:23:47,738 [PokemonCatchWorker] [INFO] [pokemon_fled] Psyduck fled.

2016-08-08 12:23:49,401 [PokemonCatchWorker] [INFO] [threw_pokeball] Used Greatball, with chance 46.80 (51 left)

2016-08-08 12:23:50,754 [PokemonCatchWorker] [INFO] [pokemon_caught] Captured Psyduck! [CP 495] [Potential 0.18] [2/0/6] [+150 exp]
`

PR #3118, opened 2016-08-08
Change word usage: "fled" to "escaped"

DeviousFusion:
Short Description: 

Fixes:
- 
- 
- #3117 

"fled" is confusing to lot of people and is easily confused with pokemon vanishing. "escaped" is a better term.


TheSavior:
:+1:

PR #3132, opened 2016-08-08
Create config.json

GitPupper:
### Description:
This mainly fixes the log spam that says it's keeping the best of 3, but never recycles. So, I added in some code, and bam, fixed it! It recycles everything *but* two of each by default.

***

### Fixes:
- Recycling

***

### New Config:
http://pastebin.com/UStttHL2

### DIFF
--- a/configs/config.json
+++ b/configs/config.json
@@ -1,0 +1,125 @@
+{
+    "auth_service": "google",
+    "username": "email@gmail.com",
+    "password": "password",
+    "location": "34.0087376, -118.5004127",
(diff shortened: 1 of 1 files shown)


leadboots5:
You should improve the existing config examples if you think you have something to add. not create a new config file.

GitPupper:
@leadboots5 Sorry, I'm new to PRs. I knew after I submitted it I did it wrong so I just uploaded the pastbin. I was hoping that the devs could just integrate the `"release"` portion since that's what I was trying to fix in the first place.

PR #3135, opened 2016-08-08
keep improving inventory by recycling based on type filter;

tiagoAnon:
Short Description: 

Enables inventory improving by setting a configuration for how much of each item type to keep and recycling the worst items.

Fixes:
- Added type_filter config
- Item filtering logic for berries, balls, potions and revives if type_filter is configured, overriding item_filter logic
- Added sample json

Notes:
I haven't worked with Python for 5+ years, but I hope it's up to your standards ;)


jimmywarting:
#2482 #2465 #2465 #2361

cmezh:
It's better not to silently override item_filter, but throw exception. Why to keep config parts that doesn't work (overridden)?

configs/config.json.type_filter.example
```diff
@@ -0,0 +1,129 @@
...
+            "evolve_above_iv": 0.8,
+            "logic": "or",
+            "evolve_speed": 20,
+            "use_lucky_egg": false
+        }
+      },
+      {
+        "type": "RecycleItems",
+        "config": {
+          "type_filter": {
```
TheSavior: Perhaps this doesn't need its own config. You could add it to the other one and change this key to

```
  "// type_filter"
```

TheSavior:
This looks great. It seems like you have merge conflicts, can you fix those?

@douglascamata Please review when you get the chance

douglascamata:
@tiagoAnon can you solve merge conflicts, please?

tiagoAnon:
I did one last change, because the feature was causing a lot of logging noise due to some lack of sync of the inventory somewhere on the bot. Changed it to keep some unallocated space empty (based on the configuration). It seems to merge OK with Dev.

douglascamata:
@tiagoAnon please, split your logic in many smaller methods. The way it is now is very complicated to read and understand.

jimmywarting:
I want type_filter

TheSavior:
@BriceSD, is there something we can do here to utilize inventory calls instead of all these checks on the response object?

BriceSD:
@tiagoAnon This feature seems awesome, please update it to use the latest items_recycle. I really want to use it :D

@TheSavior Yes we should use the new inventory here. We should also add methods in the Items class to access to the lowest ball/potion/revive/etc. available, to get every member of the type ball/berry/potion/revive/etc. These could be reuse somewhere else, like in the pokemon_catcher to choose the type of ball that should be thrown.

jimmywarting:
Maybe should make a completely new PR? it have been lot of changes to the recycle script

BriceSD:
Merged #4005. We have this feature now.

PR #3137, opened 2016-08-08
When JSON parsing fails, give a rough indication of why

cwild:
Try to report back any problems with the provided JSON config file to the user.

Here is an example of the output:

```
user@machine:~/git/PokemonGo-Bot$ python pokecli.py -cf configs/duff.json
2016-08-08 20:25:01,880 [       cli] [INFO] PokemonGO Bot v1.0
configs/duff.json:2:20: Error: Can not decode value starting with character u'\u201c'
   |  At line 2, column 20, offset 22
configs/duff.json:2:25: Info: Recovering parsing after character u','
   |  At line 2, column 25, offset 27
configs/duff.json: has errors
2016-08-08 20:25:01,946 [       cli] [CRITICAL] Error with configuration file
```
Made possible thanks to the demjson module.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -21,3 +21,4 @@ gpxpy==1.1.1
 mock==2.0.0
 timeout-decorator==0.3.2
 raven==5.23.0
+-e git+https://github.com/dmeranda/demjson.git@5bc65974e7141746acc88c581f5d2dfb8ea14064#egg=demjson
--- a/requirements.txt
+++ b/requirements.txt
@@ -21,4 +21,4 @@ gpxpy==1.1.1
 mock==2.0.0
 timeout-decorator==0.3.2
 raven==5.23.0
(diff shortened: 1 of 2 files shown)


requirements.txt
```diff
@@ -21,3 +21,4 @@ gpxpy==1.1.1
 mock==2.0.0
 timeout-decorator==0.3.2
 raven==5.23.0
+-e git+https://github.com/dmeranda/demjson.git@5bc65974e7141746acc88c581f5d2dfb8ea14064#egg=demjson
```
solderzzc: Can we use official package instead of grab from git?

cwild: Thanks for the feedback, I shall update this momentarily.

solderzzc:
Confirmed, works as expected.
Thanks for the PR.
👍

PR #3143, opened 2016-08-08
Added Run-Loop

devn0ll:
Short Description: 

modified run script wich let you run the boot in a loop(if it crashes it restarts)


run_loop.sh
```diff
@@ -0,0 +1,24 @@
...
+        echo -e "Please create one or use another config file"
+        echo -e "./run.sh [path/to/config/file]"
+        exit 1
+    fi
+fi
+while [ true  ]
+do
+python pokecli.py --config ${config}
+echo "##################################"
+echo "##Exit two times to end the loop##"
```
solderzzc: Change 'Exit' to  [Ctl+C] would be better?

devn0ll: Yeah sounds good

douglascamata: why don't you edit run.sh with these contents?

devn0ll: unsure if everyone wants to run theri bot forever ?

solderzzc:
Thanks

ZeroWolfe:
the code keep launching pokecli.py every 10 second

devn0ll:
bashe is used to wait until the task is done and will continue then as long as you don't use &
so basicly it will do the follow
start HERE
print the message (when this is done)
wait one second (when this is done)
launch the bot (when this is done [means if it crashes or the python script is stopped ])
wait 10 seconds (when this is done)
go to HERE

PR #3155, opened 2016-08-08
MoveToMapPokemon will not run if total amount of balls is too low

pmquan:
* Add ability to skip the MoveToMapPokemon task if total amount of balls is too low

Sometimes, MoveToMapPokemon task will eat your balls very quickly without enough spin_fort. With this setting, MoveToMapPokemon will not run unless you have total amount of balls (combination of Pokeballs, Greatballs and Ultraballs) higher than your desired number. Otherwise, It will run the MoveToFort task.

### DIFF
--- a/configs/config.json.map.example
+++ b/configs/config.json.map.example
@@ -58,6 +58,7 @@
           "address": "http://localhost:5000",
           "max_distance": 500,
           "min_time": 60,
+          "min_ball": 1,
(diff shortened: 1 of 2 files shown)


pmquan:
I found it not catching VIP pokemons when the amount of balls is low but usable. Will fix it in another PR.

PR #3157, opened 2016-08-08
Made paths to .json files absolute so pokecli.py can be called from CRON

Quantra:
Made paths to .json files absolute so pokecli.py can be called from CRON


TheSavior:
Thanks, this should also help with windows support.

:+1:

PR #3161, opened 2016-08-08
Time based tasks

douglascamata:

### DIFF
--- a/pokemongo_bot/__init__.py
+++ b/pokemongo_bot/__init__.py
@@ -8,6 +8,7 @@ import os
 import random
 import re
 import sys
+import struct
 import time
 import Queue
 import threading
@@ -31,7 +32,9 @@ from pokemongo_bot.websocket_remote_control import WebsocketRemoteControl
 from worker_result import WorkerResult
 from tree_config_builder import ConfigException, MismatchTaskApiVersion, TreeConfigBuilder
 from sys import platform as _platform
-import struct
+
+
+
 class PokemonGoBot(object):
     @property
     def position(self):
@@ -441,10 +444,15 @@ class PokemonGoBot(object):
 
         # Check if session token has expired
         self.check_session(self.position[0:2])
+        start_tick = time.time()
 
         for worker in self.workers:
             if worker.work() == WorkerResult.RUNNING:
                 return
+
+        end_tick = time.time()
+        if end_tick - start_tick < 5:
+             time.sleep(5 - (end_tick - start_tick))
 
     def get_meta_cell(self):
         location = self.position[0:2]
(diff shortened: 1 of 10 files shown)


pokemongo_bot/cell_workers/catch_visible_pokemon.py
```diff
@@ -41,7 +41,21 @@ def work(self):
             self.bot.cell['wild_pokemons'].sort(
                 key=
                 lambda x: distance(self.bot.position[0], self.bot.position[1], x['latitude'], x['longitude']))
-            return self.catch_pokemon(self.bot.cell['wild_pokemons'].pop(0))
+
+            for pokemon in self.bot.cell['wild_pokemons']:
```
TheSavior: Hmm, it doesn't look like this got split out from the other PRs. Was the intention to make this PR just the time based tasks?

TheSavior:
It seems like the main benefit of adding time based tasks is that it is easier than having tasks be able to check whether they should run quickly, and ideally without making unnecessary requests to the api. If we have tasks that don't run every tick, we are more likely to get into a case where bags fill up because we haven't run the cleanup tasks yet.

pokemongo_bot/__init__.py
```diff
@@ -441,11 +444,16 @@ def tick(self):
 
         # Check if session token has expired
         self.check_session(self.position[0:2])
+        start_tick = time.time()
 
         for worker in self.workers:
             if worker.work() == WorkerResult.RUNNING:
                 return
 
+        end_tick = time.time()
```
TheSavior: Right now we are already limiting our api wrapper to making at most two request to the api per second. Set here: https://github.com/PokemonGoF/PokemonGo-Bot/blob/master/pokemongo_bot/api_wrapper.py#L47

That is probably sufficient control for throttling to not need to have a minimum time for an entire tick.

TheSavior:
This may not be necessary anymore with the improvements that we have just made. It would be valuable to re-evaluate how much improvement we get and if it is worth the issues we would have for not following the way the ticks have worked previously. We would also run into issues with our bag getting full inside the loop.

Quantra:
I've been running the xp-improvements branch over night, restarting every 30mins without issues.

I like this feature a lot.  The first thing I did was make a cell worker to just spit out some updated stats every 3 minutes.

Even if the improvement in xp efficiency isn't great I still think this is very useful feature.

TheSavior:
@Quantra You should try the current dev. We incorporated a lot of the improvements. I'd be curious to hear what the results are for you on the current dev vs the xp-improvements branch.

Quantra:
@TheSavior Have current dev running well now.  No problems after 30 mins.  xp/h increased by 50-100% over xp-improvements branch.  Great!

However evolves aren't happening correctly as per my config.  Just one evolve at start up then nothing.  Unrelated so will investigate in an issue if required.

pokemongo_bot/cell_workers/catch_visible_pokemon.py
douglascamata: oh crap, I'll remove this

douglascamata:
I'll close this because we have a better solution implemented.

TheSavior:
@Quantra Woah, dev is that much better than xp-improvements? That is awesome and very surprising. Thanks for checking!

PR #3162, opened 2016-08-08
catching every single pokemon nearby

douglascamata:


z4ppy:
I mean pop isn't need here ;)

douglascamata:
@z4ppy you're right. I'll remove.

TheSavior:
This may not be necessary anymore with the improvements that we have just made. It would be valuable to re-evaluate how much improvement we get and if it is worth the issues we would have for not following the way the ticks have worked previously. We would also run into issues with our bag getting full inside the loop.

TheSavior:
Yeah, I updated these to properly return control to the tick, so we will spin all the forts before moving to the next task in the list. I'll go ahead and close both PRs then.

PR #3163, opened 2016-08-08
catch lured pokemon in all forts nearby

douglascamata:

### DIFF
--- a/pokemongo_bot/cell_workers/catch_lured_pokemon.py
+++ b/pokemongo_bot/cell_workers/catch_lured_pokemon.py
@@ -1,9 +1,11 @@
 # -*- coding: utf-8 -*-
 from __future__ import unicode_literals
 
-from pokemongo_bot.cell_workers.utils import fort_details
-from pokemongo_bot.cell_workers.pokemon_catch_worker import PokemonCatchWorker
 from pokemongo_bot.base_task import BaseTask
+from pokemongo_bot.worker_result import WorkerResult
+from pokemongo_bot.constants import Constants
+from pokemongo_bot.cell_workers.utils import fort_details, distance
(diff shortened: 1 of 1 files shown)


pokemongo_bot/cell_workers/catch_lured_pokemon.py
```diff
@@ -12,39 +13,52 @@ class CatchLuredPokemon(BaseTask):
     def work(self):
         lured_pokemon = self.get_lured_pokemon()
         if lured_pokemon:
-            self.catch_pokemon(lured_pokemon)
+            for pokemon in lured_pokemon:
```
TheSavior: Wow, it looks like the code that was here was really bad. It only checked the nearest fort for lured pokemon? That is totally broken. It looks like the new code goes through all nearby forts to find all lured pokemon. Could we take that win separately from catching all the pokemon here?

I'm proposing we just leave this `work()` function the way it is in this PR. We will get lots of improvements from the fix to `get_lured_pokemon`. It would be good to see how much of the behavior we need that is.

mjmadsen:
Can I ask why we have separate files for visible and lured pokemon in the first place? Seems this should be one and the logic merged, as if we have a config to catch any pokemon, we probably want to catch all pokemon in the area. Sorry if this is off topic, but seems related to @TheSavior 's comment about refactoring get_lured_pokemon.

TheSavior:
It shouldn't actually cause any negative effects having two tasks for now, but I don't think I'd be opposed to merging them in a future PR.

TheSavior:
@douglascamata I changed this to check all forts for lured pokemon, but only catch one. If there are more to catch it returns RUNNING, otherwise SUCCESS. Feel free to merge if you are comfortable.

PR #3166, opened 2016-08-08
Configurable minimum balls to run MoveToMap task

pmquan:
* Bot only runs MoveToMap task when it has more than the minimum amount of balls (pokeballs+superballs) or the target is VIP's pokemon.

### DIFF
--- a/pokemongo_bot/cell_workers/move_to_map_pokemon.py
+++ b/pokemongo_bot/cell_workers/move_to_map_pokemon.py
@@ -82,6 +82,7 @@ class MoveToMapPokemon(BaseTask):
         self.pokemon_data = self.bot.pokemon_list
         self.unit = self.bot.config.distance_unit
         self.caught = []
+        self.min_ball = self.config.get('min_ball', 1)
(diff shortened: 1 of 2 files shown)


bigkraig:
:+1:

pokemongo_bot/cell_workers/move_to_map_pokemon.py
```diff
@@ -249,8 +249,14 @@ def work(self):
 
         pokemon = pokemon_list[0]
 
+
+        min_ball = 1
+        #Load the min amount of ball from configuration file
+        if self.config['min_ball']:
```
TheSavior: Can you move this into the initialize function? 

`self.min_ball = self.config.get('min_ball', 1)`

pmquan: Thanks for your suggestion. I already pushed a new commit. I don't know how to merge 3 differents commits into a single one ;(

TheSavior:
:+1:

PR #3180, opened 2016-08-09
Linux Bash Easy installer script

helloitu:
Short Description: 
 This very simple script easy configure the bot for Noob users.

Fixes:
- 
- 
-


easy-installer.sh
```diff
@@ -0,0 +1,48 @@
+#!/bin/bash
```
solderzzc: Should be renamed to 'Linux' ?

solderzzc:
easy-installer.sh should rename to linux-easy-installer.

nikhil-pandey:
Why not use the default config and replace the values there using `sed`? This way will have to maintain additional config.

solderzzc:
Is that better to work out a combine version with this pr
https://github.com/PokemonGoF/PokemonGo-Bot/pull/3263

simonsmh:
Of course sed is much better. #3263

helloitu:
okay, congratulations on arrogance ^^, closed request.

Issue #3190, opened 2016-08-09
pokemon_catch_worker.py boundary error


mjmadsen:
2016-08-08 21:27:35,253 [PokemonCatchWorker] [INFO] [pokemon_appeared] A wild Golbat appeared! [CP 908] [Potential 0.82] [A/D/S 14/15/8]
2016-08-08 21:27:37,901 [PokemonCatchWorker] [WARNING] [softban] Failed to use berry. You may be softbanned.
** run result omitted **
Traceback (most recent call last):
  File "pokecli.py", line 521, in <module>
    main()
  File "pokecli.py", line 95, in main
    bot.tick()
  File "/home/jimbo/PokemonGo-Bot/pokemongo_bot/__init__.py", line 446, in tick
    if worker.work() == WorkerResult.RUNNING:
  File "/home/jimbo/PokemonGo-Bot/pokemongo_bot/cell_workers/catch_lured_pokemon.py", line 17, in work
    self.catch_pokemon(pokemon)
  File "/home/jimbo/PokemonGo-Bot/pokemongo_bot/cell_workers/catch_lured_pokemon.py", line 65, in catch_pokemon
    return_value = worker.work()
  File "/home/jimbo/PokemonGo-Bot/pokemongo_bot/cell_workers/pokemon_catch_worker.py", line 114, in work
    self._do_catch(pokemon, encounter_id, catch_rate_by_ball, is_vip=is_vip)
  File "/home/jimbo/PokemonGo-Bot/pokemongo_bot/cell_workers/pokemon_catch_worker.py", line 332, in _do_catch
    'success_percentage': self._pct(catch_rate_by_ball[current_ball]),
IndexError: list index out of range

### Other Information
Git Commit: 4b16e9da985e689d57beae8bf18878f4d872bdc4
Python Version: Python 2.7.12

It looks like we assign best_ball to ultra_ball before checking that we have them. Either way, we have a boundary error occurring somewhere in here. Haven't spent much time looking at this file, will take a closer look.

# note
I'm not actually soft-banned. This is a legit error in our source.

PR #3197, opened 2016-08-09
Fix for #3190

bigkraig:
Short Description: 

Recent refactor can return an empty list

Fixes:
when we would normally return the empty list, return the original list


mjmadsen:
I think we have the problem in a few other files. I'll keep an eye on my logs.

Good, and quick, work!

Issue #3082, opened 2016-08-08
[TransferPokemon] [FollowSpiral] | Loop | Log Spam


GitPupper:
### Logs
***
***
2016-08-08 10:10:13,146 [PokemonGoBot] [INFO] Found encrypt.so! Platform: darwin Bot root directory: /Users/bobbyobrien/PokemonGo-Bot
2016-08-08 10:10:13,146 [PokemonGoBot] [INFO]
2016-08-08 10:10:13,488 [PokemonGoBot] [INFO] [bot_start] Starting bot...
2016-08-08 10:10:14,777 [CollectLevelUpReward] [INFO] [level_up_reward] Received level up reward: []
2016-08-08 10:10:14,777 [IncubateEggs] [INFO] [next_egg_incubates]  Next egg incubates in 2.64 km
2016-08-08 10:10:14,972 [TransferPokemon] [INFO] [keep_best_release]  Keeping best 4 Pidgey, based on cp
2016-08-08 10:10:14,973 [TransferPokemon] [INFO] [keep_best_release]  Keeping best 4 Rattata, based on cp
2016-08-08 10:10:15,155 [FollowSpiral] [INFO] [position_update] Walking from (34.0117, 118.5001, 0) to (34.0117, 118.5001, 0) (0.0 m)
2016-08-08 10:10:15,421 [TransferPokemon] [INFO] [keep_best_release]  Keeping best 4 Pidgey, based on cp
2016-08-08 10:10:15,421 [TransferPokemon] [INFO] [keep_best_release]  Keeping best 4 Rattata, based on cp
2016-08-08 10:10:15,602 [FollowSpiral] [INFO] [position_update] Walking from (34.0117, 118.5001, 0) to (34.012333049997544, 118.5001, 0) (70.3919467627 m)
2016-08-08 10:10:18,694 [TransferPokemon] [INFO] [keep_best_release]  Keeping best 4 Pidgey, based on cp
2016-08-08 10:10:18,695 [TransferPokemon] [INFO] [keep_best_release]  Keeping best 4 Rattata, based on cp
2016-08-08 10:10:20,076 [TransferPokemon] [INFO] [keep_best_release]  Keeping best 4 Pidgey, based on cp
2016-08-08 10:10:20,076 [TransferPokemon] [INFO] [keep_best_release]  Keeping best 4 Rattata, based on cp
[...]

Issue #3064, opened 2016-08-08
Don't log TransferPokemon with keep_best_iv if rules don't match any


DiegoRBaquero:
### Expected Behavior
Not logging about 3rd+ Pokemon based on IV because none should be released based on rules.

### Actual Behavior
Logging about releasing 3rd+ Pokemon when they don't match release rules:
````
[TransferPokemon] [INFO] [keep_best_release] Keeping best 2 Eevee, based on iv
[TransferPokemon] [INFO] [keep_best_release] Keeping best 2 Horsea, based on iv
````

This keeps showing up every 4 lines.

It does release others normally based on rules:
````
[TransferPokemon] [INFO] [keep_best_release] Keeping best 2 Eevee, based on iv
[TransferPokemon] [INFO] [keep_best_release] Keeping best 2 Pidgey, based on iv
[TransferPokemon] [INFO] [future_pokemon_release] Releasing Pidgey (CP 284/IV 0.38) based on rule: CP < 100 OR IV < 0.8
[TransferPokemon] [INFO] [pokemon_release] Exchanged Pidgey [CP 284] [IV 0.38] for candy.
````
### Steps to Reproduce
Having 3+ Pokemons that shouldn't be released and this config:
`"release": {"any": {"release_below_cp": 100, "release_below_iv": 0.8, "logic": "or", "keep_best_iv": 2} }`

### Other Information
OS: Mac OS
Git Commit: 5b3fe3d187b6b5fdfdede00f064bf7bcbd5a484d
Python Version: 2.7.12

GitPupper:
Same problem. It's because the bot doesn't transfer pokemon for candy anymore. To fix this I just edited the config to keep best 0. It doesn't show it in the logs, but it still does't transfer pokemon.

TheSavior:
Fixed in https://github.com/PokemonGoF/PokemonGo-Bot/pull/3223

Issue #2790, opened 2016-08-07
TransferPokemon keep_best_release 


AcorpBG:
### Expected Behavior
TransferPokemon should not call or log non stop.
2016-08-07 15:20:25,217 [TransferPokemon] [INFO] [keep_best_release] Keeping best 4 Zubat, based on cp and iv
2016-08-07 15:20:25,218 [TransferPokemon] [INFO] [future_pokemon_release] Releasing Zubat (CP 10/IV 0.71) based on rule: CP < 225 OR IV < 0.8
2016-08-07 15:20:25,640 [TransferPokemon] [INFO] [pokemon_release] Exchanged Zubat [CP 10] [IV 0.71] for candy.
Check keep_best_release  then release extras.

### Actual Behavior
2016-08-07 15:20:25,217 [TransferPokemon] [INFO] [keep_best_release] Keeping best 4 Zubat, based on cp and iv
2016-08-07 15:20:25,218 [TransferPokemon] [INFO] [future_pokemon_release] Releasing Zubat (CP 10/IV 0.71) based on rule: CP < 225 OR IV < 0.8
2016-08-07 15:20:25,640 [TransferPokemon] [INFO] [pokemon_release] Exchanged Zubat [CP 10] [IV 0.71] for candy.
2016-08-07 15:20:28,945 [TransferPokemon] [INFO] [keep_best_release] Keeping best 3 Drowzee, based on cp and iv
2016-08-07 15:20:38,960 [TransferPokemon] [INFO] [keep_best_release] Keeping best 3 Drowzee, based on cp and iv
2016-08-07 15:20:49,143 [TransferPokemon] [INFO] [keep_best_release] Keeping best 3 Drowzee, based on cp and iv
2016-08-07 15:20:49,966 [TransferPokemon] [INFO] [keep_best_release] Keeping best 3 Drowzee, based on cp and iv
2016-08-07 15:20:50,698 [  SpinFort] [INFO] [spun_pokestop] Spun pokestop п═я┐п╦пҐп╦ 2. Experience awarded: 50. Items awarded: {u'Razz Berry': 1, u'Pokeball': 1, u'Super Potion': 1}
2016-08-07 15:21:01,985 [TransferPokemon] [INFO] [keep_best_release] Keeping best 3 Drowzee, based on cp and iv
2016-08-07 15:21:02,548 [RecycleItems] [INFO] [item_discarded] Discarded 1x Super Potion (maximum 0).
[...]

Issue #2787, opened 2016-08-07
Release config - Keep best IV and release below params


antoineneff:
### Expected Behavior

I would like to keep one best pokemon based on their IV unless they have more than 1200CP and 0.9 IV, in that case I want to keep them all.

### Actual Behavior

It does work actually but I have some logging errors : 
```
2016-08-07 14:02:26,712 [TransferPokemon] [INFO] [keep_best_release] Keeping best 1 Vaporeon, based on iv
2016-08-07 14:02:27,316 [MoveToFort] [INFO] [moving_to_fort] Moving towards pokestop Carol P. Johnson Memorial - 0.18km
2016-08-07 14:02:34,899 [TransferPokemon] [INFO] [keep_best_release] Keeping best 1 Vaporeon, based on iv
2016-08-07 14:02:35,433 [MoveToFort] [INFO] [moving_to_fort] Moving towards pokestop Carol P. Johnson Memorial - 0.17km
2016-08-07 14:02:43,713 [TransferPokemon] [INFO] [keep_best_release] Keeping best 1 Vaporeon, based on iv
2016-08-07 14:02:44,255 [MoveToFort] [INFO] [moving_to_fort] Moving towards pokestop Carol P. Johnson Memorial - 0.16km
2016-08-07 14:02:52,431 [TransferPokemon] [INFO] [keep_best_release] Keeping best 1 Vaporeon, based on iv
2016-08-07 14:02:53,020 [MoveToFort] [INFO] [moving_to_fort] Moving towards pokestop Carol P. Johnson Memorial - 0.15km
2016-08-07 14:03:00,815 [TransferPokemon] [INFO] [keep_best_release] Keeping best 1 Vaporeon, based on iv
2016-08-07 14:03:01,340 [MoveToFort] [INFO] [moving_to_fort] Moving towards pokestop Carol P. Johnson Memorial - 0.14km
2016-08-07 14:03:09,565 [TransferPokemon] [INFO] [keep_best_release] Keeping best 1 Vaporeon, based on iv
2016-08-07 14:03:10,138 [MoveToFort] [INFO] [moving_to_fort] Moving towards pokestop Carol P. Johnson Memorial - 0.13km
2016-08-07 14:03:17,495 [TransferPokemon] [INFO] [keep_best_release] Keeping best 1 Vaporeon, based on iv
```

### Steps to Reproduce

This is my config, I guess it's not what is expected, but I don't know how to write it then : 
```
"release": {
[...]

PR #3223, opened 2016-08-09
Fix flooding of keep_best_release

phphong:
### Short Description: 
Only log `'keep_best_release'` when there are some pokemons to release. The previous version keeps flooding `'keep_best_release'` whenever there are some pokemons which is not in the `best_pokemons` list but still don't satisfy the condition to be released.

This fixes the log flooding issue in #2790, #2753, #2559, #3064, #2787, #3082, #2621

### DIFF
--- a/pokemongo_bot/cell_workers/transfer_pokemon.py
+++ b/pokemongo_bot/cell_workers/transfer_pokemon.py
@@ -50,15 +50,16 @@ class TransferPokemon(BaseTask):
                                                                         True)]
 
                     if transfer_pokemons:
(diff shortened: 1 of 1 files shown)


dezayner:
Work fine on me

TheSavior:
:+1:

paulinhoneto:
transfer pokemon ?

paulinhoneto:
[PokemonCatchWorker] [INFO] [pokemon_inventory_full] Your Pokemon inventory is full! Could not catch!

PR #3236, opened 2016-08-09
Keeping a cache of gym information

TheSavior:
Before this change, I was fetching the details of 16 gyms every tick. Since we throttle api requests to two a second, this was 8 seconds of requests every tick to fetch gym information. Now that we are caching it, we essentially remove the entirety of that time. This should have significant impacts to xp/hr.

@douglascamata

### DIFF
--- a/pokemongo_bot/__init__.py
+++ b/pokemongo_bot/__init__.py
@@ -25,6 +25,7 @@ from event_manager import EventManager
 from human_behaviour import sleep
 from item_list import Item
 from metrics import Metrics
+from gym_cache import GymCache
 from pokemongo_bot.event_handlers import LoggingHandler, SocketIoHandler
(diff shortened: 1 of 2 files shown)


TheSavior:
It was mentioned in slack that we don't actually need to get gym details since we don't do anything with gyms. I don't see any reference to these fields in the bot. It looks like we just grab this data so that we can write it to the json file so we can render it on the map. 

I think these are silly requests to be making when we have stated we don't want any gym related code in the bot. It seems like a pretty irrelevant UI feature. @douglascamata thoughts?

douglascamata:
@TheSavior we can safely remove them, IMHO.

TheSavior:
@douglascamata PR here: https://github.com/PokemonGoF/PokemonGo-Bot/pull/3245

PR #3259, opened 2016-08-09
Return an empty list if no pokemon are available.

poupas:
Short Description: The current dev branch is crashing when trying to catch lured pokemon.

The changes introduced in 4c95259 expose this bug.

The line that triggers the bug is the following:
https://github.com/PokemonGoF/PokemonGo-Bot/commit/4c95259c25543b483a539a928ae67e7e590778f3#diff-900c950b761cc9227d14af60ae8ee987R16


TheSavior:
:+1: Thanks!

PR #3263, opened 2016-08-09
Powerful setup.sh

simonsmh:
Short Description: 

Very powerful run.sh with lots of function.
1.install(make .so)
2.update
3.config generator
4.config backup
5.run loop make it never get down

It should run like 
run.sh *.json or other opinion. 
See -help.


run.sh
```diff
@@ -1,26 +1,145 @@
...
+sed -i "s/GOOGLE_MAPS_API_KEY/$gmapkey/g" configs/config.json
+echo "edit configs/config.json to modify any other config."
+}
+
+function Pokebotinstall () {
+cd $pokebotpath
+echo "1.Debian/Ubuntu"
+echo "2.Centos/RedHat"
+echo "3.Mac os"
+echo "4.Other"
```
terbolous: This can generally be probed with stuff like `lsb_release` or `uname`

run.sh
```diff
@@ -1,26 +1,145 @@
-#!/usr/bin/env bash
+#!/bin/bash
+pokebotpath="/root/PokemonGo-Bot"
+backuppath="/root/PokemonGo-Bot/backup"
```
terbolous: hard coded paths :-(

simonsmh: it can be edited.
Let me fix it.

run.sh
simonsmh: I can't fix it because I don't have mac or centos devices. It's just working. Can you help me to fix it? Thx.

solderzzc:
run.sh is a script to run configuration, we need separate setup to setup.sh.

simonsmh:
@solderzzc Name it pokebot.sh is better.

simonsmh:
Does it really need so much shell? :P

solderzzc:
yes, I have two scripts running by ./run.sh, so we'd better keep it for other users.

run.sh
```diff
@@ -1,26 +1,17 @@
 #!/usr/bin/env bash
-
-# Starts PokemonGo-Bot
-config=""
-
-if [ ! -z $1 ]; then
```
solderzzc: this one is still needed for multiple configurations.

simonsmh:
Is it OK? I edited run.sh setup.sh install.sh and removed setup.py which is not fit to use. The usage of run.sh is different but easier.

run.sh
simonsmh: OK I will revert it.

simonsmh:
Maybe every new user here should use this new install.sh or setup.sh to make them use this bot in a very short time. The guide of how to install may should be updated too. :p

run.sh
```diff
@@ -1,26 +1,21 @@
 #!/usr/bin/env bash
-
-# Starts PokemonGo-Bot
-config=""
-
+pokebotpath=$(pwd)
+filename=""
 if [ ! -z $1 ]; then
```
solderzzc: This is needed.

solderzzc:
great, waiting for the document :) we need add doc folder

PR #3264, opened 2016-08-09
Allow UpdateTitleStats to emit events instead of rewriting the console

mmnsgo:


TheSavior:
@douglascamata, please review.

pokemongo_bot/cell_workers/update_title_stats.py
```diff
@@ -70,6 +70,8 @@ def __init__(self, bot, config):
         self.min_interval = self.DEFAULT_MIN_INTERVAL
         self.displayed_stats = self.DEFAULT_DISPLAYED_STATS
 
+        self.bot.event_manager.register_event('update_title', parameters=('title'))
```
douglascamata: Good suggestion to move the task's specific event registration to inside the task itself. I think we should follow this everywhere, @TheSavior. 

@mmnsgo feel free to open another PR moving the registration of all events used only inside a task to the task class itself.

douglascamata:
👍

PR #3277, opened 2016-08-09
Dev

Raiven66:
New config values for sniping only filtered pokemon. Reasoning of this feature is to reduce sniping to only important stuff and minimize inhuman behavior - teleporting.

if "snipe_high_prio_only" is true, only VIP pokemon and pokemon with priority higher than "snipe_high_prio_threshold" will be sniped

### DIFF
--- a/configs/config.json.map.example
+++ b/configs/config.json.map.example
@@ -61,7 +61,9 @@
           "min_time": 60,
           "min_ball": 50,
           "prioritize_vips": true,
-          "snipe": false,
+          "snipe": true,
+          "snipe_high_prio_only": true,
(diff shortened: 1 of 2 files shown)


DimaVIII:
Somehow this is a bit to much with VIP and snipe_high_prio_only.
So you have two list you need keep track on.
In my opinion the VIP should be removed from the snipe function as you already have a full list of pokemons where you set the priority manually for each.

Raiven66:
@DimaVIII the whole Move To Map Pokemon feature already includes a prioritize_vips config. With regard to this, I added the VIPs in this PR as well. What probably could be done is to use the prioritize_vips flag to control this as well.

PR #3312, opened 2016-08-09
Log stats on terminal

sergiopalacio:
Short Description: 
Update title stats not working on Git bash users 
Fixes:
- Show a terminal log

EDIT: Maybe add a config parameter to enable this behavior?
EDIT2: Config enable already pushed

If you want to enable stats logging on terminal just add a "terminal_log" parameter set to true on your config file.
Aswell, if you want to disable the title changing, just add an "terminal_title" set to false and it will disable this behavior.
Fe: 
```
{
       "type": "UpdateTitleStats",
       "config": {
           "min_interval": 120,
           "stats": [
             "uptime",
             "xp_earned",
             "xp_per_hour",
             "stardust_earned"
           ],
           "terminal_log": true,
           "terminal_title": false
       }
     },
```

### DIFF
--- a/pokemongo_bot/cell_workers/update_title_stats.py
+++ b/pokemongo_bot/cell_workers/update_title_stats.py
@@ -18,10 +18,22 @@ class UpdateTitleStats(BaseTask):
         "type": "UpdateTitleStats",
         "config": {
             "min_interval": 10,
-            "stats": ["login", "uptime", "km_walked", "level_stats", "xp_earned", "xp_per_hour"]
+            "stats": ["login", "uptime", "km_walked", "level_stats", "xp_earned", "xp_per_hour"],
         }
     }
 
+    You can set a logging on terminal mode like this:
+
+    Example logging on console (and disabling title change):
+    {
+        "type": "UpdateTitleStats",
+        "config": {
+            "min_interval": 10,
+            "stats": ["login", "uptime", "km_walked", "level_stats", "xp_earned", "xp_per_hour"],
+            "terminal_log": true,
+            "terminal_title": false
+        }
+    }
     Available stats :
     - login : The account login (from the credentials).
     - username : The trainer name (asked at first in-game connection).
@@ -53,9 +65,6 @@ class UpdateTitleStats(BaseTask):
     """
     SUPPORTED_TASK_API_VERSION = 1
 
-    DEFAULT_MIN_INTERVAL = 10
-    DEFAULT_DISPLAYED_STATS = []
-    TERMINAL = False
 
     def __init__(self, bot, config):
         """
@@ -68,13 +77,14 @@ class UpdateTitleStats(BaseTask):
         super(UpdateTitleStats, self).__init__(bot, config)
 
         self.next_update = None
-        self.min_interval = self.DEFAULT_MIN_INTERVAL
-        self.displayed_stats = self.DEFAULT_DISPLAYED_STATS
-        self.terminal = self.TERMINAL
+
+        self.min_interval = int(self.config.get('min_interval', 120))
+        self.displayed_stats = self.config.get('stats', [])
+        self.terminal_log = self.config.get('terminal_log', False)
+        self.terminal_title = self.config.get('terminal_title', True)
 
         self.bot.event_manager.register_event('update_title', parameters=('title'))
         self.bot.event_manager.register_event('log_stats',parameters=('title'))
-        self._process_config()
 
     def initialize(self):
         pass
@@ -91,8 +101,11 @@ class UpdateTitleStats(BaseTask):
         # If title is empty, it couldn't be generated.
         if not title:
             return WorkerResult.SUCCESS
-        self._update_title(title, _platform)
-        if(self.terminal is True):
+
+        if self.terminal_title is True:
+            self._update_title(title, _platform)
+
+        if self.terminal_log is True:
             self._log_on_terminal(title)
         return WorkerResult.SUCCESS
 
@@ -147,15 +160,6 @@ class UpdateTitleStats(BaseTask):
 
         self.next_update = datetime.now() + timedelta(seconds=self.min_interval)
 
-    def _process_config(self):
-        """
-        Fetches the configuration for this worker and stores the values internally.
-        :return: Nothing.
-        :rtype: None
-        """
-        self.min_interval = int(self.config.get('min_interval', self.DEFAULT_MIN_INTERVAL))
-        self.displayed_stats = self.config.get('stats', self.DEFAULT_DISPLAYED_STATS)
-        self.terminal = self.config.get('terminal', self.TERMINAL)
 
     def _get_stats_title(self, player_stats):
         """
--- a/pokemongo_bot/cell_workers/update_title_stats.py
+++ b/pokemongo_bot/cell_workers/update_title_stats.py
@@ -71,7 +71,7 @@ class UpdateTitleStats(BaseTask):
         self.displayed_stats = self.DEFAULT_DISPLAYED_STATS
 
         self.bot.event_manager.register_event('update_title', parameters=('title'))
-
+        self.bot.event_manager.register_event('log_stats',parameters=('title'))
         self._process_config()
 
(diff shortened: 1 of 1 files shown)


douglascamata:
@sergiopalacio task configuration is the way, definitely.

TheSavior:
Yeah, this is great. We definitely want this to be configurable and I think it should probably default to off.

sergiopalacio:
Hey, just right now added the option "terminal" which enables this behavior if set true on cfg, if not set, it will remain on his default false value. @douglascamata @TheSavior

pokemongo_bot/cell_workers/update_title_stats.py
```diff
@@ -69,6 +70,7 @@ def __init__(self, bot, config):
         self.next_update = None
         self.min_interval = self.DEFAULT_MIN_INTERVAL
         self.displayed_stats = self.DEFAULT_DISPLAYED_STATS
+        self.terminal = self.TERMINAL
```
douglascamata: This seems a bit repetitive. Why shouldn't you initialise default values here directly without going through these constants?

pokemongo_bot/cell_workers/update_title_stats.py
```diff
@@ -90,7 +92,8 @@ def work(self):
         if not title:
             return WorkerResult.SUCCESS
         self._update_title(title, _platform)
-        self._log_on_terminal(title)
+        if(self.terminal is True):
```
douglascamata: Parenthesis are not necessary.

pokemongo_bot/cell_workers/update_title_stats.py
```diff
@@ -152,6 +155,7 @@ def _process_config(self):
         """
         self.min_interval = int(self.config.get('min_interval', self.DEFAULT_MIN_INTERVAL))
         self.displayed_stats = self.config.get('stats', self.DEFAULT_DISPLAYED_STATS)
+        self.terminal = self.config.get('terminal', self.TERMINAL)
```
douglascamata: Here you rewrite all the configs again. IMHO, this code should  go to `__init__.py` and instead of the constants as default we can have actual values of these constants.

sergiopalacio:
@douglascamata @TheSavior Should I'm thinking about I should add parameter like "title_off" , which will  let us disable the terminal title changes, since users which use Git bash never get approach of this task. Or maybe just someone prefer only terminal logging? 

Sorry if I missed something with my english :)

pokemongo_bot/cell_workers/update_title_stats.py
sergiopalacio: Thanks, swiching between languages made me use parenthesis.

crvfts:
UpdateTitleStats is now showing in my logger every 2-3 seconds. I only want it on the titlebar like before. 

How may I fix this?

sergiopalacio:
@crvfts Just remove terminal parameter from config

crvfts:
@sergiopalacio that was the first thing I tried, it did not work.

sergiopalacio:
@crvfts I cannot reproduce your problem mate, maybe bad pull?

douglascamata:
@sergiopalacio did you take a look to my comments? When you fix them I'll merge this.

sergiopalacio:
@douglascamata Yes, I'm writing the final code now, what about a param to swich off "terminal title changing"?

douglascamata:
@sergiopalacio I think a parameter to set it on, like `terminal_title` is better. When you finish, please ping me.

sergiopalacio:
@douglascamata Ok, all done. I refactored the code, as you said, it was redundant. And is more clear having get configs on init you're right. 
Changed the terminal parameter to terminal_log (default false) 
And added a new parameter terminal_title (default true) 
Thanks for reviewing so fast

pokemongo_bot/cell_workers/update_title_stats.py
```diff
@@ -67,12 +77,14 @@ def __init__(self, bot, config):
...
 
-        self.bot.event_manager.register_event('update_title', parameters=('title'))
+        self.min_interval = int(self.config.get('min_interval', 120))
+        self.displayed_stats = self.config.get('stats', [])
+        self.terminal_log = self.config.get('terminal_log', False)
+        self.terminal_title = self.config.get('terminal_title', True)
 
-        self._process_config()
+        self.bot.event_manager.register_event('update_title', parameters=('title'))
+        self.bot.event_manager.register_event('log_stats',parameters=('title'))
```
douglascamata: both these two `register_event` calls needs a comma in the parameters tuple, because of this:

```
>>> ('title')
'title'

>>> ('title',)
('title',)
```

In short words: the first one is a normal string, the second one is a tuple with one single string inside.

douglascamata:
@sergiopalacio i'm more small thing and good to go :D

sergiopalacio:
@douglascamata hahaha I'm starting with python, and this kind of things are important for me ;) Check to see if you have more suggestions :dancer:

pokemongo_bot/cell_workers/update_title_stats.py
```diff
@@ -89,7 +101,12 @@ def work(self):
         # If title is empty, it couldn't be generated.
         if not title:
             return WorkerResult.SUCCESS
-        self._update_title(title, _platform)
+
+        if self.terminal_title is True:
```
douglascamata: This can be refactored to `if self.terminal_title`, you don't need to explicitly check if it's `True`. Same to the other `if` right below this.

douglascamata:
@sergiopalacio the tuple trick is not so clear. It doesn't happen with lists though (using squared brackets).

pokemongo_bot/cell_workers/update_title_stats.py
sergiopalacio: Oh I thought it didn't work on phyton, even tried :P nice wait for it

sergiopalacio:
@douglascamata Refactor done. Any though? :D

douglascamata:
👍 

LGTM

douglascamata:
@sergiopalacio thanks for the contribution!

sergiopalacio:
@douglascamata It's only the beginning :100:

Kassadd:
Thanks a lot, very useful.

One question, why update_title has and emit_event too? If you set terminal_log to false and terminal_title to true, is because you don't want the log spammed.

h3k1ru:
Will be good to have pokemon per hour too..

TCB13:
This is great! However I would like very much to see all the information that appears when the user logs in, like stardust, etc... That would be a great addition.. Why not just copy the code that generates that information?

sergiopalacio:
@TCB13 Yep, you have a complete list of which params you can show on pokemongo_bot/cell_workers/update_title_stats.py refer to "Available stats" line 37

@Kassadd Yep, you're right, it wasn't my own code and did not touch _update_title method. If I remember, somewhere on the code, when script ends it logs stats, so maybe it could be related?

Moonlight-Angel:
@sergiopalacio The issue @crvfts talked about is real. You're using an event with updating the terminal title and this event is handled in the LoggerHandler, which logs it to the terminal.
You shouldn't use an event when updating the title.

I'm preparing a PR fixing this issue, containing some refactoring, adding missing unit tests and renaming the worker since his name isn't appropriate anymore (logging stats into terminal isn't updating the title 😕).

siembz:
Is it possible to get some colour on the stats ?

BasvanH:
On terminal means you should see the stats between the rest of the info being logged on the CLI while running the bot? Because I'm not seing any stat info. Grabbed dev a few minutes ago, and added type "UpdateLiveStats" with "terminal_log" set to true.

Moonlight-Angel:
@BasvanH Wasn't able to reproduce your issue. `terminal_log` logs the stats as expected.
If you still have the issue, open an issue and give more information about your OS etc.

BasvanH:
@Moonlight-Angel , I dont see the stats then running MoveToFort, I do see them running FollowPath.

CentOS 7, python 2.7.5

PR #3317, opened 2016-08-09
Keep extra pokemon in our bag so we can do smarter things later

thebigjc:
Short Description: 
Don't release pokemon unless we have limited space based on an optional config. Allows us to do smarter things later, like keep enough to evolve in batch with an egg.

### DIFF
--- a/pokemongo_bot/cell_workers/recycle_items.py
+++ b/pokemongo_bot/cell_workers/recycle_items.py
@@ -9,7 +9,7 @@ class RecycleItems(BaseTask):
     SUPPORTED_TASK_API_VERSION = 1
 
     def initialize(self):
-        self.min_empty_space = self.config.get('min_empty_space', None)
+        self.min_empty_space = int(self.config.get('min_empty_space', 0))
         self.item_filter = self.config.get('item_filter', {})
         self._validate_item_filter()
 
@@ -27,7 +27,7 @@ class RecycleItems(BaseTask):
         total_bag_space = self.bot.player_data['max_item_storage']
         free_bag_space = total_bag_space - items_in_bag
 
-        if free_bag_space > 0 and isinstance(self.min_empty_space, int) and free_bag_space >= self.min_empty_space:
+        if free_bag_space >= self.min_empty_space:
             return
 
         self.bot.latest_inventory = None
--- a/pokemongo_bot/cell_workers/recycle_items.py
+++ b/pokemongo_bot/cell_workers/recycle_items.py
@@ -1,129 +1,77 @@
 import json
 import os
-
-from pokemongo_bot import inventory
 from pokemongo_bot.base_dir import _base_dir
 from pokemongo_bot.base_task import BaseTask
-from pokemongo_bot.services.item_recycle_worker import ItemRecycler
 from pokemongo_bot.tree_config_builder import ConfigException
-from pokemongo_bot.worker_result import WorkerResult
 
-DEFAULT_MIN_EMPTY_SPACE = 6
 
 class RecycleItems(BaseTask):
     SUPPORTED_TASK_API_VERSION = 1
 
-    """
-    Recycle undesired items if there is less than five space in inventory.
-    You can use either item's name or id. For the full list of items see ../../data/items.json
-
-    It's highly recommended to put this task before move_to_fort and spin_fort task in the config file so you'll most likely be able to loot.
-
-    Example config :
-    {
-      "type": "RecycleItems",
-      "config": {
-        "min_empty_space": 6,           # 6 by default
-        "item_filter": {
-          "Pokeball": {"keep": 20},
-          "Greatball": {"keep": 50},
-          "Ultraball": {"keep": 100},
-          "Potion": {"keep": 0},
-          "Super Potion": {"keep": 0},
-          "Hyper Potion": {"keep": 20},
-          "Max Potion": {"keep": 50},
-          "Revive": {"keep": 0},
-          "Max Revive": {"keep": 20},
-          "Razz Berry": {"keep": 20}
-        }
-      }
-    }
-    """
-
     def initialize(self):
-        self.items_filter = self.config.get('item_filter', {})
         self.min_empty_space = self.config.get('min_empty_space', None)
+        self.item_filter = self.config.get('item_filter', {})
         self._validate_item_filter()
 
     def _validate_item_filter(self):
-        """
-        Validate user's item filter config
-        :return: Nothing.
-        :rtype: None
-        :raise: ConfigException: When an item doesn't exist in ../../data/items.json
-        """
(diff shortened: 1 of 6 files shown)


pokemongo_bot/cell_workers/transfer_pokemon.py
```diff
@@ -7,7 +7,24 @@
...
+        self.min_empty_space = self.config.get('min_empty_space', None)
+
+    def get_pokemon_slot_left(self):
+        return self.bot._player["max_pokemon_storage"] - self.bot.get_inventory_count('pokemon')
+
     def work(self):
+        if self.min_empty_space is not None:
+            free_poke_space = self.get_pokemon_slot_left()
+            if free_poke_space >= self.min_empty_space:
+                print free_poke_space
```
TheSavior: You should remove this print

pokemongo_bot/cell_workers/transfer_pokemon.py
```diff
@@ -7,7 +7,24 @@
...
+                self.emit_event(
```
TheSavior: I imagine this is going to log **all** the time. Do we really need an event for this? @douglascamata What do you think?

pokemongo_bot/cell_workers/transfer_pokemon.py
thebigjc: Fixed

TheSavior:
Looks like some bad commits got merged in here

pokemongo_bot/cell_workers/transfer_pokemon.py
thebigjc: I followed the model from https://github.com/PokemonGoF/PokemonGo-Bot/blob/dev/pokemongo_bot/cell_workers/recycle_items.py#L32 which also is chatty. Happy to remove this or do something less chatty (i.e. only every N calls)

thebigjc:
bleh - sorry - bad pull - will fix

TheSavior:
This looks great. Just curious on @douglascamata's thoughts on the event

thebigjc:
Thanks!

knobunc:
LGTM.  It's the same pattern as with the items.

However, I find them both a little chatty... perhaps we can turn down the frequency of the messages?

douglascamata:
As we talked in Slack, the message should only be displayed when at least ONE is recycled by the task.

thebigjc:
@douglascamata @knobunc @TheSavior Thanks for the feedback - removed the spammy messages as discussed.

joaodragao:
I think you should separate between `min_empty_space` for bag storage and Pokemon storage. Also, the `min_empty_space` is defined under `config` of `RecycleItem`.

bstpierr:
👍

pokemongo_bot/cell_workers/recycle_items.py
```diff
@@ -27,16 +27,8 @@ def work(self):
...
-            if free_bag_space >= self.min_empty_space:
-                    self.emit_event(
-                        'item_discard_skipped',
-                        formatted="Skipping Recycling of Items. {space} space left in bag.",
-                        data={
-                            'space': free_bag_space
-                        }
-                    )
-                    return
+        if self.min_empty_space is not None and free_bag_space >= self.min_empty_space:
```
XtrmJosh: Might be worth switching this to if `isinstance(self.min_empty_space, int)` in the interest of also not returning when people accidentally break the syntax of config files...

pokemongo_bot/cell_workers/transfer_pokemon.py
```diff
@@ -10,7 +10,16 @@
...
     SUPPORTED_TASK_API_VERSION = 1
 
+    def initialize(self):
+        self.min_empty_space = self.config.get('min_empty_space', None)
+
+    def get_pokemon_slot_left(self):
+        return self.bot._player["max_pokemon_storage"] - self.bot.get_inventory_count('pokemon')
+
     def work(self):
+        if self.min_empty_space is not None and self.get_pokemon_slot_left() >= self.min_empty_space:
```
douglascamata: Check `inventory.py` to learn how to check for pokemon bag space properly without making extra requests.

thebigjc:
Feedback from @XtrmJosh and @douglascamata incorporated!

douglascamata:
@thebigjc I'll take a look tomorrow's morning. Too late for me, sorry.

thebigjc:
No worries. Shouldn't take more than a minute.

douglascamata:
@thebigjc solve merge conflicts please, this LGTM already!

thebigjc:
@douglascamata done! Ready for you to merge!

pokemongo_bot/cell_workers/recycle_items.py
```diff
@@ -27,16 +27,8 @@ def work(self):
...
-            if free_bag_space >= self.min_empty_space and items_in_bag < total_bag_space:
-                    self.emit_event(
-                        'item_discard_skipped',
-                        formatted="Skipping Recycling of Items. {space} space left in bag.",
-                        data={
-                            'space': free_bag_space
-                        }
-                    )
-                    return
+        if isinstance(self.min_empty_space, int) and free_bag_space >= self.min_empty_space and items_in_bag < total_bag_space:
```
TheSavior: You could change line 12 to `int(self.min_empty_space = self.config.get('min_empty_space', 0))` and remove the `isinstance` check here

TheSavior: Why do you need the check for `items_in_bag < total_bag_space`, doesn't that just mean that the bag isn't full? Wouldn't we want to run if the bag is full?

thebigjc:
@TheSavior Addressed your two points

pokemongo_bot/cell_workers/recycle_items.py
```diff
@@ -27,16 +27,8 @@ def work(self):
...
-            if free_bag_space >= self.min_empty_space and items_in_bag < total_bag_space:
-                    self.emit_event(
-                        'item_discard_skipped',
-                        formatted="Skipping Recycling of Items. {space} space left in bag.",
-                        data={
-                            'space': free_bag_space
-                        }
-                    )
-                    return
+        if free_bag_space >= self.min_empty_space:
```
TheSavior: If people don't set the min_empty_space, then this will always be true, and thus skip recycle_items. That probably isn't the expected default behavior.

TheSavior:
@thebigjc Why did you close this? Do we not need this anymore?

thebigjc:
@TheSavior Mostly frustration. I had a 4 line code change doing something exactly the same way as running code that turned into a 25 comment back and forth refactoring code I hadn't planned on touching, and the delays on the back-and-forths resulted in many merge conflicts as much more complex changes were coming through. There are much more interesting features to work on than something this simple and stacking all these changes into this PR isn't how I would like to spend my limited time.

Issue #3313, opened 2016-08-09
!dev - CatchLuredPokemon infinite loop


Philipp59:
testing with 0f2bddd26675c5232b2e9f4e83dc6525c41a842d
bags are not full and config is fine 

not sure whether its just session stale or a problem with task. 
Anyone got this also and can help testing  ?
```
2016-08-09 18:09:47,725 [CatchLuredPokemon] [INFO] [lured_pokemon_found] Lured pokemon at fort Wooden Bridge (f2dae5c651c0411c923a6e455b6fa2ad.16)
2016-08-09 18:09:48,399 [CatchLuredPokemon] [INFO] [lured_pokemon_found] Lured pokemon at fort Alexander The Great Rock (0c2581656bd848ba83dadd7b16ea5b86.16)
2016-08-09 18:09:48,400 [CatchLuredPokemon] [INFO] [lured_pokemon_found] Lured pokemon at fort Wooden Bridge (f2dae5c651c0411c923a6e455b6fa2ad.16)
2016-08-09 18:09:49,271 [CatchLuredPokemon] [INFO] [lured_pokemon_found] Lured pokemon at fort Alexander The Great Rock (0c2581656bd848ba83dadd7b16ea5b86.16)
2016-08-09 18:09:49,272 [CatchLuredPokemon] [INFO] [lured_pokemon_found] Lured pokemon at fort Wooden Bridge (f2dae5c651c0411c923a6e455b6fa2ad.16)
2016-08-09 18:09:50,187 [CatchLuredPokemon] [INFO] [lured_pokemon_found] Lured pokemon at fort Alexander The Great Rock (0c2581656bd848ba83dadd7b16ea5b86.16)
2016-08-09 18:09:50,194 [CatchLuredPokemon] [INFO] [lured_pokemon_found] Lured pokemon at fort Wooden Bridge (f2dae5c651c0411c923a6e455b6fa2ad.16)
2016-08-09 18:09:50,883 [CatchLuredPokemon] [INFO] [lured_pokemon_found] Lured pokemon at fort Alexander The Great Rock (0c2581656bd848ba83dadd7b16ea5b86.16)
2016-08-09 18:09:50,884 [CatchLuredPokemon] [INFO] [lured_pokemon_found] Lured pokemon at fort Wooden Bridge (f2dae5c651c0411c923a6e455b6fa2ad.16)
[...]

Issue #3287, opened 2016-08-09
Arrives at the pokemon but continues without catch it. 


muratcancicek:
Please check configuration at http://jsonlint.com/ before posting an issue.

### Expected Behavior

On the map config, moving and arriving at the pokemon and catch it.

### Actual Behavior

 arriving at the pokemon but, doesn't catch it, just move another target. My bag and pokeballs are okay and the both catches other pokemons regularly.  

### Steps to Reproduce

catch the the pokemon

### Other Information
OS: 
Git Commit: (run 'git log -n 1 --pretty=format:"%H"' and paste it here)
Python Version: (run 'python -V' and paste it here)

muratcancicek:
Enabling snipe temporarily solved my problem.

muratcancicek:
@PlanetoftheMapes For it doesn't even say  "encountered X", just arrived at X then it continues moving. 
Is sniping dangerous for banning?

Happyrain67:
Same here.
Walking but never catching.

Happyrain67:
Ok I have fixe it with a good config.

Juste place task "CatchVisiblePokemon" and "CatchLuredPokemon" AFTER "MoveToMapPokemon".
And I have place task "MoveToFort" in last.

This work !
The Bot Walking, Encoutered and catch pokemon.

p1n9:
Hi I had the same issue and I corrected it by reordering the sequence of tasks and by adding the missing parameters from the file [config.json.map.example](https://github.com/PokemonGoF/PokemonGo-Bot/blob/dev/configs/config.json.map.example) (dev version)

Now it works like a charm =)

Issue #3271, opened 2016-08-09
!dev - balls left don't using lower ball when empty


Philipp59:
great balls are empty but pokeballs are available and the bot does not switch
using 59f55fd38f7eb0d2677c650b039b0d9277fceba2
```
2016-08-09 14:07:43,605 [PokemonCatchWorker] [INFO] [pokemon_caught] Captured Nidoran F! [CP 602] [Potential 0.58] [1/12/13] [+210 exp]
2016-08-09 14:07:43,605 [PokemonCatchWorker] [INFO] [gained_candy] You now have 2934 Nidoran F candy!
2016-08-09 14:07:50,646 [PokemonCatchWorker] [INFO] [pokemon_appeared] A wild Pidgey appeared! [CP 336] [Potential 0.56] [A/D/S 7/15/3]
2016-08-09 14:07:54,924 [PokemonCatchWorker] [INFO] [threw_pokeball] Used Greatball, with chance 45.74 (0 left)
2016-08-09 14:08:05,665 [PokemonCatchWorker] [INFO] [pokemon_appeared] A wild Pidgey appeared! [CP 336] [Potential 0.56] [A/D/S 7/15/3]
2016-08-09 14:08:09,589 [PokemonCatchWorker] [INFO] [threw_pokeball] Used Greatball, with chance 45.74 (0 left)
2016-08-09 14:08:16,696 [PokemonCatchWorker] [INFO] [pokemon_appeared] A wild Pidgey appeared! [CP 336] [Potential 0.56] [A/D/S 7/15/3]
2016-08-09 14:08:20,014 [PokemonCatchWorker] [INFO] [threw_pokeball] Used Greatball, with chance 45.74 (0 left)
2016-08-09 14:08:27,335 [PokemonCatchWorker] [INFO] [pokemon_appeared] A wild Pidgey appeared! [CP 336] [Potential 0.56] [A/D/S 7/15/3]
2016-08-09 14:08:31,467 [PokemonCatchWorker] [INFO] [threw_pokeball] Used Greatball, with chance 45.74 (0 left)
2016-08-09 14:08:39,748 [PokemonCatchWorker] [INFO] [pokemon_appeared] A wild Pidgey appeared! [CP 336] [Potential 0.56] [A/D/S 7/15/3]
2016-08-09 14:08:43,640 [PokemonCatchWorker] [INFO] [threw_pokeball] Used Greatball, with chance 45.74 (0 left)
[...]

PR #3332, opened 2016-08-09
If you have no other pokeballs, be wiling to use ultra balls

TheSavior:
Right now I have 0 pokeballs, 0 greatballs, and 60 ultra balls. The bot was refusing to catch anything because it only wanted to use ultraballs on VIPs. If we have no other balls, we should be willing to use ultra balls.

### DIFF
--- a/pokemongo_bot/cell_workers/pokemon_catch_worker.py
+++ b/pokemongo_bot/cell_workers/pokemon_catch_worker.py
@@ -276,7 +276,7 @@ class PokemonCatchWorker(BaseTask):
     def _do_catch(self, pokemon, encounter_id, catch_rate_by_ball, is_vip=False):
         # settings that may be exposed at some point
         berry_id = ITEM_RAZZBERRY
-        maximum_ball = ITEM_ULTRABALL if is_vip else ITEM_GREATBALL
+        maximum_ball = ITEM_ULTRABALL
         ideal_catch_rate_before_throw = 0.9 if is_vip else 0.35
 
         berry_count = self.bot.item_inventory_count(berry_id)


h3k1ru:
Then eventually you will end up with 0 Ultraballs as well. If you came to 0 Pokeballs that's mean your pokestop/pokemon ratio is bad and need improve.

Dwite:
I think this will use ultraball even if there are other pokeballs when Pokemon catch rate is very low, but I'm not 100% sure about it, maybe extend condition?

TheSavior:
The lowest available pokeball is chosen with this logic:

```
current_ball = ITEM_POKEBALL		             current_ball = ITEM_POKEBALL
             while items_stock[current_ball] == 0 and current_ball < maximum_ball:		             while items_stock[current_ball] == 0 and current_ball < maximum_ball:
                 current_ball += 1
```

Dwite:
Ok

leanderjanssen:
Personally I would like this behaviour to be configurable through configuration.
I would like to save my ultraballs for my defined VIPs.

AcorpBG:
I run my bot with this change since the catch was reworked. Yes it will use more ultraballs the best way to deal with this is to be able to set both max_ball_non_vip and ideal_catch_rate_before_throw_non_vip

TheSavior:
The challenge for me is that I don't get very many low level pokeballs from pokestops anymore, so I'm completely out of pokeballs that it is willing to use, but it still tries to catch pokemon and complains that there are no available balls. While it is definitely important to save some to be able to catch VIPs, if it isn't willing to use ultraballs for regular pokemon, the bot essentially stops doing anything.

achretien:
Agreed wit @TheSavior 

```
PokeBalls: 78 | GreatBalls: 0 | UltraBalls: 200
```

Philipp59:
@TheSavior 
```
current_ball = ITEM_POKEBALL
while items_stock[current_ball] == 0 and current_ball < maximum_ball:
  current_ball += 1
```
will this also be a fix for  #3313 ?

AcorpBG:
maximum_ball = ITEM_ULTRABALL
         ideal_catch_rate_before_throw = 0.9 if is_vip else 0.5
I'm running it like this right now after lvl 20 you get alot more great and ultra balls with this config I'm forcing it to use more great/ultra balls and berries and also it is way more effective in catching

h3k1ru:
I think we should be able to specify minimum catch rate for every ball via config. And for berries too. Also berry must be thrown every catch attempt, not only first one, because it boost only one attempt.

TheSavior:
I think the problems you guys are bringing up with configuring pokemon catching is valuable, but I'm not sure it is relevant to this PR. This PR simply allows us to catch pokemon when we still have balls left. Changing the logic of how we pick balls is something we should tackle separately.

DimaVIII:
@TheSavior If you fix this can you please also investigate this two issues https://github.com/PokemonGoF/PokemonGo-Bot/issues/3271 & https://github.com/PokemonGoF/PokemonGo-Bot/issues/3287 they appear after the last changes as there been so many pokeball changes and it's damaging the bot. There are many infinity loops and so on. Thank you!

TheSavior:
@douglascamata Please review

licklake:
there should be an option to reserve (lets say 10 for e.g.) ultraballs for v.i.p pokemon only

TheSavior:
It sounds like there is more logic that needs to go into this to satisfactorily cover everyone's use cases than I am willing to work on for this right now. In the short term I can modify my config to keep more low level pokeballs.

PR #3331, opened 2016-08-09
Improve and update pokemon.json

amal:
## Short Description:

Improve and update pokemon.json

1. Unminify for simplier updates and usage
2. [Add family id to "Next Evolution Requirements"](https://github.com/PokemonGoF/PokemonGo-Bot/pull/2462/commits/c2c6e600b88fe427b0ac0d215c5cf39b5d952ce8) (by @achretien from #2462). It can help when doing maths with candies for evolution.
3. [Add missing Previous evolution(s)](https://github.com/PokemonGoF/PokemonGo-Bot/pull/2462/commits/b01bcec55a26b2b9183488d61e0ada935cf4e0c3) (by @achretien from #2462)
4. Add BaseAttack, BaseDefense, BaseStamina, CaptureRate, FleeRate, Fast/Special Attack(s)

## Explanation:

With this changes we can add calculation of **moveset perfection**, **cp perfection** ([more accurate then simple iv perfection](https://github.com/jabbink/PokemonGoBot/issues/469)) and many others (**best tiers calculation**, **attack/defense dps**, etc..)

It will allow to significantly improve logic in many places like catching, evolving & releasing pokemons.

I'll add mentioned calculations in this next pool request, when this will be merged.

## Additional info:

This pull includes and complements #2462

PR #3336, opened 2016-08-09
Add fast & charged attacks data

amal:
## Short Description:

Add data for fast & charged attacks (originally by @iananass in #2117).
Name, type, damage, duration (in milliseconds), energy (addition for qiuck, cost for charged), dps.

## Explanation:

With this data (and other from #3331) we can calculate **moveset perfection** and other interesting info (**best tiers**, **attack/defense rates**, etc..)

It will allow to significantly improve logic in many places like catching, evolving, releasing & renaming pokemons.

I'll add mentioned calculations in this next pool requests, when this will be merged.
This is a second PR in series :)
Previous: #3331

## Additional info:

This pull includes json data (without logic) from #2117 (Show pokemon moves when catching)


TheSavior:
:+1:

licklake:
I would love to see more logic in the bot. Great work!

amal:
Next PR: #3455

PR #3341, opened 2016-08-09
Fix evolution error in pokemon.json

jebabin:
Fix evolution error in pokemon.json


jebabin:
That the first time I use github and obviously I did not do what I wanted.. I made a change to pokemon.json and want to submit it for change in dev, could someone guide me with the right steps to follow ?

PR #3352, opened 2016-08-09
changes to improve event system based on new web ui devs requests

douglascamata:
- added encounter_id, latitude, longitude and pokemon_id to pokemon caught, vanished and appeared events.

- data structure of events broadcasted by the bots now is the following:

```
{
    "event": event_name,
    "account": bot_account,
    "data": event_parameters
}
```

- position update event being sent at debug level for each step in the `StepWalker`

- `get_player_info` remote command now includes player and inventory info

@brantje: use this PR to keep working on the web ui and we discuss further modifications here.

### DIFF
--- a/brantje.py
+++ b/brantje.py
@@ -1,0 +1,9 @@
+# coding: utf-8
+from socketIO_client import SocketIO
+s = SocketIO('localhost', 4000)
+def echo(msg):
+    print msg
+    
+s.on('get_player_info:d.camata@gmail.com', echo)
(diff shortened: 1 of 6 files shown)


brantje.py
```diff
@@ -0,0 +1,9 @@
+# coding: utf-8
```
TheSavior: Did this mean to be committed?

Jasperrr91:
I can haz merge?

brantje:
:+1:

brantje.py
douglascamata: I'll remove in future commit, sorry :P

PR #3357, opened 2016-08-09
Patch 1

MasterCATZ:
Inventory would fill up from Greatball's


GitPupper:
Or just read the wiki...https://github.com/PokemonGoF/PokemonGo-Bot/wiki/Item-ID's

TheSavior:
Looks like the PR content isn't correct. Feel free to create a new PR if you need.

PR #3391, opened 2016-08-10
Master

jaytothedubs:
j


TheSavior:
This PR doesn't make sense. We don't accept PRs to master, and it doesn't make sense to pull PRs from master either. Please base branches on dev and open them pointed to dev. That way you shouldn't have such large diffs.

PR #3400, opened 2016-08-10
Allow social event exchange anonymous between clients, MQTT broker will hide  users' ip and personal information.

solderzzc:
## Short Description:
Push the kinds of sample code in for discussion or contribution.
Exchange pokemon go information will help everyone to have better experiment.

### DIFF
--- a/pokemongo_bot/event_handlers/social_handler.py
+++ b/pokemongo_bot/event_handlers/social_handler.py
@@ -1,6 +1,4 @@
 # -*- coding: utf-8 -*-
-from __future__ import unicode_literals
-
 from pokemongo_bot.event_manager import EventHandler
 import thread
 import paho.mqtt.client as mqtt
@@ -25,11 +23,12 @@ class MyMQTTClass:
         self._mqttc.publish(channel, message)
     def connect_to_mqtt(self):
         self._mqttc.connect("test.mosca.io", 1883, 60)
-        self._mqttc.subscribe("$GOF/Social/#", 0)
+        self._mqttc.subscribe("$GOF/All/#", 0)
     def run(self):
         self._mqttc.loop_forever()
 class SocialHandler(EventHandler):
-    def __init__(self):
+    def __init__(self, bot):
+        self.bot = bot
         self.mqttc = MyMQTTClass()
         self.mqttc.connect_to_mqtt()
         thread.start_new_thread(self.mqttc.run)
@@ -40,6 +39,9 @@ class SocialHandler(EventHandler):
         else:
             message = '{}: {}'.format(event, str(data))
         if event == 'catchable_pokemon':
-            self.mqttc.publish("$GOF/Social/Catchable", message)
+            self.mqttc.publish("$GOF/All/Catchable", str(data))
+            print data
+            if data['pokemon_id']:
+                self.mqttc.publish("$GOF/PokeID/"+str(data['pokemon_id']), str(data))
             #print 'have catchable_pokemon'
-            #print message
+            print message
--- a/pokemongo_bot/event_handlers/social_handler.py
+++ b/pokemongo_bot/event_handlers/social_handler.py
@@ -1,0 +1,45 @@
+# -*- coding: utf-8 -*-
+from __future__ import unicode_literals
+
+from pokemongo_bot.event_manager import EventHandler
+import thread
+import paho.mqtt.client as mqtt
+class MyMQTTClass:
+    def __init__(self, clientid=None):
+        self._mqttc = mqtt.Client(clientid)
+        self._mqttc.on_message = self.mqtt_on_message
+        #self._mqttc.on_connect = self.mqtt_on_connect
+        #self._mqttc.on_publish = self.mqtt_on_publish
+        #self._mqttc.on_subscribe = self.mqtt_on_subscribe
+    #def mqtt_on_connect(self, mqttc, obj, flags, rc):
+        #print "rc: "+str(rc)
+    def mqtt_on_message(self, mqttc, obj, msg):
+        print msg.topic+" "+str(msg.qos)+" "+str(msg.payload)
+    #def mqtt_on_publish(self, mqttc, obj, mid):
+        #print "mid: "+str(mid)
+    #def mqtt_on_subscribe(self, mqttc, obj, mid, granted_qos):
+    #    print "Subscribed: "+str(mid)+" "+str(granted_qos)
+    #def mqtt_on_log(self, mqttc, obj, level, string):
+    #    print string
+    def publish(self, channel, message):
+        self._mqttc.publish(channel, message)
+    def connect_to_mqtt(self):
+        self._mqttc.connect("test.mosca.io", 1883, 60)
+        self._mqttc.subscribe("$GOF/Social/#", 0)
+    def run(self):
+        self._mqttc.loop_forever()
+class SocialHandler(EventHandler):
+    def __init__(self):
+        self.mqttc = MyMQTTClass()
+        self.mqttc.connect_to_mqtt()
+        thread.start_new_thread(self.mqttc.run)
+    def handle_event(self, event, sender, level, formatted_msg, data):
+        #sender_name = type(sender).__name__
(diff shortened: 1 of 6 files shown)


solderzzc:
the code is using publish test MQTT broker for testing:
http://moxd.io/2015/10/public-mqtt-brokers/

solderzzc:
> MQTT is a machine-to-machine (M2M)/"Internet of Things" connectivity protocol. It was designed as an extremely lightweight publish/subscribe messaging transport. It is useful for connections with remote locations where a small code footprint is required and/or network bandwidth is at a premium. For example, it has been used in sensors communicating to a broker via satellite link, over occasional dial-up connections with healthcare providers, and in a range of home automation and small device scenarios. It is also ideal for mobile applications because of its small size, low power usage, minimised data packets, and efficient distribution of information to one or many receivers
[Official MQTT](http://mqtt.org)

pokemongo_bot/event_handlers/social_handler.py
```diff
@@ -0,0 +1,45 @@
+# -*- coding: utf-8 -*-
+from __future__ import unicode_literals
+
+from pokemongo_bot.event_manager import EventHandler
+import thread
+import paho.mqtt.client as mqtt
+class MyMQTTClass:
```
solderzzc: Reference from:
https://eclipse.googlesource.com/paho/org.eclipse.paho.mqtt.python/+/1.1/examples/sub-class.py

maestroi:
How will this be better for everyone?

TheSavior:
It is a neat idea to have the bot be able to report the seen pokemon to each other. However, I'm not sure this is the right approach to implement it. I imagine the value propositions for something like this are:
1. Crowd source a live map
2. Create a reliable data source for bots to snipe from. 

I don't think a queue is the right approach for this. If clients want to search for specific pokemon to snipe, they wouldn't be able to do so. They would have to process each message that the queue receives. If the central server is getting every pokemon, then the local clients will be spending forever iterating through that queue trying to find the ones they care about. I think a queue will cause any bot using this to hit 100% cpu utilization very quickly with the number of bots we have running.

I think `social` is the wrong name for this and hides what it is actually doing. Since this is reporting data to a central server, it is **extremely** important to get this right.

Instead of a event handler, I think this should be written as tasks. Either one or two tasks

If one task: It would be similar to MoveToMapPokemon, but instead of connecting to an instance of a map, it would connect to a central server that it can use to search for specific pokemon. As part of using this task it could be a good citizen and report back the pokemon visible for the bot.

If two tasks: One task that is written to send all the visible pokemon to the central server, and another to snipe pokemon from the list.

Either way, it seems like this is a great thing that should start as a plugin in a different repo. We don't risk our user's privacy, it is very clear that it is opt in, and we wouldn't have to be so worried about the central server. 

@douglascamata Do you have any other additional thoughts?

maestroi:
@TheSavior aah this explains it way better, than i can see the value. but if its filtered on only certain pokemon for sniping and maybe a max range?
Its better manageable i guess, we could use the socket to send filtered data at first only high value pokemon in cp or iv than you will get a better crowd sourced map i guess.

Thank you for the explanation

pokemongo_bot/event_handlers/social_handler.py
```diff
@@ -0,0 +1,45 @@
...
+    #def mqtt_on_publish(self, mqttc, obj, mid):
+        #print "mid: "+str(mid)
+    #def mqtt_on_subscribe(self, mqttc, obj, mid, granted_qos):
+    #    print "Subscribed: "+str(mid)+" "+str(granted_qos)
+    #def mqtt_on_log(self, mqttc, obj, level, string):
+    #    print string
+    def publish(self, channel, message):
+        self._mqttc.publish(channel, message)
+    def connect_to_mqtt(self):
+        self._mqttc.connect("test.mosca.io", 1883, 60)
```
mmnsgo: this should be in the config file.

solderzzc:
@TheSavior  you are right. 
1. name it correctly 
2. MQTT is a prove of concept, get rid of the server setup at the implementing stage
3. Channel can be setup in MQTT, publish different kinds of pokemon into different channel is simple
4. Finally, need a better voting and anti spam system, I think there's good open source out there. Just need figure it out.

pokemongo_bot/event_handlers/social_handler.py
solderzzc: yep, welcome to contribute on the branch.

douglascamata:
We need to discuss a lot more about how will the server process everything to be able to choose a message broker. 

IMHO, whatever sends data to the server needs to be an async event handler. Why event handler? Because we already have pokemon encounter events. They have pokemon name, id, expiration timestamp, lat and lng (and I probably can include IV and CP), which is all the information we need for a crowdsourced map.

For the message broker, I believe we should use RabbitMQ and JSON as we already do. We enqueue data and have many "workers" in the server dumping them to a database with geospatial capabilities and very fast write speeds, like MongoDB. 

Then we can use a microframework, like Flask or Hug (**hug is awesome**, thx @JSchwerberg), to serve a simple API for snipping and to tell users what place is best to farm which pokemon. 

We can also release a dump of our database every month, for example, to allow anyone to do geospatial statistical analysis in our data. 

What you guys think?

TheSavior:
@douglascamata I totally agree with that. One thing I am very concerned about is that this current implementation doesn't just write the pokemon to a queue, but all of the bots are writing to the same queue and *reading* from the same queue. So every bot has to process all of the sightings from *every* other bot. If we are reporting to some central server, that central server needs to be a bit smarter, just as you are proposing.

I would prefer us to partner with some other existing project that is handling snipe data so that we don't have to build and manage that ourselves. We'd also hopefully avoid having to pay for hosting.

I think releasing the dataset to the community would be really valuable and help promote community goodwill.

douglascamata:
@TheSavior you misunderstood something. We would put found pokemon in a queue (RabbitMQ) and query the API (flask/hug/webserver) for pokemon to snipe. We would be listening for everyone's published encounters. We listen for the encounter that we request. For example, if you want to snipe Dratini, you will tell the server and it won't notify you about anything else (unless you request it).

TheSavior:
Exactly, I'm agree with that implementation. I just wanted to make it clear that that is *not* what this PR is doing

solderzzc:
@TheSavior  The design is same as you discussed, the implementation is in progress, my knowledge is not enough to finish all the pieces. That's why I post this branch and PR for discussing and contribution. Doesn't mean just what done in the code.

JSchwerberg:
+1 on https://github.com/PokemonGoF/PokemonGo-Bot/pull/3400#issuecomment-238953246 

My two cents:

The way I envision it, we would have the individual bots sending their encounter data along with current Lat/Long to either ZeroMQ or RabbitMQ.  There is a consumer -- likely a celery worker -- subscribed to that message queue; it handles the objects and drops the data into Couchbase/Mongo (both of which already do geospatial lookups in their default implementations.

We can use [Marshmallow](https://marshmallow.readthedocs.io/en/latest/) to model out the objects, possibly even implementing it into the bot as well as the consumer and API, to guarantee data consistency across the document store; best part is that if we're using [hug](http://hug.rest) for the API, it has special handling for Marshmallow schemas.

It would be a huge undertaking, but using something like Marshmallow to enforce data consistency -- and to make our objects portable between the bots, the backend servers, the API, etc -- will make this really easy to scale and extend going forward.

TheSavior:
@JSchwerberg that sounds right to me. However, there are many sites right now that already aggregate this data and provide a nice api for fetching live sniping data. One example is this: http://pokesnipers.com/

These websites already have all that infrastructure set up, they are paying for the servers, and they are handling the scaling. I would really like to not need for us to be building that ourselves so we can focus on the bot. If these sites didn't exist and weren't already widely used by other bots as well for sniping data then we'd be in a situation. I don't feel like we would be providing any additional value by reinventing the wheel here.

TheSavior:
Closing this in favor of: https://github.com/PokemonGoF/PokemonGo-Bot/issues/3672

PR #3412, opened 2016-08-10
Return RUNNING if there are more forts to spin

TheSavior:
## Short Description:
Previously, SpinForts only looked at a single fort. It would spin it, and then the task list would continue, causing us to walk, possibly out of range of other forts.

Now when there are more forts to spin, we return RUNNING to cause the tick loop to start over.

This fix should have a relatively large impact if you are in a place with lots of pokestops close together.

### DIFF
--- a/pokemongo_bot/cell_workers/spin_fort.py
+++ b/pokemongo_bot/cell_workers/spin_fort.py
@@ -27,10 +27,12 @@ class SpinFort(BaseTask):
         return self.ignore_item_count or self.bot.has_space_for_loot()
 
     def work(self):
-        fort = self.get_fort_in_range()
+        forts = self.get_forts_in_range()
 
-        if not self.should_run() or fort is None:
+        if not self.should_run() or len(forts) == 0:
             return WorkerResult.SUCCESS
+
+        fort = forts[0]
 
         lat = fort['latitude']
         lng = fort['longitude']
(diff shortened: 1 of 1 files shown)


TheSavior:
@douglascamata Please review

sffetlio:
I've been using it for a few hours and is good so far.

mjmadsen:
Ran this for about 5 hours and seemed to have a negative impact on distance traveled and overall XP/hr. Took the change out and comparing results against current dev branch.

TheSavior:
@mjmadsen This PR essentially only makes the bot spin all of the poke stops around you that it can before walking more. Right now it will only spin 1, and then walk away.

douglascamata:
@TheSavior I review tomorrow morning, too late now.

achretien:
If I understand the code, when returning RUNNING, the bot will redo all the task before the spin task? Instead of redoing the task and continue with the list.

achretien:
Could we keep the id of the still running task at the level of the loop to restart the loop at the same task? And also add a param to the task saying continue for example?

douglascamata:
@achretien when RUNNING is returned the tick is immediately finished and restarted.

douglascamata:
@achretien doing this is the same as a loop inside the task and we don't want loops inside tasks to interact with stuff nearby.

douglascamata:
off topic: @TheSavior my slack app doesn't load... what about yours?

achretien:
Ok so if we have move to fort before spinning fort in the tasks, we could
still move out of the range of all the available stops?

douglascamata:
@achretien yes, of course. That is a simple broken logic in task configuration. You only SpinForts AFTER MoveToFort.

TheSavior:
Is there anything left to do here? Or do you want to merge it?

douglascamata:
👍 

@TheSavior LGTM!

mjmadsen:
@TheSavior Found out it was the number of bots I was running concurrently. My bad!

PR #3413, opened 2016-08-10
Improve printing items

xnning:
## Short Description:

Before:
`Items awarded: {u'Pokeball': 2, u'Revive': 1}`
After:
`Items awarded: {'Pokeball': 2, 'Revive': 1}`


TheSavior:
I think this was an object on purpose since it is getting emitted as an event. @douglascamata do you have any insight here?

douglascamata:
@TheSavior @xnning well, imho, this is not necessary. It's clearly the same thing. What you think, @TheSavior?

TheSavior:
I think I'd prefer leaving it as an object so that it can be sent along in events that things can use.

PR #3417, opened 2016-08-10
Give the possibility to disable a task

achretien:
We can now disable a task from the config file.

Add `enabled` true by default in base task class

### DIFF
--- a/configs/config.json.example
+++ b/configs/config.json.example
@@ -7,35 +7,30 @@
     "libencrypt_location": "",
     "tasks": [
       {
-        "type": "HandleSoftBan",
-        "config": {
-          "enabled": true
-        }
+        "type": "HandleSoftBan"
       },
       {
-        "type": "CollectLevelUpReward",
-        "config": {
-          "enabled": true
-        }
+        "type": "CollectLevelUpReward"
       },
       {
         "type": "IncubateEggs",
         "config": {
-          "enabled": true,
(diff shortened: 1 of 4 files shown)


configs/config.json.example
```diff
@@ -7,23 +7,35 @@
     "libencrypt_location": "",
     "tasks": [
       {
-        "type": "HandleSoftBan"
+        "type": "HandleSoftBan",
+        "config": {
```
TheSavior: I don't think we should add these to the examples. It is clutter and people will think it is required. Maybe add some other task and say enabled false, just to give an example.

pokemongo_bot/tree_config_builder.py
```diff
@@ -61,7 +61,8 @@ def build(self):
                 )
 
             instance = worker(self.bot, task_config)
-            workers.append(instance)
+            if instance.enabled:
```
TheSavior: Please add a unittest for this case.

achretien: done!

configs/config.json.example
achretien: I added the NicknamePokemon task with enabled:false and removed the config from other tasks

TheSavior:
This is great, thanks!

:+1:

vicch:
With this feature all `.json.*.example` can be combined into one.

PR #3438, opened 2016-08-10
Small PEP8 changes

mmnsgo:
Removed default values for 
class PokemonGoBot - update_web_location
class BaseTask - emit_event
class IncubateEggs - _check_inventory
class EventManager - register_event - emit

used autopep8 to fix styling issues
```
$ autopep8 --in-place --aggressive --aggressive -r --exclude "lib,bin,share,docs,include" .
```


vicch:
"Small" changes, `Files changed 54` `+1,460 -815` :smile:

TheSavior:
@mmnsgo It is a bit infeasible for us to be able to merge a PR like this. It would essentially require us to stop development to get it in otherwise it will constantly have conflicts. 

I love these changes and would be super happy getting the code to have a more consistent style and being able to enforce it. I think what might be a more reasonable approach would be to fix a couple files at a time, and change `pylint-recursive.py` to lint those files for all errors. That way we can make sure once we fix these files, nobody will be able to merge PRs if they break the code style in those files. We'd also be able to clean up the repo iteratively instead of freezing it while we make these changes.

How do you feel about that?

Unfortunately, do to the size of this PR and the merge conflicts, I need to close this one.

mmnsgo:
@TheSavior I understand, thank you for the suggestion. 
I took a look at the repo history, ~40 files out the ~60 I changed have been committed recently.
```
git diff --stat @{10.days.ago}
```

would you consider a pr that includes only files that didn't change recently?

TheSavior:
I don't think I'd be excited about a PR that was changing the formatting of any file unless it went all the way and we updated the linter that runs with travis to enforce that we don't break the style in the future.

But yeah, doing that for a bunch of files that aren't typically modified at the same time seems reasonable.

PR #3440, opened 2016-08-10
Don't show Inventory full event if we set "ignore_item_count"

achretien:
Don't show Inventory full event log if we set "ignore_item_count"


TheSavior:
:+1:

PR #3455, opened 2016-08-10
Better inventory:  attacks & movesets,  IV CP perfection,  pokemon level,  etc..

amal:
## Short Description:

Added information about attacks and movesets to the new inventory system.
Also added more metrics and info for pokemons.

## Explanation:

1. New static helpers added:  **FastAttacks**,  **ChargedAttacks**,  **LevelToCPm**
2. New instance classes added:
   - **Attack**, **ChargedAttack**: Information about each attack<br>
     id, name, type, duration, energy, dps<br>
     damage_with_stab, dps_with_stab (STAB - Same-type attack bonus)<br>
     Also calculated perfection rate of the attack in it's type
   - **Moveset**: Information about specific moveset of the pokemon<br>
    fast_attack, charged_attack, raw dps, dps for attack, dps for defense<br>
    Also calculated moveset perfection rate for attack and for defense (between all movesets of the specific pokemon)
3. Additional information for pokemons:
   - pokemon types list
   - lists of fast/charged attacks
   - list of movesets
   - maximum possible CP
   - current pokemon level
   - nickname, hp, max hp
   - basic values (Attack, Defense, Stamina) fixed for each pokemon
   - internal values (IVs)
   - current fast/charged attacks + current moveset
   - **IV CP perfection** - kind of IV perfection percent but calculated using weight of each IV in CP of the best final evolution of current pokemon.<br>
     Attack plays an important role in CP, and different pokemons have different base value, that's means 15/14/15 is better than 14/15/15 for lot of pokemons, and if one pokemon's base def is more than base sta, 15/15/14 is better than 15/14/15.<br>
     So IV CP perfection tends to be more accurate than simple IV perfection.
   - calculated exact value of current CP (not rounded)
   - percent of maximum possible CP
   - family_id, first/prev evolution ID, next/last evolution IDs

With this additions we can also calculate **best tiers**, **attack/defense** ratings, etc..

All this allows to significantly improve logic in many places like catching, evolving, releasing & renaming pokemons.

This is a third PR in series :)
Previous: #3336, #3331


mmnsgo:
is there any config file change to consider to tweak/tune/use these changes?

amal:
@mmnsgo Nope.  This PR contains only data & metrics.  Tasks/workers should be changed later to use new data.

solderzzc:
@TheSavior  @douglascamata  how do you think about ?

data/level_to_cpm.json
```diff
@@ -0,0 +1,81 @@
+{
+  "1":    0.094,
```
TheSavior: What is CPM?

TheSavior: Nvm. CP Multiplier: http://pokemongo.gamepress.gg/cp-multiplier

TheSavior:
This is amazing. Do you feel like it might make sense to add some tests for some of this logic? If this is going to be as core to things as we all want it to be, having some tests to verify that this works as expected and doesn't end up breaking down the line would probably be valuable. I imagine if you stopped working on this and we found a problem that looks like it is related to this code, we would have no idea what the right fix is. Some tests would solve that problem.

Totally up to you though, this PR looks great and I'd happily merge it as is. I'd just be even happier with tests. :)

@douglascamata Can you take a look at this please?

data/level_to_cpm.json
earthchie: Please don't use these CPM. Better use cp_multiplier and additional_cp_multiplier given from APIs instead.

TotalCPM = cp_multiplier + additional_cp_multiplier

amal: @earthchie cp_multiplier and additional_cp_multiplier are used when available (for pokemons).
this data is used to get CPM at specific level or get level of pokemon by it's CPM

earthchie: Then we need to do some test whether these values still accurate or not. 
Niantic (or pgoapi) added cp_multiplier on Jul 29, and additional_cp_multiplier on Aug 9. 
If it really from Niantic. I believe they've nerf/buff some Pokemons and these static CPM gonna be useless.

amal:
@TheSavior I added many many comments with notes links for sources of formulas/calculations everywhere on the code to make everything clearer.  All code is tested during the day (for correctness) on few of my accounts with different pokemons.

Also I added some asserts in the code for cases like unknown attack or moveset – it will show if json data should be improved.

May be I can add more asserts in some places. I don't know what exactly to cover with tests here ))
Probably I can add some test cases with expected results of calculations. We need something like that?

Also I am relatively new in Python and have not yet written any unit tests in it :)
Can you point to some example of how to format tests? Especially helpfull would be if example is in this project.

TheSavior:
@amal Having runtime asserts is definitely better than nothing. I haven't gone through this code close enough to know exactly what would be valuable to test, I think you are the best person to know that here. 

If you are looking for examples, there are a bunch of files in the repo currently that have filenames that end in _test.

amal:
@TheSavior Thanks! Started writing tests. Hope to finish for an hour or so.

pokemongo_bot/inventory.py
```diff
@@ -158,52 +454,263 @@ def has_next_evolution(self):
...
+    def __str__(self):
+        return self.name
+
+    def __repr__(self):
+        return self.name
 
     def can_evolve_now(self):
-        return self.has_next_evolution() and self.candy_quantity >= self.evolution_cost
+        return self.has_next_evolution() and \
+               self.candy_quantity > self.evolution_cost
```
achretien: Should be self.candy_quantity >= self.evolution_cost

pokemongo_bot/inventory.py
```diff
@@ -158,52 +454,263 @@ def has_next_evolution(self):
...
+        # Individial Values of the current pokemon (different for each pokemon)
+        self.iv_attack = data.get('individual_attack', 0)
+        self.iv_defense = data.get('individual_defense', 0)
+        self.iv_stamina = data.get('individual_stamina', 0)
+
         self._static_data = Pokemons.data_for(self.pokemon_id)
         self.name = Pokemons.name_for(self.pokemon_id)
-        self.iv = self._compute_iv()
-        self.in_fort = 'deployed_fort_id' in data
-        self.is_favorite = data.get('favorite', 0) is 1
```
achretien: This variable is used in TransferTask

pokemongo_bot/inventory.py
```diff
@@ -158,52 +454,263 @@ def has_next_evolution(self):
...
+
+        # Individial Values of the current pokemon (different for each pokemon)
+        self.iv_attack = data.get('individual_attack', 0)
+        self.iv_defense = data.get('individual_defense', 0)
+        self.iv_stamina = data.get('individual_stamina', 0)
+
         self._static_data = Pokemons.data_for(self.pokemon_id)
         self.name = Pokemons.name_for(self.pokemon_id)
-        self.iv = self._compute_iv()
-        self.in_fort = 'deployed_fort_id' in data
```
achretien: Also used in TransferTask

pokemongo_bot/inventory.py
```diff
@@ -249,11 +790,21 @@ def candies(refresh=False):
     return _inventory.candy
 
 
-def pokemons(refresh=False):
-    if refresh:
-        refresh_inventory()
+def pokemons():
```
achretien: refresh in used in TransferTask

achretien:
Just added some comments, I think you broke compatibility with the TransferTask.

pokemongo_bot/inventory.py
```diff
@@ -219,16 +726,50 @@ def refresh(self):
...
         self.bot.latest_inventory = None
-        inventory = self.bot.get_inventory()['responses']['GET_INVENTORY']['inventory_delta']['inventory_items']
+        inventory = self.bot.get_inventory()['responses']['GET_INVENTORY'][
+            'inventory_delta']['inventory_items']
         for i in (self.pokedex, self.candy, self.items, self.pokemons):
             i.refresh(inventory)
 
-        user_web_inventory = os.path.join(_base_dir, 'web', 'inventory-%s.json' % (self.bot.config.username))
-        with open(user_web_inventory, 'w') as outfile:
-            json.dump(inventory, outfile)
```
achretien: This part is used to update the current web ui.

amal:
@achretien Many thanks for attention! Indeed inadvertently made some incorrect edits. Now fixed and everything should be fine.

amal:
@TheSavior Tests added. It was a really good decision to do it :)

Found and fixed some [minor inconsistency in data](https://github.com/PokemonGoF/PokemonGo-Bot/pull/3455/commits/ff1ae0680a6e6b331d1eaf7d91735be7c87f0aae) + few logic caveats.

Also returned inadvertently deleted pieces of code (thanks to @achretien)

douglascamata:
@amal this PR is awesome! I'll review with more time and patience tomorrow because it's a bit late already and I was busy.

amal:
@douglascamata thank you!

TheSavior:
I'm going to merge this. It seems like we aren't using this very heavily yet so the risk if an issue occurs is pretty low. And I have confidence that if something does come up right away that you will be able to handle it @amal.

Thanks again for contributing. I look forward to see how far these improvements go.

rajulbhatnagar:
> Exception: Unexpected moveset [Confusion, Signal Beam] for `#48` Venonat
is what I get after this. Issue  #3548

bixuanzju:
@rajulbhatnagar Seems the Special Attacks for Venonat is wrong, should be

```diff
modified   data/pokemon.json
@@ -1993,10 +1993,9 @@
       }
     ],
     "Special Attack(s)": [
-      "Dazzling Gleam",
+      "Signal Beam",
       "Psybeam",
-      "Poison Fang",
-      "Shadow Ball"
+      "Poison Fang"
     ],
     "BaseAttack": 108,
     "BaseDefense": 120,
```

amal:
@rajulbhatnagar, @bixuanzju problem is fixed with #3550

PR #3467, opened 2016-08-10
UpdateTitleStats -> UpdateLiveStats, new stat, refactoring

Moonlight-Angel:
## Short Description:
After https://github.com/PokemonGoF/PokemonGo-Bot/pull/3312 was merged, the worker's name wasn't appropriate anymore.
He's now called `UpdateLiveStats`.
I also made some refactoring, unit tests and added a new stat to display (`captures_per_hour`, as requested in https://github.com/PokemonGoF/PokemonGo-Bot/pull/3312#issuecomment-238723561).

## Changelog
* Renamed UpdateTitleStats to UpdateLiveStats
* Cleaned worker documentation
* Added documentation for terminal_log and terminal_title
* Fixed https://github.com/PokemonGoF/PokemonGo-Bot/pull/3312#issuecomment-238672978
* Made some refactoring
* Added captures_per_hour stat that shows estimated pokemon captures per hour
* Added a captures_per_hour method in `metrics.py`
* Added unit tests for features added in https://github.com/PokemonGoF/PokemonGo-Bot/pull/3312
* Added unit tests for captures_per_hour

## Side notes/thoughts
* Updating the worker's name involves updating the configuration for a lot of people, but on the other hand, it's probably better than having a bad-named worker that does more than what it claims.
* This PR does a lot of things. If you want me to split it, just ask. 👍

### DIFF
--- a/pokemongo_bot/cell_workers/update_live_stats.py
+++ b/pokemongo_bot/cell_workers/update_live_stats.py
@@ -26,12 +26,12 @@ class UpdateLiveStats(BaseTask):
     }
 
     min_interval : The minimum interval at which the stats are displayed,
-                   in seconds (defaults to 10 seconds).
+                   in seconds (defaults to 120 seconds).
                    The update interval cannot be accurate as workers run synchronously.
     stats : An array of stats to display and their display order (implicitly),
-            see available stats below.
-    terminal_log : Logs the stats into the terminal.
-    terminal_title : Displays the stats into the terminal title.
+            see available stats below (defaults to []).
+    terminal_log : Logs the stats into the terminal (defaults to false).
+    terminal_title : Displays the stats into the terminal title (defaults to true).
 
     Available stats :
     - login : The account login (from the credentials).
--- a/pokemongo_bot/cell_workers/update_live_stats.py
+++ b/pokemongo_bot/cell_workers/update_live_stats.py
@@ -155,7 +155,7 @@ class UpdateLiveStats(BaseTask):
             stdout.write("\033]0;{}\007".format(title))
             stdout.flush()
         elif platform == "win32":
-            ctypes.windll.kernel32.SetConsoleTitleA(title)
+            ctypes.windll.kernel32.SetConsoleTitleA(title.encode())
         else:
             raise RuntimeError("unsupported platform '{}'".format(platform))
         self._compute_next_update()
--- a/pokemongo_bot/cell_workers/update_live_stats.py
+++ b/pokemongo_bot/cell_workers/update_live_stats.py
@@ -6,27 +6,17 @@ from pokemongo_bot.base_task import BaseTask
 from pokemongo_bot.worker_result import WorkerResult
 from pokemongo_bot.tree_config_builder import ConfigException
 
-class UpdateTitleStats(BaseTask):
+
+class UpdateLiveStats(BaseTask):
     """
-    Periodically updates the terminal title to display stats about the bot.
+    Periodically displays stats about the bot in the terminal and/or in its title.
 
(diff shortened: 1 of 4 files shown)


mmnsgo:
can you stop the task (skip tick) if both terminal_log and terminal_title are false?
it would let everyone keep the task in their config without slowing down the bot when not required.

Moonlight-Angel:
@mmnsgo Fair point. Done. 👍

sergiopalacio:
@Moonlight-Angel Really nice mate, you're right about naming and it was something I was worried about. On my repo I currently removed the event log on update title method. Could you add the default values on terminal_log and terminal_title docs?

Moonlight-Angel:
@sergiopalacio Done. 👍

mmnsgo:
@sergiopalacio maybe something like this would create even less confusion?
```
            "log_stats": true,
            "update_cli": true
```

Moonlight-Angel:
Not sure @mmnsgo. CLI is a more technical term.

sergiopalacio:
@mmnsgo @Moonlight-Angel IMHO Since he is refactoring class name it could be 
`"update_title": true, 
"log_on_terminal: false
`

mmnsgo:
@Moonlight-Angel, true, what about:
```
"display_stats": true,
"log_stats: true,
```

Moonlight-Angel:
My vote goes in favor of @sergiopalacio's proposition. Looks to be more verbose.
@mmnsgo What does "display" mean ? Logging the stats into the console is some kind of display too.

mmnsgo:
I imagine a situation where there're 2 groups of people

group one: looks at logs for debugging, etc - they should enable log_stats
group two: looks at one or more interface, they're using (or developing) a task that reads the event update_title and behaves accordingly (in your case, is the task itself) - they should enable display_stats regardless the medium (cli, webpage, screen, watch, etc)

if display_stats is enabled, log_stats is true by default.
if log_stats is false, display_stats is ignored and considered false.

pokemongo_bot/cell_workers/update_live_stats.py
```diff
@@ -56,16 +56,9 @@ class UpdateTitleStats(BaseTask):
...
-    min_interval : The minimum interval at which the title is updated,
-                   in seconds (defaults to 10 seconds).
-                   The update interval cannot be accurate as workers run synchronously.
-    stats : An array of stats to display and their display order (implicitly),
-            see available stats above.
     """
     SUPPORTED_TASK_API_VERSION = 1
 
-
     def __init__(self, bot, config):
```
TheSavior: Tasks shouldn't be implementing __init__. BaseTasks's __init__ calls `initialize` which tasks should implement to do most of this logic. You should be able to check out other tasks to see how they do it.

Moonlight-Angel:
@TheSavior The reason I'm using init is to comply with the linter saying the properties aren't initialized in the `__init__` function.
If you really want me to do the initialization inside the `initialize` method, I'll do it.

pokemongo_bot/cell_workers/update_live_stats.py
Moonlight-Angel: @TheSavior The reason I'm using init is to comply with the linter saying the properties aren't initialized in the `__init__` function.
If you really want me to do the initialization inside the `initialize` method, I'll do it.

JayMaree:
How can I download these changes into my current PokemonGo folder on my server?

TheSavior:
:+1:

[Approved with PullApprove](https://pullapprove.com/PokemonGoF/PokemonGo-Bot/pull-request/3467/?utm_source=github-pr&utm_medium=comment-badge&utm_campaign=PokemonGoF/PokemonGo-Bot)

PR #3487, opened 2016-08-10
remove unused code in evolve_pokemon worker

leadboots5:
## Short Description:
Previously IV was computed in each worker. Now its fetched from inventory. This was left over and not called in the worker at all.

NOTHING!


TheSavior:
:+1:

PR #3496, opened 2016-08-10
Modify pokemon_catch_worker.py to use Inventory class and fix #3411

### LINKED ISSUE #3411: Berries and balls count not always updated
3 berry throws and always "have 2". Last one with error.
screenshot_2016-08-10-14-45-58

One more haopening
screenshot_2016-08-10-14-58-19

Same thing happens with Ultraballs. Probably, not only with Ultra- ones.
screenshot_2016-08-10-16-42-16

cmezh:
## Short Description:
Modify pokemon_catch_worker.py to use Inventory class and fix #3411 

Fix #3411


pokemongo_bot/cell_workers/pokemon_catch_worker.py
```diff
@@ -283,11 +283,12 @@ def _do_catch(self, pokemon, encounter_id, catch_rate_by_ball, is_vip=False):
...
         ideal_catch_rate_before_throw = 0.9 if is_vip else 0.35
 
-        berry_count = self.bot.item_inventory_count(berry_id)
-        items_stock = self.bot.current_inventory()
-
         while True:
 
+            self.bot.latest_inventory = None
+            berry_count = self.bot.item_inventory_count(berry_id)
+            items_stock = self.bot.current_inventory()
```
Anakin5: It would be better to modify the item count when spinning fort or using the item. And then only refresh when we need to recycle.

cmezh: Do yiu mean that it this case we should manually modify remaining balls/berries amount? Because we need to know exact amount before pokemon catch/retry.

cmezh: BTW, what if no SpinFort task specifed in config? For bot just walking and catching.

Anakin5: If every tasks update inventory when they remove or add an item, then you only need to request inventory at start.
Let's change the Spin fort to update the inventory.

cmezh: It's not what I'm talking about.
Look: we have task CatchVisiblePokemon, it will call PokemonCatchWorker 3 times if we have 3 pokemons to catch. There are no SpinFort between these calls, and even if we use SpinFort for updating inventory, all 3 catches will be done (or fail) with the completely same inventory.
Also if we use not PokemonCatchWorker for inventory updating, but CatchVisiblePokemon, we will get non-updated inventory between catch tries.

Anakin5: You have the initial amount of berries when the bot start. Remove one in inventory after using it. Add the new ones you collect.
Whatever is the number of Pokemons you plan to catch, you account of berries will be up to date.

cmezh: It doens't look correct if we manually change inventory data.

TheSavior:
We definitely don't want to cause the api to refetch all of the inventory each time we need to throw a pokeball. 

Perhaps @achretien or @amal have a better idea of how inventory should be handled here.

cmezh:
@TheSavior We have two possibilities: either refetch inventory or manually modify self.bot.latest_inventory instead of modifying local variables with amounts. If the second way is good enough, I will do it this way.

achretien:
We are moving from self.bot.latest_inventory to the global Inventory class. We used it in the Transfer task and the evolve task.

The goal is to move everything related to the inventory to this class

cmezh:
@achretien Ok, I will take a look at it.

cmezh:
I hope now it's ok.

achretien:
It should be good as is. In a next PR I will remove the Pokemon class to use the one in the inventory. Then we will be able to add the caught Pokemon to the inventory without refreshing it.

cmezh:
After latest commit tsted for 30 min, caught 29 pokemons without any troubles.

pokemongo_bot/inventory.py
```diff
@@ -108,7 +108,13 @@ class Items(_BaseInventoryComponent):
     STATIC_DATA_FILE = os.path.join(_base_dir, 'data', 'items.json')
 
     def count_for(self, item_id):
-        return self._data[item_id]['count']
+        try:
+            return self._data[item_id]['count']
+        except KeyError:
+            return 0
+
+    def decrement_count(self, item_id):
```
TheSavior: Is it worth adding a test for this to prove that we can increment, decrement, and call count_for and get results that we expect?

TheSavior: Would we ever want to decrement by more than one? Like when we transfer a bunch of items? If so, perhaps an argument for amount that defaults to 1?

BriceSD: With #2482 I added methods to increment/decrement. Also it’s not using raw data from api anymore.
Please update your fork and take a look at it

cmezh:
Thank you for your great work, @BriceSD! I've changed my code according to your commit.
Sorry for a lot of commits, I'm new to Git...

TheSavior:
@BriceSD, when you give the thumbs up, @douglascamata or I can merge.

BriceSD:
:+1:

douglascamata:
👍

cmezh:
Thank you!

douglascamata:
@cmezh thank you for the contribution 👍

PR #3524, opened 2016-08-10
Improved items display for loot on level up and pokestop

Jamezin:
# Short Description:
Improved item display when either leveling up or when collecting loot from pokestops.

pokestop finished
pokestop2 finished
levelup finished
levelup2 finished

#3413


leadboots5:
I think we used to have multiple lines and it was changed to using a single line so it takes up less space. Trying to go full circle

brantje:
You emit the same event twice, not needed, as this will make things listening to the socked the pokestop is spinned twice in a very short time, this can be handled by the (color) Logger.
So a :-1:  from me.

Jasperrr91:
What @brantje says. Gives more trouble than it does good.

Issue #3387, opened 2016-08-10
[Suggestion] Save Player State/Information In File


pacpac42:
Hi, would it be possible to save the player state/information in a json/text file in the data folder? 
The state information should include list of pokemons (most important) with their IVs, and everything that is currently printed at startup. This will also remove the need to print that at startup, and will allow all of us to inspect our lists of pokemons and other information without having to run the client. Thank you. 

E.g.
```
Pokemon      CP        IV          A        D        S
Psyduck      515       0.16        10       2        3
Pidgey       189       0.36        0        13       2
Staryu       201       0.38        7        2        15
...

Item       #
Pokeball   5
...
```

Ph4i1ur3:
it kind of already is in the /web folder. Would be nicer to have it organized as you demonstrated. For the time being the web server provides the information you are asking for, and can be run separately from the bot

pacpac42:
Ah I missed that. I agree that it would be great if all that information could be easily exported, but this does not seem to be very necessary at this point in time. Thanks for explaining.

omeryagmurlu:
Made a plugin
https://github.com/omeryagmurlu/PokemonGo-Bot-exportfile-plugin

pacpac42:
Awesome and thank you! It looks like there are also online tools to convert json to csv easily so i'll close this ticket.

PR #3529, opened 2016-08-10
Export to file

omeryagmurlu:
## Short Description:
Adds a worker to export some useful information, currently only csv format.

## Fixes:
- https://github.com/PokemonGoF/PokemonGo-Bot/issues/3387

### DIFF
--- a/configs/config.json.export.example
+++ b/configs/config.json.export.example
@@ -15,7 +15,8 @@
       {
         "type": "ExportFile",
         "config": {
-          "file_type": "csv"
+          "file_type": "csv",
+          "relative_path": "export.csv"
         }
       },
       {
--- a/pokemongo_bot/__init__.py
+++ b/pokemongo_bot/__init__.py
@@ -461,10 +461,7 @@ class PokemonGoBot(object):
         )
 
         # Export File
-        self.event_manager.register_event(
-            'file_exported',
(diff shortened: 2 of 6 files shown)


Nihisil:
@TheSavior what do you think. Should this PR be merged to the dev, or maybe better to keep this feature as separate plugin?

TheSavior:
I think this is a great example of something that should be a plugin. At least to start. If we find this is very popular and wants to be a built in task, we can reconsider later.

@omeryagmurlu Feel free to follow the documentation for creating plugins. https://github.com/PokemonGoF/PokemonGo-Bot/blob/master/docs/plugins.md

omeryagmurlu:
Ok, will make this a plugin.

Is there a way for plugins to have events? I registered the event in PokemonGoBot class, which I can't do with a plugin.

TheSavior:
What do you need to emit that event? Is it just to write to the console?

@douglascamata Do you have any thoughts on how plugins should handle this? I imagine theoretically at least, tasks could register there events in initialize using:

```
self.bot.event_manager.register_event('file_exported')
```

douglascamata:
This task and this plugin aren't really necessary. One can enable websocket and remote control and create a simple script that dumps the necessary data every now and then.

PR #3531, opened 2016-08-10
Ensure recycling happens if bag is over capacity.

brdude:
## Short Description:
Ensures you that item Recycling happens if you have more items than the total bag capacity.

When you level up, you are awarded items which can cause the bag to be over the capacity.


abusalman:
+1

solderzzc:
Yes, we need this merged.

PR #3542, opened 2016-08-11
run.bat for windows

sia84:
## Short Description:
Adding run.bat for windows with a persistent loop to restart pokecli in case it dies


GitPupper:
Here's the same thing for macOS and Linux.
[PokeBot.zip](https://github.com/PokemonGoF/PokemonGo-Bot/files/412409/PokeBot.zip)

PR #3556, opened 2016-08-11
Cache recent forts (for forts.max_circle_size)

Quantra:
## Short Description:

added bool option to cache recent forts -crf --forts.cache_recent_forts (default true)

saves recent_forts in data/recent-forts-{username}.json on spin

loads recent_forts from same file on start up

bot doesn't start a new recent_forts on every reset

-Bot forgets recent_forts upon reset
-
-


douglascamata:
Can you have all the events level in this PR to `debug`? I don't think any of them looks like information a normal user would care about. Also please fix merge conflicts. Thanks!

Quantra:
@douglascamata updated as per your comment =]

douglascamata:
@Quantra can you try to think of a good way of not writing to a file all the time? I don't think this is a good thing to do :/

Quantra:
@douglascamata off the top of my head I can only come up with dumb time based writes but I like the current precision.  After a reset I don't try and spin the same fort again.  I will think on it though.

Could do something like https://github.com/PokemonGoF/PokemonGo-Bot/commit/a5e91315ed82581a901fbf5019a3708e5e7a471b and pass it out to another thread if that would help?  But I see that code isn't being used anymore?

Quantra:
@douglascamata I figured something like this might would allow caching on exit/crash.

pokecli.py
```
def main():
    bot = False
  
    try:
        #  run the bot
       .....
       ......
    except GeocoderQuotaExceeded:
        raise Exception("Google Maps API key over requests limit.")
    except Exception as e:
        # always report session summary and then raise exception
        if bot:
            report_summary(bot)
        raise

    finally:
        if bot:
           # CACHE FORTS HERE
```

But I can only kill via cron so never reach the finally statement and in testing I accidentally ran 2 bots at the same time and now have some kind of ban :(

No more testing for me at least today but I leave this idea here.  I can at least confirm the finally statement is reached when the bot is closed with ctrl-c

Quantra:
@douglascamata I have changed the behaviour such that the caching happens if the bot is stopped by `KeyboardInterrupt` SIGTERM or an Exception.

The SIGTERM handling will be unreliable when using multiple threads i.e. the recently commented out update_web_location_worker.  Are there any plans to bring this back or other multi threading?

douglascamata:
@Quantra no, sir. We're migrating to a web ui with websockets communication. Writting location files for the web ui won't be necessary.

douglascamata:
@Quantra looks good. Did you test it?

Quantra:
@douglascamata yes i've tested with SIGTERM, ctrl-c and causing an error and the forts cache works as expected.

Have been running both with and without fort_caching with no problems.

douglascamata:
👍

Quantra:
@douglascamata after some more research I found pgoapi uses subprocess.  So SIGINT isn't a reliable way to close the bot running with python2.7.

If SIGINT is sent whilst pgoapi has a thread open that thread takes the SIGINT and pokecli.py will keep running.  So we can't guarantee the caching will happen here.

It's not critical for caching forts but I don't advise caching last-location or anything important like this.

For myself restarting bots with CRON I send SIGINT, wait a bit then send SIGKILL to ensure I don't have overlapping bots.  ~99% of the time forts will be cached properly.

PR #3572, opened 2016-08-11
Adding support for disabling sleeps intended for app delay parity

TheSavior:
We believe the likelihood that Niantic will detect and ban botters will be via the length of time between specific requests is very low.

We have lots of sleeps in the bot that make things run much slower. Some people want the added delay to feel more safe with the bot, others want this bot to crank out as much xp/hr as possible. 

This PR adds a flag that will enable or disable all sleeps intended for app delay parity. This PR disables those delays be default.

### DIFF
--- a/configs/config.json.example
+++ b/configs/config.json.example
@@ -93,6 +93,7 @@
     "location_cache": true,
     "distance_unit": "km",
     "reconnecting_timeout": 15,
+    "enable_app_emulation_delay": false,
     "evolve_captured": "NONE",
     "catch_randomize_reticle_factor": 1.0,
     "catch_randomize_spin_factor": 1.0,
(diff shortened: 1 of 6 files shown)


TheSavior:
@douglascamata Please take a look at this. We will wrap more sleeps in this flag in a follow up PR. I wanted to keep this PR from being too large.

Nihisil:
Maybe we need to change all `time.sleep()` to the `human_behaviour.sleep()` and inside the last function we can process `enable_app_emulation_delay` flag.

TheSavior:
Hmm, that could be a reasonable approach. Is there anywhere we use `human_behavior.sleep` that isn't intended to emulate app delays?

douglascamata:
I agree with @Nihisil. Changing all `time.sleep` calls to our own `sleep` implementation will give us much better control.

TheSavior:
Sounds good, I'll go with that approach

PR #3578, opened 2016-08-11
Quick and possibly dirty fix for the annoying debuging messages in co…

Plotin:
This is just a quick and dirty fix for the annoying debug messages in colored logger. It is in no way intended to add proper logging levels support to the logger, just don't print the debug messages any more ;-)


Plotin:
#3532 seems to do the same...

ryanb93:
Can we close this as #3532 has been chosen as the fix for this issue.

Plotin:
Oki doki

PR #3583, opened 2016-08-11
Improve Docker Image and `docker-compose.yml`

lawliet89:
- Use the non-`onbuild` variant of the python base image to better make use of Docker image caching when rebuilding images
 - Update some volumes in `docker-compose.yml` for pogoweb

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -3,7 +3,8 @@ services:
   bot1-pokego:
     build: .
     volumes:
-     - ./configs/config.json:/usr/src/app/configs/config.json
+     - ./configs:/usr/src/app/configs
+     - ./web:/usr/src/app/web
     stdin_open: true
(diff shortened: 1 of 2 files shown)


docker-compose.yml
```diff
@@ -3,18 +3,19 @@ services:
   bot1-pokego:
     build: .
     volumes:
-     - ./configs/config.json:/usr/src/app/configs/config.json
+     - ./configs:/usr/src/app/configs
+     - ./web:/usr/src/app/web
```
lawliet89: This is needed because when building the image `bot1-pokego`, `/usr/src/app/web` is a volume and anything that is copied into there is ignored.

So we mount `./web` as a volume to include the `pogoweb` files in the container, and also include the json files added in there by the bot.

PR #3587, opened 2016-08-11
Added string format to step_walker

guitared:
## Short Description:
Added string format for location in step_walker.py

### DIFF
--- a/pokemongo_bot/step_walker.py
+++ b/pokemongo_bot/step_walker.py
@@ -59,6 +59,7 @@ class StepWalker(object):
             'position_update',
             sender=self,
             level='debug',
+            formatted='To {current_position} from {last_position}',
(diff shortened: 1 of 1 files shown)


TheSavior:
I don't think this is the right fix. This should only be printed out in debug mode, but it is getting printed even when not in debug mode. @douglascamata how do we fix this?

ryanb93:
@TheSavior I think this is because of the coloured_logging_handler using sys.stdout.write and the 'level' parameter being passed through isn't being used at all.

douglascamata:
The coloured logging handler should be fixed to use the logging module correctly as the normal logging handler uses.

douglascamata:
#3637 will fix this properly by fixing the coloured log handler implementation.

PR #3591, opened 2016-08-11
Trying to set up appveyor for windows based CI

TheSavior:


TheSavior:
I got stuck here: 
```
python pylint-recursive.py

======================================================================
ERROR: test_api_call_throttle_should_fail (tests.api_wrapper_test.TestApiWrapper)
----------------------------------------------------------------------
  File "C:\Python27\lib\site-packages\timeout_decorator\timeout_decorator.py", line 66, in new_function
    old = signal.signal(signal.SIGALRM, handler)
AttributeError: 'module' object has no attribute 'SIGALRM'
======================================================================
ERROR: test_api_call_throttle_should_pass (tests.api_wrapper_test.TestApiWrapper)
```

terbolous:
SIGARLM is not available on windows: http://stackoverflow.com/questions/8420422/python-windows-equivalent-of-sigalrm

TheSavior:
Thanks for that.

@douglascamata @terbolous Any ideas on how we should be changing this part of the code to support running on windows?

DayBr3ak:
I guess you could skip them if the environment is windows. I can't think of something else on a short term

DayBr3ak:
@TheSavior I've coded an alternative for windows, should I do a PR on your appveyor branch ?

douglascamata:
I'd just ignore Windows, imho.

TheSavior:
I think ignoring the tests on windows is an okay solution. @douglascamata surely you aren't saying we shouldn't care about windows for this bot at all, right?

douglascamata:
@TheSavior usually I don't provide support for Windows in stuff I develop because: 

1. it's really a pain in the ass to setup lots of things
2. I don't have a Windows machine to test and don't want to spin up  a VM just for this because of Point 1.

TheSavior:
I agree, I think the community needs to be the one that gets us to be able to support windows. It seems like there are plenty of people who are using windows based on the installation scripts I am seeing getting created. If we get appveyor set up, then we ideally wouldn't have to worry much about things.

douglascamata:
@TheSavior yes yes, we can have appveyor, but I cannot help with it :/

k4n30:
@DayBr3ak & @TheSavior - Not to be pushy, but who's going to finish it? I like the idea of it (running on windows until I can be bothered to push off to my *nix server)

TheSavior:
I've been hoping the auto installers that people have been working on would make this easier. 

@nivong, do you have any idea how I can programmatically install the protoc library and the c++ compiler? Without that, this PR is essentially not possible.

nivong:
Nope not really give me a week to see if I can figure it out. Btw why would got want this what are the pros and cons

TheSavior:
@nivong I assume you are asking about the pros/cons to using appveyor.

Appveyor is essentially Travis but it runs on windows instead of mac / linux. So the big assumption here is that we actually care about windows users and want the bot to work for them. I'm making this assumption based on the effort that has gone into the setup instructions for windows.

Pros:
 - We can feel more comfortable that if appveyor fails, the PR broke windows compatibility, something we never would know without appveyor until some user complains.
 - If we can make the setup scripts include all dependencies, then we can validate the installer for all platforms we support.

Cons:
 - A few years back when I last used Appveyor it was slower than travis.
 - If we don't want to support windows users, Appveyor is not worth the overhead for making sure we work on a platform we don't support.

danielsdian:
@TheSavior regarding installing the protoc library, you can do that using git. Thats how we did it on the installation batch. We also created a script in that batch that downloads and installs git, python and the c++ compiler. It even checks which architecture you are running and downloads the correct software.

danielsdian:
Regarding the pros and cons of supporting Windows, I'd say we do it, since the bot already works with it without much trouble.

k4n30:
@danielsdian this is not bot related, this is testing PRs automatically before they are merged to ensure it passes tests. Right now these tests are done on linux, and this PR is about covering the same tests from a windows perspective - [Background reading](https://docs.travis-ci.com/user/for-beginners)

k4n30:
@TheSavior given up?

TheSavior:
We never figured it out, so it shouldn't have been merged. When it was merged it caused every build for every PR to fail.

k4n30:
there's no sad face option so I had to downvote

Issue #3590, opened 2016-08-11
use curl instead of wget as fallback


jimmywarting:
I tried to run `setup.sh -e` on mac 

```
./setup.sh: line 17: wget: command not found
tar: Error opening archive: Failed to open 'pgoencrypt.tar.gz'
```

wget don't come pre installed on mac, curl dose

simonsmh:
Wget should be installed by default not curl.
Which system do you use?

simonsmh:
OK i see you are on mac.

simonsmh:
Does `curl -O http://pgoapi.com/pgoencrypt.tar.gz` work?

jimmywarting:
> Does curl -O http://pgoapi.com/pgoencrypt.tar.gz work?

yes

nivong:
Thanks for the report, will be fixed in next PR.

simonsmh:
https://github.com/PokemonGoF/PokemonGo-Bot/pull/3592

simonsmh:
I don't have mac so I don't know how shell actually worked on mac. Hope it works or anyone have a better solution.
:(

nivong:
https://github.com/PokemonGoF/PokemonGo-Bot/pull/3593

kestel:
Curl may be better, but curl it's not in base system on FreeBSD for example, I think we need additional checks and OS detection.

simonsmh:
Close plz. #3593

lightyoruichi:
brew install wget

PR #3592, opened 2016-08-11
Update setup.sh to fix for wget: command not found on mac

simonsmh:
A stupid fix for wget: command not found on mac
:(
https://github.com/PokemonGoF/PokemonGo-Bot/issues/3590

### DIFF
--- a/setup.sh
+++ b/setup.sh
@@ -14,7 +14,12 @@ pip install -r requirements.txt
 
 function Pokebotencrypt () {
 echo "Start to make encrypt.so"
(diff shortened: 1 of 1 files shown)


nivong:
this is not how you would do this

jimmywarting:
Guessing you should do feature detection instead of os detection (like we say on web development)

use wget if it exist
else use curl if it exist

nivong:
this is how you would do it:
https://github.com/PokemonGoF/PokemonGo-Bot/pull/3593

Issue #1815, opened 2016-07-30
pgoapi.exceptions.PleaseInstallProtobufVersion3


Cephis:
### Expected Behavior
start pokemonGo-bot

### Actual Behavior
python pokecli.py
Traceback (most recent call last):
  File "pokecli.py", line 38, in <module>
    from pgoapi.exceptions import NotLoggedInException
  File "/home/cephis/PokemonGo-Bot/src/pgoapi/pgoapi/__init__.py", line 17, in <module>
    raise PleaseInstallProtobufVersion3()
pgoapi.exceptions.PleaseInstallProtobufVersion3

### Steps to Reproduce
I install the last releases, made installs as described in the walkthrough and start with python pokecli.py in a prompt. I tried to install protobuf 3 especially but I don't know how to do it.

### Other Information
OS: linux mint rosa
Git Commit: (run 'git log -n 1 --pretty=format:"%H"' and paste it here)

rguedes:
run: `pip install -r requirements.txt`

Cephis:
already done

samjack:
`pip uninstall pgoapi` & `pip install -r requirements.txt`

Cephis:
sudo pip uninstall pgoapi & pip install -r requirements.txt
[1] 4520
Obtaining pgoapi from git+https://github.com/tejado/pgoapi.git@81e786cabf027a1c8fbd1e9a07e1c11aa3d8ee8b#egg=pgoapi (from -r requirements.txt (line 1))
  Updating ./src/pgoapi clone (to 81e786cabf027a1c8fbd1e9a07e1c11aa3d8ee8b)
Can't uninstall 'pgoapi'. No files were found to uninstall.
error: impossible d'ouvrir .git/FETCH_HEAD : Permission non accordée

  Complete output from command /usr/bin/git fetch -q:
  
----------------------------------------
Cleaning up...
Command /usr/bin/git fetch -q failed with error code 255 in /home/cephis/PokemonGo-Bot/src/pgoapi
Storing debug log for failure in /home/cephis/.pip/pip.log
[1]+  Fini                    sudo pip uninstall pgoapi

samjack:
did you try `./install.sh`

Cephis:
yes, i do it, it worked

sgitkene:
maybe you mix python3 with python2?

try
`pip2 install -r requirements.txt`
and change python in run.sh to python2

Cephis:
[...]

PR #3600, opened 2016-08-11
remove Debian python-protobuf dependency

HongxuChen:
## Short Description:

Debian `python-protobuf` dependency removed (install.sh/setup.sh);
docs/installation.md not synced

- during installation of pgoapi, protobuf will automatically be a
dependency (confirmed with `pip show pgoapi`)

- Even for Ubuntu 16.04 LTS, python-protobuf version is
2.6.1-1.3, which is too old
  + confirmed with `apt-cache show python-protobuf`
  + see issue #1815
  (https://github.com/PokemonGoF/PokemonGo-Bot/issues/1815)

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -5,8 +5,6 @@ RUN echo $timezone > /etc/timezone \
     && ln -sfn /usr/share/zoneinfo/$timezone /etc/localtime \
     && dpkg-reconfigure -f noninteractive tzdata
 
-RUN apt-get update \
-    && apt-get install -y python-protobuf
 RUN cd /tmp && wget "http://pgoapi.com/pgoencrypt.tar.gz" \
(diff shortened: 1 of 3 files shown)


HongxuChen:
The problem for the issue before was that after `python-protobuf` (through apt-get/apt/aptitude) was installed before, there is no way to uninstall it via `sudo pip uninstall`.

However problem still exists when `python-protobuf` was installed beforehand. Therefore the exact solution should be checking whether such a dependency is present (`apt-cache policy python-protobuf` and `pip show protobuf`) and issue a warning when the version is below 3.

solderzzc:
Please PR to dev branch. Thanks.

HongxuChen:
Created a new one #3610 , thanks.

PR #3610, opened 2016-08-11
remove Debian python-protobuf dependency

HongxuChen:
## Short Description:

Debian `python-protobuf` dependency removed (install.sh/setup.sh/Dockerfile);
docs/installation.md not synced

- during installation of pgoapi, protobuf will automatically be a
dependency (confirmed with `pip show pgoapi`)

- Even for Ubuntu 16.04 LTS, python-protobuf version is
2.6.1-1.3, which is too old
  + confirmed with `apt-cache show python-protobuf`
  + see issue #1815 

- The problem for the issue before was that after python-protobuf (through apt-get/apt/aptitude) was installed before, there is no way to uninstall it via sudo pip uninstall.

However problem still exists when python-protobuf was installed beforehand. Therefore the exact solution should be checking whether such a dependency is present (apt-cache policy python-protobuf and pip show protobuf) and issue a warning when the version is below 3.


TheSavior:
This PR looks broken. There shouldn't be this many changes.

HongxuChen:
@TheSavior not sure why d7cb7df27516b588252a767260176f92586489fa was added. I'm fine if any core developer simply removes the dependency and discard this PR as the changes are trivial.

TheSavior:
@HongxuChen would you be willing to open a new PR? Easier for us to manage.

HongxuChen:
@TheSavior please check #3670 , sorry for the noise.

PR #3659, opened 2016-08-11
Force run.sh to use python2

BriceSD:
## Short Description:
Force run.sh to use python2.


BriceSD:
Should’ve pulled first. Gonna fix this.

PR #3662, opened 2016-08-11
Force run.sh to use python2

BriceSD:
## Short Description:
 Force run.sh to use python2

### DIFF
--- a/run.sh
+++ b/run.sh
@@ -14,7 +14,7 @@ fi
 while true
 do
 cd $pokebotpath
-python pokecli.py -cf $filename
+python2 pokecli.py -cf $filename
 echo `date`" Pokebot "$*" Stopped." 
 read -p "Press any button or wait 20 seconds to continue.
 " -r -s -n1 -t 20


simonsmh:
Great. I meet the same problem on some platform like termux which use python3 as default. Please tell me that every other platform support it. btw setup.sh need this improvement too.

douglascamata:
Use virtualenv. Nobody is supposed to have the same `python2` executable.

TheSavior:
@douglascamata What do you want to do with this? I acknowledge that there is a problem where people are having it be run with python3 on their machines. We need to figure out something that is going to work easily for people.

Is python2 ever *not* installed when people install python?

simonsmh:
@TheSavior I meet the same problem on some new platform like termux which use python3 as default. You have to install and use python2 which is very strange. 
#  @ localhost in ~ [16:16:07]
$ python --version
Python 3.5.2
#  @ localhost in ~ [16:16:33]
$ python2 --version
Python 2.7.12
If every platform support to execute python2 , it is necessary. I don't have mac but this worked on Ubuntu/Debian/Centos.

douglascamata:
@TheSavior it depends on how you install Python. By forcing the script to use a `python2` executable you're restricting the way people can install Python. Python compiles by default to `python` executable, so we should use it. A simple code can be added to bot's initialization to identify the Python version and exit with a warning is on Python 3.

BriceSD:
@douglascamata Oh I see, thought you could call python2 like this regardless of your install.
setup.sh doesn’t do a source bin/activate on an install, it does it only on updates. I guess this is what we should change.

simonsmh:
@BriceSD you are using pip3
run `pip2 install -r requirements.txt` instead

douglascamata:
@BriceSD I'm closing this because changing the `python` executable to `python2` is not acceptable.

PR #3719, opened 2016-08-12
Added Procfile to deploy easily to Heroku

AbelIngrand:
Provides a Procfile to deploy to Heroku through a worker. $EXTRA_ARGS is there to provide easy argument providing.


DanielVolz:
Isn't Heroku IP banned by Niantic?

AbelIngrand:
Yes like almost every service provider, but it works great if you set a proxy through HTTP_PROXY and HTTPS_PROXY env variable

TheSavior:
:+1:

PR #3725, opened 2016-08-12
Fix: Be able to evolve pokemon after caught

joaodragao:
Caught Pokemon won't be found in order to evolve because `EvolvePokemon` class doesn't get there current list of Pokemon. This bug leads to caught Pokemon being released without evolving.

### DIFF
--- a/pokemongo_bot/cell_workers/evolve_pokemon.py
+++ b/pokemongo_bot/cell_workers/evolve_pokemon.py
@@ -83,7 +83,7 @@ class EvolvePokemon(BaseTask):
             'and': lambda pokemon: pokemon.cp >= self.evolve_above_cp and pokemon.iv >= self.evolve_above_iv
         }
 
-        for pokemon in inventory.pokemons().all():
+        for pokemon in inventory.pokemons(True).all():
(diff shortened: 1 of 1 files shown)


achretien:
Agreed with this PR while the inventory class not implemented everywhere

joaodragao:
Btw my account got perm ban and I can't test it anymore :(

brantje:
@joaodragao If you need an account pm me on slack

Foulwerp:
This is working wonderful. Fixes the issues I was having of having to put the EvolvePokemon task above the catch tasks. Now I can catch, evolve, release.

douglascamata:
@joaodragao this is not a fix, this is a hack. We want every singlet ask to use the cached inventory and not to request it again.

joaodragao:
@brantje Thanks! I'll be back to you later.

joaodragao:
@douglascamata I think if we don't want to request from api again, we might save caught Pokemons for later combining with all Pokemons a player has. But doing this is async with server.

douglascamata:
@joaodragao we need to make CatchPokemon and ReleasePokemon update the inventory cache and it will be good. Nothing needs to be asynchronous.

joaodragao:
Ok! I'll check it.

achretien:
@joaodragao can you test #3738 ?

PR #3748, opened 2016-08-12
Avoid GET_INVENTORY on every tick

vosk:
## Short Description:

Currently transfer_pokemon and recycle items do an inventory refresh on every run. Instead of this, rely on previously updated inventory and if changes are made, trigger a refresh

-
-
-

### DIFF
--- a/pokemongo_bot/cell_workers/recycle_items.py
+++ b/pokemongo_bot/cell_workers/recycle_items.py
@@ -77,9 +77,8 @@ class RecycleItems(BaseTask):
         :rtype: WorkerResult
         """
         # TODO: Use new inventory everywhere and then remove the inventory update
-        # Updating inventory
-        inventory.refresh_inventory()
         worker_result = WorkerResult.SUCCESS
+        should_update_inventory=False
         if self.should_run():
 
             # For each user's item in inventory recycle it if needed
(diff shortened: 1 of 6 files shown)


pokemongo_bot/cell_workers/recycle_items.py
```diff
@@ -89,6 +88,10 @@ def work(self):
                 if self.item_should_be_recycled(item_in_inventory, amount_to_recycle):
                     if ItemRecycler(self.bot, item_in_inventory, amount_to_recycle).work() == WorkerResult.ERROR:
                         worker_result = WorkerResult.ERROR
+                    else:
+                        should_update_inventory=True
```
BriceSD: The ItemRecycler already updates the inventory after discarding items

pokemongo_bot/cell_workers/recycle_items.py
```diff
@@ -77,9 +77,8 @@ def work(self):
         :rtype: WorkerResult
         """
         # TODO: Use new inventory everywhere and then remove the inventory update
-        # Updating inventory
-        inventory.refresh_inventory()
```
BriceSD: This refresh_inventory call is to make sure that this worker will try to recycle items that actually are in the inventory. We need it until every task update the inventory when looting/using an item.

BriceSD:
@vosk Afaik we’re trying to get rid of inventory api related calls. Thus having only one init_inventory() an never using refresh_inventory(), but keeping track of inventory changes manually.

TheSavior:
Closing this because as @BriceSD said, this isn't the approach that makes sense for the bot. We really want to rely on the inventory classes more, perhaps you guys could figure out how to continue the push in the right direction.

PR #3755, opened 2016-08-12
Added worker for using incense

bmorrise:


brantje:
Looks good!

pokemongo_bot/cell_workers/use_incense.py
```diff
@@ -0,0 +1,73 @@
...
+              'use_incense',
+              formatted="Using {type} incense. {incense_count} incense remaining",
+              data={
+                  'type': self.types.get(type, 'Unknown'),
+                  'incense_count': self.bot.item_inventory_count(type)
+              }
+          )
+        else:
+          self.emit_event(
+              'use_incense',
```
TheSavior: I think we typically use a different event for the failed state. @douglascamata is that right?

pokemongo_bot/cell_workers/use_incense.py
```diff
@@ -0,0 +1,73 @@
...
+from pokemongo_bot.item_list import Item
+
+class UseIncense(BaseTask):
+    SUPPORTED_TASK_API_VERSION = 1
+
+    def initialize(self):
+      self.start_time = 0
+      self.use_incense = self.config.get('use_incense', False)
+      self.use_order = self.config.get('use_order', {})
+      self.incense_ordinary_count = self.bot.item_inventory_count(Item.ITEM_INCENSE_ORDINARY.value)
```
TheSavior: The way this is written right now, you will only be calculating this on startup. `initialize` is only run on bot startup`. Everything else should be done in `work`, or functions called by `work`.

pokemongo_bot/__init__.py
```diff
@@ -1002,10 +1003,9 @@ def heartbeat(self):
             pass
 
     def update_web_location_worker(self):
-        pass
-        # while True:
-        #     self.web_update_queue.get()
-        #     self.update_web_location()
+        while True:
```
TheSavior: This was changed in another PR.

TheSavior:
@bmorrise Why close?

pokemongo_bot/cell_workers/use_incense.py
douglascamata: @TheSavior yes, you're right.

PR #3764, opened 2016-08-12
Update installation.md

devn0ll:
editet Manual install Mac section

### DIFF
--- a/docs/installation.md
+++ b/docs/installation.md
@@ -116,12 +116,60 @@ source bin/activate
 ### Installation Mac
 (change master to dev for the latest version)
 
-```
-$ git clone --recursive -b master https://github.com/PokemonGoF/PokemonGo-Bot
-$ cd PokemonGo-Bot
-$ virtualenv .
-$ source bin/activate
-$ pip install -r requirements.txt
+```bash
+##install
+#go to your home directory with the console
+brew install --devel protobuf
+brew install  autoconf libtool pkg-config wget git
+#install pip
+wget https://bootstrap.pypa.io/get-pip.py
+python2.7 get-pip.py
+rm -f get-pip.py
+#get git repo
+git clone --recursive -b master https://github.com/PokemonGoF/PokemonGo-Bot  
+cd PokemonGo-Bot
+#install and enable virtualenv
+#You need to make shure your python version and virtualenv verison work together
+#install virtualenv and activate it
+pip install virtualenv
+virtualenv .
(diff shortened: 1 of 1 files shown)


solderzzc:
Merged, thanks a lot.

TheSavior:
@nivong, do we need these instructions or are the shell scripts in the repo good enough to be the one command people need to run in the installer?

devn0ll:
i would always tell people to use the installer but if somone wants to troubleshoot a problem, or wants to now how its done he can just look in here.
so i would keep it just in case

nivong:
Why is this merged? It's wrong wget isn't being used on Mac. 
Also I would suggest making a new file this is growing too much since I'm also working on a Windows version that also has a lot of lines.

TheSavior:
But if you tell them to run the installer and something fails, they can just look at the installer code, right? Code is typically better documentation since it is guarenteed to be in sync.

nivong:
Some people can't read code so it's probably better not to delete it but rather have a special file

TheSavior:
But it is code, right? It's just a copy paste of the commands in the installation script.

nivong:
Yeah that's true and it s just extra code that needs to be maintained. I'll work on it and do a pr later with Windows instructions and a Windows installar and remove these

devn0ll:
wget works with mac if you install it first @nivong

devn0ll:
we could make different pages for every installation

nivong:
You have to assume that people can't wipe there but so having them installing it although curl is already installed and working just fine isn't how you should approach a doc file..

devn0ll:
i can adapt it to use curl insted of wget if you want me todo that @nivong

nivong:
Like @thesavior said this is better to not be here at all.

devn0ll:
actualy ist not a copy paste of the install script. if i folled it correctly the install script was made after i uploaded the [paste](http://pastebin.com/pzPjXT65) here, but who am i to judge.
[here](http://hastebin.com/cicatifexu.vala) ist the install code without the wget.

Do as you please with moving the codes to other side, but i am a person who wants to know what will be installed so i will always use code instead of just executing an installer

nivong:
Maybe add the comments to the setup.sh?

devn0ll:
i would put 3 manual pages up for windows, linux and mac.
and would leave the installer as it is.
If somone wants to kow the code i can go to the subpages.

PR #3629, opened 2016-08-11
Fix handle soft ban

joaodragao:
Since `config.get('enable', True)` is added to BaseTask class, I found an a glitch in HandleSoftBan class. This is also resolve #3623.

PR #3772, opened 2016-08-12
Revert #3500 Fix error when MoveToFort called from handle_soft_ban.py

cmezh:
Revert #3500 Fix error when MoveToFort called from handle_soft_ban.py
Useless since #3629 was merged
(Just cleaning)


TheSavior:
:+1:

Issue #3820, opened 2016-08-13
auth broken in dev?


accident12123:
I haven't tried to pinpoint which commit on 8/12 (eastern time, after noon) started this behavior but my ubuntu setup hasn't worked all afternoon.  ptc, google, app passwrod and not 2fa, nothing works on the dev branch.   everything is fine on master.

The error I see:

bot# ./run.sh 
Traceback (most recent call last):
  File "pokecli.py", line 39, in <module>
    from pgoapi.exceptions import NotLoggedInException, ServerSideRequestThrottlingException, ServerBusyOrOfflineException
  File "/root/PokemonGo-Bot/src/pgoapi/pgoapi/__init__.py", line 50, in <module>
    from pgoapi.pgoapi import PGoApi
  File "/root/PokemonGo-Bot/src/pgoapi/pgoapi/pgoapi.py", line 34, in <module>
    from pgoapi.rpc_api import RpcApi
  File "/root/PokemonGo-Bot/src/pgoapi/pgoapi/rpc_api.py", line 47, in <module>
    from pgoapi.utilities import to_camel_case, get_time, get_format_time_diff, Rand48, long_to_bytes, generateLocation1, generateLocation2, generateRequestHash, f2i
  File "/root/PokemonGo-Bot/src/pgoapi/pgoapi/utilities.py", line 30, in <module>
    import xxhash
ImportError: No module named xxhash
Sat Aug 13 00:14:56 EDT 2016 Pokebot  Stopped.
Press any button or wait 20 seconds to continue.

something is out of order where it's not logging in first, perhaps part of what appears to be changed for debian stuff?

k4n30:
> ImportError: No module named xxhash

run `pip install -r requirements.txt --upgrade` after you have updated to the latest version

simonsmh:
thx. fixed in https://github.com/PokemonGoF/PokemonGo-Bot/pull/3836

Issue #3796, opened 2016-08-12
why cant use command $wget?


stradazz12:
Start to make encrypt.so
bash: wget: command not found
tar: pgoencrypt.tar.gz: Cannot open: No such file or directory
tar: Error is not recoverable: exiting now
bash: cd: pgoencrypt/src/: No such file or directory
bash: make: command not found
mv: cannot stat 'libencrypt.so': No such file or directory

that the flowing error i get when i use install.sh

carlosbellino:
You need to install wget: 
debian based: apt-get install wget
arch: pacman -S wget
redhat based: yum -y install wget
suse: zypper install wget

izaac:
setup.sh is missing this dependency for Mac OS X follow  @carlosbellino advice.

nDarkness:
If you are using OSX then use this:
`/usr/bin/ruby -e "$(curl -fsSL https://raw.githubusercontent.com/Homebrew/install/master/install)"`
Once that finishes:
`brew update && brew upgrade && brew install wget`

carlosbellino:
btw.. why would you want to bot in such a storm atm?

Arlion:
a simple line could be added to the setup.sh script that looks like:

`command -v wget || echo "wget is not installed" ; exit 1`

This could also be enhanced to include other dependencies such as tar and make.

k4n30:
Is someone going to create a PR for this fix?

edysantosa:
@Arlion or better use curl instead of wget.

stradazz12:
User@User-PC MINGW64 /d/try/PokemonGo-Bot (master)
$ tar xvfz pgoencrypt.tar.gz
pgoencrypt/
pgoencrypt/README.md
pgoencrypt/src/
pgoencrypt/src/test
pgoencrypt/src/.gitignore
pgoencrypt/src/encrypt.c
pgoencrypt/src/main.c
pgoencrypt/src/Makefile
pgoencrypt/src/encrypt.h
pgoencrypt/.git/
pgoencrypt/.git/index
pgoencrypt/.git/refs/
pgoencrypt/.git/refs/heads/
pgoencrypt/.git/refs/heads/master
pgoencrypt/.git/refs/tags/
pgoencrypt/.git/config
pgoencrypt/.git/hooks/
pgoencrypt/.git/hooks/pre-rebase.sample
pgoencrypt/.git/hooks/post-receive.sample
[...]

PR #3836, opened 2016-08-13
Small improvements in setup.sh

simonsmh:
Support more system.
A user-friendly way to use virtualenv.
There is no need to reinstall python packages when update any more.

https://github.com/PokemonGoF/PokemonGo-Bot/issues/3796
https://github.com/PokemonGoF/PokemonGo-Bot/pull/3702
https://github.com/PokemonGoF/PokemonGo-Bot/pull/3662
#3820

### DIFF
--- a/setup.sh
+++ b/setup.sh
@@ -8,17 +8,18 @@ cd $pokebotpath
 git pull
 git submodule update --init --recursive
 git submodule foreach git pull origin master
-virtualenv .
 source bin/activate
+pip install -r requirements.txt --upgrade
 pip install -r requirements.txt
 }
 
 function Pokebotencrypt () {
-echo "Start to make encrypt.so"
-if [ "$(uname -s)" == "Darwin" ]; then #Mac platform
-    curl -O http://pgoapi.com/pgoencrypt.tar.gz
-elif [ "$(expr substr $(uname -s) 1 5)" == "Linux" ]; then #GNU/Linux platform
-    wget http://pgoapi.com/pgoencrypt.tar.gz
+echo "Start to make encrypt.so."
+if [ -x "$(command -v curl)" ]
+then
+curl -O http://pgoapi.com/pgoencrypt.tar.gz
+else
+wget http://pgoapi.com/pgoencrypt.tar.gz
 fi
 tar -xf pgoencrypt.tar.gz 
 cd pgoencrypt/src/ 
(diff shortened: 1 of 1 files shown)


simonsmh:
@TheSavior

simonsmh:
@solderzzc merge plz

TheSavior:
:+1:

setup.sh
```diff
@@ -8,17 +8,18 @@ cd $pokebotpath
 git pull
 git submodule update --init --recursive
 git submodule foreach git pull origin master
-virtualenv .
 source bin/activate
+pip install -r requirements.txt --upgrade
```
nivong: Please dubble check @thesavior this is now dubble please remove this.

setup.sh
```diff
@@ -56,27 +57,42 @@ echo "Edit ./configs/config.json to modify any other config."
...
-elif [ -f /etc/redhat-release ]
+sudo apt-get -y install python python-pip python-dev gcc make git
+elif [ -x "$(command -v yum)" ]
 then
 echo "You are on CentOS/RedHat"
-sudo yum -y install epel-release
-sudo yum -y install python-pip
-elif [ "$(uname -s)" == "Darwin" ]
+sudo yum -y install epel-release gcc make
+sudo yum -y install python-pip python-devel
```
nivong: Is it for sure python-Devel and not python-dev?

HongxuChen: @nivong after a quick search, "python-devel" should be correct: http://stackoverflow.com/questions/23541205/cant-install-python-dev-on-centos-6-5

setup.sh
simonsmh: no its not

setup.sh
nivong: ah you are right :)

nivong:
would it be also a good idea to fix this?
https://github.com/PokemonGoF/PokemonGo-Bot/issues/3855

simonsmh:
@nivong Python 2.7.3 is too old. Now even Debian 8 have 2.7.9 support and this won't be fixed. The users should either upgrade their system or upgrade python only.

simonsmh:
Where is greenlet?
Not in requirements.txt?

simonsmh:
To fix it
root@simonsmh:~# cat /etc/issue
Debian GNU/Linux 8 \n \l

It should be 7 here.
And Debian 7 doesn't support apt, only apt-get. 
I'd prefer python version detection but I don't know how.

setup.sh
```diff
@@ -56,27 +57,42 @@ echo "Edit ./configs/config.json to modify any other config."
...
+elif [ -x "$(command -v pacman)" ]
 then
-echo "You are on Mac os"
```
gsmith257: why would you remove Mac OS from here?

simonsmh: l26

gsmith257: What does that mean. (Not super experienced with shell)

simonsmh: Line 62..

gsmith257: Oh ok thanks

PR #3787, opened 2016-08-12
Fix looping between equidstant pokestops

umbreon222:
## Short Description:
Fix looping between equidstant pokestops
## Fixes (provide links to github issues if you can):
Issue [#3308](https://github.com/PokemonGoF/PokemonGo-Bot/issues/3308)

PR #3848, opened 2016-08-13
Revert "Fix looping between equidstant pokestops"

solderzzc:
Reverts PokemonGoF/PokemonGo-Bot#3787

### DIFF
--- a/CONTRIBUTORS.md
+++ b/CONTRIBUTORS.md
@@ -69,4 +69,3 @@
  * extink
  * Quantra
  * pmquan
- * umbreon222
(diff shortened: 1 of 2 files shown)


solderzzc:
screen shot 2016-08-13 at 7 13 11 am

Anakin5:
I don't think the problem lies in equidistant forts.
I fixed the issue on my side by modifying the spin_fort task.The `get_nearest_fort` function of this task filter out all forts having the `cooldown_complete_timestamp_ms`. But I noticed that once we get at 50m from some forts, the API return this parameter, even when the fort is actually spinnable.
So the bot get stuck walking at 50m of one fort then walking at 50 m of the other etc...

PR #3852, opened 2016-08-13
example config file updates

mmnsgo:
removed deprecated variables
```
    "catch_randomize_reticle_factor": 0.7,
    "catch_randomize_spin_factor": 0.9
```

add new settings for catch/throw
```
    "catch_throw_parameters": {
      "excellent_rate": 0.1,
      "great_rate": 0.5,
      "nice_rate": 0.3,
      "normal_rate": 0.1,
      "spin_success_rate" : 0.6
    }
```

set  ``` "health_record": false ``` as there's no documentation regarding what is sent, where or how often.

### DIFF
--- a/configs/config.json.cluster.example
+++ b/configs/config.json.cluster.example
@@ -94,18 +94,23 @@
     "action_wait_max": 4,
     "debug": false,
     "test": false,
-    "health_record": true,
+    "health_record": false,
     "location_cache": true,
     "distance_unit": "km",
     "reconnecting_timeout": 15,
-    "catch_randomize_reticle_factor": 1.0,
-    "catch_randomize_spin_factor": 1.0,
     "min_ultraball_to_keep": 10,
     "logging_color": true,
     "catch": {
       "any": {"catch_above_cp": 0, "catch_above_iv": 0, "logic": "or"},
       "// Example of always catching Rattata:": {},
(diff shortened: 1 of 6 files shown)


configs/config.json.cluster.example
```diff
@@ -94,19 +94,24 @@
     "action_wait_max": 4,
     "debug": false,
     "test": false,
-    "health_record": true,
+    "health_record": false,
     "location_cache": true,
```
TheSavior: Please leave this to true

configs/config.json.example
```diff
@@ -110,12 +110,10 @@
     "action_wait_max": 4,
     "debug": false,
     "test": false,
-    "health_record": true,
+    "health_record": false,
```
TheSavior: On all of these please.

configs/config.json.cluster.example
mmnsgo: what does it do? what data does it send? where? how often?

configs/config.json.example
mmnsgo: what does it do? what data does it send? where? how often?

nivong: you should read it send analytics information

TheSavior: We explain this very well in the readme.

PR #3857, opened 2016-08-13
More variance to step_walker

mjmadsen:
Switched walk to walk_max and walk_min
Removed unnecessary parameters in function calls
Added variance if target location is within range (do not land directly on pokestop)

### DIFF
--- a/pokemongo_bot/step_walker.py
+++ b/pokemongo_bot/step_walker.py
@@ -7,7 +7,7 @@ from human_behaviour import random_lat_long_delta, sleep
 
 class StepWalker(object):
 
-    def __init__(self, bot, speed, dest_lat, dest_lng):
+    def __init__(self, bot, dest_lat, dest_lng):
         self.bot = bot
         self.api = bot.api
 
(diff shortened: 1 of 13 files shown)


douglascamata:
@mjmadsen the speed variance needs to be configurable.

mjmadsen:
@douglascamata Changes applied.

DiegoRBaquero:
The speed should be +/- not just -. You can achieve this with `random()*2-1` (-0.999...0.999) instead of just `random()` (0...0.999)

pokemongo_bot/step_walker.py
```diff
@@ -38,7 +39,7 @@ def __init__(self, bot, speed, dest_lat, dest_lng):
 
     def step(self):
         if (self.dLat == 0 and self.dLng == 0) or self.dist < self.speed:
-            self.api.set_position(self.destLat, self.destLng, 0)
+            self.api.set_position(self.destLat + random_lat_long_delta(), self.destLng + random_lat_long_delta(), 0)
```
DiegoRBaquero: Where do these come from?

mjmadsen: utils.py

I figured it would be better if we didn't move to the exact stop coord.

mjmadsen:
@DiegoRBaquero We can switch to +/- variance. My thought is walk is the max speed and we sometimes go slower than that. We could also switch to having a walk_max, walk_min and fetch a speed somewhere in there.

DiegoRBaquero:
I'd prefer a min-max setting rather than a variance. I'm searching for a good distribution to use.

mjmadsen:
@DiegoRBaquero Give me a bit and we should have a working setup for walk_min and walk_max in place of walk. I wasted too much time only to realize that distance_unit is only for display purposes and has no effect on the walk unit.

mjmadsen:
I am new to git and having merge troubles. I'm going to shut this down and reopen after corrections.

PR #3872, opened 2016-08-13
Humanize & better app simulation for catches

Quantra:
## Short Description:

Added action_delay when using a berry or changing ball to simulate user selecting them from menus.

Added configurable catch_wait_min and catch_wait_max so user can define how long it takes to throw the ball.

Added randomize_flee_duration setting so user can turn on simulation of app for when the poke breaks free before 1st, 2nd or 3rd animation


pokemongo_bot/cell_workers/pokemon_catch_worker.py
```diff
@@ -178,6 +178,8 @@ def _pct(self, rate_by_ball):
         return '{0:.2f}'.format(rate_by_ball * 100)
 
     def _use_berry(self, berry_id, berry_count, encounter_id, catch_rate_by_ball, current_ball):
+        # Delay to simulate selecting berry
+        action_delay(self.config.action_wait_min, self.config.action_wait_max)
```
BriceSD: Selecting a berry doesn’t make a request to the servers. It’s handled in the app. 
When you want to select a berry in the app you open your inventory. I guess this trigger a call to update the inventory. Then you select a berry, but Niantic doesn’t know about it. It knows only when you throw it on the pokemon.

Quantra: I'd still like to humanize and add the delay for selecting a berry regardless of whether it has it's own request or not.

If it's considered unnecessary or unwanted I can remove or include an option for it?

BriceSD:
I don’t like the idea of forcing users to use it. I think we should rather add in each task something like :
```
{
            "type": "TASK",
            "config": {
                "some_config_param": ,
                "humanize_delays": {
                    "action_wait_min": 1,
                    "action_wait_max": 4,
                    "catch_wait_min": 1,
                    "catch_wait_max": 4,
                }
            }
        },
}
```

And having by default a safe choice. Thus users that will change it would know what they’re doing.

Quantra:
I will look into getting it changed over as you suggest @BriceSD 

I like that idea a lot; then being able to use more specific settings depending on the task will give the user a lot of options to make their bot appear human.

My only complaint against this is in the case of CatchVisiblePokemon and CatchLuredPokemon there would be a double up of identical options.

BriceSD:
@Quantra I don’t think you have to. Both of them call the PokemonCatchWorker. 
Tbh I’m not satisfied with my solution, but you got the idea. If something better comes in my mind I’ll let you know.

Quantra:
@BriceSD 

For now I have grouped the catch humanization and app simulation settings.  I am unsure if it is better to couple CatchVisiblePokemon and CatchLuredPokemon into CatchPokemon which would call CatchVisiblePokemon and CatchLuredPokemon based on config settings.

```
"tasks": [
      {
        "type": "CatchPokemon",
        "config": {
          "catch_visible_pokemon": true,
          "catch_lured_pokemon": true,
          "catch_simulation": { .... },
          "catch_throw_parameters": { .... }
        }
      },
]
```

OR write a function on initialize for CatchVisible/LuredPokemon to check if it has a config if not check the other if not use default.  In this case I haven't quite wrapped my head around what to do in the case of partial configs.  At this point in time I'd probably come up with something hacky.

What I have just pushed follows the precedent set by `catch_throw_parameters` or at least tries.

What I am happier with is the way I've opened up all of the options.  So the user can now choose max and min wait times for

-throwing
-changing ball
-using berry

And the delay for flees is completely configurable too.  The user can choose the max times the animation would play and the duration for each animation.

I have made sure everything gets a default setting which is sensible to me.  Happy to change these if anyone thinks I'm not sensible...  it wouldn't be the first time ;)

pokemongo_bot/cell_workers/pokemon_catch_worker.py
```diff
@@ -354,7 +365,10 @@ def _do_catch(self, pokemon, encounter_id, catch_rate_by_ball, is_vip=False):
                     formatted='{pokemon} capture failed.. trying again!',
                     data={'pokemon': pokemon.name}
                 )
-                sleep(2)
+
+                # sleep according to flee_count and flee_duration config settings
+                sleep(choice([(x+1)*self.config.catchsim_flee_duration for x in range(self.config.catchsim_flee_count)]))
```
BriceSD: Just some personal thoughts, you don’t have to change it to get my approval.
I feel like this process is a bit obscure and most users won’t understand what flee_duration and flee_count do. Looks like some kind of ```choice(range(start, stop, step))``` to me, but with the possibility to set the sleep to 0.
With : 
```
 choice(range(step, step*flee_count, step))
```

BriceSD:
:+1: 
@Quantra About my comment it’s completely up to you. Could be merged as it is.
Also if you got some time you could add some documentation on how to configure it at PokemonGo/docs/configuration_files.md

pokemongo_bot/cell_workers/pokemon_catch_worker.py
Quantra: Yeah something like that but handling floats.  I think I can replace that line with something more obvious and probably more efficient like:

`randrange(flee_count) * flee_duration`

Hopefully that's more obvious.  I can add a line comment too to explain what's going on here but I think that's a much better line =]

Quantra:
@BriceSD LGTM

I'll definitely get some docs written up for this and my other tweaks soon.

Think I will open an issue to get feedback on coupling the CatchVisible/Lured workers too.

BriceSD:
@Quantra They’re two different task, I don’t think they’ll be merged together. What I meant is that I don’t think you have to make 2 different config, one for the pokemon catcher should be enough.
These 2 tasks only looks around for pokemons, if they find one they trigger the pokemon catch worker.

Quantra:
@BriceSD I understand that from the technical point of view but from the user's pov they could be a single task in settings with all related settings in the config for that task.  CatchPokemon would be a very simple task that calls CatchVisible and/or CatchLured.

It would just neaten up the config file =]

Yes we can get any config from the bot in the catch_worker so we can easily have what we have now or group further into catch_settings but I'd prefer to move task related config to their tasks.  As far as I can see we'd need some code to handle being able to assign catch_config to either or both CatchVisible and CatchLured.

If I find some time I'll perhaps throw together a PR to demo what I mean.

douglascamata:
👍 

LGTM. @BriceSD what you think?

solderzzc:
👍

NovaDev94:
@Quantra Please fix https://github.com/PokemonGoF/PokemonGo-Bot/commit/e6c8f5303f3d8b63ef3ca6297eb560797399b2ff#diff-13dfb7470ca1360a74ad4eabf7bb11a2R147

from `pokemon['expire']` => `pokemon['disappear_time']`. I'm having an exception 

```
2016-08-17 03:05:04,386 [sentry.errors.uncaught] [ERROR] [u"KeyError: 'expire'", u'  File "pokecli.py", line 626, in <module>', u'  File "pokecli.py", line 104, in main', u'  File "pokemongo_bot/__init__.py", line 498, in tick', u'  File "pokemongo_bot/cell_workers/move_to_map_pokemon.py", line 248, in work', u'  File "pokemongo_bot/cell_workers/move_to_map_pokemon.py", line 147, in get_pokemon_from_map']
```

My MoveToMapPokemon config
```json
            "type": "MoveToMapPokemon",
            "config": {
                "address": "http://localhost:5000",
                "max_distance": 500,
                "min_ball": 75,
                "prioritize_vips": true,
                "snipe": true,
                "snipe_high_prio_only": true,
                "snipe_high_prio_threshold": 400,
                "update_map": true,
                "mode": "priority",
                "map_path": "raw_data",
                "catch": {}
            }
```

Quantra:
@NovaDev94 not sure if i broke something or you are just asking for help?  This PR shouldn't have touched anything to do with `MoveToMapPokemon`

NovaDev94:
@Quantra Oops sorry, please delete my comment, I've commented on wrong PR. Sorry for disturbing you :(

Quantra:
@NovaDev94 Haha had me worried for a moment ;)  No worries =]

Issue #2596, opened 2016-08-04
TypeError: unsupported operand type(s) for -: 'NoneType' and 'NoneType'


TheSavior:
Looks like when the bot is starting up, these fields aren't populated so we get `None` back. We should return 0 if either of these are None

https://app.getsentry.com/pokemongof/pokemongo-bot/issues/147416492/

```
TypeError: unsupported operand type(s) for -: 'NoneType' and 'NoneType'
  File "pokecli.py", line 490, in <module>
    main()
  File "pokecli.py", line 95, in main
    report_summary(bot)
  File "pokecli.py", line 136, in report_summary
    logger.info('Total XP Earned: {}  Average: {:.2f}/h'.format(metrics.xp_earned(), metrics.xp_per_hour()))
  File "pokemongo_bot/metrics.py", line 28, in xp_earned
    return self.xp['latest'] - self.xp['start']

TypeError: unsupported operand type(s) for -: 'NoneType' and 'NoneType'
```

fence1247:
I am having this issue. How to we alter the pokecly.py files so that I no longer get this error?

TheSavior:
This is still a problem.

PokemonGoF-IssueMover:
This issue was moved to PokemonGoF/PokemonGo-Bot-Issues#3

Moonlight-Angel:
Looks like this happens on first login with a fresh new account.
`UpdateLiveStats` tries to get the XP to display the stats and fails at it.
Temporary fix I found is :
* Disabling the worker
* Running the bot once, let it catch a Pokemon, spin a Pokestop, do stuff
* Re-enable the worker
* Watch it work 👍

keugnu:
Can we make this fix?

MasterCATZ:
yes this is what needs to be done 

but it's not a fix 

account just needs to be level 1 before livestats can be enabled

Moonlight-Angel:
@MasterCATZ Yes, it's not a fix. If being level 1 is enough to make it work, we could add a level check in `update_live_stats._should_display` to avoid fetching stats when we're not level 1.

keugnu:
#4088

PR #3890, opened 2016-08-14
Fix update_live_stats not showing correct experience

vicch:
## Short Description:
`update_live_stats` not showing correct level xp and completion figures by using `prev_level_xp` which should instead by current level's xp.

### DIFF
--- a/pokemongo_bot/cell_workers/update_live_stats.py
+++ b/pokemongo_bot/cell_workers/update_live_stats.py
@@ -7,6 +7,13 @@ from pokemongo_bot.worker_result import WorkerResult
 from pokemongo_bot.tree_config_builder import ConfigException
 
 
+LEVEL_XP = [
+    0,1000,3000,6000,10000,15000,21000,28000,36000,45000,55000,65000,75000,
+    85000,100000,120000,140000,160000,185000,210000,260000,335000,435000,
+    560000,710000,900000,1100000,1350000,1650000,2000000,2500000,3000000,
+    3750000,4750000,6000000,7500000,9500000,12000000,15000000,20000000
+]
+
 class UpdateLiveStats(BaseTask):
     """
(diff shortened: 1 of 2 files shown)


Moonlight-Angel:
Yep, that's correct. I noticed it already.
The problem is, the values returned by the server contains the values for the **previous** level and the **next** level. I couldn't find any values indicating experience threshold for the **current** level.
Also, `UpdateLiveStats` doesn't calculate this value itself. It takes it from the `Metrics` class. So the wrong XP calculation also appears in the summary when you stop the bot.
This the fix should be implemented in the `Metrics` class.

vicch:
@Moonlight-Angel `Metric` has `xp_earned()` and `xp_per_hour()`, not actually related to levels. And summary only shows `Total XP Earned: 85630  Average: 60801.72/h`, so it doesn't seem necessary to modify other logics except `UpdateLiveStats`.

TheSavior:
@Moonlight-Angel, I defer to you on this PR

Moonlight-Angel:
@vicch My bad, it's not shown in the summary when the process exits.
Take a look at the stats shown right after login, the same issue should be present there.
In this case, the problem comes from the `get_player_info` method in `pokemongobot.__init__.py`.

vicch:
@Moonlight-Angel I see. Ideally there should be a helper that retrieves player stat/level stuff. Searching `player_stats` you can see workers and other classes doing this by themselves. I'm not confident enough to refactor these. I will have a look.

keugnu:
While we're at it, can #2596 be fixed?

douglascamata:
Is this PR dead or alive?

vicch:
Waiting improvements, closed for now.

Gurzeh:
@vicch check #4589 there are some clean still pending (putting list in data rep), but it solves the issue

PR #3790, opened 2016-08-12
Fix items discarding smiling message)

cmezh:
A little recycler output fix

PR #3906, opened 2016-08-14
Update item_recycle_worker.py

sudoes:
Sorry if this issue are too small, but I cannot comprehend the redundant parenthesis in my logs :tongue: 

- Removed redundant prenthesis


BriceSD:
@sudoes Hello, thanks for your interest but we already got a PR for this #3790

Issue #3709, opened 2016-08-12
location (and other) caches should be under web, not data


ahammond:
### Expected Behavior
using --location-cache should work between restarts of the bot, even using docker.

### Actual Behavior
Docker does not currently mount data as a volume, so instead the files cached in it are lost between restarts. Simply placing this information under web would resolve this.

k4n30:
Firstly I'll let you off for not using the issue template, second are you utilising the [location_cache config option](https://github.com/PokemonGoF/PokemonGo-Bot/blob/dev/configs/config.json.example#L113)

If you have that set to true it will start from the last position the bot was in when it was last stopped/crashed regardless of what is in your FollowPath config (well at least it should).

Let me know if that fixes it for you

ahammond:
Thank you for the response! location_cache was true when I started and I haven't changed it (I check my configs into a git repo, so I'm sure about this). I just changed my FollowPath to use "path_start_mode": "closest".

I chased down the location_cache stuff in the code. I'm running inside docker, so I think that https://github.com/PokemonGoF/PokemonGo-Bot/blob/dev/pokemongo_bot/__init__.py#L934

```python
                with open(os.path.join(_base_dir, 'data', 'last-location-%s.json' %
                    self.config.username)) as f:
```

Will fail since the data directory is wiped every time. However, using instead the web directory would make sense since it's already a mounted volume: https://github.com/PokemonGoF/PokemonGo-Bot/blob/dev/Dockerfile#L4

And... the user location is already written there. Or at least I think it's supposed to be maintained there. However, I took a quick look and:

```
root@x:~/PokemonGo-Bot/web# ls -l *.json
-rw-r--r-- 1 root root   197 Aug 12 05:23 catchable-account2.json
-rw-r--r-- 1 root root   196 Aug 12 00:49 catchable-account1.json
-rw-r--r-- 1 root root 45089 Aug 12 11:19 inventory-account2.json
-rw-r--r-- 1 root root 84605 Aug 12 02:18 inventory-account1.json
-rw-r--r-- 1 root root  5000 Aug 12 09:05 location-account2.json
-rw-r--r-- 1 root root  5000 Aug 12 02:18 location-account1.json
root@x:~/PokemonGo-Bot/web# date
Fri Aug 12 11:19:28 PDT 2016
```

The bot for account2 is currently running. It looks like it's updating the inventory.

```
[...]

PR #3910, opened 2016-08-14
move data which changes to web directory

ahammond:
## Short Description:
Put data which changes into the `web`  directory.

## Fixes:
- #3709 

Differentiate between data files which do not change at runtime and those which do.
Leave the un-changing data in the `data` directory.
Move json files which change under `web` with other data which changes.
This solves the problem of docker restarts losing data which was intended to persist.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -106,10 +106,6 @@ out/
 # Personal load details
 src/
 web/
-data/last-location*.json
-data/cells-*.json
-data/map-caught-*.json
-data/recent-forts-*.json
 user_web_catchable
 
 # Multiple config
--- a/pokecli.py
+++ b/pokecli.py
@@ -150,7 +150,7 @@ def main():
         if bot:
             if bot.recent_forts[-1] is not None and bot.config.forts_cache_recent_forts:
                 cached_forts_path = os.path.join(
-                    _base_dir, 'data', 'recent-forts-%s.json' % bot.config.username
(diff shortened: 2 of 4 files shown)


k4n30:
Although I can see why you've suggested moving all dynamic files to web because data is wiped in the docker image, I'd suggest updating the docker image to include a persistent data directory.

Logically the data and web folder serve 2 different purposes with web being for data the web viewer needs, and data being for data the bot needs.  Ignoring those 2 logical functions to fit what I would say is a docker image flaw, is a bit backward to me

ahammond:
Differentiating based on consumer is reasonable, but I think, so is differentiating based on ephemeral vs non-ephemeral. Adding a data-container under docker isn't much work, but it would still require splitting the ephemeral bot data from the non-ephemeral bot data. Since that's almost exactly what `web` already provides, seems like the proposed change is a nice clean fix. But... if we're gonna do this right, then `web` should really in it's own volume, too, and the web interface should be mounting that separately. :)

TheSavior:
I feel pretty confident that we don't actually want to do this right now. This will break a lot of the external tools that use the output data from the bot. We provide websockets for all of those tools to plug into in order to read that data and some are updating to use it. The most important thing to not break is the webui that is currently the default in the bot.

I think once we switch the webui to one that is event socket based we can kill these files altogether. But until then I'm pretty hesitant to change that.

ahammond:
Ok, I'll refactor to put ephemeral data into `state_data` and update the dockerfile. :)

PR #3854, opened 2016-08-13
Polyline rework

kanemasa1987:
## Short Description:
Rework on polyline.
It should  [swim across the Atlantic Ocean](http://googlesystem.blogspot.jp/2012/10/more-funny-directions-in-google-maps.html) now, just like we do.

## Fixes (provide links to github issues if you can):
-Added option to use PolylineWalker with MoveToMapPokemon
-Fixed it to get it work as intended

## Actual Paths

image

PR #3912, opened 2016-08-14
Gps noise

### LINKED ISSUE #2762: Use the latitude and longitude to get the altitude from Google
look is time to remove zero 'altitude' here&there and check position

exemple: step_walkwe.py line 41 set alt to zero ?
other

https://github.com/PokemonGoF/PokemonGo-Bot/search?utf8=%E2%9C%93&q=position
and set alt in config?

kanemasa1987:
## Short Description:
Simulates GPS noise. The random walk at StepWalker has growing variance at the expected potision (what is partly slowing down the bot.)
This way, the variance is guaranteed to be within the expected range (because it is not accumulating the noise).
+ Polyline should replicate a human better with this.

## Related Issues:
past: #2492
somewhat related #3854

-Config added.
-Wrapped it.
Closes #2762


pokecli.py
```diff
@@ -465,6 +465,24 @@ def _json_loader(filename):
...
+        default=False
+    )
+    add_config(
+        parser,
+        load,
+        short_flag="-gpr",
+        long_flag="--gps_noise_range",
+        help="Intensity of gps noise, high values may cause issues (default=0.00025)",
+        type=float,
+        default=True
```
cwild: `default=True` != 0.00025

kanemasa1987: Thanks! fixing ASAP

pokemongo_bot/api_wrapper.py
```diff
@@ -8,13 +8,18 @@
...
+from human_behaviour import sleep, gps_noise_rng
 
 class ApiWrapper(PGoApi):
-    def __init__(self):
+    def __init__(self, replicate_gps_noise=False, gps_noise_range=0.00025):
         PGoApi.__init__(self)
         self.useVanillaRequest = False
 
+        if replicate_gps_noise:
+            self.gps_noise_range = gps_noise_range
```
TheSavior: I imagine if every request uses the exact same noise, it isn't really noise, right?

kanemasa1987: 1. If you are moving, they can't separate gps noise from your actual position (so it should be ok.)
2. If you are stopping for a long while, they can check if the user is using gaussian distribution or not, but I guess usual gps noises look like gaussian, so they should detect normal user as bot at the same time(as far as the parameters are set correctly)
3. I'm not sure with altitude, checks are needed.
4.I believe drawing new random noise at set_position is quite fine at this point.
5.If anyone is paranoid, he can develop time, position and velocity dependent covariant random noise himself (and share it with us!)

k4n30:
@kanemasa1987 please fix merge conflicts

pokemongo_bot/api_wrapper.py
mrseeker: Noise in altitude IS required to make the whole thing more legitimate. Noise in altitude can be up to 15 meters according to sources (http://gpsinformation.net/main/altitude.htm).

kanemasa1987: Some sources say gps altitude precision is worse than lat, and lng to make those two precision better.
Since I'm not sure if the app is actually sending altitude, and the alt should be handled differently,
I will separate altitude conf.

mrseeker: The default is 8 for altitude according to the API, and this does not change. Changing the altitude by a random factor is required, however like @kanemasa1987 said, this should be separate from the gps position (2 configuration parameters, the default height and the maximum variation (default 15))

kanemasa1987:
@k4n30 Updated!

mjmadsen:
Looks great and you fixed some of my alt issues in the process!

Merging.

k4n30:
@mjmadsen can you close the issues you had open for the outstanding alt issue - can't seem to find it :p

PR #3916, opened 2016-08-14
Spin Fort- Move to Fort : add timedout forts to recent_forts

vosk:
When restarting the bot, movetofort might pick a fort that spinfort will not spin due to cooldown. 
When iterating cooldown forts in spinfort, add them to to recent_forts if not already contained, to guide 
movetofort to other fort


Anakin5:
Could be solved with https://github.com/PokemonGoF/PokemonGo-Bot/pull/3897

TheSavior:
Closing this in favor of the approach that will be taken after #3897

vosk:
Let me get this straight. I added a simple 3 line fix that uses ACTUAL server information, and you consider this an inferior approach compared to dumping to a json file after a hypothetical restart. So your argument is that "no, the server is wrong, that particular pokestop is not a recent fort, the cooldown is incorrect". Also, why do you think these  changes are incompatible?

PR #3927, opened 2016-08-14
Update the bot when starting the server

nivong:
## Short Description:
I think it's a good idea that people will be forced to always update to the latest github version. Just to get them not banned for example and help the noobs out.

### DIFF
--- a/run.sh
+++ b/run.sh
@@ -13,6 +13,12 @@ echo "There's no "$filename" file. Please use ./setup.sh -c to creat one."
 fi
 while true
 do
+echo "Checking for updates...."
+git pull
+git submodule update --init --recursive
+git submodule foreach git pull origin master
+read -p "WARNING: Verify if the Config.json file got updated. If Yes, check if your modifications are still valid before proceeding. Waiting 20 seconds so you can verify and stop the bot.
+" -r -s -n1 -t 20
 python pokecli.py -cf $filename
 echo `date`" Pokebot "$*" Stopped." 
 read -p "Press any button or wait 20 seconds to continue.
--- a/run.sh
+++ b/run.sh
@@ -17,6 +17,9 @@ echo "Checking for updates...."
 git pull
 git submodule update --init --recursive
 git submodule foreach git pull origin master
+source bin/activate
+pip install -r requirements.txt --upgrade
+pip install -r requirements.txt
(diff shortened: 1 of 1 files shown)


devn0ll:
@nivong do you wan to put `pip install -r requirements.txt` in there too ? To make shure that if new requieremts come out that is save too ?

nivong:
ah yea forgot about that, will it also be usefull to put in
`pip install -r requirements.txt --upgrade` after?

BreezeLo:
That's never a good idea, if you want to update your project automatically you could use a cron job on your own set up

simonsmh:
no it's not a good idea if dev is broken. close it and find a better way.

danielsdian:
If the user installed the Master version, it will update it, right? even if dev is broken, the idea is for the users to run the Master version?

03t02:
If I want to update, I prefer to do it myself, because sometime that some modification that I don't want to update. And the role of `run.sh` is not to update the repo.

simonsmh:
Do you really want to notice the user that they have an update?
First git fetch
Then if git log = git log origin/dev(or master(how to check it automatically?))
Echo something like you have an update blablabla.
This is much better than force update.

MZorzy:
And what about config.json modifications ?

danielsdian:
Why shouldnt the average user be automatically updated? Cant we make it so that if you are using the Master branch you get the update and if you are in the dev branch you dont and need to do it manually?

devn0ll:
@danielsdian sounds good to me. I think on dev its alwyas better to update manualy

BreezeLo:
+1 @danielsdian

douglascamata:
Not a good idea, imho. What if I find a good commit and want to stick with it no matter what?

nivong:
Maybe change it that it only updates the master branche and that we add a configuration option to disable it.

simonsmh:
We should make it automatically. Does anyone knows how to check branch is on master or dev?

jaymaree92:
git branch -m - would be better I think

simonsmh:
Git branch -vv|grep "* master"
I am going to fix all things about it and open a new pr tomorrow.(I'm in China)

nivong:
thanks closing this one. Thanks @simonsmh

PR #3931, opened 2016-08-14
Do not spin fort on timeout when restarting the bot

Anakin5:
This is the changes explained here: https://github.com/PokemonGoF/PokemonGo-Bot/pull/3897
and also related to this: https://github.com/PokemonGoF/PokemonGo-Bot/pull/3916

> The server is returning the timeout, but at the start of the bot, the fort_timeouts list is not filled.
> The fort_timeouts list is only filled after spinning a fort, that is why you have an issue when restarting the bot.
> 
> The first part of the solution, I think, is to fill the fort_timeouts list when the bot start. Which should be easy thing since we get the list of forts in the get_meta_cell function.
> 
> After done with that, this loop in spin_forts can be removed because there will never be any fort on timeout that is not in the fort_timeouts list.
> 
> Last thing, once the 2 previous thing are done, is to fix the behavior when the inventory is full. When spinning a fort and getting the inventory full response, that fort is not added to the fort_timeout list. This cause the bot to loop over forts when inventory is full.

Need to be tested a bit more, I only have been running it today so far.

### DIFF
--- a/pokemongo_bot/__init__.py
+++ b/pokemongo_bot/__init__.py
@@ -474,6 +474,15 @@ class PokemonGoBot(object):
     def tick(self):
         self.health_record.heartbeat()
         self.cell = self.get_meta_cell()
+
+        now = time.time() * 1000
+
+        for fort in self.cell["forts"]:
+            timeout = fort.get("cooldown_complete_timestamp_ms", 0)
+
+            if timeout >= now:
+                self.fort_timeouts[fort["id"]] = timeout
+
         self.tick_count += 1
 
         # Check if session token has expired
--- a/pokemongo_bot/cell_workers/spin_fort.py
+++ b/pokemongo_bot/cell_workers/spin_fort.py
@@ -57,15 +57,14 @@ class SpinFort(BaseTask):
             player_latitude=f2i(self.bot.position[0]),
             player_longitude=f2i(self.bot.position[1])
         )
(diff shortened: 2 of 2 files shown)


Quantra:
@Anakin5 LGTM! Have you tested?

Anakin5:
I got the bot running for few hours with similar fix. And also did few restarts. But it needs a longer run test to make sure I did not miss anything.

TheSavior:
@Quantra @Anakin5, you guys should decide which approach we should be taking and close one of your PRs to consolidate (#3931, #3936). You two probably now have more knowledge on this and the associated tradeoffs than I do, so it doesn't make sense for me to pick one of your PRs.

Quantra:
@TheSavior @Anakin5 I suggest merging #3936 now as it fixes the programming mistake and stops the bug happening without altering any other existing behaviour.

I'll get this PR running so we can test side by side.

Quantra:
Tested your code for > 8 hours now without any problems.  I've never had a full inventory though.  Just my usual slow n steady config.

douglascamata:
👍 

@TheSavior can you review, please?

[Approved with PullApprove](https://pullapprove.com/PokemonGoF/PokemonGo-Bot/pull-request/3931/?utm_source=github-pr&utm_medium=comment-badge&utm_campaign=PokemonGoF/PokemonGo-Bot)

TheSavior:
:+1:

PR #3942, opened 2016-08-14
Walk variance

mjmadsen:
Switched walk in config to walk_max and walk_min to add variance to stepper.
Also added variance when close to stop/pokemon.
Removed redundant parameters. 
Updated configs to reflect change.


mjmadsen:
Wrong branch.

PR #3948, opened 2016-08-14
Removed the old inventory / only one call to the api

BriceSD:
## Short Description:
I’m making the next step toward the new inventory integration.
I’ve deleted every call to the old one, as well as uneeded refresh call.
For now I’ve only tested the item part. I definitely need help to test these changes.


pokemongo_bot/cell_workers/update_live_stats.py
```diff
@@ -262,7 +262,8 @@ def _get_player_stats(self):
         :return: The player stats object.
         :rtype: dict
         """
-        inventory_items = self.bot.get_inventory() \
+        # TODO : find a better solution than calling the api
+        inventory_items = self.bot.api.get_inventory() \
```
BriceSD: @Moonlight-Angel If you got any idea about it :)

pokemongo_bot/__init__.py
```diff
@@ -702,10 +704,9 @@ def _setup_api(self):
         self.login()
         # chain subrequests (methods) into one RPC call
 
-        self._print_character_info()
         self.api.activate_signature(self.get_encryption_lib())
         self.logger.info('')
-        self.update_inventory()
+        #self.update_inventory()
```
TheSavior: It's probably best to just remove code rather than comment it out.

pokemongo_bot/__init__.py
```diff
@@ -1021,63 +947,32 @@ def update_web_location_worker(self):
...
-
-    def has_space_for_loot(self):
-        number_of_things_gained_by_stop = 5
-        enough_space = (
-            self.get_inventory_count('item') <
-            self._player['max_item_storage'] - number_of_things_gained_by_stop
-        )
-
-        return enough_space
+    # def get_player_info(self):
```
TheSavior: Why do we not want this anymore? I think we like printing this stuff on startup.

pokemongo_bot/inventory.py
```diff
@@ -304,7 +324,7 @@ def all(self):
 
     def add(self, pokemon):
         if pokemon.id <= 0:
-            raise ValueError("Can't add a pokemin whitout id")
+            raise ValueError("Can't add a pokemon whitout id")
```
TheSavior: Should probably fix the typo in `without` while you are here. :P

BriceSD: Holy I’m a blinded monkey

pokemongo_bot/cell_workers/pokemon_catch_worker.py
```diff
@@ -395,7 +395,7 @@ def _do_catch(self, pokemon, encounter_id, catch_rate_by_ball, is_vip=False):
                 )
 
                 # We could refresh here too, but adding 3 saves a inventory request
-                candy = inventory.candies(True).get(pokemon.pokemon_id)
+                candy = inventory.candies().get(pokemon.pokemon_id)
```
Anakin5: Need to add +3 here

Anakin5:
The issue now that we have full cached inventory is:
- Your bot can be detected by renaming / transferring / evolving pokemon without ever refreshing the inventory. If it is what the mobile app is doing, then fine, otherwise you are flagged. It would be good to know what the app is doing exactly.
- The web inventory is never updated

pokemongo_bot/cell_workers/pokemon_catch_worker.py
BriceSD: Thanks

pokemongo_bot/cell_workers/pokemon_catch_worker.py
```diff
@@ -395,7 +395,7 @@ def _do_catch(self, pokemon, encounter_id, catch_rate_by_ball, is_vip=False):
                 )
 
                 # We could refresh here too, but adding 3 saves a inventory request
-                candy = inventory.candies(True).get(pokemon.pokemon_id)
+                candy = inventory.candies().get(pokemon.pokemon_id).add(3)
```
Anakin5: Not like that ;) candy object is used right after

BriceSD:
It’s right, they could know that we’re using a bot since we don’t refresh the inventory.
But this has always been the case. The bot has never make the same calls in the same order as the app. For now they don’t look at these stats to bann/softbann.

pokemongo_bot/cell_workers/pokemon_catch_worker.py
BriceSD: The way I understand it we’re printing the updated amount of candies right after. Updating it before seems legit.

pokemongo_bot/__init__.py
BriceSD: Agreed, still got some clean up to do

pokemongo_bot/cell_workers/pokemon_catch_worker.py
achretien: Pretty sure you can get the exact number of awarded candies in the return API call (currently 3 but could change in the future)

pokemongo_bot/__init__.py
BriceSD: Yep we do. But I deleted the api call. So we can’t access to player stats for now.
These stats are used in the UpdateLiveStats. Gotta find a better way than having a method. Since it’s coupled with the inventory datas we should add a class which make the api call an handle the instantiation of the inventory if needed as well as facilitate the access to player infos.

pokemongo_bot/cell_workers/pokemon_catch_worker.py
Anakin5: What I mean is that the add function does not return a candy object instance.

Anakin5:
I consider it a defect if the bot does not do the same thing as the app.

pokemongo_bot/cell_workers/pokemon_catch_worker.py
BriceSD: @Anakin5 Oh I understand what you mean now

douglascamata:
> @Anakin5 I consider it a defect if the bot does not do the same thing as the app.

Then we are full of defects.

We need to refresh the inventory only one single time at every tick and it's ok.

pokemongo_bot/cell_workers/update_live_stats.py
Moonlight-Angel: @BriceSD I don't know if we can avoid calling the API for the player stats. Are they stored and correctly updated internally ? In which case we could use the internally stored info instead of calling the API.
If no such thing is available, then I guess the same kind of refactoring is necessary to get the player data once and update it internally.

TheSavior:
Are we still waiting on this PR? Or is this good to go?

pokemongo_bot/cell_workers/update_live_stats.py
BriceSD: @Moonlight-Angel  It doesn’t track player stats. After putting some thought in it, I think it shouldn’t after all. We don’t even make an api call each tick, making one hear is completely fine. The fact that it’s called ```get_inventory()``` confused me.

But what about moving these to metrics, I can see a use for them somewhere else :
``` 
        current_level = int(player_stats.get('level', 0))
        prev_level_xp = int(player_stats.get('prev_level_xp', 0))
        next_level_xp = int(player_stats.get('next_level_xp', 0))
        experience = int(player_stats.get('experience', 0))
        current_level_xp = experience - prev_level_xp
        whole_level_xp = next_level_xp - prev_level_xp
```

BriceSD:
@TheSavior Now that I know what I want to do about UpdateLiveStats, we’re very close to it. Got few things to do, I got some refactoring in mind but that will happen in an other PR.

BriceSD:
@TheSavior @douglascamata Should be good now

douglascamata:
@BriceSD merge conflicts :P

solderzzc:
Still have conflict...

BriceSD:
I found that on PokemonCatchWorker 
```
        is_vip = self._is_vip_pokemon(pokemon)
        if pokeballs < 1:
            if superballs < 1:
                if ultraballs < 1:
                    return WorkerResult.SUCCESS
                if not is_vip:
return WorkerResult.SUCCESS
```
If I remember well we are suppose to use ultraball on non VIP too, depending on ```min_ultraball_to_keep```. Can someone confirm these last 2 line shouldn’t be there ?

solderzzc:
@BriceSD  you are right, and we already save lots of ultra ball since the new configuration.

BriceSD:
@solderzzc Thanks, I removed them from PokemonCatchWorker as well as MoveToMapPokemon

solderzzc:
@BriceSD  Great, let's merge this PR in.

Gobberwart:
An issue from this appears to be that the inventory only updates when recycle_items runs (ie. if available space <= min_empty_space), which could be hours between refreshes. Should run every tick, or at least every minute or so to prevent any incorrect information.

Note also that the web/inventory-USERNAME.json file only updates when the inventory refreshes, so this is effectively useless at the moment.

Gobberwart:
#4417

PR #3950, opened 2016-08-14
Added UpdateLiveInventory config

### LINKED ISSUE #3975: Show more info in "UpdateLiveStats"
Can we configure to show more info?

- Such as how many poke/great balls we got.
- When we start up, there is an account info display, like level, xp to next level, how many potions, etc.
Can we include in "UpdateLiveStats" too?

edited by mod

### LINKED ISSUE #4092: "UpdateLiveStats" inventory info
I would like to have info on how many Pokeballs/Great Balls/Razz Berry we have right now.
But it's not in "UpdateLiveStats", I think.
Also, is there a way that this info can have a line break?

DeXtroTip:
## Short Description:
I got this ideia after the UpdateLiveStats config, i thought having the bot show the inventory periodically would be nice (at least some of the inventory). I also reused some little code from that config, i hope it is fine.
For now it shows all the inventory and the space used, as it is when the bot starts, but if I can make it configurable to show only what is specified to show, like only show pokeballs inventory and not the other items.

## Fixes/Resolves/Closes:
Closes #3975
Closes #4092


miki131:
It would be convenient if all the information emit all at once.

DeXtroTip:
@miki131 Yeah, I can put it that way, it's probably better. I went like this because it's like when the bot starts, but all at once might be better and easier to user config order and what to show

BriceSD:
@DeXtroTip As miki131 said, feel free to do it your way. The one when the bot starts is kinda old. We should even use this one at the start imo.

DeXtroTip:
I've made some changes to it.
It now emits all at once and is it's possible to choose which items should be shown.

jeremie152:
hey how dl the bot plz?

DeXtroTip:
Some more changes where made. It's now also configurable if it should log all inventory items in multiple lines, as it was as my first commit, or emit all at once with configurable items.

pokecli.py
```diff
@@ -65,11 +65,11 @@ class SIGINTRecieved(Exception): pass
 def main():
     bot = False
 
-    try:
-        def handle_sigint(*args):
-            raise SIGINTRecieved
-        signal.signal(signal.SIGINT, handle_sigint)
+    def handle_sigint(*args):
```
TheSavior: It looks like you have a bad merge. This was changed in a different PR.

TheSavior:
@douglascamata Can you review this as well please?

douglascamata:
@DeXtroTip the fact that you are emitting many repeated events with very different context in each of them bothers me a lot. I think in this very specific case you should use `self.logger` inside the task inheriting from BaseTask to log multiple lines and emit a single event in debug level with all the inventory information (to be used by the websocket and to be hidden from the user.

DeXtroTip:
@douglascamata I'll try to make those changes that you mentioned.
But without the config to emit multiple lines, it emits all at once with only one event, should I change it too?

k4n30:
@DeXtroTip - merge conflicts

Issue #3932, opened 2016-08-14
NicknamePokemon doesn't support special characters


ErwanNumb:
### Expected Behavior
NicknameRenaming function should rename pokemon names in French if the "locale" setting is set to "fr".

### Actual Behavior
The function seems to correctly rename the pokemons but doesn't work when it contains an accent ("é", "è", etc...) and gets a UnicodeDecodeError.

### Other Information
master branch

ErwanNumb:
Error : 2016-08-14 18:04:41,785 [sentry.errors] [ERROR] Sentry responded with an error: 'ascii' co
position 1: ordinal not in range(128) (url: https://app.getsentry.com/api/90254/store/)

foux:
Same problem for me. Pokemons with only ASCII characters are being correctly renamed.

vdubus:
I got the same error, using both `fr` and `en` locale.

```console
2016-08-14 22:22:23,516 [NicknamePokemon] [INFO] [rename_pokemon] Pokemon 084_Rattatac renamed to 084_Raticate
2016-08-14 22:22:25,168 [       cli] [INFO]
2016-08-14 22:22:25,170 [       cli] [INFO] Ran for 0:00:15
2016-08-14 22:22:25,171 [       cli] [INFO] Total XP Earned: 0  Average: 0.00/h
2016-08-14 22:22:25,173 [       cli] [INFO] Travelled 0.00km
2016-08-14 22:22:25,174 [       cli] [INFO] Visited 0 stops
2016-08-14 22:22:25,176 [       cli] [INFO] Encountered 0 pokemon, 0 caught, 1 released, 0 evolved, 0 never seen before
2016-08-14 22:22:25,178 [       cli] [INFO] Threw 0 pokeballs
2016-08-14 22:22:25,180 [       cli] [INFO] Earned 0 Stardust
2016-08-14 22:22:25,180 [       cli] [INFO]
2016-08-14 22:22:25,183 [       cli] [INFO] Highest CP Pokemon:
2016-08-14 22:22:25,184 [       cli] [INFO] Most Perfect Pokemon:
Traceback (most recent call last):
  File "pokecli.py", line 586, in <module>
    main()
  File "pokecli.py", line 104, in main
    bot.tick()
  File "E:\Dev\GitHub\PokemonGo-Bot\pokemongo_bot\__init__.py", line 482, in tick
    if worker.work() == WorkerResult.RUNNING:
[...]

PR #3951, opened 2016-08-14
#3932 - Fix nicknaming task exception with non ascii characters

vdubus:
Fix ticket #3932 when the nicknaming task throw an exception with
pokemon's name containing non ascii characters.


BriceSD:
:+1:

douglascamata:
👍

PR #3966, opened 2016-08-15
Bring together CatchVisiblePokemon & CatchLuredPokemon in config

Quantra:
## Short Description:
***Not ready to merge***
Before I polish this up I wanted to put it here for review and to ask a couple of questions.

This PR brings together CatchVisiblePokemon and CatchLuredPokemon task in the config by using CatchPokemon which accepts booleans for catch_visible and catch_lured.

The user now has one catch related task with catch related config options.

```
"tasks": [
	    {
	      "type": "CatchPokemon",
	      "config": {
	        "catch_visible_pokemon": true,
	        "catch_lured_pokemon": true,
	        "min_ultraball_to_keep": 5,
	        "catch_throw_parameters": {
	          "excellent_rate": 0.1,
	          "great_rate": 0.5,
	          "nice_rate": 0.3,
	          "normal_rate": 0.1,
	          "spin_success_rate" : 0.6
	        }
	      }
	    }
]
```

My idea is move as many options to their related task's config as possible such that the config file is easier to navigate and groups of options will be removed when the user removes a task they don't wish to run.

My questions are... is this a good idea?  Are there any technical reason we might need some options outside of tasks?

In the case of options in a task config is it possible for me to add command line options for them?  Finally what options should be command line options and what shouldn't?


douglascamata:
@Quantra we have plans to completely drop command line setup for tasks, it should only be there for bot's configs (not task configs), so...

Quantra:
Still needs docs, example configs and deprecation message.

I've tested it and made sure the default config options are picked up correctly and that if CatchVisible or CatchLured are in tasks it switches to using a default version of CatchPokemon.  The caveat here is if the user only had CatchVisible or CatchLured the deprecation behaviour will enable both.

I've also changed the default settings for catch_parameters to match those in the example config.

pokemongo_bot/tree_config_builder.py
```diff
@@ -37,6 +38,15 @@ def build(self):
...
             task_config = task.get('config', {})
 
+            if task_type in ['CatchVisiblePokemon', 'CatchLuredPokemon']:
+                if deprecated_pokemon_task:
+                    continue
+                else:
+                    deprecated_pokemon_task = True
+                    task_type = 'CatchPokemon'
+                    task_config = {}
+                    # Best way to show deprecation message?
```
Quantra: @douglascamata

k4n30:
https://github.com/PokemonGoF/PokemonGo-Bot/pull/4520#issuecomment-241598619

Issue #3915, opened 2016-08-14
[FeatureRequest] config file in hjson


popurax:
### Problem
the bot has gone, so has edited config.json settings.
For debugging purposes, during the `tasks:[]` 100 elements of place trying to comment out and did.
But, without commented out on the json, such as ` "comment":"", ` .
Converts multiple lines commented out and json so I would like to [hjson](https://hjson.org/) can be use.

### Proposal
+ Implementation of [hjson](https://hjson.org/)
 - Preprocesser of json
 - Multiple lines can be commented out
 - There are libraries for Python

### Other Infomation
OS: Ubuntu16.04
Branch: dev

k4n30:
Please move your feature request/s to our feature hub - http://feathub.com/PokemonGoF/PokemonGo-Bot

k4n30:
Feathub may be turned off in favour of better github tags, issue templates and voting structures. Please feel free to continue using this issue (but don't delete the Feathub request just yet)

k4n30:
https://github.com/PokemonGoF/PokemonGo-Bot/issues/3843#issuecomment-239708680

rawgni:
@k4n30 should close this. mentioned in #3967 that we are staying with json.

PR #3967, opened 2016-08-15
config file in hjson #3915

rawgni:
## Short Description:
use hjson instead of json to parse config file

- #3915


douglascamata:
We already discussed this, it's not going to happen.

Issue #3953, opened 2016-08-14
Add support for FreeBSD - encrypt.so file path detection


Zapskii:
Hi,
Is there any chance FreeBSD support can be added for encrypt.so file detection?

In `pokemongo_bot/__init__.py`

replace:
`if _platform == "linux" or _platform == "linux2" or _platform == "darwin":`

with:
`if _platform == "linux" or _platform == "linux2" or _platform == "darwin" or _platform == "freebsd10":`

Thanks

BriceSD:
@Zapskii you could make a PR :)

keugnu:
I would love FreeBSD support

PR #3982, opened 2016-08-15
Update __init__.py - FreeBSD10 encrypt.so path detection

Zapskii:
Ref: https://github.com/PokemonGoF/PokemonGo-Bot/issues/3953

The change adds support for FreeBSD10 into the encrypt.so path detection routine


eggins:
Seems unbreakable and just extra features. Merged.

cliffalbert:
See my reported issue #3998, related to this one, but with added security to not crash when unsupported OS is being used. Would also be wise to add sunos5 as an supported OS.

PR #3943, opened 2016-08-14
step_walker variance 

mjmadsen:
Switched walk in config to walk_max and walk_min to add variance to stepper.
Also added variance when close to stop/pokemon.
Removed redundant parameters.
Updated configs to reflect change.

Updated for conflicts (#3857).

PR #3997, opened 2016-08-15
Randomized Walkspeed

BreezeLo:
Randomizes walk speed between 1% and config.walk_offset_percent_max if config.walk_random = true


mjmadsen:
#3943 is similar. I would appreciate input.

jimmywarting:
Haven't looked at the code, but judging by the configuration i would prefer #3943 min/max walk step then using %

mjmadsen:
@jimmywarting the initial code used a fraction to alter set walking speed. Per a request, I switched to a max and min setup.

BreezeLo:
Matter of preference I suppose, figured this was the shortest and cleanest way to do it, it would work in either case though the min/max step walk values in #3943 are good as well. In this case 1% is the minimum

mjmadsen:
You could remove the bool and just have folks set percentage to 0 if they do not want variance - but you'd need to make the minimum variance percentage 0.

BreezeLo:
I set the minimum to 1% since we are using a switch in the config for true/false ("walk_random", true), if set to false it's back to default (0)

BreezeLo:
imported the random module since importing just the function wasn't enough (for uniform)

TheSavior:
This might not be necessary now that #3943 has been merged. I'll let you guys decide and @BreezeLo, please close this PR if not necessary anymore.

pokemongo_bot/step_walker.py
```diff
@@ -2,7 +2,7 @@
 
 from cell_workers.utils import distance
 from human_behaviour import random_lat_long_delta, sleep
-
+import random
```
Quantra: I prefer the flat `import random` because it makes using the functions like `random.uniform()` or `random.choice()` more obvious what they are.  Other files in this project do `from random import uniform` then `uniform()`

BreezeLo:
Just saw @TheSavior, will close in about ~4 hours, just want to see if anyone else has any input on this one - I'm satisfied with #3943

Quantra:
@BreezeLo There's more than one way to solve most problems =]

in #3943 
`walk_min = 2`
`walk_max = 4`

is the same as in here
`walk = 200`
`walk_offset_percent_max = 0.02`

Live dangerously I want to set my walk to 200 =P

BreezeLo:
That's because there is a mistake here, should of been 

`self.speed = speed + speed * random.uniform(0.01, self.bot.config.walk_offset_percent_max)` ;) updated

Issue #3999, opened 2016-08-15
ImportError: No module named xxhash


gsmith257:
I am currently using the dev branch:
`hidden-home:PokemonGo-Bot GrantSmith$ git branch
* dev
`
I have read all the issues about this and someone made a pull request to fix it and after that I reinstalled the bot and also reinstalled the requirements. It still gives me the error that it been giving me for the past 4 days. It worked fine until the end of friday and the gave me the xxhash error.

### Expected Behavior:
`(PokemonGo-Bot) hidden-home:Documents GrantSmith$ ./pokemongobot.sh
2016-08-12 17:11:14,986 [       cli] [INFO] PokemonGO Bot v1.0
2016-08-12 17:11:14,998 [       cli] [INFO] Configuration initialized
2016-08-12 17:11:14,998 [pokemongo_bot.health_record.bot_event] [INFO] Health check is enabled. For more logrmation:
2016-08-12 17:11:14,999 [pokemongo_bot.health_record.bot_event] [INFO] https://github.com/PokemonGoF/PokemonGo-Bot/tree/dev#analytics
2016-08-12 17:11:15,247 [requests.packages.urllib3.connectionpool] [INFO] Starting new HTTP connection (1): www.google-analytics.com
2016-08-12 17:11:15,999 [PokemonGoBot] [INFO] [set_start_location] Setting start location.
2016-08-12 17:11:15,999 [PokemonGoBot] [INFO] [x] Coordinates found in passed in location, not geocoding.
2016-08-12 17:11:16,000 [PokemonGoBot] [INFO] [location_found] Location found: 34.0522300, -118.2436800 (34.05223, -118.24368, 0.0)
2016-08-12 17:11:16,000 [PokemonGoBot] [INFO] [position_update] Now at (34.05223, -118.24368, 0)
2016-08-12 17:11:16,002 [PokemonGoBot] [INFO] [login_started] Login procedure started.
2016-08-12 17:11:20,990 [PokemonGoBot] [INFO] [login_successful] Login successful.
2016-08-12 17:11:22,804 [PokemonGoBot] [INFO] 
`

### Actual Behavior:
`hidden-home:Documents GrantSmith$ ./pokemongobot.sh
Traceback (most recent call last):
  File "pokecli.py", line 39, in <module>
    from pgoapi.exceptions import NotLoggedInException, ServerSideRequestThrottlingException, ServerBusyOrOfflineException
[...]

PR #4023, opened 2016-08-15
add xxhash to requirements.txt

jrjhealey:
# Please Note (you may remove this section before opening your PR):
We receive lots of PRs and it is hard to give proper review to PRs. Please make it easy on us by following these guidelines:

## Short Description:
Add xxhash to requirements.txt

#3999 

A number of issues have pointed to needing xxhash, and have been resolved by installing it explicitly after the fact with `pip install xxhash`. latest example was #3999.

Has this been excluded from requirements.txt for a specific reason?

Seems to me this should be in the requirements.txt no? Incidentally I don't know what version of xxhash pip installs by default with the above command because investigating xxhash.__version__ returns no attribute - so others might know?

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -23,3 +23,4 @@ timeout-decorator==0.3.2
 raven==5.23.0
 demjson==2.2.4
 greenlet==0.4.9
+xxhash==0.6.1


douglascamata:
@jrjhealey it doesn't need to be explicitly in the requirements because it's declared as a dependency of other stuff we install.

jrjhealey:
Do we have any insight why people are having issues with it even when running
`pip install -r requirements.txt`?

Unless we're being lied to and they're installing manually :p

cmezh:
@jrjhealey There are no pip log in #3999, so I'm sure that OP have done something wrong.
I had no troubles with pip install on 3 different machines.

jrjhealey:
Fair enough!

Issue #4021, opened 2016-08-15
Bot crashes as it starts


yweitkemper:
### Expected Behavior

Start bot
### Actual Behavior

Bot starts logins in. after 5 seconds of waking gives me this error 
```
Traceback (most recent call last):
  File "pokecli.py", line 609, in <module>
    main()
  File "pokecli.py", line 104, in main
    bot.tick()
  File "/Users/naturalhealingways/PokemonGo-Bot/pokemongo_bot/__init__.py", line 483, in tick
    if worker.work() == WorkerResult.RUNNING:
  File "/Users/naturalhealingways/PokemonGo-Bot/pokemongo_bot/cell_workers/follow_spiral.py", line 117, in work
    if dist <= 1 or (self.bot.config.walk > 0 and step_walker == None):
AttributeError: 'Namespace' object has no attribute 'walk'
2016-08-15 11:53:23,128 [sentry.errors] [ERROR] Sentry responded with an error: 'ascii' codec can't decode byte 0x9c in position 1: ordinal not in range(128) (url: https://app.getsentry.com/api/90254/store/)
Traceback (most recent call last):
  File "/Users/naturalhealingways/PokemonGo-Bot/lib/python2.7/site-packages/raven/transport/threaded.py", line 174, in send_sync
    super(ThreadedHTTPTransport, self).send(data, headers)
  File "/Users/naturalhealingways/PokemonGo-Bot/lib/python2.7/site-packages/raven/transport/http.py", line 47, in send
    ca_certs=self.ca_certs,
  File "/Users/naturalhealingways/PokemonGo-Bot/lib/python2.7/site-packages/raven/utils/http.py", line 66, in urlopen
    return opener.open(url, data, timeout)
  File "/Users/naturalhealingways/PokemonGo-Bot/lib/python2.7/site-packages/future/backports/urllib/request.py", line 494, in open
    response = self._open(req, data)
  File "/Users/naturalhealingways/PokemonGo-Bot/lib/python2.7/site-packages/future/backports/urllib/request.py", line 512, in _open
    '_open', req)
  File "/Users/naturalhealingways/PokemonGo-Bot/lib/python2.7/site-packages/future/backports/urllib/request.py", line 466, in _call_chain
    result = func(*args)
  File "/Users/naturalhealingways/PokemonGo-Bot/lib/python2.7/site-packages/raven/utils/http.py", line 46, in https_open
    return self.do_open(ValidHTTPSConnection, req)
[...]

PR #4025, opened 2016-08-15
Fix bugs for #3943

alexyaoyang:
## Improvements:
- Changed references of walk to walk_min in follow path and spiral workers.
- Swap walk_min and walk_max if walk_min > walk_max

### DIFF
--- a/pokemongo_bot/cell_workers/follow_path.py
+++ b/pokemongo_bot/cell_workers/follow_path.py
@@ -126,7 +126,7 @@ class FollowPath(BaseTask):
             lng
         )
 
-        if dist <= 1 or (self.bot.config.walk > 0 and is_at_destination):
+        if dist <= 1 or (self.bot.config.walk_max > 0 and is_at_destination):
             if (self.ptr + 1) == len(self.points):
                 self.ptr = 0
                 if self.path_mode == 'linear':
--- a/pokemongo_bot/cell_workers/follow_path.py
+++ b/pokemongo_bot/cell_workers/follow_path.py
@@ -126,7 +126,7 @@ class FollowPath(BaseTask):
             lng
         )
 
-        if dist <= 1 or (self.bot.config.walk_max > 0 and is_at_destination):
+        if dist <= 1 or (self.bot.config.walk_min > 0 and is_at_destination):
             if (self.ptr + 1) == len(self.points):
                 self.ptr = 0
                 if self.path_mode == 'linear':
--- a/pokemongo_bot/cell_workers/follow_spiral.py
+++ b/pokemongo_bot/cell_workers/follow_spiral.py
@@ -114,7 +114,7 @@ class FollowSpiral(BaseTask):
                 }
             )
 
-        if dist <= 1 or (self.bot.config.walk_max > 0 and step_walker == None):
+        if dist <= 1 or (self.bot.config.walk_min > 0 and step_walker == None):
             if self.ptr + self.direction >= len(self.points) or self.ptr + self.direction <= -1:
                 self.direction *= -1
             if len(self.points) != 1:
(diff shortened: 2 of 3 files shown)


pokemongo_bot/cell_workers/follow_path.py
```diff
@@ -126,7 +126,7 @@ def work(self):
             lng
         )
 
-        if dist <= 1 or (self.bot.config.walk > 0 and is_at_destination):
+        if dist <= 1 or (self.bot.config.walk_max > 0 and is_at_destination):
```
BriceSD: Why ```walk_max``` and not ```walk_min``` ? 
Having both seems more appropriate to me : ```self.bot.config.walk_max > 0 && self.bot.config.walk_min > 0```

pokemongo_bot/cell_workers/follow_spiral.py
```diff
@@ -114,7 +114,7 @@ def work(self):
                 }
             )
 
-        if dist <= 1 or (self.bot.config.walk > 0 and step_walker == None):
+        if dist <= 1 or (self.bot.config.walk_max > 0 and step_walker == None):
```
BriceSD: Why ```walk_max``` and not ```walk_min``` ? 
Having both seems more appropriate to me : ```self.bot.config.walk_max > 0 && self.bot.config.walk_min > 0```

BriceSD:
Why ```walk_max``` and not ```walk_min``` ? 
Having both seems more appropriate to me : ```self.bot.config.walk_max > 0 && self.bot.config.walk_min > 0```

BriceSD:
@mjmadsen Please review, need to fix asap.

Setsu-BHMT:
The intention of the original code was to check if the bot was supposed to walk at all; given that interpretation it might be more appropriate to check walk_min, though it's probably a good idea to rethink the logic here. Checking both shouldn't be necessary as logically if walk_min > 0 then walk_max must be >0 or ==0; this is assuming of course that we have a check somewhere to make sure walk_max is greater than or equal to walk_min (do we have this?)

alexyaoyang:
Yup, from what I can see, those if condition checks if user's config allows the bot to walk at all.

Thus checking for one of them should be sufficient.

@Setsu-BHMT You are right, I just tried, we do not have that check in place. I'll add a swap if min > max.

BriceSD:
#4021

douglascamata:
The bot should not support teleporting anymore, as it will result is super fast ban. We should only support walking "normally", so those checks can be entirely removed, imho.

alexyaoyang:
@douglascamata I think checking if walking speed is > 0 is not related to teleporting?

douglascamata:
@alexyaoyang walk speed should be required and bot should use the default value of 4.16 if it's not provided in the config. If walk speed is 0, we need to bump it up to 4.16.

alexyaoyang:
@douglascamata I'm not sure if we are on the same issue here. 

One issue is checking if walk_min is > 0 in the respective follow walkers, I would recommend this condition check to stay put. 

Another issue is to disable teleporting, which I think is not related to this PR.

Final issue is bumping of walking speed of 0, which I have just committed. I took the liberty to randomize the bumped up walking speed between [3.0,5.0)

mjmadsen:
#4046 Didn't see this PR before posting. Either PR should suffice.

TheSavior:
I merged #4046. I'm really surprised this one doesn't look like it has merge conflicts. I'll let you close this one if you feel like the other was sufficient.

alexyaoyang:
Sure, closing this PR right now.

CyberMew:
Maybe some of the changes not present in the other PR could be picked over.

alexyaoyang:
@CyberMew I'll create a new PR for the changes not present in #4046.

PR #4031, opened 2016-08-15
Enhancement/sleep schedule check if current sleep is over

alexyaoyang:
- Currently, if we set up the sleep schedule task such that the bot should be sleeping now and start the bot, the bot will not sleep and continue to run until next sleep cycle before starting to sleep.
- Changed it to start sleeping whenever it should be sleeping, no matter when the bot is started.
- Added config field "skip_current_sleep_cycle" (defaults to true)

### DIFF
--- a/pokemongo_bot/cell_workers/sleep_schedule.py
+++ b/pokemongo_bot/cell_workers/sleep_schedule.py
@@ -73,14 +73,14 @@ class SleepSchedule(BaseTask):
         )
 
     def _calculate_current_sleep(self):
-        current_sleep = self._next_sleep - timedelta(days=1)
+        self._current_sleep = self._next_sleep - timedelta(days=1)
         current_duration = self._get_next_duration()
-        self._current_end = current_sleep + timedelta(seconds = current_duration)
+        self._current_end = self._current_sleep + timedelta(seconds = current_duration)
 
     def _should_sleep_now(self):
         if datetime.now() >= self._next_sleep:
             return True
-        if datetime.now() < self._current_end:
+        if datetime.now() >= self._current_sleep and datetime.now() < self._current_end:
             self._next_duration = (self._current_end - datetime.now()).total_seconds()
             return True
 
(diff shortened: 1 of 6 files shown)


pokemongo_bot/cell_workers/sleep_schedule.py
```diff
@@ -60,6 +61,8 @@ def _process_config(self):
             timedelta(
                 hours=duration_random_offset.hour, minutes=duration_random_offset.minute).total_seconds())
 
+        self.skip_current_sleep_cycle = self.config.get('skip_current_sleep_cycle', True)
```
Quantra: This shouldn't change during `work()` set it in `initialize()`

Quantra:
I tried and failed to come up with a neat solution for this.  Yours looks good at first glance =]

pokemongo_bot/cell_workers/sleep_schedule.py
alexyaoyang: It is in `_process_config()`, not `work()` :)

alexyaoyang:
@Quantra Thanks 👍

pokemongo_bot/cell_workers/sleep_schedule.py
Quantra: doh! skimmed over the `@@ -60,6 +61,8 @@ def _process_config(self):`  Sorry!  Carry on ;)

TheSavior:
If you are adding `skip_current_sleep_cycle` so that you can support not changing behavior, I really appreciate that. However, I think in this case the current behavior isn't expected by people so I'd actually be fine with not having that config and always sleep when it is meant to sleep.

Up to you though, if you think we should have the config, then I'm totally on board with that.

alexyaoyang:
@TheSavior Yes, I was thinking of the user testing the bot functions before enabling the SleepSchedule but `enabled` will be sufficient. Removed `skip_current_sleep_cycle ` from this PR.

TheSavior:
👍

PR #4033, opened 2016-08-15
shell improvement #2

simonsmh:
shell improvement.
When using run.sh you will receive a notice that you have update.
try to add support for Cygwin.
setup.sh reset will force sync to user based branch not dev.


TheSavior:
:+1:

PR #4048, opened 2016-08-16
Deprecate action_wait_min/max in favour of task specific options

Quantra:
## Short Description:
***Need help testing***

This pull request deprecates `action_wait_min` and `action_wait_max` in favour of task specific options.

RecycleItems gets `recycle_wait_min` and `recycle_wait_max`

TransferPokemon and PokemonOptimizer get `transfer_wait_min` and `transfer_wait_max`

The plan will be to have configurable delays for all tasks so the user can tweak them as they feel best.

Everything here should work but I need help confirming it does.  TransferPokemon appears to be working but my fresh account isn't recycling items yet and I don't use PokemonOptimizer.

If you can help me test please set the min/max options to the same value and when the action happens check the delay (in log) is this +/-30% for jitter.  Please delete the options and run the bot for a while to confirm default values are being used.


BriceSD:
:+1:

solderzzc:
👍

RedViper9:
#4147 
Three commas missing from config.json.example.

PR #4050, opened 2016-08-16
User configurable delay when spinning fort

Quantra:
## Short Description:
***Needs example configs updating & docs***

This PR allows the user to set min and max time spent spinning a pokestop and replaces the previous 2 second pause.

```
      {
        "type": "SpinFort"
        "config": {
          "spin_wait_min": 2,
          "spin_wait_max": 3
        }
      }
```

With #3872 and #4048 everything should have a user configurable delay in it's own task config.

### DIFF
--- a/configs/config.json.example
+++ b/configs/config.json.example
@@ -87,6 +87,10 @@
       },
       {
         "type": "SpinFort"
+        "config": {
+          "spin_wait_min": 2,
+          "spin_wait_max": 3
+        }
       },
       {
         "type": "MoveToFort",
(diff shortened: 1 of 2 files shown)


MZorzy:
Why not have a special humanpauseconfig.json with all pause and let config.json small ?

Quantra:
@MZorzy I think all task related config should be set in the task's config.  I don't want to look somewhere else in the config or another file.

BriceSD:
@Quantra You made a PR in master, open an other one in dev please

PR #4058, opened 2016-08-16
Added heartbeat threshold, it's not reasonable to send heartbeat to server every second.

solderzzc:
## Short Description:
Send heartbeat to server every second is not smart.

Added 'heartbeat_threshold' option and set default to 10.

### DIFF
--- a/configs/config.json.example
+++ b/configs/config.json.example
@@ -6,6 +6,7 @@
     "gmapkey": "GOOGLE_MAPS_API_KEY",
     "encrypt_location": "",
     "websocket_server": false,
+    "heartbeat_threshold": 10,
     "tasks": [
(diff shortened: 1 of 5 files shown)


cmezh:
Is this option should be user-configurable?

solderzzc:
yes, it is, I need update config.example.

cmezh:
Hmm... I can't imagine the reason why users would like to change this option. Or maybe I not fully understand it.

solderzzc:
They don't need to as always they use default value :)

k4n30:
@solderzzc should there be a check for if the user sets a value too low? ie less than 5 (to prevent bans maybe?) Happy to be told no

PR #4070, opened 2016-08-16
Check for tutorial completion

net8q:
## Short Description:

* Check in player_data for tutorial completion and complete the tutorial if needed.
* Add a config.nickname defaulting to config.username to be able to set a nickname in the config file
* Maybe some more tests needed (config.username may not be a good default value, do nothing until user sets nickname may be better)

### DIFF
--- a/configs/config.json.cluster.example
+++ b/configs/config.json.cluster.example
@@ -22,7 +22,15 @@
         }
       },
       {
-        "type": "CollectLevelUpReward"
+        "type": "CompleteTutorial",
+        "config": {
+          "enabled": false,
+          "// set a name": "",
+          "nickname": ""
+        }
+      },
+      {
+         "type": "CollectLevelUpReward"
       },
       {
         "type": "IncubateEggs",
--- a/configs/config.json.example
+++ b/configs/config.json.example
@@ -22,7 +22,15 @@
         }
       },
       {
-        "type": "CollectLevelUpReward"
+        "type": "CompleteTutorial",
+        "config": {
+          "enabled": false,
+          "// set a name": "",
+          "nickname": ""
+        }
+      },
+      {
+         "type": "CollectLevelUpReward"
       },
       {
         "type": "IncubateEggs",
--- a/configs/config.json.map.example
+++ b/configs/config.json.map.example
@@ -22,7 +22,15 @@
         }
       },
       {
-        "type": "CollectLevelUpReward"
+        "type": "CompleteTutorial",
+        "config": {
+          "enabled": false,
+          "// set a name": "",
+          "nickname": ""
+        }
+      },
+      {
+         "type": "CollectLevelUpReward"
       },
       {
         "type": "IncubateEggs",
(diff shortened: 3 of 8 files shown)


k4n30:
This is awesome if it works (I'm assuming you have tested it).  What about a check for team at level 5?

net8q:
I first used and modified the pgoapi example script for TOS (which actually does not work, the mark_tutorial_complete takes and array of int, not an int)

I created an account and checked the MITM traffic of the app to verify the first capture calls, and the complete sequence.

I think set_avatar will not work with the current API, it uses a nested dict so it may need https://github.com/tejado/pgoapi/pull/214 (I have merged it on my env)

Team choice would be a separate commit, but it's easy to do I think

pokemongo_bot/__init__.py
```diff
@@ -712,13 +712,122 @@ def _setup_api(self):
...
+
+        return True
+
+    def _encounter_tutorial(self):
+        # You just need to call the API with the pokemon you choose
+        # Probably can't get MewTwo as first pokemon though
+        first_pokemon_id=random.choice([1,4,7])
+        response_dict=self.api.encounter_tutorial_complete(pokemon_id=first_pokemon_id)
+        try:
+            if response_dict['reponses']['ENCOUNTER_TUTORIAL_COMPLETE']['status'] == 1:
```
net8q: Damn typo

net8q:
Tested some more and fixed the typos in tutorial encounter, which was working but returned an error anyway

solderzzc:
The PGOAPI changes need to be merged on this repo:
https://github.com/keyphact/pgoapi

solderzzc:
It's better to have a Worker class based on BaseTask.

mjmadsen:
@k4n30 If we do team assignment, it should be enabled/disabled in the config.

pokemongo_bot/__init__.py
```diff
@@ -712,13 +712,123 @@ def _setup_api(self):
...
 
+        self._check_tutorial_state()
         self._print_character_info()
         self.api.activate_signature(self.get_encryption_lib())
         self.logger.info('')
         self.update_inventory()
         # send empty map_cells and then our position
         self.update_web_location()
 
+    def _check_tutorial_state(self):
```
TheSavior: This should definitely not live in `__init__`. Can you move this to another file / class with this code and call it from here.

net8q:
I prepared the config for setting team too, but it still feel strange to set team in a task named CompleteTutorial

net8q:
Well, only think that would be needed to set the team would be to check for level > 5, and that would need access to playerstats in inventory. I'd wait until Inventory implements PlayerStats in order to avoid throwing another API call to inventory.

solderzzc:
> This branch has conflicts that must be resolved
We can leave the team assign there, people can do it by themselves if the account can be used on mobile before banned :)

solderzzc:
👍 
Thanks for the PR.

PR #4095, opened 2016-08-16
heartbeat_threshold using time delta

Quantra:
## Short Description:
@solderzzc 

This PR changes the behaviour of heartbeat_threshold such that it measures the time since it was last run and only runs if this exceeds the set threshold.

I've done this because ticks are of variable lengths and with the changes I've been making to add configurable delays they can sometimes be >10 sec between ticks for me.


solderzzc:
@Quantra  great PR, thanks a lot.

PR #4100, opened 2016-08-16
adds log to file

### LINKED ISSUE #2729: Writing Logs to external file/s
### Expected Behavior
We used to be able to separate output log, info & exception info to different logs (with for example supervisor) std_err & std_out.

```
while true; do
pkill -f pokecli
eval "python pokecli.py --config configs/config.json >> logs/log.txt" &
eval "python pokecli.py --config configs/config1.json >> logs/log1.txt" &
sleep 1800
done
```

I'm using this bash script for restarting the bots every 30 minutes. It used to keep the logs in the separate files but i guess after the changes to the logging logic it doesn't do the same. I call my bash script with this command.
"nohup ./run.sh >> logs/restart.txt &"
it was logging the exceptions into "restart.txt" and other stuff into respective log files but not anymore. Everything goes into restart.txt right now.

Can we add a config option like "pathToLogFile: './logs/log.txt'" and keep the logs in there? It may work as now when this config is not specified or false.

Edited by mod: To clarify other user wanted features

IceBreeze:
## Short Description:
Adds logging to file.

The filename is built using the account name so the file starts after the initialization of the configuration file, but the console log starts as usual.

- closes #2729


k4n30:
@IceBreeze you have merge conflicts that you will need to resolve

@douglascamata , @bstpierr & @TheSavior no love for this PR?

douglascamata:
@IceBreeze @k4n30 unfortunately this is not a good implementation. I have something in mind to enable task configuration for logger, like logging level, type (file or terminal), etc. This can be easily implemented in BaseTask class.

IceBreeze:
@douglascamata @k4n30 this was a quick response to a request received in the issue #2729.
I don't know exactly what you have in mind and my knowledge of the project is limited to a couple of hours, so don't worry :)

I also had some doubts about to rewrite the initialization, but this task requires more times to familiarize with the project.

k4n30:
@douglascamata should we close this then? or merge and update later?

solderzzc:

PR #4105, opened 2016-08-16
re-Wording followPath to more readable, recycle_items.should_run should accept <= for min_empty_space: 0

fake-or-dead:
- re-wording followPath 'position_update' for more readable

- Inventory expect to be -1 for items much more than inventory limits to make config `min_empty_space: 0` working _// refactor now_

edit:
- change recycle_items.should_run instead of return un-expected value from Inventory.Items.get_space_left()


BriceSD:
I support your idea of making ```min_empty_space: 0``` work. But if you have 0 space left, the inventory should return 0. The issue comes from recycle_items.py. Hint : take a look at ```should_run()```.

fake-or-dead:
@BriceSD oh ok

I change `should_run:` logic to 
`if inventory.Items.get_space_left() <= (DEFAULT_MIN_EMPTY_SPACE if self.min_empty_space is None else self.min_empty_space):`

instead of return -1 from  `inventory.Items.get_space_left()`

PR #4109, opened 2016-08-16
User configurable delay when spinning fort

Quantra:
## Short Description:
Same as #4050 but request merge to correct branch.

This PR allows the user to set min and max time spent spinning a pokestop and replaces the previous 2 second pause.

```
      {
        "type": "SpinFort"
        "config": {
          "spin_wait_min": 2,
          "spin_wait_max": 3
        }
      }
```

With #3872 and #4048 everything should have a user configurable delay in it's own task config.

### DIFF
--- a/configs/config.json.cluster.example
+++ b/configs/config.json.cluster.example
@@ -75,6 +75,10 @@
       },
       {
         "type": "SpinFort"
+        "config": {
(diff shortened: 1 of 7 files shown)


BriceSD:
:+1: 
One between each item recycling and one between each step could be interesting imo

PR #4129, opened 2016-08-17
SQLite backend for bot/plugin use

cwild:
## Provide SQL backend for bot

Some things would be better stored in a database. Successful login attempts perhaps (as the example in this first commit).
Longer term, a Pokemon capture log could be stored.

**Wanting to gather collective thoughts on this. Feedback very welcome**

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -106,6 +106,7 @@ out/
 # Personal load details
 src/
 web/
+data/*.db
 data/last-location*.json
 data/cells-*.json
 data/map-caught-*.json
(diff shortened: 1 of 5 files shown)


TheSavior:
@douglascamata Please review and provide thoughts based on your socket server and logging context.

BreezeLo:
got catch limit / pokestop limit logic ready here, waiting for when this is approved, looks good to me.

cwild:
Are people happy with how this was implemented? I was looking for feedback to see if it would be suitable.

Gurzeh:
@BreezeLo hi man. Do you know the limit number of Pokestop / day?

Issue #4084, opened 2016-08-16
Keep best 2 attack pokemon


Tagada22:
Hello 

Its possible to keep pokemon with the bestet attack 

Exemple :

Dragonite  :

Dragon claw :  35
Dragon : dragon pulse : 65 

Transfert pokemon 

Dragon claw : 35

Hyper beam : 120 

Keep pokemon 

http://www.pokemongodb.net/2016/05/dragonite.html?m=1

Combined this setting and iv 100% we can have the most powerful pokemon.

Maybe options is already implanted or next feature ?

Anakin5:
You can use "dps" as a sorting order in Pokemon optimizer. Check if it fits your need.

k4n30:
What determines the best attack in your mind @Tagada22 ? DPS as Anakin mention, or highest number is best?

tolinwei:
Gym fighting is not just a matter of attack value, some secondary moves have longer "Dodge Window", that allows players to dodge the still easier; some moves have lower attack value but higher DPS, which generates more damage.
Check out the link here: https://pokemongo.gamepress.gg/pokemon/149
For Dragonite, **Dragon Claw** has lowest "attack value" - 35, but it has hightest DPS, which is the best for attacking.

Foulwerp:
You can use dps_defense and dps_attack not sure how accurately these are calculated though.

dps_defense I believe would be the best for defending a gym, and the other is attacking a gym. 

I think the AI for gym defense needs a rework, Since they don't dodge and can attack only every 2 seconds makes most gyms with enough time a push over.

```
                if keep_best_dps >= 1:
                    dps_limit = keep_best_dps
                    best_dps_pokemons = sorted(group, key=lambda x: (x.moveset.dps, x.iv, x.cp), reverse=True)[:dps_limit]
                    best_pokemon_ids |= set(pokemon.id for pokemon in best_dps_pokemons)
                    order_criteria = 'Moveset DPS'   
                    
                if keep_best_dps_defense >= 1:
                    dps_limit = keep_best_dps_defense
                    best_dps_pokemons = sorted(group, key=lambda x: (x.moveset.dps_defense, x.iv, x.cp), reverse=True)[:dps_limit]
                    best_pokemon_ids |= set(pokemon.id for pokemon in best_dps_pokemons)
                    order_criteria = 'Defensive Moveset DPS'   
                    
                if keep_best_dps_attack >= 1:
[...]

PR #4144, opened 2016-08-17
Keep by best hp max

### LINKED ISSUE #3958: Keep heaviest pokemon
### Edited by moderator:
Submitting a request to have an option in the config to keep pokemon by their stats.
Example: Recycle pokemon by size.

***
#### Original Post:

Hi,

Is there a way to keep pokemons based on their weight or size instead of only iv or cp ?

Thanks

supercourgette:
## Short Description:

Add the possibility to keep pokemon by best max hp. I believe the IV is already dependent of the hp (and therefore hp max). therefore I'm not quite sure of the utility of this commit... Yet, since it was requested, I did it.

- Resolve #3958 ?


supercourgette:
Not sure why there is my old commit on the update wiki...

Foulwerp:
Can we use a list of criteria to keep based so the user can pick keep stats from most to least importance. Instead of adding a new method for each one. What if I find moveset.defense_perfection, cp, then iv the order I want them in.

For example.

```"Ekans": { "keep_best": "defensive_moveset_dps,cp,iv"}```

Instead of forcing players to choose one of the few keep bests they can define their own, on what they think is best?

k4n30:
+1 to @Foulwerp 's idea, makes it more customisable and prevents the incoming xx number of feature requests

supercourgette:
You right, I'll have a look at it.

supercourgette:
I did the changes, but I didn't have time to test them a lot though. It seems to work, but I only tried for "moveset.defense_perfection, hp_max" and not for the other (my bot don't have a lot of pokemons...). 

I removed what I did for keep_best_hp_max, so if people used it in their configuration, it will not work anymore.

And I'm not sure why there is the "Merge remote-tracking branch 'upstream/dev' into dev" here... :/

solderzzc:
This is really good, just could you please also look into the pokemon_optimizer.py? 
This is not applied to pokemon_optimizer.py.

supercourgette:
Probably answer #4084

supercourgette:
@solderzzc I'll have a look at it. pokemon_optimizer is a bit more confusing though, I will need some time to understand it.

PR #4145, opened 2016-08-17
Assign VIP status based on NCP

RedViper9:
## Short Description:
Added NCP support to pokemon_catch_worker.py.
Added logic to compare based on three variables.

Since NCP is effectively pokemon level / trainer level this seems like an improvement over absolute CP for determining catch cut offs.

I was only looking to specify anything above a certain CP% as a VIP, but this should allow you to avoid catching low CP pokemon as well.

Using a three variable comparison allows for more intricate decisions to be made by the pokemon_catch_worker. If this is a reasonable addition, it can be added to transfer_worker as well, though PokemonOptimizer largely deprecates the TransferPokemon task.

### DIFF
--- a/configs/config.json.cluster.example
+++ b/configs/config.json.cluster.example
@@ -5,6 +5,8 @@
     "location": "SOME_LOCATION",
     "gmapkey": "GOOGLE_MAPS_API_KEY",
     "encrypt_location": "",
+    "websocket_server": false,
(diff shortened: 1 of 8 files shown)


solderzzc:
> This branch has conflicts that must be resolved

RedViper9:
Screwed up with GitHub.
Mistakes were made. Lessons were learned. Will open new PR.

PR #4151, opened 2016-08-17
Fixed Config.json.X.example files

RedViper9:
## Short Description:
Added missing commas and generally homogenized the different config files.
## Fixes/Resolves/Closes (please use correct syntax):
-Fixes #4048 
-Resolves #4147

Issue #4147, opened 2016-08-17
config.map.json erorr affter update last version


cbogithub:
```
2016-08-17 16:23:14,519 [       cli] [INFO] PokemonGO Bot v1.0
configs/config.test1.json:45:8: Error: Values must be separated by a comma
   |  At line 45, column 8, offset 1168
   |  Object started at line 43, column 6, offset 1122
configs/config.test1.json:90:10: Error: Values must be separated by a comma
   |  At line 90, column 10, offset 2913
   |  Object started at line 71, column 18, offset 2154
configs/config.test1.json:117:7: Warning: Strict JSON does not allow a final comma in an object (dictionary) literal
   |  At line 117, column 7, offset 3676
   |  Object started at line 96, column 17, offset 3048
configs/config.test1.json:121:8: Error: Values must be separated by a comma
   |  At line 121, column 8, offset 3733
   |  Object started at line 119, column 6, offset 3694
configs/config.test1.json:172:6: Warning: Object contains duplicate key: u'// Zubat'
   |  At line 172, column 6, offset 5315
   |  Object started at line 161, column 15, offset 4728
configs/config.test1.json: has errors


```

```
{
    "auth_service": "ptc",
    "username": "xxxxx",
    "password": "xxxxx",
    "location": "-33.860983, 151.212546",
    "gmapkey": "",
    "encrypt_location": "",
    "websocket_server": false,
    "heartbeat_threshold": 10,
    "tasks": [
      {
        "type": "HandleSoftBan"
      },
      {
        "type": "SleepSchedule",
        "config": {
          "enabled": false,
          "time": "22:54",
          "duration":"7:46",
          "time_random_offset": "00:24",
          "duration_random_offset": "00:43"
        }
      },
      {
        "type": "CollectLevelUpReward"
      },
      {
        "type": "IncubateEggs",
        "config": {
          "longer_eggs_first": true
        }
      },
      {
        "type": "UpdateLiveStats",
        "config": {
          "enabled": false,
          "min_interval": 10,
          "stats": ["username", "uptime", "stardust_earned", "xp_earned", "xp_per_hour", "stops_visited"],
          "terminal_log": true,
          "terminal_title": true
        }
      },
      {
        "type": "TransferPokemon"
        "config": {
	        "transfer_wait_min": 1,
	        "transfer_wait_max": 4
        }
      },
      {
        "type": "NicknamePokemon",
        "config": {
          "enabled": false,
[...]

PR #4159, opened 2016-08-17
fix example config

rawgni:
- #4147 
- #4151


Quantra:
Thank you. My struggles when linting those files late at night have been merged in.  This will sort out the validation but I should probably do something with those indents... bah

PR #4161, opened 2016-08-17
Documenting docker-compose with Tor

mcrowson:
updated the installation docs to discuss the docker-compose_tor.yml. Swapped rm with down for removing docker-compose containers so that they are stopped first if the containers were launched in detached mode.


nivong:
👍

douglascamata:
👍

Issue #4062, opened 2016-08-16
Request: Ignore pokemon when have no ball inside inventory (if not, hard ban)


ipenpen:
**Problem:**
  When number of poke ball inside inventory become 0, and catchable pokemon appear, my character always try to catch the same pokemon. This action will repeat until pokemon disappear. And next day my account get hard ban ....

**Request feature:**
  Character will ignore any pokemon when number of ball inside inventory is 0.

**How to fix:**
  Add validate number of poke ball inside inventory to "pokemon_catch_worker.py" file.

Hope it help!

p/s: i donnot know how to add feature request label, some one help me plz.

avexus:
Well, this kind of behavior is suspicious.

iruy:
Is it be possible to automatically remove from task configuration: `CatchLuredPokemon` and `CatchVisiblePokemon`? And force `FollowSpiral` together with `MoveToFort` and `SpinFort` if not present. 
This would help a lot with `FollowPath` task which doesn't work very well together with `MoveToFort`

youberser:
It happened to me few times... I solved it by having a crontab and 2 config files.
Each 30 min the bot stop catching pokemons and just farm pokestops for 5 mins

antistes:
This problems exists since a while. It maybe only related to FollowPath mode. There was never really a fix beside the introduction of using ultraballs if no other ball is available. The bot stops moving when a lured or wild pokemon appears and there is no ball available. Without moving it will try to catch it forever and dont spin a fort to get new balls.

PR #4000, opened 2016-08-15
Fix no pokeballs log spamming

umbreon222:
## Short Description:
This fixes the no_pokeballs spam in console by checking pokeball count AoT. Also, fixes a possible bug in equidistant pokestops that could occur.

## Fixes (provide links to github issues if you can):
Closes #3630

PR #4168, opened 2016-08-17
stop catching pokemon if we do not have pokeballs

rawgni:
## Short Description:
check the number of pokeballs in inventory before trying to catch pokemon.

- #4062 
-
-

### DIFF
--- a/pokemongo_bot/cell_workers/catch_pokemon.py
+++ b/pokemongo_bot/cell_workers/catch_pokemon.py
@@ -2,6 +2,11 @@
 from pokemongo_bot.base_task import BaseTask
 from pokemongo_bot.worker_result import WorkerResult
 from pokemongo_bot.cell_workers import CatchVisiblePokemon, CatchLuredPokemon
+from pokemongo_bot import inventory
+
+ITEM_POKEBALL = 1
+ITEM_GREATBALL = 2
+ITEM_ULTRABALL = 3
 
 
 class CatchPokemon(BaseTask):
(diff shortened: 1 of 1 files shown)


umbreon222:
Duplicate of my PR [#4000](https://github.com/PokemonGoF/PokemonGo-Bot/pull/4000)

umbreon222:
Plus this solution only works when catching lured pokemon...

rawgni:
not entirely. This PR stop the CatchLuredPokemon and CatchVisiblePokemon to be called.

solderzzc:
When the account leveling up and recycled all low level balls, should it use the Ultraball if > Number ?

umbreon222:
You're stopping the pokemon catching too low in the call chain. You should check for no pokeballs **before** you move to them and try to catch them.

rawgni:
@solderzzc sorry, but i don't understand your question.

@umbreon222 you mean too high in the call chain ? There are two tasks (**CatchPokemon** and **MoveToMapPokemon**) that leads to the execution of **PokemonCatchWorker**. I guess I miss one of them, ie. **MoveToMapPokemon**

The reason for implementing the stop at **CatchPokemon** is to prevent the bot from wasting cycle on checking on pokemons in the vicinity.

umbreon222:
@rawgni, Unless you could make it only emit once, you shouldn't emit a message saying there are no usable balls every time the catch worker is told to work because that would still cause spamming. Also, it seems to me that your PR is more efficient than mine was(checking higher in call chain) and if that's truly the case then my PR should be reverted and yours be merged instead.

Quantra:
LGTM

Can you import `ITEM_POKEBALL` etc from somewhere?  Variables in all caps should be global constants ie they are set once and never change.  I'd expect them to be defined in a single location already.

umbreon222:
@Quantra, pokemongo_bot/item_list.py?

Quantra:
@umbreon222 yes =]

solderzzc:
@rawgni Here's my situation, there's new configuration to keep Pokemon Ball, if configured to 100 mean PokeBall+Great+Ultra (Master ?)  <= 100, so when the level is low, you got much PokeBall, but when you are above level 20 or better, you will have 100 Ultra Balls.
In this case, no normal pokemons captured.

rawgni:
@solderzzc the PR actually does let you attempt to capture the pokemon as long as the number of poke balls in your inventory is > 0 (not counting master). 

The only config I see regarding this issue is _min_ultraball_to_keep_ . There check for it in **PokemonCatchWorker** task will be the one stopping you to catch if the condition match.

solderzzc:
Got it, thanks a lot.

ipenpen:
I update today, and still have this bug.
Open this issue again pls.

bug

PR #4179, opened 2016-08-17
Further SQLite efforts to ease usage via `bot.database`.

cwild:
# Further SQL backend improvements

- `PokemonGoBot`.`database` instance will now hold the underlying sqlite connection directly. No need for ther horrible looking `bot.database.backend.connection`

- PokemonCatchWorker now correctly implements BaseTask by using `initialize()`.
- PokemonCatchWorker now inherits from a Datastore also (this is in preparation for some work that @BreezeLo is doing for monitoring daily captures)

### DIFF
--- a/pokemongo_bot/cell_workers/pokemon_catch_worker.py
+++ b/pokemongo_bot/cell_workers/pokemon_catch_worker.py
@@ -7,6 +7,7 @@ from pokemongo_bot.base_task import BaseTask
 from pokemongo_bot.human_behaviour import sleep, action_delay
 from pokemongo_bot.inventory import Pokemon
 from pokemongo_bot.worker_result import WorkerResult
+from pokemongo_bot.datastore import Datastore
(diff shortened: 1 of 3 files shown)


pokemongo_bot/cell_workers/pokemon_catch_worker.py
```diff
@@ -27,31 +28,32 @@
...
-class PokemonCatchWorker(BaseTask):
+class PokemonCatchWorker(Datastore, BaseTask):
 
     def __init__(self, pokemon, bot, config):
         self.pokemon = pokemon
-        self.api = bot.api
-        self.bot = bot
-        self.position = bot.position
-        self.pokemon_list = bot.pokemon_list
+        super(PokemonCatchWorker, self).__init__(bot, config)
```
solderzzc: Why need switch to call super init from here?

PR #4188, opened 2016-08-17
Added Daily Catch Limits & Catch Log (SQL)

BreezeLo:
Added daily pokemon catch limits (per 24 hours) using config variable "daily_catch_limit". The catches are now stored to the "catch_log" table in the database (and in data/caught-username.json file) which is queried using "SELECT" to find all catches that took place in the past 24 hours. If the limit is reached, the pokemon will not be attempted to be caught.

Default is 800 (since majority of community reports bans at at 1,000 / 24 hours)

```
# check catch limits before catch
        with self.bot.database.backend.connection as conn:       
            c = conn.cursor()
            c.execute("SELECT DISTINCT COUNT(encounter_id) FROM catch_log WHERE dated >= datetime('now','-1 day')")
            
        result = c.fetchone()
        
        while True:
            max_catch = self.bot.config.daily_catch_limit
            if result[0] < max_catch:
            # catch that pokemon!
```

I use distinct on the ecounter_id just in case there is a duplicated record for what ever reason.

image

Many thanks to @cwild for the database/migrations implementation and a lot of help and discussion on slack (starting at https://github.com/PokemonGoF/PokemonGo-Bot/pull/4129)

### DIFF
--- a/configs/config.json.cluster.example
+++ b/configs/config.json.cluster.example
@@ -148,6 +148,7 @@
     "distance_unit": "km",
     "reconnecting_timeout": 15,
     "logging_color": true,
+    "daily_catch_limit": 800,
     "catch": {
       "any": {"catch_above_cp": 0, "catch_above_iv": 0, "logic": "or"},
       "// Example of always catching Rattata:": {},
--- a/configs/config.json.example
+++ b/configs/config.json.example
@@ -156,6 +156,7 @@
     "distance_unit": "km",
     "reconnecting_timeout": 15,
     "logging_color": true,
+    "daily_catch_limit": 800,
     "catch": {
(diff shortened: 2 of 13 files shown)


BreezeLo:
note: appveyor can't build because of a deleted file within my branch (catch_log.py)

mjmadsen:
@BreezeLo From the looks of this, it records the catches from all accounts and doesn't keep them separate, is that correct? Do we think the limits are per IP and/or account?

Gurzeh:
@BreezeLo 1) Great! 2) that seem an afterward check, right? What happen when you start the bot and it reaches the limit? It stops?

mjmadsen:
@BreezeLo Since you're working on the limits, could we move this to a new task? That way we can enable/disable from the config and have it all grouped together. I would also love if we add in an account field so we can then start out account switching after limit work.

PR #4189, opened 2016-08-17
List Pokemon bag at bot start

DeXtroTip:
## Short Description:
It adds an option to log the user pokemon bag when bot starts.
I know it uses a lot of lines it's not pretty but helps you to know which pokemon you have.
It has configs to select which info to show about the pokemons (CP, IV, DPS, etc.)
It's not the most useful functionality to add right now, but i just thought it could be nice.
I was also thinking to add an option to log it to a file, it may be a better way to check this Pokemon information.


pokemongo_bot/__init__.py
```diff
@@ -823,6 +825,54 @@ def _print_character_info(self):
...
+                'level': "Level {}".format(pokemon.level),
+                'hp': 'HP {}/{}'.format(pokemon.hp, pokemon.hp_max),
+                'moveset': 'Moves: {}'.format(pokemon.moveset),
+                'dps': 'DPS {}'.format(round(pokemon.moveset.dps, 2))
+            }
+            if info not in poke_info:
+                raise ConfigException("info '{}' isn't available for displaying".format(info))
+            return poke_info[info]
+
+        self.logger.info('Pokemons:')
```
cwild: I think the `s` should be dropped.

DeXtroTip: @cwild That's true, thanks! Fixed it.

configs/config.json.example
```diff
@@ -147,6 +147,16 @@
       "max_circle_size": 50,
       "cache_recent_forts": true
     },
+    "pokemon_bag": {
+      "// if 'show_at_start' is true, it will log all the pokemons in the bag (not eggs) at bot start": {},
+      "show_at_start": true,
```
RedViper9: This can be useful, but will print a great deal of spam if unwanted. Please set default value in all configs as "show_at_start": False,

pokemongo_bot/__init__.py
```diff
@@ -732,6 +732,8 @@ def _setup_api(self):
         # chain subrequests (methods) into one RPC call
 
         self._print_character_info()
+        if self.config.pokemon_bag_show_at_start and self.config.pokemon_bag_pokemon_info:
+            self._print_list_pokemon()
```
RedViper9: It is not clear what will happen if "show_at_start" is not set in config. Perhaps it would be best to set it to default false unless True is passed in by the config.

configs/config.json.example
DeXtroTip: That's true, my bad. I'll fix it asap.

pokemongo_bot/__init__.py
DeXtroTip: The default value for "show_at_start" is defined as False at pokecli.py.

k4n30:
@DeXtroTip what about updating https://github.com/PokemonGoF/PokemonGo-Bot/blob/dev/docs/configuration_files.md

Issue #4019, opened 2016-08-15
[Feature Request] Pause if earned x amount of xp and/or caught y amount of pokemon


CyberMew:
I think it would be great if we had an option to pause/stop the bot at least temporary for the day (or depending on sleep schedules) once we have met our 'quota'.

tolinwei:
+1, I read about this in other issues (and the Reddit post of course), not sure if it's WIP.

keugnu:
#3922

mjmadsen:
On the todo list as well (#4004). I'm going to begin work on this tonight, less I find another PR already up.

mjmadsen:
Contained in #3922.

mjmadsen:
#3922 Duplicate

PR #4191, opened 2016-08-17
Anti detect bot threshold

### LINKED ISSUE #3922: [Feature Request] Anti softban / Daily Limits
From a post on the pokemongodev subreddit:
Pokestop ban:
- Occurs when you search roughly 2k pokestops in a 23 hour timespan.

Pokemon ban:
- Occurs when you catch 1k pokemon in a 23 hour timespan.

Are there any included anti-measures (take a long pause/quit when getting close to the limits, for instance) or would it be easy to include?

EDIT: Source: https://www.reddit.com/r/pokemongodev/comments/4xkqmq/new_ban_types_and_their_causes/

Other things that could be limited daily

EXP per 24 hours (or user defined)
km walked per 24 hours (or user defined)

alexyaoyang:
## Features
- Let bot rest after X amount of pokemon caught
- Let bot rest after Y amount of pokestops spun
- Randomize X and Y within configured range
- Set rest duration
- Set randomize offset for rest duration
- Closes #3922 #4019

## Help needed
- Problem: Metrics doesn't reset after sleeping and it is used to check if bot should sleep again. If metrics don't reset, bot continues to sleep after waking the first time because threshold has been met.

I coded a temporary solution by raising an exception so that the bot stops completely and the metrics are reset. Works well for running with ./run.sh to restart automatically. But I don't really like this solution. I tried looking for a metrics reset of any kind but there seem to be none. 

Any suggestions are welcome!


pokemongo_bot/cell_workers/bot_threshold.py
```diff
@@ -0,0 +1,125 @@
...
+    LOG_INTERVAL_SECONDS = 600
+
+    def initialize(self):
+        self._process_config()
+        self._calculate_threshold()
+
+    def work(self):
+        if self._should_rest_now():
+            self._rest()
+            raise Exception("Resetting metrics for next run of BotThreshold")
```
cwild: This looks a bit messy to me.

alexyaoyang: @cwild That's the line I was referring to in my PR above. Thanks for pointing it out :)

solderzzc:
We have this merged https://github.com/PokemonGoF/PokemonGo-Bot/pull/4188
Please have a look

DeXtroTip:
I have a sugestion, not sure if it can work, but give it a try if you want:
1) add more 2 variables, one for pokemon one to pokestops
2) start both at 0 at init
3) subtract each of them to '_pokemon_capturated'  and '_visited_pokestop' 
4) update them with actual metrics when '_should_rest_now()' returns True

If you need help, I'll be glad do help.

BreezeLo:
as mentioned, please refer to #4188 as it's more efficient (reason I say this is because once your session is destroyed in your example, the catches are erased from memory, so each time a user restarts the bot or crashes for any reason the counts are reset to 0 which isn't practical in this case)

alexyaoyang:
Ah yes #4188 will be a better solution. Closing this. Thanks for the suggestions!

PR #4211, opened 2016-08-18
Improvements to Windows Batches 

danielsdian:
Removed the need to attach PyYAML as it is automatically installed by
the Requirements.txt and removed the virtualenv ., as it was only needed
to run multiple bots. For this, you must now do a manual installation.
Running multiple bots require further steps anyway and virtualenv was
creating some issues on Windows systems.


danielsdian:
Bad GitHub. Bad. This wasnt meant to be created! lol

PR #4244, opened 2016-08-19
show commit hash info

### LINKED ISSUE #4241: When running on dev the version reported is not accurate
When you run the dev branch it says

`2016-08-18 21:19:01,214 [       cli] [INFO] PokemonGO Bot v1.0`

Without specifying the commit SHA.

Not sure what devs or other people thinks, but in my opinion this should have the commit that is running.

I'm asking because I'm running with docker, and i'm not sure if I've to rebuild my container after I made a `git pull`

rawgni:
## Short Description:

show commit hash info in the log.

Closes #4241

### DIFF
--- a/pokecli.py
+++ b/pokecli.py
@@ -34,6 +34,7 @@ import ssl
 import sys
 import time
 import signal
+import string
 import subprocess
 from datetime import timedelta
 from getpass import getpass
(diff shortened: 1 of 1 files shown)


solderzzc:
What if there's no git in Desktop release?

rawgni:
@solderzzc maybe we could check whether git exists.
if it doesn't, don't do anything.

what do you think ?

solderzzc:
@rawgni  it's hard to deal with multi platform for the check, or just parse the command response ?

PR #4245, opened 2016-08-19
Start to use the signature interface.

solderzzc:
Start to pass signature from bot to pgoapi.
Will keep working on this branch till ready to merge.

### DIFF
--- a/pokemongo_bot/api_wrapper.py
+++ b/pokemongo_bot/api_wrapper.py
@@ -64,7 +64,50 @@ class ApiRequest(PGoApiRequest):
         return True
 
     def _call(self):
-        signature = Signature()
+        # Need fill in the location_fix
+        location_fix = Signature.LocationFix()
+
+        sensor_info = Signature.SensorInfo(
+            timestamp_snapshot=20111,
+            magnetometer_x=-0.04073212668299675,
+            magnetometer_y=0.02653736248612404,
+            magnetometer_z=-0.10395454615354538,
+            angle_normalized_x=26.402721405029297,
+            angle_normalized_y=-35.71720886230469,
+            angle_normalized_z=-29.149093627929688,
+            accel_raw_x=0.43482502851815374,
+            accel_raw_y=0.8037761094798027,
+            accel_raw_z=0.08975112865596838,
+            gyroscope_raw_x=0.5030811429023743,
+            gyroscope_raw_y=0.33783626556396484,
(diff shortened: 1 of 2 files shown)


MZorzy:
don't like sensor data is randomly generated (#4249) . better than fixed or zero, but they must have some sense - correlation with other sensor

Calcyfer:
@MZorzy It's better than nothing at all, of course. In the future this can be calculated based on information from the walker.

But I think in the cat-and-mouse game that is avoiding bans, it's more likely an empty value will trigger one than non-logical values. Sensors aren't always accurate :)

solderzzc:
@MZorzy yes, Motion detection cost many cpu usage, so even a random value is far more better than no-value :)

net8q:
I've studied sensors info and I came to the following conclusion : 

Magnetometer info should be calculated from coordinate with something like https://pypi.python.org/pypi/geomag/ it gives you bx (North), by (East), bz (downward) component of the magnetic field at any given point on earth using the same model than the one used on Android or iOS. Most critical point here is that downward component increase as you go from equator to poles. It can be one of the sanity check niantic may use.

Accelerometer info includes gravity, so if your phone is flat on a table, x and y will be near 0, and z will be downward 9.8m/s² (Thats how your phone knows which way is up when you rotate to switch between landscape and portrait)

To return coherent values in the sensor info, a model needs to be written which takes as input : 
* Coordinate (altitude is a plus)
* Direction (assuming the phone is facing approximatively towards the direction you're going)
* Pitch (like you're walking with your phone upward at an average angle, a flat phone on a desk is sign of a spoofer)
* Maybe speed (If we want to include linear acceleration data, but I don't think Niantic can correlate between GPS speed and accelerometer at this level)

Then some vector maths and rotation matrix needs to be done to convert Magnetic field and gravitational field info from the Earth reference to the Phone reference with the given orientation/pitch. 

And raw values needs be calibrated to match the value range the device would return.

Noise can be added then :
* Magneto does not have much noise according to what I've seen on my phone, but it's not very precise either. Values are stable in a short span of time.
* Gyro and accelerometer have much noise, especially if you think that somebody is shaking the phone as he is walking

Fortunately barometer and temperature sensors are still rare, so we won't need to plug into a weather service

solderzzc:
@net8q  really nice writing, let me merge this one in and keep fine tuning this part.

PR #4250, opened 2016-08-19
Configurable altitude range 

mjmadsen:
- Added a min and max altitude range in config
- Added alt variance to calls to set_position in API

I did not change test files, or add a set_position wrapper as I initially intended. This does not also currently look up alt for each long/lat. It is purely random and is a very basic start. At the least, they can't query their data for alt == 0 and ban us all in one quick sweep. 

Big note, I am still not sure what unit altitude needs to be sent as. For the time being, I put a default 0.75 to 2.5.

### DIFF
--- a/configs/config.json.cluster.example
+++ b/configs/config.json.cluster.example
@@ -160,6 +160,8 @@
     },
     "walk_max": 4.16,
     "walk_min": 2.16,
+    "alt_min": 0.75,
+    "alt_max": 2.5,
     "debug": false,
     "test": false,
     "health_record": true,
--- a/configs/config.json.example
+++ b/configs/config.json.example
@@ -179,6 +179,8 @@
     },
     "walk_max": 4.16,
     "walk_min": 2.16,
+    "alt_min": 0.75,
+    "alt_max": 2.5,
     "debug": false,
     "test": false,
     "health_record": true,
--- a/configs/config.json.map.example
+++ b/configs/config.json.map.example
@@ -413,6 +413,8 @@
     },
     "walk_max": 4.16,
     "walk_min": 2.16,
+    "alt_min": 0.75,
+    "alt_max": 2.5,
     "debug": false,
     "test": false,
     "health_record": true,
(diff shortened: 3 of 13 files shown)


solderzzc:
Thanks a lot for the PR ! let me merge it.

mjmadsen:
@solderzzc Fantastic! Thank you for reviewing and merging so quickly.

MZorzy:
just gitpull it, start bot and alt still zero ? (lat e long edited):
2016-08-19 10:05:06,953 [PokemonGoBot] [INFO] [x] Coordinates found in passed in location, not geocoding.
2016-08-19 10:05:06,954 [PokemonGoBot] [INFO] [location_found] Location found: xx.495530, xx.603577 (xx.49553, xx.603577, 0.0)
2016-08-19 10:05:06,954 [PokemonGoBot] [INFO] [position_update] Now at (xx.49553, xx.603577, 0)
2016-08-19 10:05:06,954 [PokemonGoBot] [INFO] [login_started] Login procedure started.

mjmadsen:
@MZorzy Because your saved position last had an alt of 0. It is possible that my edits didn't save alt to account locations cache. I'll take a gander.

supergithubo:
@mjmadsen I noticed a sudden elevation on each step: 

`"alt": 16.070621059253707` to `"alt":  82.59619883853987`. 

There's a sudden 66 units (meters? feet?) of elevation since we use random alt between min and max altitude. I think the bot will be detected. It's like the user jumps very high when playing. Your thoughts?

mjmadsen:
@supergithubo It's meters. Our initial thought was at least putting something out so we're not sticking out like a sore thumb with an alt of 0 all the time. I suppose we could generate the alt randomly somewhere else and then just rely on noise. I'd prefer if we could fetch an alt from our long/lat and just use noise as we move around. 

You could look up alt for your starting coords, then change your min/max based off that number.

supergithubo:
I'm just a bit cautious on how they will detect the bot from elevating in very high altitude then suddenly goes down to low altitude on each step. I guess I'll just set min/max with a very short range in between so the bot elevates smoothly. Thanks @mjmadsen

PR #4265, opened 2016-08-19
Small fixes for altitude

mjmadsen:
- Changed default altitude values. I'm pretty sure we're supposed to send it in meters
- Small fixes where alt was missing


solderzzc:
Thanks a lot.

PR #4269, opened 2016-08-19
Impose minimum on walk/alt minimums

mjmadsen:
- walk_min must be 1 or greater
- alt_min must be 0 or greater


MZorzy:
walk_min must be ZERO or greater, it can stop, sitdown an wait for catch only
from docs : "Set the maximum walking speed (1 is about 1.5km/hr)"

mjmadsen:
This was a choice made because of the discussion from #4097.

We can lower the minimum, but we don't want very close to zero values.

PR #4272, opened 2016-08-19
Existing check on login table / migrations

BreezeLo:
Just a check to make sure the table doesn't already exist before creating


BreezeLo:
closed, one moment

PR #4274, opened 2016-08-19
Minor fix

DeXtroTip:
## Short Description:
Just a minor fix at emit event.

### DIFF
--- a/pokemongo_bot/__init__.py
+++ b/pokemongo_bot/__init__.py
@@ -387,7 +387,7 @@ class PokemonGoBot(Datastore):
         )
         self.event_manager.register_event(
             'next_egg_incubates',
-            parameters=('km_needed', 'distance_in_km',)
+            parameters=('km_needed', 'distance_in_km', 'eggs', 'eggs_inc')
         )
         self.event_manager.register_event('incubator_already_used')
         self.event_manager.register_event('egg_already_incubating')


pokemongo_bot/__init__.py
```diff
@@ -387,7 +387,7 @@ def _register_events(self):
         )
         self.event_manager.register_event(
             'next_egg_incubates',
-            parameters=('km_needed', 'distance_in_km',)
+            parameters=('km_needed', 'distance_in_km', 'eggs', 'eggs_inc')
```
mjmadsen: The call to next_egg_incubates in incubate_eggs.py only sends one parameter currently.

DeXtroTip: That's true, did not notice that.

mjmadsen:
I would love if we could configure how often we display the hatch status and and listing all eggs in progress of being hatched, if you'd like to expand this.

DeXtroTip:
@mjmadsen That's sounds a nice idea! I'll look into it.

BreezeRo:
seems to use the two extra args now

`                self.emit_event(
                    'next_egg_incubates',
                    formatted='Next egg ({km_needed} km) incubates in {distance_in_km:.2f} km (Total eggs: {eggs}, Incubating: {eggs_inc})',
                    data={
                        'km_needed': self.used_incubators[0]['km_needed'],
                        'distance_in_km': km_left,
                        'eggs': len(self.eggs),
                        'eggs_inc': len(self.used_incubators)
                    }
                )`

solderzzc:
@BreezeRo  Yes, this fixed no egg incubated log issue.

PR #4288, opened 2016-08-19
Revert "Change log color for UpdateLiveStats and UpdateLiveInventory"

BreezeLo:
Reverts PokemonGoF/PokemonGo-Bot#4230

it was too bright and too spammy in magenta color, perhaps the default live stat update timer should be set higher (like 1 minute), or using a different color


alexyaoyang:
@BreezeLo Sure, I'm open to suggestions from others. For myself, setting the interval to 120 and magenta color works very well.

BreezeLo:
solid point, I just noticed update stats are off by default anyway, no problem :dancer:

PR #4289, opened 2016-08-19
Salt hash for device_id

BreezeLo:
Salts the hash with a user defined variable (hash_salt in config) + username for device_id generated by @joelgreen in PR https://github.com/PokemonGoF/PokemonGo-Bot/pull/4287/files


surewhynotmon:
Just out of curiousity, is the whole salt stuff really more random than uuid.uuid4().hex and os.urandom(16).encode('hex')? Since it is all going to be as random as hashlib.md5 anyways. What's the idea of using a salt?

PR #4293, opened 2016-08-19
Dev

Qcasares:
* pass config into ApiWrapper and use md5 hash of username for device_id

* update to pgoapi which increments version number from 1.1.6 to 1.1.7

* include location fix in the signature, note in the future we will need a batch of location fixes per second

* use @ instead of /commit

* new api with google login fix


solderzzc:
oops, no. Don't need PR to master to keep things easier. 
Close it for now.

PR #4295, opened 2016-08-19
Berry usage threshold

alexyaoyang:
## Possible scenario
- If user always has excess berries when clearing inventory, adjust the berry threshold higher so that berries are used more often.
- Users who always have very limited berries can tweak the berry threshold lower and save more berries for VIP pokemon.


alexyaoyang:
@solderzzc Ping

solderzzc:
@alexyaoyang  I like the default value, most time that's reasonable from submit. Then users even don't need modify it in config.json

alexyaoyang:
@solderzzc  I'm happy with the defaults too. Just figured this may be useful for some 👍

Issue #4291, opened 2016-08-19
 [CompleteTutorial] [ERROR] KeyError while setting tutorial state


surewhynotmon:
### Expected Behavior
Complete tutorial 

### Actual Behavior
 [CompleteTutorial] [ERROR] KeyError while setting tutorial state

### Your FULL config.json (remove your username, password, gmapkey and any other private info)
{"config": {"// set a name": "", "nickname": "NAME", "enabled": true}, "type": "CompleteTutorial"},

### Output when issue occurred
 [CompleteTutorial] [ERROR] KeyError while setting tutorial state

### Steps to Reproduce
Run a new account with tutorial task

### Other Information
OS: Ubuntu
 
Git Commit: 
729389638a9f64015761e1987bf14d5c9a8eb434
Python Version: 
2.7.11

net8q:
Could you post complete output (censuring personal info)

surewhynotmon:
@net8q 

2016-08-19 17:52:21,452 [CompleteTutorial] [INFO] Completed legal screen
2016-08-19 17:52:24,275 [CompleteTutorial] [INFO] Completed avatar selection
2016-08-19 17:52:36,910 [CompleteTutorial] [INFO] Completed first capture
2016-08-19 17:52:36,911 [CompleteTutorial] [INFO] Trying to set testingbot1111 as nickname
2016-08-19 17:52:42,382 [CompleteTutorial] [INFO] Name changed to testingbot1111
2016-08-19 17:52:42,648 [CompleteTutorial] [INFO] Completed first time experience
2016-08-19 17:53:02,055 [CompleteTutorial] [ERROR] KeyError while setting tutorial state
2016-08-19 17:53:07,423 [CompleteTutorial] [ERROR] KeyError while setting tutorial state

I can't reproduce it after restarting though.
I added a new log.error line before it...since just reporting KeyError without the full error message is annoying. I'd have to create a new account to test it out.

net8q:
If the bot does not go through the tutorial after a restart, it means the tutorial is complete, as we are fetching tutorial state from the API. 
I may be that on the second pass, the tutorial state is an old cached value not showing that the tutorial is complete

surewhynotmon:
@net8q 

2016-08-19 22:08:06,463 [CompleteTutorial] [INFO] Trying to set NAME as nickname
[...]

PR #4296, opened 2016-08-19
Only go through CompleteTutorial once at launch

net8q:
## Short Description:
Run only the worker once as we only need to check tutorial once.

#4291


solderzzc:
Nice!~

PR #4297, opened 2016-08-19
condition logic for device_id hash/salt

BreezeRo:


pokemongo_bot/api_wrapper.py
```diff
@@ -37,10 +37,13 @@ def __init__(self, config=None):
             ApiWrapper.DEVICE_ID = hashlib.md5(key_string + salt).hexdigest()
             with open(did_path, "w") as text_file:
                 text_file.write("{0}".format(salt))
-        else:
+        elif self.config is not None and os.path.exists(did_path) == True:
```
bstpierr: Could this instead have the ifs be indented like so:

```
if self.config is not None:
    # Common to having the config
    if os.path.exists(did_path):
        # Only config and path exists
    else:
        # Only config and path doesn't exist
else:
    # Common no config
    if os.path.exists(did_path):
        # Only no config and path exists
    else:
        # Only no config and path doesn't exist
```

This makes all cases more obvious and easier to read IMO.

If there are more common things to run between the path existing or not, you can invert the ifs.

BreezeRo:
updated please revise

solderzzc:
Looks good for me.

solderzzc:
👍

solderzzc:
Close this for now :)  lol

neohunter:
Where should I put the

os.path.join(_base_dir, 'data', 'deviceid-%s.txt' % self.config.username)

Why is not on the documetnation and why is not an option to set on the config.json? this is like the most easy way for hiantic to ban users...

PR #4313, opened 2016-08-20
Pokemon optimizer round 3

Anakin5:
- Make sure optimizer try its best to reach the minimum evolve count if we want to use lucky eggs
- Do not transfer favorite pokemon and pokemon in fort
- Add database logging stuff
- Update pokemon optimizer config example
- Update pokemon.json


BreezeRo:
Looks good to me

PR #4331, opened 2016-08-20
reordered moves alphabetically

bruno-kenji:
Reordered moveset alphabetically, as @Anakin5 suggested in #4323

Not closing the issue, since people are apparently having a weird behaviour:
Missing moveset messages for Pokemon that **already has those movesets included** in pokemon.json.


bruno-kenji:
Actually nevermind, it's just that master is outdated. dev is ok now.

Issue #2673, opened 2016-08-07
UnicodeDecodeError: 'ascii' codec can't decode byte 0xc3 in position 1


InuYashaMX:
### Actual Behavior
2016-08-07 06:04:30,659 [       cli] [INFO] Most Perfect Pokemon: 
Traceback (most recent call last):
  File "pokecli.py", line 493, in <module>
    main()
  File "pokecli.py", line 129, in main
    raise e
UnicodeDecodeError: 'ascii' codec can't decode byte 0xc3 in position 1: ordinal not in range(128)

### Other Information
Python 2.7.9
OS: 
Git Commit: (run 'git log -n 1 --pretty=format:"%H"' and paste it here)
Python Version: (run 'python -V' and paste it here)

DimaVIII:
Same issue current dev commit; OSX

ftykazaf:
same issue happen after caught a pokemon, dev, osx

kanemasa1987:
I think this is the same issue I am having.

It's due to the pokemongo_bot/event_manager.py trying to print fort_name.
https://github.com/PokemonGoF/PokemonGo-Bot/blob/master/pokemongo_bot/event_manager.py#L61

The fort_name is utf-8 and causing the problem.

quick monkey patch would be replacing the line with

        if u'fort_name' in data:
            s = data[u'fort_name']
            data[u'fort_name'] = s.decode('utf8')
        formatted_msg = formatted.format(**data)

but there should be some better way...

SGColdSun:
Add the following in pokecli.py:

On line, 33 where it says import ssl
add a new line under it that says
import sys

Then add the 2 following lines 
reload(sys)
sys.setdefaultencoding('utf8')

before the if statement a few lines below that.


Here's an example of the complete file:
http://pastebin.com/cUhSeAh2

InuYashaMX:
its works thank you SGColdSun

chinmaygupta28:
Even after doing that I get this error now.

UnicodeDecodeError: 'utf8' codec can't decode byte 0x9c in position 1: invalid start byte
[...]

PR #4334, opened 2016-08-20
Fix for unicode decode error #2673

Napator:
Short Description:

Solution made by SSGColdSun which is the only one worked in my case (ubuntu 16.04 with python and pip up to date).

Fixes :
-UnicodeDecodeError: 'ascii' codec can't decode byte 0xc3 in position 1 #2673

### DIFF
--- a/pokecli.py
+++ b/pokecli.py
@@ -32,6 +32,8 @@ import logging
 import os
 import ssl
 import sys
+reload(sys)
+sys.setdefaultencoding('utf8')
 import time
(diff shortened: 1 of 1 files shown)


nivong:
lol i thought i already did this, yes please merge this asap @TheSavior

TheSavior:
I'm new to python, so I need to go learn what reload does. But why would you reload a module right after loading it and *then* change the encoding?

PR #4344, opened 2016-08-20
Transfert pokemon only when we don't have free slots

Segflow:
## Short Description:
Normally, we don't transfert pokemon after each catch, a more human-like behavior is to transfert pokemon only when we don't have enough slot.
This PR adds this feature.

### DIFF
--- a/docs/configuration_files.md
+++ b/docs/configuration_files.md
@@ -54,6 +54,7 @@ The behaviors of the bot are configured via the `tasks` key in the `config.json`
   * `max_revives_keep`: Default `None` | Maximum amount of revives to keep in inventory
 * SpinFort
 * TransferPokemon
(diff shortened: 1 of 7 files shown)


k4n30:
@Segflow please update https://github.com/PokemonGoF/PokemonGo-Bot/blob/dev/docs/configuration_files.md as well

Segflow:
@k4n30 done.

docs/configuration_files.md
```diff
@@ -54,6 +54,7 @@ The behaviors of the bot are configured via the `tasks` key in the `config.json`
   * `max_revives_keep`: Default `None` | Maximum amount of revives to keep in inventory
 * SpinFort
 * TransferPokemon
+  * `min_free_slot`: Default `5` | Once the pokebag has less empty slots than this amount, the transfer process is triggered. | Big values (i.e 9999) will trigger the transfer porcess after each catch.
```
k4n30: process*

PR #4350, opened 2016-08-20
Polyline, Altitude bugfix and Polyline support for MoveToFort

kanemasa1987:
## Short Description:
1. Fixes bug for PolylineWalker, current version teleports at specific conditions.
2. lowered config value for gps xy range (the average "norm" is now 20m, rather than the std)
3. Some GPS related messages were wrong, fixed them.
4. Now that polyline is fixed, added support for MoveToFort


kanemasa1987:
added fix for #4250

mjmadsen:
Nice work.

PR #4352, opened 2016-08-20
Display current candy count after evolution and transfer

Anakin5:
- Display current candy count after evolution and transfer

Sorry to review my copy about transfer and evolution logging. I think candy count is one of the most relevant thing we can display when performing those actions.

### DIFF
--- a/pokemongo_bot/__init__.py
+++ b/pokemongo_bot/__init__.py
@@ -342,7 +342,7 @@ class PokemonGoBot(Datastore):
         )
         self.event_manager.register_event(
             'pokemon_evolved',
-            parameters=('pokemon', 'iv', 'cp', 'ncp', 'dps', 'xp')
+            parameters=('pokemon', 'iv', 'cp', 'xp')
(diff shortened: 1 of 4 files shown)


pokemongo_bot/__init__.py
```diff
@@ -342,7 +342,7 @@ def _register_events(self):
         )
         self.event_manager.register_event(
             'pokemon_evolved',
-            parameters=('pokemon', 'iv', 'cp', 'ncp', 'dps', 'xp')
```
solderzzc: why need remove ncp/dps display, the merge issue?

Anakin5: It is just that there is so many attributes that we could display, it is hard to content everyone. Also moves and DPs are changing at evolve so it's kind of wrong info.
So I take another approach, which is sticking with the classical iv/cp and add the rewards XP,candies.

solderzzc:
got it, thanks for the PR.

Issue #3588, opened 2016-08-11
Detect config changes on the fly


Dracon23:
New bot user, following development for two weeks now...

I'm still tinkering with the bot, tweaking config here and there... which leads me to my request...

Is it possible to have to bot monitor the config and actually reload it when parts like release, recycling config or tasks to run have changed **_without_** having to restart?

fari-99:
wow, this is a good feature, but if you do that, I will recommend setting the update for every cycle for every like 15 minutes. because if you check it every cycle it will overhead. I think haha. or you can just use keyboard input to check the config. but hey that's a good feature.

Dracon23:
There are possibilities to do this in python on linux, don't know how mac and windows compare.

fari-99:
hoo, so you get "notice" the file is changing, and then restart the bot. love to see that happen. if i want to just farming pokeball, i need to delete catch task, after get enough pokebal i put the task back.

Dracon23:
@fari-99 exactly my thought.

HecticSeptic:
This is possible on Windows, one of the Po-Go-Map projects had this enabled, though it may have been due to debugging being enabled.

Dracon23:
I haven't touched windows in over two years, let alone did stuff to it (aka wrote something) at all. So I wouldn't know

Just wanted to give some input :F

Dracon23:
Necro does this, so it's def possible on win, just may not be in python

raxxon:
Just attach it to the relog cycle. Bot already has a login recycle event every 30 minutes or so, just re-read configuration when logging back in.

Dracon23:
I wouldn't want to wait for a relog cycle to be honest... I'm currently investigating a use case for a different app I work on, this shouldn't be too hard from what I gather...

k4n30:
Please move your feature request/s to our feature hub - http://feathub.com/PokemonGoF/PokemonGo-Bot

k4n30:
https://github.com/PokemonGoF/PokemonGo-Bot/issues/3843#issuecomment-239708680

sohje:
I already did this with with following changes:
Config_changed
```python
def initialize(bot, config):
    tree = TreeConfigBuilder(bot, config.raw_tasks).build()
    bot.workers = tree
    bot.metrics.capture_stats()
    bot.health_record = BotEvent(config)
    bot.config_file = config_file

def check_config_modification(config_file):
[...]

Issue #3506, opened 2016-08-10
Segregate Config to module based


asthemic:
Separate bot/pokemon/map config from authentication/location config. 

What do you think? 

- I'm not keen on having the password in plaintext in the same file as the rest of the settings. 
- Can have a separate config for emulated device settings? 
- Splitting them will let me mess around with the config whilst I'm at work on my break. :)
- Also makes requesting new users post their config easier because now you can by default just copy the example one when the config doesn't exist so it doesn't contain their account details.

k4n30:
Please move your feature request/s to our feature hub - http://feathub.com/PokemonGoF/PokemonGo-Bot

k4n30:
https://github.com/PokemonGoF/PokemonGo-Bot/issues/3843#issuecomment-239708680

bstpierr:
I would also like this feature. It would also make developers' lives easier since they would only have to update one config example when adding or changing config keys.

PR #4359, opened 2016-08-20
Config changes

sohje:
## Short Description:
Detect config changes on the fly

- https://github.com/PokemonGoF/PokemonGo-Bot/issues/3588
- Some cleanup
- Need some tests on Win systems!
- Maybe we need some delay between checks?

### p.s. 
Im totally new to this project, so im not sure that bot logic isnt broken. But it works fine for me - OSx, bots running perfectly.


doctordo0m:
IMHO this is far too much overhead for most of the people. This should definetly be an option in the config file, to be started or not (maybe even with a check frequency?)

sohje:
@doctordo0m I dont think this option should be added to config.json file cause of self-reference. Maybe CLI option with delay param?

k4n30:
@sohje merge conflicts needs resolving

pokecli.py
```diff
@@ -32,18 +32,18 @@
 import os
 import ssl
 import sys
-reload(sys)
-sys.setdefaultencoding('utf8')
```
nivong: why delete this?

sohje:
Omg.. wrong click and F$#K

danielsdian:
Would be nice if we could "Modularize" the config file in order to make it wasier to modify. Like, a section for GPS configs (GPS location, noise, altitude), another for pokemon management (capture, evolution and release), another for movement logic (Walk speeds, Spiral, cluster), stuff like that.

k4n30:
@danielsdian different Feature request found here #3506 

This FR is for detecting config changes after bot start

k4n30:
@sohje why the close?

PR #4381, opened 2016-08-20
Added FollowPath and UpdateLiveStats Settings

dsync89:
Added FollowPath and UpdateLiveStats description. The UpdateLiveStats options are taken from the source code.


TheSavior:
:+1:

PR #4400, opened 2016-08-20
Softban/Permban Slack Notifications

BreezeRo:
if "slackname" is set in config you will receive softban/permban notifications/alerts to your PM on Slack Chat!


Segflow:
This have nothing to do with the bot core itself.
What you need is a plugin, that do the job.

Issue #3773, opened 2016-08-12
[FeatureRequest] Go home


andreasMikk:
Hi,

Sorry if this is not the right place to post this but just wanted to share an idea.
I would like to suggest a Go home then quit feature.
If you are like me you will bot in your own town so you can stop the bot and wait a reasonable amount of time before you open the app at home.
It would be a nice feature if you can add your home location and get the bot to walk home and then quit.

Thanks :)

Regards
Andreas M

mico:
agree

RedViper9:
For the time being you can get behavior similar to this using the MoveToMapPokemon feature. You will need to set "update_map": false, and you may need to increase the range depending on how far away your bot is from "home". If you start the PokemonGo-Map with your home as it's coordinates, your bot should go to the area of your home and catch pokemon on the way. If you set the search radius "steps" to something small, the bot won't wander too far away from your house.

DimaVIII:
You can use this app on ypur iPhone for GPS-spoofing. No Jailbreak needed. 
https://www.youtube.com/watch?v=VVChSH45EQg

k4n30:
Please move your feature request/s to our feature hub - http://feathub.com/PokemonGoF/PokemonGo-Bot

jrjhealey:
I like this idea a lot, for those people who only want to bot a little and still play legitimately in the intervening time, or to open the app and check up on what the bot has been up to.

k4n30:
https://github.com/PokemonGoF/PokemonGo-Bot/issues/3843#issuecomment-239708680

k4n30:
going to consider this resolved by #4443

lntn:
I can't see how wake_up_at_location can help on this request. What I need is actually "sleep at location" whereby the bot auto move to a certain location then stop

RedViper9:
@Intn Let me know if I'm right in this.
@k4n30 
The request is to have the bot come back to a home area, presumably to avoid softbans resulting from switching from bot to mobile app. The wake_up_at_location option may resolve that if the bot is interrupted after sleep, but provides no benefit (in terms of avoiding softbans) compared with disabling the bot for a period of time.

lntn:
@RedViper9 you're right, it's to avoid the "teleport" effect.
Without this we will need to wait for a while before we can launch the mobile app to enjoy our catches.

lntn:
@jrjhealey yeah that's true. May I know what is the benefit for botting in US instead of UK?

[...]

PR #4442, opened 2016-08-21
Wake up at location after sleep schedule

### LINKED ISSUE #4292: Create small random change of position to resume the Bot after some time off
When you stop the Bot (by more than X hours for example) does not return the exact same position. It is strange to stay a few hours without running the bot and it resumes in the same previous position.

supercourgette:
## Short Description:

Wake up at a given location after the sleep of SleepSchedule is finished.

wake_up_at_location: (lat, long | lat, long, alt | "") the location at which the bot wake up. Note that an empty string ("") will not change the location.

- Closes #4292 (partially )
- Provide some workaround  #3773 (by choosing to wake up at home)


supercourgette:
Heu don't merge yet.

PR #4443, opened 2016-08-21
Wake up at location after sleep schedule

### LINKED ISSUE #4292: Create small random change of position to resume the Bot after some time off

supercourgette:
## Short Description:

Wake up at a given location after the sleep of SleepSchedule is finished.

wake_up_at_location: (lat, long | lat, long, alt | "") the location at which the bot wake up. Note that an empty string ("") will not change the location.

- Closes #4292 (partially )
- Provide some workaround  #3773 (by choosing to wake up at home)


pokemongo_bot/cell_workers/sleep_schedule.py
```diff
@@ -60,6 +64,22 @@ def _process_config(self):
         self.duration_random_offset = int(
             timedelta(
                 hours=duration_random_offset.hour, minutes=duration_random_offset.minute).total_seconds())
+        
+        self.wake_up_at_location = False
+        wake_up_at_location = self.config.get("wake_up_at_location", "")
```
solderzzc: If no setup in config file, should not have wake_up_at_location processing.

supercourgette:
I noted that something removed the Random Pause configuration from the docs. Should I send again?

solderzzc:
The random pause is very needed, reset to start point will prevent bot traveling all the world.
So if you can get that part back, will be very good.

pokemongo_bot/cell_workers/sleep_schedule.py
supercourgette: It is because in def work(), I had to check if wake_up_at_location is setup. I presumed it would not make a difference using a var or checking in config_file. Will fix.

supercourgette:
Updated the randomPause docs while I was at it. :)

solderzzc:
There should be altitude and a GPS noise.

supercourgette:
The altitude was already there (even if it wasn't visible in the example configs). The GPS noise is applied automatically in the api_wrapper > def set_position (unless I'm missing something :) ).

solderzzc:
Thanks for the pr!

PR #4446, opened 2016-08-21
Pokemon optimizer improvements

Anakin5:
- Refresh web inventory when pokemon inventory change
- Take into account the candy rewarded at evolution. This is to fix a borderline case. ex:

If we have 23 candies and 2 Pidgey (at 12 evolution cost), the previous formula give us that only one of them can be evolved:
(23 + 2) / (12 + 1) = 1

The fixed formula return that both can be evolved (which is correct):
(23 + 2 - 1) / 12 = 2


pokemongo_bot/cell_workers/pokemon_optimizer.py
```diff
@@ -29,13 +29,13 @@ def initialize(self):
         self.max_pokemon_storage = inventory.get_pokemon_inventory_size()
         self.last_pokemon_count = 0
 
-        self.config_transfer = self.config.get("transfer", False)
-        self.config_evolve = self.config.get("evolve", False)
+        self.config_transfer = self.config.get("transfer", True)
```
mjmadsen: I'd prefer things to be off by default.

Anakin5:
done

pokemongo_bot/cell_workers/pokemon_optimizer.py
```diff
@@ -34,8 +34,8 @@ def initialize(self):
         self.config_evolve_time = self.config.get("evolve_time", 20)
         self.config_evolve_for_xp = self.config.get("evolve_for_xp", True)
         self.config_evolve_only_with_lucky_egg = self.config.get("evolve_only_with_lucky_egg", False)
-        self.config_evolve_count_for_lucky_egg = self.config.get("evolve_count_for_lucky_egg", 90)
-        self.config_may_use_lucky_egg = self.config.get("may_use_lucky_egg", False)
+        self.config_evolve_count_for_lucky_egg = self.config.get("evolve_count_for_lucky_egg", 92)
+        self.config_may_use_lucky_egg = self.config.get("may_use_lucky_egg", True)
```
mjmadsen: Could we revert this as well?

Anakin5:
done

solderzzc:
Thanks for the PR.

PR #4467, opened 2016-08-21
Increased time outs to not getting banned

nivong:
Changed the times out we should also add a random time out to the envolve time #TODO
(this CAN be merged)


Anakin5:
It is way too much I think. It does not take 10 seconds to make one transfer. By hand and without rushing, you are at one every 3 seconds.
Time for evolve is the same as transfer plus the animation time. It is not 35 seconds.

mjmadsen:
Would prefer this be merged to dev first.

@Anakin5 is correct, IMO. I could see 3-6 as a more reasonable default value.

Calcyfer:
Just did a quick test using the app based on 10 evolves and transfers : 
Evolve : 22 - 26 seconds
Transfer : 2 - 5 seconds

k4n30:
Please create PR against dev

PR #4470, opened 2016-08-21
Add .get_alt() method to Polyline class 

th3w4y:
## Short Description:
Use Google Maps Elevation api to get the elevations along the path of the polyline
## Fixes
- Implements .get_alt() method


MZorzy:
i think is better pick altitude at start.
then use some "noise" when walk, and upgrade altitude after x time, so low use google API (will not work after 2500 query?)

th3w4y:
@MZorzy this uses only one API request per Polyline object created
Plus to mention we have 2500 Direction requests and 2500 Elevations requests, thus they consume a different quota since is a different API

PR #4472, opened 2016-08-21
Update live stats emit's all stats (handy for socket listeners)

brantje:
Current `UpdateLiveStats` only emits a string.
Rather than emitting all stats. 
This PR fixes that problem.


k4n30:
Hahaha travis got serverbusy or offline

brantje:
Saw it, was wondering wat failed the check xD

solderzzc:
@brantje, you should fix the CI build first :) lol, you get a tick.

PR #4483, opened 2016-08-21
Improve IncubateEggs

DeXtroTip:
## Short Description:
Now it shows all eggs currently incubating, ordered from the first going to hatch to the last.
Added "min_interval" setting at config file to print eggs info periodically. It will still show if an egg hatched or was incubated outside this interval.


solderzzc:
Oh, I need revert this since the CI build failed.

k4n30:
@DeXtroTip Care to recreate as this was reverted

solderzzc:
I think this my mistake ? I fixed another issue caused the CI thing. Any advise ?

solderzzc:
OK, I didn't revert this one, just created a revert PR and then closed it.

PR #4487, opened 2016-08-21
Add iv setting to rename pokemon

developerllazy:
## Short Description:
add the function to setting the iv, only rename the pokemon if the iv is above the value according to the settings. It try to minimize the number of pokemon to be renamed.

-
-
-


BreezeRo:
Looks good, please add to the other versions of config.json and update the config document with the new variable.

Anakin5:
That would be nice to be a bit more generic. What about cp ? attack ? defense ? stamina etc... should we add a parameter for all of them ?
Maybe the parameter could be a dictionary holding the minimum values for all of that:
```
"min_values": {
    "iv": 0.8,
    "cp": 1200
}
```

solderzzc:
CI build failed

solderzzc:

PR #4492, opened 2016-08-21
Added some variance in max amount of pokemon bag before releasing

Napator:
For more human behaviour.

### DIFF
--- a/configs/config.json.cluster.example
+++ b/configs/config.json.cluster.example
@@ -35,7 +35,8 @@
       {
         "type": "IncubateEggs",
         "config": {
-          "longer_eggs_first": true
+          "longer_eggs_first": true,
+          "min_interval": 120
         }
       },
       {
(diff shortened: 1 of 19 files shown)


solderzzc:

BriceSD:
Lets say min_free_slot = 0. In other words, transfer pokemons once my bag is full. It’ll take a lot of try before transferring, you’ve 5% chance to do so.
Same issue when 0 ≤ min_free_slot ≤ 20.

Do you use every possible option ? Personally I don’t like to be forced to use one. What about adding this to the config file ?

Anakin5:
Seems to me that  it is going to rand a new number at every loop, so random distribution  is very bad and you are going to release at your configured value or very close to it most of the time.

solderzzc:
@Anakin5 so what should we do? Any further suggestion?

Anakin5:
Rand a number in the init and after work is done. That way, you will hold onto that number.
Also, check you always stay between 0 and inventory size.

PR #4503, opened 2016-08-21
Pokemon list candies

DeXtroTip:
## Short Description:
Added option to show candies for each pokemon when printing all pokemon at start.


solderzzc:
nice to have.

Issue #4337, opened 2016-08-20
Web inventory doesn't get updated


NovaDev94:
### Expected Behavior

`web/inventory-<username>.json` file get updated after Pokemon catched, Fort spinned,...

### Actual Behavior

`web/inventory-<username>.json` is only written on bot start, in the `Inventory` constructor (calling `self.refresh()`). The file doesn't get updated after that.  

### Other Information

I got this unexpected behavior after #4238 is merged. Maybe some calls is missing to update the web inventory, or that's the new expected behavior?

Anakin5:
That is expected but that would be nice to do some thing about it.
The inventory.json was supposed to hold the full response from the api, but since we do not call the api anymore, it is not updated.

Which lead to other questions:
- Can we change the format of the inventory.json to store our internal inventory instead of the api one ?
- Would it require changes in the web module ?
- What is the status of the web interface anyway ? I do not yet understand why there is some websocket, event logger, database etc... since there is no interface to communicate with them.

k4n30:
@douglascamata & @brantje ^

NovaDev94:
I think the only way to monitor the "catched" Pokemon now is to use the web interface since the desktop version is pretty much broken for now.  

Could we just provide a config for "refreshing the inventory & dump to file" with interval? That config will be disabled by default for sure.

Gobberwart:
The web interface is broken as well. api inventory refresh is now only called once during recycle, and then only if min_empty_space is exceeded. In my case, that's about once per 90 minutes.

Should be refreshing inventory once per bot tick? Or at least every few minutes.

NovaDev94:
@Gobberwart The web interface still works fine for me (with the purpose of managing my Pokemon list and also the items).  

It's sad that I can't do it anymore with the new cached inventory :(.

BriceSD:
We’re moving away from writing in files. With our new map we won’t need it anymore. That being said, it’s not ready yet. I guess there is quite a lot of people who want this. Refreshing the inventory at every tick could be a temporary solution.

Gobberwart:
I've included this in my PR #4513 as part of the item recycler update. Once (if) approved/included, the web inventory should update every tick, without needing api call.

PR #4507, opened 2016-08-21
Update web inventory at every tick

BriceSD:
## Short Description:
Now update again web inventory
## Fixes/Resolves/Closes :
- #4337

### DIFF
--- a/pokecli.py
+++ b/pokecli.py
@@ -116,6 +116,7 @@ def main():
 
                 while True:
                     bot.tick()
+                    inventory.refresh_inventory()
 
             except KeyboardInterrupt:
                 bot.event_manager.emit(


pokecli.py
```diff
@@ -116,6 +116,7 @@ def get_commit_hash():
 
                 while True:
                     bot.tick()
+                    inventory.refresh_inventory()
```
solderzzc: Is this trigger API protocol call to contact server?

BriceSD: Yes it does. Once we got our new map we’ll get rid of this.

solderzzc: You mean the local file server or The Niantic server?

BriceSD: Niantic server ```inventory = self.bot.api.get_inventory()```

solderzzc: The mobile app contact server every 10s as know as the heart, we should not change the api sequence. Even now, we have much work to refactor the call sequence.

BriceSD: An other solution could be to add something like ```inventory.to_json()```. This way we wouldn’t need to make a call to the server.
But this seems pretty annoying to do. Definitely not on top of my priority list, I won’t do it myself.

solderzzc: ok, so we are good here :)

BriceSD: Could also move this to the heartbeat method. It’s up to you.

solderzzc:
We'd better compare the MITM package with real app. To understand what do they including in the heartbeat.

pokecli.py
k4n30: Should be in the heartbeat function, not everyone uses the webui and not everyone should have to suffer a potential ban by doing api calls every tick because of a small minority who want a fancy gui

BriceSD:
Closed in favour of #4513

Issue #4466, opened 2016-08-21
Altitude still hardcoded to 8.0


NatNew0:
Hi,

I noticed that the altitude parameter is still hardcoded to 8.0. 
I've set     "gps_default_altitude": 209.3,
and turned on Debug mode and look for 

latitude: xx.xxxxxx
longitude: yy.yyyyyy
altitude: 8

--> It's always using 8 (or is the debug output wrong ?)
The settings seems to be hardcoded from the api here: "./pgoapi/pgoapi/rpc_api.py:        request.altitude = 8  # not as suspicious as 0"

mjmadsen:
I've been a bit curious if changes dealing with altitude were being applied to API requests. From conversations with the pgoapi folks, it accepts and sends altitude, rather than using a static value. 

It is quite likely we have output errors. I'll stick this as a bug and hopefully we can find something.

kanemasa1987:
@NatNew0 
Can you try changing this line?
https://github.com/PokemonGoF/PokemonGo-Bot/blob/dev/pokecli.py#L553

I think this is the only 8 we have.

mjmadsen:
@kanemasa1987 We have to have some value in there as a default. It's only used to initialize our self.alt, which is later generated at random or pulled from somewhere (geocoder, etc).

NatNew0:
I was wondering if the value we set in the config.json is really  used.
I changed the default value in pokecli.py to 188.3:
    add_config(
        parser,
        load,
        long_flag="--gps_default_altitude",
        help="Initial altitude (default=8.0)",
        type=float,
        default=188.0

in the config.json there's a value of 180.3 set.
then the debug startup looks like:

2016-08-21 13:33:00,458 [pgoapi.pgoapi] [DEBUG] Set Position - Lat: xx1.xxxxxxxxxx Long: yy1.yyyyyyyyy Alt: 152.151282236
2016-08-21 13:33:00,459 [PokemonGoBot] [INFO] [login_started] [97mLogin procedure started.[0m
2016-08-21 13:33:00,459 [pgoapi.pgoapi] [DEBUG] Set Position - Lat: xx1.xxxxxxxxx Long: yy1.yyyyyyyyyy Alt: 180.3
2016-08-21 13:33:00,459 [pgoapi.pgoapi] [DEBUG] Auth provider: ptc

After this the requests look like:
requests {
  request_type: DOWNLOAD_SETTINGS
[...]

PR #4518, opened 2016-08-22
Update to latest pgoapi

alexyaoyang:
## Fixes
- #4466
- Latest pgoapi checks for user set altitude before defaulting it to 8.


solderzzc:
Oh, great patch !

PR #4519, opened 2016-08-22
Dev Only

oralunal:
## Short Description:
These lines maybe work in dev version but master version didn't updated yet and these tasks make the bot throw error

- Removed lines that work in dev but don't work master.


solderzzc:
We need batch process the dev to master merge.

PR #4520, opened 2016-08-22
abandon-lured-pokemons-when-not-enough-pokeball

leanhdaovn:
- Abandon lured pokemons when there's not enough pokeball
- Refactor code for checking remaining pokeball quantity

### DIFF
--- a/pokemongo_bot/base_task.py
+++ b/pokemongo_bot/base_task.py
@@ -1,5 +1,21 @@
 import logging
+from pokemongo_bot import inventory
+from pokemongo_bot.inventory import Pokemon
 
+ITEM_POKEBALL = 1
+ITEM_GREATBALL = 2
+ITEM_ULTRABALL = 3
+
+LOGIC_TO_FUNCTION = {
+    'or': lambda x, y, z: x or y or z,
+    'and': lambda x, y, z: x and y and z,
+    'orand': lambda x, y, z: x or y and z,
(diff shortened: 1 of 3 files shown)


mjmadsen:
Note commit 695ca70 no longer uses catch_visible_pokemon and catch_lured_pokemon. There is also a ball check.

leanhdaovn:
I think that commit works just fine @mjmadsen.
Only the edge case of min_ultraball_to_keep for vip pokemons hasn't been handled.

k4n30:
@leanhdaovn not only is there merge conflicts that needs resolving, but you have made significant changes to base task which IMHO is bad as no other tasks that are pulling from base task need these changes and will need testing against all those who inherit from base task.

You should move all your changes to the function of which you are trying to fix, which is in this case the "[catch pokemon](https://github.com/PokemonGoF/PokemonGo-Bot/blob/dev/pokemongo_bot/cell_workers/catch_pokemon.py)" worker which merged catch lured and catch visible pokemon recently.

@Quantra Are you planning on doign the rest of the deprecation of catch lured, and catch visible files as they are going to cause issues. My suggestion is to fully deprecate them, rather than allow users to use both options depending on their config, as changes will need to be made to both options

leanhdaovn:
@k4n30 I think we can close this PR since the commit @mjmadsen mentioned above addressed the problem

PR #4523, opened 2016-08-22
Consolidate catch workers

### LINKED ISSUE #4524: Catch worker edge bug
With the merger of the catch worker tasks, we have an error that will cause two catch attempts in one call to work if we have only one visible pokemon and at least one lured pokemon.

This could be the cause that folks see fewer pokemon over time (server throttle). 

PR #4523

mjmadsen:
Consolidated catch workers to one file.
Fixed bug that caused us to catch two pokemon in one call to work if we have a single visible pokemon and at least one lured pokemon available.

Closes #4524

### DIFF
--- a/pokemongo_bot/cell_workers/catch_pokemon.py
+++ b/pokemongo_bot/cell_workers/catch_pokemon.py
@@ -15,7 +15,6 @@ class CatchPokemon(BaseTask):
             return WorkerResult.ERROR
 
         pokemon = []
-
         if self.config.get('catch_visible_pokemon', True):
             pokemon = get_visible_pokemon()
         if self.config.get('catch_lured_pokemon', True):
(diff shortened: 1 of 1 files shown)


mjmadsen:
I suppose we should move the list builder and sorting to init and pop the list every time we send it to the worker. Sound reasonable?

solderzzc:
I am not sure about the current logic. Is this true:
1. Get obj map
2. run into worker and get pokemon list
3. saved to catch list, catch in catch worker

mjmadsen:
@solderzzc the logic for building each part (catchable, wild and lured) is the same as it was in the lured/visible workers, except I tossed them together as the actual catch function from each is the same.

solderzzc:
let's merge and get more test result.

mjmadsen:
@solderzzc But yes, that seems to be the general flow.

PR #4530, opened 2016-08-22
not always snipe vip pokemon if using snipe_high_prio_threshold

anhvn01457:
## Short Description:
- I think we don't need to always snipe vip pokemon, if that pokemon is needed for sniping, it should have higher priority than threshold. By that we can sure vip pokemon is till catch with Berry & Best Ball but no need to snipe (avoid ban)

### DIFF
--- a/pokemongo_bot/cell_workers/move_to_map_pokemon.py
+++ b/pokemongo_bot/cell_workers/move_to_map_pokemon.py
@@ -271,7 +271,7 @@ class MoveToMapPokemon(BaseTask):
 
         if self.config['snipe']:
             if self.snipe_high_prio_only:
-                if self.snipe_high_prio_threshold < pokemon['priority'] or pokemon['is_vip']:
+                if self.snipe_high_prio_threshold < pokemon['priority']:
                     self.snipe(pokemon)
             else:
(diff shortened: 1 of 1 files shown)


solderzzc:
Snipe and vip should not merged in configuration.
The snipe score is very clear.

anhvn01457:
@solderzzc how about the case when I want Dratini is Vip to secure catch rate but don't want to snipe on Dratini?

solderzzc:
Hmm, I don't get it. But I realized one reason for this commit, `catch with Berry & Best Ball`

anhvn01457:
Yes, it is. I want that pokemon is catch with Berry & Best Ball but not snipe on it

solderzzc:
Oh, I was thinking the config is in the reverted way.

solderzzc:
Why you just don't enable the Snipe task?

anhvn01457:
I want snipe on other pokemons.
Example:
my vip pokemon is Dratini, Dragonite because I want to catch them with berry & best ball. But I only want snipe on Dragonite

solderzzc:
I see, that's reasonable.

PR #4531, opened 2016-08-22
Secure catch rate if pokemon is vip

anhvn01457:
## Short Description:
- Secure hit rate with VIP pokemon

### DIFF
--- a/pokemongo_bot/cell_workers/pokemon_catch_worker.py
+++ b/pokemongo_bot/cell_workers/pokemon_catch_worker.py
@@ -426,7 +426,7 @@ class PokemonCatchWorker(Datastore, BaseTask):
             )
 
             hit_pokemon = 1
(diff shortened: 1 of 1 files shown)


pokemongo_bot/cell_workers/pokemon_catch_worker.py
```diff
@@ -426,7 +426,7 @@ def _do_catch(self, pokemon, encounter_id, catch_rate_by_ball, is_vip=False):
             )
 
             hit_pokemon = 1
```
solderzzc: Hmm, what's the meaning of `hit_pokemon` ?

anhvn01457: I think that variable is for throw ball right to pokemon or not

PR #4546, opened 2016-08-22
Fix for catch worker no usable ball loop

### LINKED ISSUE #4533: No usable pokeballs found
When "No usable pokeballs found" happens, the bot just hanging there and not tried to move and spin the fort.

mjmadsen:
If we run out of pokeballs while in the catch worker, we were getting into an endless loop.

Now returning an error to stop the process.

Closes #4533


mjmadsen:
Wrong again.

PR #4563, opened 2016-08-22
Print ignored pokemon

mjmadsen:
From @helloj

We weren't processing the ignore list.


mjmadsen:
Merging since this is just an output update.

PR #4567, opened 2016-08-22
Better edge case output on catcher worker

mjmadsen:
- Now indicate if we're trying an ultra if we have more than min.
- State no balls if we're completely out.


mjmadsen:
Merging (just an output change)

PR #4574, opened 2016-08-22
Escape slash and double quotes characters for password in setup.sh

hlung:
Passwords may contain special characters that may break bash `sed` command (slashes) or json format (double quotes). This allow those kinds of password to be used safely.

PS. My bash skill is not strong so any advice is appreciated! :)


solderzzc:
Thanks for the PR>

PR #4575, opened 2016-08-22
Fix teleporting when Google API directions quota reached

th3w4y:
## Short Description:
Polyline improvements

## Fixes/Resolves
- Fix teleporting when API quota reached
- Add unittest tests for the Polyline class (including pickles for directions and elevation responses)
- Add requests-mock to requirements.txt
- Remove the old ad-hoc polyline_generator_tester.py


solderzzc:
Good style, less coding, more testing code :)

Issue #4559, opened 2016-08-22
Treat as VIP missing Pokemon in Pokedex


Gurzeh:
### Short Description
Treat all missing (in Pokedex) Pokemon as VIP (or secure catch) ?

### Possible solution
Obvious: update config, but a programmatic solution would be more elegant

### How it would help others
For those who want to fill their Pokedex (and others)

mjmadsen:
I like the idea. We could probably add something to append those missing from our dex (which we already fetch) to the VIP list. Shouldn't be TOO hard.

PR #4580, opened 2016-08-22
Treat as VIP missing Pokemon in Pokedex

sohje:
## Short Description:
Treat as VIP missing Pokemon in Pokedex
- #4559 

Check pokemon_id in pokedex. If not seen -> treat as VIP.


solderzzc:
Oh? This is cleaver. Waiting for CI-build.

mjmadsen:
Can we add this to the config?

PR #4582, opened 2016-08-22
Split payload body into two variables (latitude/longitude) 

BreezeRo:
Just the coords for now ;) may parse JSON object later instead.


BreezeRo:
updated - for sessionID from encounter_id. merging..

PR #4587, opened 2016-08-22
SessionID / Latitude & Longitude stuff for map-chat

BreezeRo:
Basically, a user runs a google map on localhost with simplehttpserver, and this map is fed with messages through mqtt. The messages are read and variables extracted to display them onto the map.

The idea is to allow the bots to communicate with each other without compromising source of data.


BreezeRo:
closing and reopening with fresh branch

PR #4581, opened 2016-08-22
Step walker changes altitude gradually

alexyaoyang:
## Features
- Step walker now supports altitude, changes altitude gradually +/- some noise.
- FollowPath path.json now supports altitude too.
- Supports old path.json files without altitude too.

PR #4596, opened 2016-08-23
Map-Chat / Crowd Sourced Data

BreezeRo:
This goes back to PR https://github.com/PokemonGoF/PokemonGo-Bot/pull/3400 wish should of been discussed a bit more as I don't think people realized the scope at which @solderzzc was talking of going with this is HUGE!! Here is a working example (with his previous PR + help of course) of a crowd sourced map. 

Basically the idea is to use a relay system (mqtt here) to recieve messages from the client (containing pokemon spawn coordinates in this example). Each client subscribes to the topic (in this case "chat" and pokemon coordinates), and this information is relayed across the network of all bots.

Think "torrent". In this way the bots are effectively able to communicate with eachother, and as demonstrated in our example, plot the pokemon spawns on a map using data gathered from all bots that are broadcasting. The source of the coordinates is stripped away from the message, so no one knows where it came from.

As it stands now we are using a test mqtt broker, which is perfect considering we are in development branch. The idea is that the community will host or create their own mqtt brokers (relay points or "hubs"), and transmit the data to those brokers and receive them the same way as in this example. This will allow us to display all kinds of information on the map such as pokemon spawn locations, high ban count zones, lured forts, ect.

In this working example the bot client transmits coordinates of catchable pokemon to the relay server, strips identifying data, and these coordinates are then readable by all other bot clients to work with.

image

To view the map and chat/communicate

`cd map-chat`
`python -m SimpleHTTPServer 8080`

And open 127.0.0.1:8080 in your browser. To disable the feature as it's presented here, change "enable_social" in config

### DIFF
--- a/pokemongo_bot/__init__.py
+++ b/pokemongo_bot/__init__.py
@@ -25,7 +25,7 @@ from event_manager import EventManager
 from human_behaviour import sleep
 from item_list import Item
 from metrics import Metrics
-from pokemongo_bot.event_handlers import LoggingHandler, SocketIoHandler, ColoredLoggingHandler
+from pokemongo_bot.event_handlers import LoggingHandler, SocketIoHandler, ColoredLoggingHandler, SocialHandler
 from pokemongo_bot.socketio_server.runner import SocketIoRunner
 from pokemongo_bot.websocket_remote_control import WebsocketRemoteControl
 from pokemongo_bot.base_dir import _base_dir
@@ -117,7 +117,8 @@ class PokemonGoBot(Datastore):
             handlers.append(ColoredLoggingHandler())
         else:
             handlers.append(LoggingHandler())
-
+        if self.config.enable_social:
+            handlers.append(SocialHandler(self))
         if self.config.websocket_server_url:
             if self.config.websocket_start_embedded_server:
                 self.sio_runner = SocketIoRunner(self.config.websocket_server_url)
@@ -1059,14 +1060,14 @@ class PokemonGoBot(Datastore):
             possible_coordinates = re.findall(
                 "[-]?\d{1,3}[.]\d{3,7}", location_name
             )
-            if len(possible_coordinates) >= 2:
+            if len(possible_coordinates) == 2:
                 # 2 matches, this must be a coordinate. We'll bypass the Google
                 # geocode so we keep the exact location.
                 self.logger.info(
                     '[x] Coordinates found in passed in location, '
                     'not geocoding.'
                 )
-                return float(possible_coordinates[0]), float(possible_coordinates[1]), (float(possible_coordinates[2]) if len(possible_coordinates) == 3 else self.alt)
+                return float(possible_coordinates[0]), float(possible_coordinates[1]), self.alt
 
         geolocator = GoogleV3(api_key=self.config.gmapkey)
         loc = geolocator.geocode(location_name, timeout=10)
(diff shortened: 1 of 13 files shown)


pokemongo_bot/__init__.py
```diff
@@ -1059,14 +1060,14 @@ def get_pos_by_name(self, location_name):
...
-            if len(possible_coordinates) >= 2:
+            if len(possible_coordinates) == 2:
                 # 2 matches, this must be a coordinate. We'll bypass the Google
                 # geocode so we keep the exact location.
                 self.logger.info(
                     '[x] Coordinates found in passed in location, '
                     'not geocoding.'
                 )
-                return float(possible_coordinates[0]), float(possible_coordinates[1]), (float(possible_coordinates[2]) if len(possible_coordinates) == 3 else self.alt)
+                return float(possible_coordinates[0]), float(possible_coordinates[1]), self.alt
```
alexyaoyang: @BreezeRo Could you change line 1063 and 1070 back? This was merged in PR #4581.

BreezeRo:
@alexyaoyang restored, sorry about that, branches were minutes apart

alexyaoyang:
@BreezeRo All is good 👍

MasterCATZ:
perfect , exactly what I have been waiting for , shared scanner data 

now all that will be needed is an verification system so if multiple bots are is getting false spawns it can be removed from the swarm ? ( ie someone falsifying data )

BreezeRo:
@MasterCATZ absolutely, through use of the encounter_id + coordinates we can somewhat verify if the coordinate is legit, next step is to perhaps ensure that the long/lat is actually valid and not in the ocean before allowing it to be read. I'm sure some guys around here got some good ideas to input on this. Very good point you bring up.

Nik0lashka:
@BreezeRo could you please reduce logging, cause people will turn off this feature and we will loose part of data, collected from enabled-by-default config #4602

TheCannings:
I'm getting an error of:

Oops! Something went wrong.
This page didn't load Google Maps correctly. See the JavaScript console for technical details.

I check the console and see:

Google Maps API warning: NoApiKeys https://developers.google.com/maps/documentation/javascript/error-messages#no-api-keys

Is this working off the google maps api key in the config.json or should we be specifying this elsewhere?

hobbes3:
Same as @TheCannings. Where do I put the Google Maps key?

newjanson:
yap ^

danielsdian:
When this goes live, we need to modify the PokemonGo-Bot-Start batch to start this webserver too. Right now, it only starts the bot and the PokemonGo-Bot Map.

BreezeRo:
The key message is only a warning and can usually be ignored, let me see if I can reproduce it

danielsdian:
FYI, the chat option doesnt work on Microsoft Edge. Maybe should add a warning about it. you can see the map and the dots, but the text boxes doesnt pop up.

danielsdian:
Also, on Chrome all the messages are popping up as "undefined".

BreezeRo:
thats normal Daniels, please read the original OP, this is a work in progress ;)

danielsdian:
I know, just providing some feedback.

daemonsyn:
Anyone with this error?

`    self.mqttc.publish("pgomapgeo/"+geo_hash+"/"+str(data['pokemon_id']), str(data['latitude'])+","+str(data['longitude'])+","+str(data['encounter_id'])+","+str(data['pokemon_id'])+","+str(data['expiration_timestamp_ms'])+","+str(data['pokemon_name']))
KeyError: 'pokemon_name'
`

nivong:
when creating something like this don't forget to update the readme and the other docs..

MZorzy:
can this be used for snip pokemon ? without run a local map scanner, read other bot account same ip no good)

PR #4603, opened 2016-08-23
Extend enable/disable tasks

mjmadsen:
Added enable or disable for other tasks.


solderzzc:
The base has enable option ?

PR #4607, opened 2016-08-23
Added a config option for "Treat as VIP missing Pokemon in Pokedex"

Calcyfer:
## Short Description:
Adds a config option for #4580 in the CatchPokemon task : `treat_unseen_as_vip`, defaulting to true (current behaviour).
Added info to the docs.

### DIFF
--- a/configs/config.json.cluster.example
+++ b/configs/config.json.cluster.example
@@ -115,6 +115,7 @@
 	        "min_ultraball_to_keep": 5,
 	        "berry_threshold": 0.35,
 	        "vip_berry_threshold": 0.9,
+	        "treat_unseen_as_vip": true,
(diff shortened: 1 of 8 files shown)


sohje:
Seems good.
But do we really need another config option? Throw and lose 1 berry isnt a big deal if we encounter new pokemon and its transparent for user. Maybe we should leave it as built-in and dont clog main config file?

Calcyfer:
Better to have more config options than users opening issues because they don't like the change.

solderzzc:
@sohje  let's merge it? It doesn't break things.

sohje:
@solderzzc Yeap, sure.

solderzzc:
great. Thanks for pr and review.

PR #4640, opened 2016-08-24
Catcher improvements

mjmadsen:
- Removed catch_lured_pokemon.py
- Removed catch_visible_pokemon.py
- Added encounter check to catcher
- Reduced list scans in catcher
- Removed distance sort (no need)
- Added random selection order 

#4610 works but was added to depreciated files. This should accomplish the same.


mjmadsen:
CI Passed. Working fine on my tests.

PR #4641, opened 2016-08-24
Reduce reachable range

mjmadsen:
- Reduced stop max range to 35 (it was 40)
- Reduced pokemon max range to 35 (was 60)

### This needs testing. 

I've read the radius of our "range" for stops is 20, I've read it 40, etc. I think we should at least try reducing it a bit lower than 40, if that is the range.

As far as pokemon range, can any confirm it is more than a stop? Currently we have 60, which seems high. 

My thought is perhaps we were getting success results back from the server at distances that are too far away, when the app would handle this check on the client side. They run a distance check against logged encounters (fort or poke) on flagged accounts and we're banned. Whatever the range, we need to adjust and add randomness to it (some_config_min, max_range].


alexyaoyang:
Great initiative! Some test results from me, minimum Pokestop range of 38 is sufficient, 40 should be a very safe number.

Issue #4629, opened 2016-08-23
[FEATURE REQUEST] "FollowPath" with the function of "PolylineWalker"


maxprzemo:
### Short Description
I apologize in advance for my English
"FollowPath" with the function of "PolylineWalker"

### Possible solution
In the file
/PokemonGo-Bot/pokemongo_bot/cell_workers/follow_path.py

add import:
`from pokemongo_bot.walkers.walker_factory import walker_factory`

add in _process_config
'self.walker = self.config.get('walker', 'PolylineWalker')'

replace in def work(self):
step_walker = StepWalker(
to
step_walker = walker_factory(self.walker,

In the file
PokemonGo-Bot/configs/config.json.path.example
add
"walker": "PolylineWalker"
in
"FollowPath"
for example
      {
        "type": "FollowPath",
        "config": {
          "path_mode": "loop",
          "path_start_mode": "first",
          "path_file": "configs/path.example.json",
          "number_lap": 10,
          "timer_restart_min": "00:10:00",
          "timer_restart_max": "00:20:00",
          "walker": "PolylineWalker"
        }
      }

is all you need to do :)
Tests and it works
### How it would help others
Very nice effect. We choose the points on the map and the bot itself sets them route.We walk along the street. Like a real human.

Vvkmnn:
Love the idea. My current implementation walks right by pokestops, so I run out of items way too fast.

Implementing this on b5a8e1b0ba63a2a666323bac4549925f8dd3937c. I'll post results.

mjmadsen:
Merged #4644

PR #4644, opened 2016-08-24
Enable PolyWalker for FollowPath

alexyaoyang:
- Defaults to StepWalker
- #4629

screenshot 2016-08-23 at 6 48 55 pm


maxprzemo:
Thank you very much.

PR #4651, opened 2016-08-24
Reduce reachable range

mjmadsen:
Decreased stop (fork) max range to 38m (was 40m).
Decreased pokemon max range to 38m (was 60m).

Can we confirm it is 40m for stops and whatever it is for pokemon? We should also add variance to these.


alexyaoyang:
Great initiative! Some test results from me, minimum Pokestop range of 38 is just sufficient. It managed to reach a Pokestop on the biggest character circle possible.

PR #4653, opened 2016-08-24
teleporting do not update last-location

helloj:
## Short Description:
When the teleport_to avoid update last-location-% s.json.
CTRL-c interrupt occurs at this time, restart again will cause a significant shift abnormality

-
-
-


solderzzc:
OH, this is easily be detected ! thanks for the catch

Issue #4624, opened 2016-08-23
Shutdown of bot after leveling to set number in config? Request


xSidewinderx:
### Short Description
I would like to see an option to have to both either sleep or shutdown after a certain level was reached that was set by the user before starting the bot. Could even add on to this option by setting a certain level to reach, but having the bot calculate the xp needed and stretch the process of leveling over a a certain time frame to look more legit.

### Possible solution
By using the sleep-schedule already implemented into bot and config.json

### How it would help others
It would help improve the mimic of real players.

drricksanchez321:
I actually did this on my own copy..

modify collect_level_up_reward.py

add to top:
`import sys`

add to initialize function:
` self.quit_at_level = self.config.get("quit_at_level", 999)`

Add to the bottom of the _collect_level_reward function (after the level_up_reward emit):
`
            if self.current_level >= self.quit_at_level:
                sys.exit()
`
I added my own emit_event, which required changes to other files which I excluded here for brevity. 
then you can modify your config for the task to add a level cap:

` {
        "type": "CollectLevelUpReward",
	"config":{
	 "quit_at_level":28
	}
    }
`

alexyaoyang:
@drricksanchez321 Do you want to submit a PR with your changes? This could benefit a lot of people on top of not having to change the files back after every bot update 👍

drricksanchez321:
I would if I had a clue how to use git for more than pressing Download ZIP on the dev branch...  .NET background, TFS user here. 

The change is trivial anyhow -- and also not sure sys.exit() is the best to use, or if there's another preferred method to bubble up a workflow termination request from within a task.

mjmadsen:
Merged.

PR #4656, opened 2016-08-24
Stops bot after a set level limit

alexyaoyang:
## Features
- Defaults to -1, which means no level limit
- Uses same api call as collect level up reward
- #4624


solderzzc:
yes, very needed. Have you checked with the new map ?

alexyaoyang:
@solderzzc Just checking it out, looks very useful! Great job @BreezeRo!

PR #4658, opened 2016-08-24
Revert "Stops bot after a set level limit"

solderzzc:
Reverts PokemonGoF/PokemonGo-Bot#4656
Task name change cause many report, need keep the name.


alexyaoyang:
Ah I see, okay will keep the name and push another PR.

Issue #4583, opened 2016-08-22
Emulate phone battery


jaimerson:
### Short Description
Bot could simulate battery depletion as it walks around and either take a break when it ends or "find a place to charge" and chill there, evolving pokémon or something until charged enough to continue. 

### Possible solution
Use the existing random breaks feature to take the breaks, and add config options for minimum battery (which should be random) needed to leave on a journey. One could also config a home location to go back to and charge. The break time would be calculated based on the current position and speed.

### How it would help others
It would make the bot behavior more human-like, thus making it harder to detect as a bot (considering behavior analysis is the only way to detect)

mustii82:
Great Idea but is this important for Bot Detection ?

jaimerson:
@mustii82 well, how could we know what methods are used? At the very least it would make it more belieavable as a human

sediyama:
I think it would be more human like if the bot at least wait for a while before starting moving and from time to time, stop moving for a certain period of time.

Nik0lashka:
@sediyama #4202

sediyama:
@nik0lashka the feature that you pointed is not the same. What I am suggesting is to stay put, at the same place, logged in, searching for Pokemons and spining Forts. The feature at #4202, the bot logout.
Everybody playing stops for a while and keeps the phone running, it is what I am trying to suggest.

mjmadsen:
Does #4672 meet your request?

jaimerson:
@mjmadsen I guess so. Is there a config option to sometimes have an alive pause right after the random break? This would be perfect to simulate charging

mjmadsen:
Not quite, the different sleepers are completely independent.

PR #4672, opened 2016-08-24
Add RandomAlivePause task

cmezh:
This task keeps bot alive while it paused. Placed on the top of task list it drops execution of following tasks.
During pause it calls heartbeat and set_location for noising coordinates (of course, if noise is set in config).

Based on RandomPause task.
Inspired by (and also closes) #4583

### DIFF
--- a/configs/config.json.cluster.example
+++ b/configs/config.json.cluster.example
@@ -10,6 +10,16 @@
     "enable_social": true,
     "tasks": [
       {
+        "//NOTE: This task MUST be placed on the top of task list": {},
+        "type": "RandomAlivePause",
+        "config": {
+          "min_duration": "00:00:10",
+          "max_duration": "00:10:00",
+          "min_interval": "00:05:00",
+          "max_interval": "01:30:00"
+        }
+      },
+      {
         "type": "HandleSoftBan"
       },
       {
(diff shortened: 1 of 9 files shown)


J-A-B-S:
So your saying that during paused time, It will move to a fort or something like that but without actually send the info to server and then when pause is over it will be at the location it was last at during the paused time?

cmezh:
Nope. During the pause bot will not execute tasks set in task list after
RandomAlivePause. That's why RandomAlivePause must be placed on the top of
task list.

24 авг. 2016 г. 17:13 пользователь "J-A-B-S" <notifications@github.com>
написал:

mhdasding:
Exactly what i want, thought of doing it myself but thanks to you i don't have to =P  👍

solderzzc:
Why need alive pause ? In iPhone, if you switch app to background, still packet to server?

cmezh:
@solderzzc Not all users switch app to background when they have a short break for coffee or talk with friends or whatever more. And, in my opinion, the more bot behaviour different, the more it's humanized.

solderzzc:
ok, since it's optional, we can merge for testing, thanks a lot. @cmezh

cmezh:
Thank you, @solderzzc :)

imetanon:
if bot was in RandomAlivePause, how to cancel or stop ?

PR #4707, opened 2016-08-25
fix(map-chat): disable dummy info window

geminiyellow:
## Short Description:

`msg.text = html_sanitize(msg.text);` return `undefined` as `'undefined'`.

## Fixes:

disable `just appeared` info window

### DIFF
--- a/map-chat/javascript/map.js
+++ b/map-chat/javascript/map.js
@@ -239,7 +239,7 @@ function displayMessageOnMap(msg, olat, olong, sessid, icostr, expir, pokenick) 
   var expiration = expir;
   var pName = '';
   if(typeof pokenick === 'undefined'){
-  var pName = " <i>just appeared! </i></b><br>lat: " + olat + " / long: " + olong;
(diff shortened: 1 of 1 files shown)


map-chat/javascript/map.js
```diff
@@ -320,9 +320,8 @@ function displayMessageOnMap(msg, olat, olong, sessid, icostr, expir, pokenick)
       }
     });
 
-    if (msg.text) {
+    if (msg.text !== 'undefined') {
```
solderzzc: display the location is not that bad...

geminiyellow: @solderzzc yep, display the location is not bad, but the infoWindow is meaningless, i dont think human can read it. you can try it by yourself. btw, the same info is output in console.debug, too. 
and, if you want a infoWindow, i think there should have a filter.just pop up what you want.

solderzzc:
I've understood what you meant, thanks a lot. 
👍

solderzzc:
I'd like to have the dummy info back, till we find a better GUI. 
Since the box shows how fast the pokemon appear. 
Or make a timeout to clean up?

PR #4708, opened 2016-08-25
Allow the setting of candy threshold

### LINKED ISSUE #4256: Ignore Pokemon based on candy
### Short Description

Ignore Pokemon based on the number of candies

### Possible solution

    candy = inventory.candies(True).get(pokemon.pokemon_id)
    if candy > candy_threshold:
        ignore()
    else:
        catch()

in config.json:

    "catch": {
          "any": {
                "candy_threshold": 200
          }
     },

### How it would help others

It will benefit collection-only users. Keep account exp/h low.

bbiiggppiigg:
## Short Description:
Allow the setting of candy threshold to ignore certain pokemon if the threshold is reached
Using a default value of 400 ( For Magikarp )

Closes #4256 

-
-
-

### DIFF
--- a/.pullapprove.yml
+++ b/.pullapprove.yml
@@ -1,9 +1,11 @@
+
 approve_by_comment: true
-approve_regex: ^Approved
-reject_regex: ^Rejected
-reset_on_push: false
+approve_regex: '^:\+1:'
+reject_regex: '^(Rejected|:-1:)'
+reject_value: -1
+reset_on_push: true
+author_approval: ignored
 reviewers:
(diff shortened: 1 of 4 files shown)


.pullapprove.yml
```diff
@@ -1,3 +1,4 @@
+
```
k4n30: Why the change?

bbiiggppiigg: I'm new to pullapprove and accidentally changed the file.
I'll redo the pull request

PR #4711, opened 2016-08-25
fix(map-chat): fix pokemon icon scale

geminiyellow:
## Short Description:

click pokemon icon will let it big

## Fixes:

- set disable marker icon correct.
- change pokemon icon size from (60,60) to (36,36)


solderzzc:
@BreezeRo  how do you think about this ?

solderzzc:
Let me merge this.

PR #4713, opened 2016-08-25
Allow the setting of candy threshold

### LINKED ISSUE #4256: Ignore Pokemon based on candy

bbiiggppiigg:
## Short Description:
Allow setting ignore threshold based on the amount of candy
Default is set to 400 for Magikarp

Closes #4256

### DIFF
--- a/pokemongo_bot/cell_workers/pokemon_catch_worker.py
+++ b/pokemongo_bot/cell_workers/pokemon_catch_worker.py
@@ -222,8 +222,8 @@ class PokemonCatchWorker(Datastore, BaseTask):
         }
 
         candies = inventory.candies().get(pokemon.pokemon_id).quantity
-        threshold = pokemon_config.get('candy_threshold', 400)
-        if( candies > threshold  ):
+        threshold = pokemon_config.get('candy_threshold', False)
+        if( threshold > 0 and candies > threshold  ):
             self.emit_event(
                 'ignore_candy_above_thresold',
                 level='info',
--- a/pokemongo_bot/cell_workers/pokemon_catch_worker.py
+++ b/pokemongo_bot/cell_workers/pokemon_catch_worker.py
@@ -221,6 +221,23 @@ class PokemonCatchWorker(Datastore, BaseTask):
             'iv': False,
         }
 
+        candies = inventory.candies().get(pokemon.pokemon_id).quantity
+        threshold = pokemon_config.get('candy_threshold', 400)
+        if( candies > threshold  ):
+            self.emit_event(
+                'ignore_candy_above_thresold',
+                level='info',
+                formatted='Amount of candies for {name} is {amount}, greater than threshold {threshold}',
+                data={
+                    'name': pokemon.name,
+                    'amount': candies ,
+                    'threshold' : threshold
+                }
+            )
+            return False
+
+
+
         if pokemon_config.get('never_catch', False):
             return False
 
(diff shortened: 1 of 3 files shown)


pokemongo_bot/cell_workers/pokemon_catch_worker.py
```diff
@@ -221,6 +221,23 @@ def _pokemon_matches_config(self, config, pokemon, default_logic='and'):
             'iv': False,
         }
 
+        candies = inventory.candies().get(pokemon.pokemon_id).quantity
+        threshold = pokemon_config.get('candy_threshold', 400)
```
solderzzc: candy_threshold should default to false since this code will be triggered when config not changed for candy_threshold.

bbiiggppiigg:
What I was originally seeking is a default upper bound for number of candies , maybe float("inf") will work as a better default value than False ?

solderzzc:
Seems 0 should be good in  your code. or -1. @bbiiggppiigg

alexyaoyang:
@bbiiggppiigg I think -1 should be default value, since new players start with 0 candies for all pokemon.

alexyaoyang:
Looks good to me 👍

helloj:
candy threshold should return `True`, or ignore to next condition,
just like
```
"any": {"candy_threshold" : 400 ,"catch_above_cp": 2000, "catch_above_iv": 0.95, "logic": "or"},
```
the logic is `"or"` mean when encountered a rare beast, I want to catch it

```
        if pokemon_config.get('never_catch', False):
            return False

        if pokemon_config.get('always_catch', False):
            return True

        if pokemon.candy_quantity < pokemon_config.get('candy_threshold', -1):  # here
            return True

        catch_ncp = pokemon_config.get('catch_above_ncp', 0.8)
        if pokemon.cp_percent > catch_ncp:
            catch_results['ncp'] = True
```

bbiiggppiigg:
@helloj 
I think this depends on the what we set the threshold for. 
For me, its like that I don't want to catch Magikarp anymore since I already 400 candies , while for you it means that you want some pokemon cause the threshold is not reached yet.  I think the definition can be discussed.

Note : 
Current implementation will not allow pokemon with high iv or high cp to be caught if there's already too many candies for it, and it doesn't fit the logic of the config file.  

Since I don't want to modify the lambda function doing the logic operations, may be I should set a flag and do the logic myself before the function returns ?

helloj:
In fact, I did the same feature called `catch_below_cancy`.
It is hoped to limit the number to catch through candy quantity, but do not want to miss the high iv pokemon by `or` logic

this is my setting
```
    "catch": {
      "any": {"catch_above_cp": 1000, "catch_above_iv": 0.9, "logic": "or"},

      "// Evolve": {},
      "Charmander": { "catch_below_candy":200, "catch_above_iv": 0.95, "logic": "or" },
      "Dratini": { "catch_below_candy":400, "catch_above_iv": 0.95, "logic": "or" },
      "Magikarp": { "catch_below_candy":400, "catch_above_iv": 0.95, "logic": "or" },
```

alexyaoyang:
@bbiiggppiigg @helloj I agree, that's a good mix of the `candy_threshold` feature and our catching right now. I think I'd still like to attempt to catch that Gyarados even though I already have 400 Magikarp candies. This is more human-like.

helloj, Have not submitted a PR? If not, please do so :)

PR #4714, opened 2016-08-25
Add random delay for animation if new to dex

### LINKED ISSUE #4700: implement application like delays to account of animation screens
### Short Description
logs for the game are being joined up into one place, anything that does not look normal from a device in timings is being flagged 
from my normal play through once a pokemon shows up nearby, it takes at least 2-3 seconds before you can throw a pokeball (animation of pokemon appearing) , then it takes at least 3-20 seconds (20ish if its a new pokedex pokemon) to allow you to transfer cancel out of the capture and allow you select another pokemon to capture. 

I don't know where the asset downloads would be occuring, e.g. when you have a new pokemon nearby or when you select it to capture.

### Possible solution
implement default timings for items such as selecting/throwing pokeball after they appear. waiting the min time after a capture for the success screen and maybe the new pokedex entry. if a badge is awarded, wait the min time for the badge to display before selecting/throwing a pokeball etc.

### How it would help others
With more realistic application timings we should be able to stand out as much as the other users of the api.

mjmadsen:
- Added a random delay if the pokemon we caught is new to our pokedex

Closes #4700

### DIFF
--- a/pokemongo_bot/cell_workers/pokemon_catch_worker.py
+++ b/pokemongo_bot/cell_workers/pokemon_catch_worker.py
@@ -78,6 +78,8 @@ class PokemonCatchWorker(Datastore, BaseTask):
         self.catchsim_berry_wait_max = self.catchsim_config.get('berry_wait_max', 3)
         self.catchsim_changeball_wait_min = self.catchsim_config.get('changeball_wait_min', 2)
         self.catchsim_changeball_wait_max = self.catchsim_config.get('changeball_wait_max', 3)
+        self.catchsim_newtodex_wait_min = self.catchsim_config.get('newtodex_wait_min', 20)
+        self.catchsim_newtodex_wait_max = self.catchsim_config.get('newtodex_wait_max', 30)
 
 
     ############################################################################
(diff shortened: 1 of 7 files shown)


rawgni:
lgtm

mjmadsen:
@rawgni you've had a chance to test?

pokemongo_bot/cell_workers/pokemon_catch_worker.py
```diff
@@ -545,6 +549,10 @@ def _do_catch(self, pokemon, encounter_id, catch_rate_by_ball, is_vip=False):
                         }, outfile)
                         outfile.write('\n')
 
+                    # if it is a new pokemon to our dex, simulate app animation delay
+                    if exp_gain >= 500:
+                        sleep (randrange(newtodex_wait_min, newtodex_wait_max))
```
rawgni: @mjmadsen the variable here should be catchsim_newtodex_wait_min and catchsim_newtodex_wait_max

mjmadsen:
@rawgni Fixed. :+1:

solderzzc:
Thanks, merging.

Issue #4687, opened 2016-08-24
New account / first bot run


mjmadsen:
Two problems identified when starting a fresh account on the bot for the first time. 

### Inaccurate stat output
2016-08-24 13:16:54,171 [       cli] [INFO] 
2016-08-24 13:16:54,171 [       cli] [INFO] Ran for 0:09:50
2016-08-24 13:16:54,171 [       cli] [INFO] Total XP Earned: 0  Average: 0.00/h
2016-08-24 13:16:54,172 [       cli] [INFO] Travelled 0.00km
2016-08-24 13:16:54,172 [       cli] [INFO] Visited 0 stops
2016-08-24 13:16:54,172 [       cli] [INFO] Encountered 0 pokemon, 0 caught, 0 released, 0 evolved, 0 never seen before
2016-08-24 13:16:54,172 [       cli] [INFO] Threw 0 pokeballs
2016-08-24 13:16:54,172 [       cli] [INFO] Earned 0 Stardust
2016-08-24 13:16:54,172 [       cli] [INFO] Hatched eggs 0
2016-08-24 13:16:54,172 [       cli] [INFO] Next egg hatches in 4.64 km
2016-08-24 13:16:54,172 [       cli] [INFO] 
2016-08-24 13:16:54,172 [       cli] [INFO] Highest CP Pokemon: Drowzee [CP: 42] [IV: 10/5/3] Potential: 0.4 
2016-08-24 13:16:54,172 [       cli] [INFO] Most Perfect Pokemon: Pidgey [CP: 10] [IV: 2/12/10] Potential: 0.53 

Everything prints properly on second run of the bot (cache issue?).

### Bad egg status output
2016-08-24 13:13:59,504 [IncubateEggs] [INFO] [next_egg_incubates] Eggs incubating: [-9000.88/-8996.0 km] (Eggs left: 3, Incubating: 1)

Think the culprit is line 240 of incubate_eggs.py for the second error. Still had bad output after restarting. Maybe it's just the first egg?

commit 38b4a09

supergithubo:
Had it also on my new account. It's just the first egg. Other eggs have the correct values.

k4n30:
@mjmadsen is this a bug? or is it more of a feature request to fix it? i.e the bot doesnt crash, and works as intended, it's just the output is not displaying properly?

mjmadsen:
[...]

PR #4717, opened 2016-08-25
fix negative incubator distance

rawgni:
## Short Description:

set the default km_walked to 0

- fixes part of #4687 
-
-


mjmadsen:
From what I've looked at, it won't fix showing real values.

mjmadsen:
It would seem the current code is using start_km_walked and target_km_walked as our TOTAL walked (already walked) as the start and target as when the egg will hatch from then. So if we have already walked 1,000m and it's a 2k egg (2,000), the target_km is 3,000.

rawgni:
@mjmadsen yeah you are right, i will close this PR.

PR #4730, opened 2016-08-25
Minimize docker image

walkerlee:
## Short Description:

Build docker image from alpine, and try to reduce docker image layers.

docker image size:
350.4 MB -> 114.8MB (Only 32.7% original image size)

```
⋊> ~ docker images                                                                                            02:01:00
REPOSITORY                    TAG                 IMAGE ID            CREATED             SIZE
walkerlee/pokemongo-bot       latest              1d6ff1a9680e        18 minutes ago      114.8 MB
svlentink/pokemongo-bot       latest              fe87d34b4e99        28 minutes ago      350.4 MB
alpine                        latest              4e38e38c8ce0        8 weeks ago         4.799 MB
```

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -12,43 +12,34 @@
 # 
 # docker build  --build-arg BUILD_REPO=YourFork/PokemonGo-Bot --build-arg BUILD_BRANCH=6a4580f .
 
-FROM python:2.7.12-alpine
+FROM alpine
 
-RUN apk add --update --no-cache alpine-sdk git
-
-ARG BUILD_BRANCH
-ENV BUILD_BRANCH ${BUILD_BRANCH:-master}
-
-ARG BUILD_REPO
-ENV BUILD_REPO ${BUILD_REPO:-PokemonGoF/PokemonGo-Bot}
+ARG BUILD_REPO=PokemonGoF/PokemonGo-Bot
+ARG BUILD_BRANCH=master
+ARG TIMEZONE=Etc/UTC
 
 LABEL build_repo=$BUILD_REPO build_branch=$BUILD_BRANCH
-
-ADD https://github.com/$BUILD_REPO/archive/$BUILD_BRANCH.tar.gz .
-RUN tar -zxvf $BUILD_BRANCH.tar.gz && mv PokemonGo-Bot-* /usr/src/app && rm $BUILD_BRANCH.tar.gz
 
 WORKDIR /usr/src/app
 VOLUME ["/usr/src/app/configs", "/usr/src/app/web"]
 
-ARG timezone=Etc/UTC
-RUN ln -snf /usr/share/zoneinfo/$TZ /etc/localtime && echo $TZ > /etc/timezone
+ADD https://raw.githubusercontent.com/$BUILD_REPO/$BUILD_BRANCH/requirements.txt .
+RUN apk -U --no-cache add python py-pip \
+    && apk --no-cache add --virtual .build-dependencies python-dev gcc make musl-dev git tzdata \
+    && cp -fa /usr/share/zoneinfo/$TIMEZONE /etc/localtime \
+    && echo $TIMEZONE > /etc/timezone \
+    && wget -q -O- http://pgoapi.com/pgoencrypt.tar.gz | tar zxf - -C /tmp \
+    && make -C /tmp/pgoencrypt/src \
+    && cp /tmp/pgoencrypt/src/libencrypt.so /usr/src/app/encrypt.so \
+    && ln -s locale.h /usr/include/xlocale.h \
+    && pip install --no-cache-dir -r requirements.txt \
+    && apk del .build-dependencies \
+    && rm -rf /var/cache/apk/* /tmp/pgoencrypt /usr/include/xlocale.h \
+    && find / -name '*.pyc' -o -name '*.pyo' -exec rm -f {} \;
 
-#setup the bot
-ADD http://pgoapi.com/pgoencrypt.tar.gz /tmp/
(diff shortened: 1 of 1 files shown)


Dockerfile
```diff
@@ -12,43 +12,34 @@
...
+    && make -C /tmp/pgoencrypt/src \
+    && cp /tmp/pgoencrypt/src/libencrypt.so /usr/src/app/encrypt.so \
+    && ln -s locale.h /usr/include/xlocale.h \
+    && pip install --no-cache-dir -r requirements.txt \
+    && apk del .build-dependencies \
+    && rm -rf /var/cache/apk/* /tmp/pgoencrypt /usr/include/xlocale.h \
+    && find / -name '*.pyc' -o -name '*.pyo' -exec rm -f {} \;
+
+RUN apk -U --no-cache add --virtual .install-dependencies wget ca-certificates tar \
+    && wget -q -O- https://github.com/$BUILD_REPO/archive/$BUILD_BRANCH.tar.gz | tar zxf - --strip-components=1 -C /usr/src/app \
```
th3w4y: very elegant 👍 
i like the .install-dependencies part and the -stripcomponents=1

walkerlee: thanks. hope it can be helpful :)

Dockerfile
```diff
@@ -12,43 +12,34 @@
...
-#setup the bot
-ADD http://pgoapi.com/pgoencrypt.tar.gz /tmp/
-RUN cd /tmp && tar zxvf pgoencrypt.tar.gz \
-    && cd pgoencrypt/src \
-    && make \
-    && cp libencrypt.so /usr/src/app/encrypt.so \
-    && cd /tmp \
-    && rm -rf /tmp/pgoencrypt*
-
-ENV LD_LIBRARY_PATH /usr/src/app
```
th3w4y: We need to set the LD_LIBRARY_PATH? don't we?

Dockerfile
nivong: Is this to make it even smaller?

th3w4y:
@walkerlee: but we need to keep ENV LD_LIBRARY_PATH /usr/src/app.. don't we? (i see you removed that part)

walkerlee:
the source code set the file_name = 'encrypt.so', so we can remove this part

https://github.com/PokemonGoF/PokemonGo-Bot/blob/master/pokemongo_bot/__init__.py#L806

Dockerfile
walkerlee: yes, and more quickly!

it alse cached requirements.txt layer.
if this file no change, it just download the latest src to image and create new image layer.

th3w4y:
@walkerlee perfect so is relative to WORKDIR super 👍

th3w4y:
@walkerlee:

I believe i found a potential problem
In previous version i had: 

ADD https://github.com/$BUILD_REPO/archive/$BUILD_BRANCH.tar.gz
this get executed every docker build and if the file did not change it does not invalidate the cache

But now you only have a cache on requirements.txt

if requirements do not change your cache will not be invalidated because the next RUN commands are not changed and cache will not be invalidated, thus you will not build a newer image

Or so is my interpretation of how the Docker build caching works...

Dockerfile
```diff
@@ -12,43 +12,34 @@
...
+    && wget -q -O- http://pgoapi.com/pgoencrypt.tar.gz | tar zxf - -C /tmp \
+    && make -C /tmp/pgoencrypt/src \
+    && cp /tmp/pgoencrypt/src/libencrypt.so /usr/src/app/encrypt.so \
+    && ln -s locale.h /usr/include/xlocale.h \
+    && pip install --no-cache-dir -r requirements.txt \
+    && apk del .build-dependencies \
+    && rm -rf /var/cache/apk/* /tmp/pgoencrypt /usr/include/xlocale.h \
+    && find / -name '*.pyc' -o -name '*.pyo' -exec rm -f {} \;
+
+RUN apk -U --no-cache add --virtual .install-dependencies wget ca-certificates tar \
```
th3w4y: check my comment on caching - i believe this would be better as an   ADD statement

walkerlee:
@th3w4y:

you're right! the line 40 RUN instructions isn’t invalidated :(

i'll try another way to fix this issue. thank you!

th3w4y:
@walkerlee i would ADD back  as it was..
this way you don't need the wget anymore.. 
the you can just do your uncompression
`cat BUILD_BRANCH.tar.gz | tar zxf - --strip-components=1 -C /usr/src/app`

walkerlee:
@th3w4y:
I'm thinking to use github API to check repo version.
let me check is this possible, if not i'll use ADD back.

th3w4y:
@walkerlee same comment as above regarding the http://pgoapi.com/pgoencrypt.tar.gz i would like to recompile if the version changed i guess...

Dockerfile
```diff
@@ -12,43 +12,34 @@
 # 
 # docker build  --build-arg BUILD_REPO=YourFork/PokemonGo-Bot --build-arg BUILD_BRANCH=6a4580f .
 
-FROM python:2.7.12-alpine
+FROM alpine
```
nivong: what is the advanced to changes this to alpine alone?

th3w4y: @nivong alpine image is about 4MB

nivong: yes but adding python and pip and stuff is the phthon:2.7.12-alpine build I think

th3w4y: alpine after apk add python is about 52.2 MB instead of 72MB for python:2.7-alpine
the biggest saving is from executing api add and api del in the same RUN stanza

maestroi:
Looks very good :+1:

th3w4y:
some inprovements in #4783

PR #4513, opened 2016-08-22
Recycle Items schedule/force

Gobberwart:
## Short Description:

Added functionality to the Item recycler (recycle_items.py) which allows users to schedule regular recycles regardless of min_empty_space setting.

## Fixes/Resolves/Closes 

-Fixes #4417 
-Fixes #4337 
-Fixes issue caused by #3948 

- Added ability to schedule a forced item recycle regardless of
min_empty_space setting. Allows individual items to be recycled based on
"keep" settings, even when bag is not nearing capacity.

- Added inventory refresh to bot tick (__init.py__) to ensure web
displays information correctly. UPDATE: Now uses cached inventory instead of api call.

PR #4737, opened 2016-08-25
Larger times for recycling in example configs

Gobberwart:
## Short Description:

Minor update to example configs to set the values for recycling schedule to a sensible value (1 to 5 minutes). Previous times were far too low for the majority of users.

- Minor issue from #4513 

Times given in previous examples were too low (0 to 1 minute). Changed
to t to 5 minutes.


alexyaoyang:
Merging since only config values and they make more sense

PR #4745, opened 2016-08-25
Decrease number of messages from MoveToFort worker

cmezh:
Overwrite previous event line (only moving_to_fort and moving_to_lured_fort events).

### DIFF
--- a/pokemongo_bot/event_manager.py
+++ b/pokemongo_bot/event_manager.py
@@ -1,5 +1,6 @@
 # -*- coding: utf-8 -*-
 from __future__ import unicode_literals
+from sys import stdout
 
 
 class EventNotRegisteredException(Exception):
@@ -24,6 +25,7 @@ class EventManager(object):
     def __init__(self, *handlers):
         self._registered_events = dict()
         self._handlers = list(handlers) or []
+        self._last_event = None
 
     def event_report(self):
         for event, parameters in self._registered_events.iteritems():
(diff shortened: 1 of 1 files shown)


Gobberwart:
I actually like having this info, so I know the bot's actually doing something at times when nothing else is happening.

cmezh:
@Gobberwart If you take look at timestamp and wait 0.5-2 seconds you can see that it's working ;)
Also I've changed in my config "distance_unit" to "m" instead of "km", so now messages are not the same.

Gobberwart:
@cmezh OK, that's kinda cool actually (I misunderstood your intention). Any chance of putting this in follow_path and follow_spiral as well?

cmezh:
@Gobberwart Yep, why not? But first this idea and its implementation must be approved by devs.

Gobberwart:
Looking forward to seeing the end result of this.

mjmadsen:
@Gobberwart @cmezh Love it. When we expand, can we add to config to set frequency? Maybe ever X events we print, or by distance, etc. Maybe just an enable/disable - sometimes I like to see each for testing walker changes.

cmezh:
If we go this way, I think, enable/disable is enough. Maybe I will
implement this.
Currently you can disable this feature if you set ```debug``` option in
config to true. But in this case you will also get other debugging
messages, not only from MoveToFort.

26 авг. 2016 г. 21:56 пользователь "Matt J Madsen" <notifications@github.com>
написал:

mjmadsen:
Working on a PR, should be up shortly.

cmezh:
@mjmadsen Nice! Sorry, I forgot that you are dev and started explaining simple things 8)

mjmadsen:
config attribute isn't passed to the event manager, or I'm missing something. Might take me a bit longer than I expected.

mjmadsen:
@cmezh Check out #4780 - I need to do some minor things before it can be merged.

Gobberwart:
Looking good! I've only tested with MoveToFort but I'm really liking it. Nice work @cmezh and @mjmadsen

PR #4748, opened 2016-08-26
Fix first egg incubating

DeXtroTip:
## Short Description:
Seems like when incubating first egg the `start_km_walked` is not found and the bot will use 9001. Maybe if we set it to 0 it might work, since after this, every other time the `start_km_walked` is found.
Fixed it.

- Fixes [#4687](https://github.com/PokemonGoF/PokemonGo-Bot/issues/4687)


k4n30:
9001? wtf

rawgni:
@mjmadsen

mjmadsen:
Let me take a look in a bit.

mjmadsen:
Took me entirely too long looking into the shelve key error. This looks good (on a 2nd run account).

Thanks for the PR!

PR #4772, opened 2016-08-26
Fix max distance when sniping

Jasperrr91:
## Short Description:
People were complaining about the sniping not adhering to the maximum distance. This was caused by a statement that didn't trigger whenever snipe was turned on.

-
-
-


maestroi:
This fixed the problem :+1:

solderzzc:
Nice to have :) very good catch. I was intending to fix this, but never thought the fix can be so clean.

solderzzc:
👍

Issue #4781, opened 2016-08-26
Separate account from base configs


mjmadsen:
We should have an account file with just the following:
`{
    "auth_service": "",
    "username": "",
    "password": "",
    "location": "",
    "gmapkey": "",
    "encrypt_location": "",
    "config_location": ""
}`

Then the config_location is our base settings. We update the configs a whole lot and if you have multiple bots, that can be a lot of editing if you want to change default values. 

Though, I still yearn for allowing us to enter a series of accounts we automatically switch between every so often. Decisions implementing this could make that easier or possibly harder. 

Feedback welcomed.

solderzzc:
yes, very needed. Since the merging of work config and account config, I've done much work on the copy between configuration files.

alexyaoyang:
Agreed, having an account file will make things a lot easier for both new users and old. Your idea of switching between series of bot accounts **might** be easier with the live update code.

solderzzc:
It can't be real! https://github.com/PokemonGoF/PokemonGo-Bot/pull/4793 works as charming

mjmadsen:
Awesome. Thanks you @rawgni

PR #4793, opened 2016-08-27
added separate config

rawgni:
- #4781 
-
-

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -126,6 +126,7 @@ configs/*
 !configs/path.example.json
 !config.json.cluster.example
 !config.json.optimizer.example
+!auth.json.example
 
 # Virtualenv folders
 bin/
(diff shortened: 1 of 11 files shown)


solderzzc:
You sure you want to change as this way? Many support ticket will be opened.
How about add a auth in current main config file to point to the auth.conf? This will keep current config file working without changes.

rawgni:
@solderzzc i believe the current config will still work without a problem, ie. only using config.json instead of having auth.json and config.json.

It basically loads auth.json and then update it with whatever is in config.json. If there is nothing in auth.json then you get whatever is in config.json

solderzzc:
@rawgni  Yes, I just understood your code. Thanks a lot.

MZorzy:
Why not move *example to an subdir like ./configs/example
And let user have only real configs files under ./configs ?

Auth.json with fixed data
Some config.json
Location.json with preferred cities (need to add altitude)

Issue #4766, opened 2016-08-26
[DEV] Bot ignores incubate eggs config


Bashin:
### Expected Behavior

Bot incubates only 10km eggs in breakable incubators with the following config and leaves them unused if no 10km is available:

"infinite": [2,5,10],
"breakable": [10]

### Actual Behavior

Eggs incubating: [0.71/5.0 km, 0.71/5.0 km]
-> 1 infinite, 1 breakable

### Your FULL config.json (remove your username, password, gmapkey and any other private info)

Unmodified config.example.js (just inserted login data and changed the egg hatching part as pasted above)

### Output when issue occurred

2016-08-26 12:40:04,582 [IncubateEggs] [INFO] [incubate] Incubating a 5.0 egg.
2016-08-26 12:40:04,769 [IncubateEggs] [INFO] [incubate] Incubating a 5.0 egg.

### Steps to Reproduce

Let the bot incubate eggs with above settings.

### Other Information
OS: WIN10 64bit
Branch: dev
Git Commit: Fix shelve key error (09403f5)
Python Version: 2.7.12

Bashin:
2016-08-26 14:15:55,105 [IncubateEggs] [INFO] [incubate] Incubating a 5.0 egg.
2016-08-26 14:15:55,308 [IncubateEggs] [INFO] [incubate] Incubating a 2.0 egg.

alexyaoyang:
I think that is the expected behavior. If you want to use the breakable incubator for other x km eggs, you have to include x into the breakable array.

etiennebechara:
This bug also happened to me, I'm using the same config:

`{
			"type": "IncubateEggs",
			"config": {
				"longer_eggs_first": true,
				"min_interval": 50,
				"infinite": [2, 5, 10],
				"breakable": [10]
			}
		},`

alexyaoyang:
@supercourgette

supercourgette:
From what I can tell there are no issues here. With the config "infinite": [2,5,10],
[...]

PR #4803, opened 2016-08-27
Bot ignores incubate eggs config #4766

rawgni:
## Short Description

egg["km"] returns decimal 2.0 , 5.0 or 10.0.
The config uses integers.

Convert egg["km"] return value to integers.

- #4766 
-
-


mjmadsen:
Ah, nice catch. Couldn't see wth was going on before. 

Thanks!

alexyaoyang:
Great catch! 👍

PR #4810, opened 2016-08-27
Seperate sorting for breakable and unbreakable incubator

developerllazy:
## Short Description:
To increase the chance to have the 10K egg, we should use up all the 2K eggs so that we can have the chance to get more egg. To cater that, the sorting of the incubators are divided into 2 settings based on the type of the incubator, i.e. breakable and unbreakable.


Anakin5:
do we really need these parameters ?
for what reason would we hatch 2 and 5 km eggs with breakable incubator ?

Issue #4800, opened 2016-08-27
Removing date/time from bot output


nirvana7777:
### Expected Behavior

Moving towards pokestop Keppel Island Marina - 0.14km

### Actual Behavior

2016-08-27 18:04:19,285 [MoveToFort] [INFO] [moving_to_fort] Moving towards pokestop Keppel Island Marina - 0.14km

### Steps to Reproduce

I have tried looking for the doc where I can fix this issue but can't find it.

### Other Information
OS: Arch 32bit
 
Branch: master
 
Git Commit: 
 
Python Version: python2
 
Any other relevant files/configs (eg: path files)

lntn:
what is the problem?

nirvana7777:
Problem is how to remove "<date> <time> " from bot line outputs. Thank you.

nirvana7777:
remove " (date) (time) " *

nirvana7777:
I want it to be " moving towards pokestop XXX - X.Xkm "

Instead of " (date) (time) moving towards pokestop XXX - X.Xkm "

lntn:
I believe date, time, component's name & action is the standard log output template and there's very little reason we should remove them.

mjmadsen:
We can mark it as a feature request. Some folks might like this!

Gurzeh:
I agree. It should be possible to remove this info - even if I DO find it useful

rawgni:
@nirvana7777 
```python
 62 logging.basicConfig(
 63     level=logging.INFO,
 64     format='%(asctime)s [%(name)10s] [%(levelname)s] %(message)s')
 65 logger = logging.getLogger('cli')
```
change line 64 to `format='%(message)s'`

lntn:
ok I'm not familiar with Python and not much with git, so if someone gonna do this, you may add a new config "simplified_log" with default value false to make it more flexible.

nirvana7777:
@rawgnii Thank you !

Gobberwart:
See PR #4832 for improved version of this

mjmadsen:
Merged to dev!

PR #3706, opened 2016-08-12
Clean logging option

extink:
## Short Description:
Adds an option to enable clean logging. This will strip all meta data from the logging messages like the name of the process, the name of the event, etc.
## Fixes
- #3704

PR #4821, opened 2016-08-27
enable simple log output

rawgni:
## Short Description:

add -sl / --simplelog to pokecli.py

- #4800 
-
-

### DIFF
--- a/pokecli.py
+++ b/pokecli.py
@@ -289,6 +289,12 @@ def init_config():
     parser.add_argument("-cf", "--config", help="Config File to use")
     parser.add_argument("-af", "--auth", help="Auth File to use")
 
+    parser.add_argument("-sl", "--simplelog", help="Simple log output", action="store_true")
+    args = parser.parse_args()
+    if args.simplelog:
+        logging.root.removeHandler(logging.root.handlers[-1])
+        logging.basicConfig(level=logging.INFO, format='%(message)s')
+
     for _config in ['auth', 'config']:
         config_arg = parser.parse_known_args() and parser.parse_known_args()[0].__dict__[_config] or None
 


Gobberwart:
This should work... ish, but I'm not a fan of this approach as it's putting logger logic into the main bot client and it should be kept separate. Additionally, it requires a command line switch (should be in config).

Suggest adding something appropriate to config, and setting the logging options in the logger class (both logging_handler and colored_logging_handler) initialization, eg.

if self.config.logging_simple:
    logging.basicConfig(level=logging.INFO, format='%(message)s')

mjmadsen:
We finally merged #3706. I think it does a similar function. I'll leave up until further review.

Gobberwart:
@mjmadsen Similar, but doesn't appear to strip date/time, just the "extra" info (process name etc.)

Needs another config option to strip d/t?

mjmadsen:
@Gobberwart good eye. It would be nice if we could add to #3706 on which parts we want to see (each a separate config var).

Gobberwart:
@mjmadsen Something like that. Maybe something like:

logging: {
    datetime: true/false,
    processinfo: true/false,
    otherstuff: true/false
}

mjmadsen:
@Gobberwart Yeah, that'd be perfect.

Gobberwart:
@mjmadsen I've done some work on this but I'm not entirely satisfied that the way I've read the config options is the best/correct way to do it. Could you please cast an expert eye over #4832

Gobberwart:
I think PR #4832 can replace this as it provides even more options in a cleaner way.

rawgni:
close in favour of #4832

Issue #4817, opened 2016-08-27
Sentry responded with an error: 'utf8' codec can't decode byte 0x9c in position


updatesvc:
### Expected Behavior
Login as normally it does

### Actual Behavior
`2016-08-27 20:06:46,812 [       cli] [INFO] PokemonGO Bot v1.0
2016-08-27 20:06:46,825 [       cli] [INFO] commit: not found
2016-08-27 20:06:46,846 [       cli] [INFO] No config argument specified, checking for /configs/config.json
2016-08-27 20:06:46,851 [       cli] [WARNING] The daily_catch_limit argument has been moved into the CatchPokemon Task
2016-08-27 20:06:46,862 [       cli] [INFO] Configuration initialized
2016-08-27 20:06:46,864 [pokemongo_bot.health_record.bot_event] [INFO] Health check is enabled. For more information:
2016-08-27 20:06:46,865 [pokemongo_bot.health_record.bot_event] [INFO] https://github.com/PokemonGoF/PokemonGo-Bot/tree/dev#analytics
2016-08-27 20:06:46,881 [requests.packages.urllib3.connectionpool] [INFO] Starting new HTTP connection (1): www.google-analytics.com
2016-08-27 20:06:47,660 [SleepSchedule] [INFO] [next_sleep] Next sleep at 2016-08-28 12:09:38.660000
2016-08-27 20:06:47,841 [PokemonGoBot] [INFO] [set_start_location] Setting start location.
2016-08-27 20:06:47,842 [PokemonGoBot] [INFO] [x] Coordinates found in passed in location, not geocoding.
2016-08-27 20:06:47,842 [PokemonGoBot] [INFO] [location_found] Location found: -33.859221,151.208805 (-33.859221, 151.208805, 8.0)
2016-08-27 20:06:47,844 [PokemonGoBot] [INFO] [position_update] Now at (-33.859221, 151.208805, 8.0)
2016-08-27 20:06:47,845 [PokemonGoBot] [INFO] [login_started] Login procedure started.
2016-08-27 20:06:51,598 [PokemonGoBot] [INFO] [login_successful] Login successful.
2016-08-27 20:06:51,631 [PokemonGoBot] [INFO] Found encrypt.dll! Platform: win32 encrypt.dll directory: C:\poke\PokemonGo-Bot
2016-08-27 20:06:51,632 [PokemonGoBot] [INFO]
Traceback (most recent call last):
[...]

PR #4822, opened 2016-08-27
Fix bot crash at start if permabanned

Gobberwart:
## Short Description:
Bot not initialised (still set to "False") before trying to access event manager caused the following error:

```
"2016-08-27 20:10:17,371 [sentry.errors.uncaught] [ERROR] [u"AttributeError: 'bool' object has no attribute 'event_manager'", u' File "pokecli.py", line 746, in ', u' File "pokecli.py", line 170, in main']"
```

Separated bot initialize and start methods.

- #4817

### DIFF
--- a/pokecli.py
+++ b/pokecli.py
@@ -80,6 +80,9 @@ def main():
 
     def initialize(config):
         bot = PokemonGoBot(config)
+        return bot
+        
+    def start_bot(bot,config):
         bot.start()
         initialize_task(bot,config)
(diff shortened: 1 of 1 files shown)


Gobberwart:
It looks like there are a number of other events that MIGHT be raised during bot startup that will cause crash if the bot is still set to "False". This PR should fix all of those.

mjmadsen:
Thanks for the PR

PR #4839, opened 2016-08-28
remove equip_badge call, iso version does not add eqiup_badge to play…

### LINKED ISSUE #4522: check_awarded_badges being called but equip_badge is not
### Short Description
The client application is storing what badges it has shown to the user so that if they log into  new device they are not spammed all of the badges. This is very similar to level up api call.

### Possible solution
Have another task created like  CollectLevelUpReward which will peridoically check what badges are avialable and if they have been collected by the user. if not go and equip the badge.

https://github.com/tejado/pgoapi/wiki/check_awarded_badges
https://github.com/tejado/pgoapi/wiki/equip_badge

### How it would help others
This may reduce the glaring possibility that when an account is used solely by the bot and not logged into the native app once and a while, the number of badges for distance walked and pokemon captured might be flagging the account to be deleted even with slow walking and capturing enabled. This was also very similar to how CollectLevelUpReward which after logged in at level 15 it would collect all 15 levels of loot at once.

duttonw:
## Short Description:
having testing the iso version on awarded badges and what changes in the player data. it seems equip_badge call is not needed as the player_data does not change. As such deletion of commented out code and message which is not required. The delay is still required as badges do take time to display and accept and if multiple badges are awarded each badge needs to be accepted before the next is displayed. Note you can capture multiple pokemon before the awards show up so it is built into the heartbeat and the heartbeat seems to be delayed till you are back on the map and have not issued another event/action.

## Resolves:
resolves #4522

### DIFF
--- a/pokemongo_bot/__init__.py
+++ b/pokemongo_bot/__init__.py
@@ -1215,20 +1215,7 @@ class PokemonGoBot(Datastore):
                         data={'badge': badgename,
                               'level' : badgelevel }
                     )
-
-                    #todo move equip badge into its own task once working
-                    #should work but gives errors :'(s
-                    #response = self.api.equip_badge(badge_type=badge)
-                    response = {'responses': "awaiting further testing on api call to equip_badge"}
-                    self.event_manager.emit(
-                        'badges',
-                        sender=self,
-                        level='info',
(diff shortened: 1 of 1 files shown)


solderzzc:
Thanks for the PR and the detail.

PR #4856, opened 2016-08-28
dont forget to update the docs when adding config changes...

nivong:
more info: https://github.com/PokemonGoF/PokemonGo-Bot/pull/4596

### DIFF
--- a/docs/manual_installation.md
+++ b/docs/manual_installation.md
@@ -67,6 +67,8 @@ mv pgoencrypt/src/libencrypt.so encrypt.so
 ```bash
 cp configs/config.json.example configs/config.json
 vi configs/config.json
+cp configs/auth.json.example configs/auth.json
+vi configs/auth.json
 ```
 ####
 - make shure your git repo is up to date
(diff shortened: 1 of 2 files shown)


docs/manual_installation.md
```diff
@@ -78,15 +80,15 @@ pip install -r requirements.txt
 ####
 - finaly start the bot
 ```bash
-./run.sh configs/config.json
```
mjmadsen: We don't allow passing them in anymore?

nivong: Why would you if the run.sh does it it self? that is only useful when you have it at a different location and that is not discussed in the manual installation.

mjmadsen: If you have more than one account.

nivong: than there should be a second doc for that information not in the manual installation that coffers the manual installation ;)

mjmadsen:
Never mind what I said. We can elaborate on that someplace else.

mjmadsen:
:+1:

Issue #4836, opened 2016-08-28
Pokemon capture percentage inaccurate/confusing.


hobbes3:
### Expected Behavior

If the chance to capture a pokemon is over 100% then it should always capture.

### Actual Behavior

It still misses sometimes.

### Your FULL config.json (remove your username, password, gmapkey and any other private info)

    {
        "auth_service": "ptc",
        "username": "xxx",
        "password": "xxx",
        "location": "37.795614,-122.393536",
        "favorite_locations":[
            {"name": "San Francisco", "coords": "37.759234,-122.421307"}
        ],
        "gmapkey": "xxx",
        "encrypt_location": "",
        "websocket_server": false,
        "heartbeat_threshold": 10,
        "enable_social": true,
        "live_config_update": {
        "enabled": false,
        "tasks_only": false
        },
        "tasks": [
        {
            "type": "TelegramTask",
            "config": {
            "enabled": false,
            "master": null,
            "alert_catch": ["all"]
            }
        },
        {
            "//NOTE: This task MUST be placed on the top of task list": {},
            "type": "RandomAlivePause",
            "config": {
            "enabled": true,
            "min_duration": "00:00:03",
            "max_duration": "00:03:00",
            "min_interval": "00:08:00",
            "max_interval": "01:30:00"
            }
        },
        {
            "type": "HandleSoftBan"
        },
        {
            "type": "RandomPause",
            "config": {
            "enabled": true,
            "min_duration": "00:00:10",
            "max_duration": "00:02:00",
            "min_interval": "00:10:00",
            "max_interval": "02:00:00"
            }
        },
        {
            "type": "CompleteTutorial",
            "config": {
            "enabled": true,
            "// set a name": "",
            "nickname": "hobbes3k"
            }
        },
        {
            "type": "CollectLevelUpReward",
            "config": {
            "collect_reward": true,
            "level_limit": -1
            }
        },
        {
            "type": "IncubateEggs",
            "config": {
            "enabled": true,
            "infinite_longer_eggs_first": false,
            "breakable_longer_eggs_first": true,
            "min_interval": 120,
            "infinite": [2,5,10],
            "breakable": [2,5,10]
            }
        },
        {
            "type": "UpdateLiveStats",
            "config": {
            "enabled": true,
            "min_interval": 10,
            "stats": ["username", "uptime", "stardust_earned", "xp_earned", "xp_per_hour", "stops_visited"],
            "terminal_log": true,
[...]

PR #4862, opened 2016-08-28
Fixes the pokemon catch rates

Jasperrr91:
## Short Description:
This keeps track of the original pokeball catch rates and resets to those before using a second berry. Solves the problem where pokeball chances keep getting multiplied upon multiple razzberry uses.

- https://github.com/PokemonGoF/PokemonGo-Bot/issues/4836


Jasperrr91:
Whoops, unclean commit. I'll open a new clean PR.

PR #4866, opened 2016-08-28
Make the tracking of user data disabled by default

dvl:
I don't think it's ok to collect user data without the explicit permission of the user.

### DIFF
--- a/configs/config.json.cluster.example
+++ b/configs/config.json.cluster.example
@@ -157,7 +157,7 @@
             "berry_wait_max": 5,
             "changeball_wait_min": 3,
             "changeball_wait_max": 5,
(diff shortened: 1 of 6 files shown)


Jasperrr91:
The data is solely used to keep track of bot usage and errors. It does not contain any personal information. I do not see any reason as to why it should be disabled.

th3w4y:
Is anybody from the PokemonGoF organisation members using the data from the analytics?

Jasperrr91:
@brantje can you give some more information?

solderzzc:
Ip information was purged from the google analysis, so no user information leaking to server side. Leave it false by default will limit our motivation when seeing no users on the using list.

PR #4870, opened 2016-08-28
execute setup.sh -u if there is a need to

rawgni:
## Short Description:

Instead of telling the user they should execute ./setup.sh -u, we did it for them.

- 
-
-

### DIFF
--- a/run.sh
+++ b/run.sh
@@ -3,38 +3,35 @@ pokebotpath=$(cd "$(dirname "$0")"; pwd)
 auth=""
 config=""
 if [ ! -z $1 ]; then
-auth=$1
+  auth=$1
 else
-auth="./configs/auth.json"
+  auth="./configs/auth.json"
 fi
 if [ ! -z $2 ]; then
-config=$2
+  config=$2
 else
-config="./configs/config.json"
+  config="./configs/config.json"
 fi
 cd $pokebotpath
 source bin/activate
 git fetch -a
-if [ "1" == $(git branch -vv |grep -c "* dev") ] && [ $(git log --pretty=format:"%h" -1) != $(git log --pretty=format:"%h" -1 origin/dev) ]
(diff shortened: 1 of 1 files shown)


PokemonGoZ:
I never executed this when I installed. Explains why my bot never worked unless I did 3,000 different things to do it. Thanks setup.sh

DBa2016:
Please introduce an additional parameter then to PREVENT this behavior. When testing a pull request, having a stable not-changing environment is very helpful.

mjmadsen:
I'll agree with @DBa2016.

rawgni:
@mjmadsen @DBa2016 so now it prompt for user input. will this be better ?

run.sh
```diff
@@ -3,38 +3,39 @@ pokebotpath=$(cd "$(dirname "$0")"; pwd)
...
+if [ "1" == $(git branch -vv |grep -c "* dev") ] && [ $(git log --pretty=format:"%h" -1) != $(git log --pretty=format:"%h" -1 origin/dev) ] ||
+  [ "1" == $(git branch -vv |grep -c "* master") ] && [ $(git log --pretty=format:"%h" -1) != $(git log --pretty=format:"%h" -1 origin/master) ]
 then
-echo "Branch dev have an update. Run ./setup.sh -u to update."
-sleep 2
-elif [ "1" == $(git branch -vv |grep -c "* master") ] && [ $(git log --pretty=format:"%h" -1) != $(git log --pretty=format:"%h" -1 origin/master) ]
-then
-echo "Branch master have an update. Run ./setup.sh -u to update."
-sleep 2
+  read -p "Branch have an update. Run ./setup.sh -u to update? y/n
```
mjmadsen: has* (know it was in original) but then yes!

mjmadsen:
Never mind my edits. It's better. :)

PR #4874, opened 2016-08-28
Implemented more granular parameters for Telegram notifications

DBa2016:
## Short Description:
"alert_catch" now can be either a list as before (alert on every catch of the corresponding pokemon) or a dictionary of dictionaries, of the form:
"alert_catch": {
   "PokemonName": { "operator": "and/or", "iv": 0.9, "cp": 1000}
}

### DIFF
--- a/pokemongo_bot/event_handlers/telegram_handler.py
+++ b/pokemongo_bot/event_handlers/telegram_handler.py
@@ -21,10 +21,22 @@ class TelegramHandler(EventHandler):
             if event == 'level_up':
                 msg = "level up ({})".format(data["current_level"])
             elif event == 'pokemon_caught':
-                if data["pokemon"] in self.pokemons or self.pokemons[0]=="all":
-                    msg = "Caught {} CP: {}, IV: {}".format(data["pokemon"],data["cp"],data["iv"])
+                if isinstance(self.pokemons, list):
+                    if data["pokemon"] in self.pokemons or "all" in self.pokemons:
+                        msg = "Caught {} CP: {}, IV: {}".format(data["pokemon"],data["cp"],data["iv"])
+                    else:
+                        return
                 else:
-                    return
+                    if data["pokemon"] in self.pokemons:
+                        trigger = self.pokemons[data["pokemon"]]
+                    elif "all" in self.pokemons:
+                        trigger = self.pokemons["all"]
+                    else:
+                        return
+                    if (not "operator" in trigger or trigger["operator"] == "and") and data["cp"] >= trigger["cp"] and data["iv"] >= trigger["iv"] or ("operator" in trigger and trigger["operator"] == "or" and (data["cp"] >= trigger["cp"] or data["iv"] >= trigger["iv"])):
+                        msg = "Caught {} CP: {}, IV: {}".format(data["pokemon"],data["cp"],data["iv"])
+                    else:
+                        return
             else:
                 return
             self.tbot.sendMessage(chat_id=master, parse_mode='Markdown', text=msg)
--- a/configs/config.json.example
+++ b/configs/config.json.example
@@ -12,7 +12,12 @@
         "config": {
           "enabled": false,
           "master": null,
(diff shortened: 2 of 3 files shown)


Gobberwart:
You know, I quite like this concept, but its association with Telegram puts me off. I really don't want (and won't use) yet another messaging application.

Can I suggest you think about separating the various information-generating code from the actual message-sending code. That way, the information-generating stuff could be re-used to send notifications via other protocols, ie. email, skype, messenger etc.

Eg.

delivery_system = telegram/skype/whatever (obviously needs supporting code)
event = level_up/catch_pokemon/awarded_badge etc.
msg = nicely formatted notification message

then..

delivery_system.send_message(msg)

Let the delivery_system code worry about how to delivery the message, while the rest just deals with handling events and passing them on.

Just my 2 cents. Hope it makes sense.

DBa2016:
@Gobberwart : the original Telegram integration does not originate from me, I only adjusted the notification config and some other things.

In principle, what you are suggesting makes sense, and it would involve the following:
1) define an interface and create an abstraction layer which the actual "information-generating" code will be using
2) create an implementation for every needed IM system, starting with Telegram (because it is already there and should not be removed without a good reason).

None of these two parts is depending nor anyhow related to this PR. You are welcome to go ahead and implement your wish - I will very unlikely do because I am okay with Telegram and don't see a need for further IMs. :)

Gobberwart:
@DBa2016 No problem, just thought I'd comment while I had the opportunity. Never know who's watching. Not something I'm planning to do anything about immediately, but if anyone wants to run with it, go nuts :)

solderzzc:
@Gobberwart  The master setting is whom you can send for. Please check the code inside and correct me if I'm wrong.

Gobberwart:
@solderzzc The master setting? Sorry, not following you.

solderzzc:
@Gobberwart this one:
  "config": {
             "enabled": false,
             "master": null,

You can set the master to your own id, or the message will not send out.

solderzzc:
@Gobberwart  sure, get it.

pokemongo_bot/event_handlers/telegram_handler.py
```diff
@@ -21,10 +21,22 @@ def handle_event(self, event, sender, level, formatted_msg, data):
...
+                        return
                 else:
-                    return
+                    if data["pokemon"] in self.pokemons:
+                        trigger = self.pokemons[data["pokemon"]]
+                    elif "all" in self.pokemons:
+                        trigger = self.pokemons["all"]
+                    else:
+                        return
+                    if (not "operator" in trigger or trigger["operator"] == "and") and data["cp"] >= trigger["cp"] and data["iv"] >= trigger["iv"] or ("operator" in trigger and trigger["operator"] == "or" and (data["cp"] >= trigger["cp"] or data["iv"] >= trigger["iv"])):
```
solderzzc: There's good logic operator code in the pokemon catch worker, would you like to use that code @DBa2016

solderzzc:
`LOGIC_TO_FUNCTION = {
    'or': lambda x, y, z: x or y or z,
    'and': lambda x, y, z: x and y and z,
    'orand': lambda x, y, z: x or y and z,
    'andor': lambda x, y, z: x and y or z
}`

Gobberwart:
Does this work with "any" or "all"?

solderzzc:
Seems the `all` is included.
elif "all" in self.pokemons:
+                        trigger = self.pokemons["all"]
+                    else:

Gobberwart:
Cheers mate, I was too lazy to look :)

solderzzc:
Let's have it merge, leave the code fine tune later.

solderzzc:
👍

solderzzc:
@Gobberwart Do you want to try the pull approval ?

solderzzc:
yes, I think so...

Gobberwart:
Yay! :)

Issue #4841, opened 2016-08-28
JSON file read/write errors / corruption causing crash


lntn:
### Expected Behavior
Bot can start normally

### Actual Behavior
Bot could not start with `ValueError("No JSON object could be decoded")`

### Your FULL config.json (remove your username, password, gmapkey and any other private info)
```
{
    "auth_service": "google",
    "username": "",
    "password": "",
    "location": "taodanin",
    "favorite_locations":[
        {"name": "home", "coords": "10.761636, 106.654087"},
        {"name": "bachdang", "coords": "10.775286, 106.706436"},
        {"name": "taodanout", "coords": "10.774686, 106.693133"},
        {"name": "taodanin", "coords": "10.774779, 106.691739"}
    ],
    "gmapkey": "",
    "encrypt_location": "",
    "websocket_server": false,
    "heartbeat_threshold": 10,
    "enable_social": false,
    "live_config_update": {
      "enabled": true,
      "tasks_only": true
    },
    "tasks": [
      {
        "//NOTE: This task MUST be placed on the top of task list": {},
        "type": "RandomAlivePause",
        "config": {
          "enabled": false,
          "min_duration": "00:00:10",
          "max_duration": "00:10:00",
          "min_interval": "00:05:00",
          "max_interval": "01:30:00"
        }
      },
      {
        "type": "HandleSoftBan"
      },
      {
        "type": "CompleteTutorial",
        "config": {
          "enabled": false,
          "// set a name": "",
          "nickname": "DurianKing"
        }
      },
      {
        "type": "CollectLevelUpReward",
        "config": {
          "collect_reward": true,
          "level_limit": -1
        }
      },
      {
        "type": "IncubateEggs",
        "config": {
          "enabled": true,
          "longer_eggs_first": true,
          "min_interval": 120,
         "infinite": [2,5,10],
         "breakable": [2,5,10]
        }
      },
      {
        "type": "UpdateLiveStats",
        "config": {
          "enabled": true,
          "min_interval": 600,
          "stats": ["uptime", "km_walked", "stops_visited", "xp_earned", "pokemon_stats", "most_perfect_pokemon"],
          "terminal_log": true,
          "terminal_title": true
        }
      },
      {
        "type": "TransferPokemon",
        "config": {
          "enabled": true,
          "min_free_slot": 9900,
          "transfer_wait_min": 5,
          "transfer_wait_max": 10
        }
      },
      {
        "type": "EvolvePokemon",
[...]

PR #4877, opened 2016-08-29
Add exceptions to json file read/writes

Gobberwart:
## Short Description:

Read/write attempts to files (inventory-USERNAME.json, location-USERNAME.json) sometimes (rarely) fail with generic IO errors, causing bot to crash. This PR adds exception handling to these read/write operations to allow bot to continue.

- #4841

### DIFF
--- a/pokemongo_bot/cell_workers/pokemon_optimizer.py
+++ b/pokemongo_bot/cell_workers/pokemon_optimizer.py
@@ -96,30 +96,7 @@ class PokemonOptimizer(Datastore, BaseTask):
             self.family_by_family_id.setdefault(family_id, []).append(pokemon)
 
     def save_web_inventory(self):
-        web_inventory = os.path.join(_base_dir, "web", "inventory-%s.json" % self.bot.config.username)
-
-        with open(web_inventory, "r") as infile:
-            ii = json.load(infile)
-
-        ii = [x for x in ii if not x.get("inventory_item_data", {}).get("pokedex_entry", None)]
-        ii = [x for x in ii if not x.get("inventory_item_data", {}).get("candy", None)]
-        ii = [x for x in ii if not x.get("inventory_item_data", {}).get("item", None)]
-        ii = [x for x in ii if not x.get("inventory_item_data", {}).get("pokemon_data", None)]
-
-        for pokedex in inventory.pokedex().all():
-            ii.append({"inventory_item_data": {"pokedex_entry": pokedex}})
-
-        for family_id, candy in inventory.candies()._data.items():
-            ii.append({"inventory_item_data": {"candy": {"family_id": family_id, "candy": candy.quantity}}})
-
-        for item_id, item in inventory.items()._data.items():
-            ii.append({"inventory_item_data": {"item": {"item_id": item_id, "count": item.count}}})
-
-        for pokemon in inventory.pokemons().all():
-            ii.append({"inventory_item_data": {"pokemon_data": pokemon._data}})
-
-        with open(web_inventory, "w") as outfile:
-            json.dump(ii, outfile)
+        inventory.update_web_inventory()
 
     def get_family_optimized(self, family_id, family):
         evolve_best = []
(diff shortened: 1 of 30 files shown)


Gobberwart:
I notice the PR includes old commits even though there are no new changes in those files. Sorry, still learning how to drive git.

Gobberwart:
@mjmadsen @Jasperrr91 Your thoughts?

pokemongo_bot/cell_workers/pokemon_optimizer.py
```diff
@@ -96,30 +96,7 @@ def open_inventory(self):
...
-
-        for item_id, item in inventory.items()._data.items():
-            ii.append({"inventory_item_data": {"item": {"item_id": item_id, "count": item.count}}})
-
-        for pokemon in inventory.pokemons().all():
-            ii.append({"inventory_item_data": {"pokemon_data": pokemon._data}})
-
-        with open(web_inventory, "w") as outfile:
-            json.dump(ii, outfile)
+        inventory.update_web_inventory()
```
Gobberwart: The original code was an exact duplicate of this method in the inventory class. No need for both.

pokemongo_bot/inventory.py
```diff
@@ -1134,9 +1139,16 @@ def update_web_inventory(self):
         if not os.path.exists(web_inventory):
             self.init_inventory_outfile()
             
-        with open(web_inventory, "r") as infile:
-            json_inventory = json.load(infile)
-            infile.close()
+        try:
```
mjmadsen: Can we move this to a function somewhere so we don't have to repeat it?

Gobberwart: Latest commit excludes the infile.close() part (because it's useless) so basically it's just one line of code broken apart for ease of debugging. Not sure what the point of moving this would be.

Gobberwart: If you mean the whole lot, including the try/except stuff... Not really. They're all slightly different.

mjmadsen: Yeah, meant the whole lot - Just a thought! Only really seeing the difference as some read, some write. Either way, not a bit deal - was just seeking to reduce redundancy.

mjmadsen:
:+1:

mjmadsen:
Any more changes you'd like before we merge?

Gobberwart:
@mjmadsen Nah I'm good. I'd click the button myself but wanted someone else to have a look.

mjmadsen:
Where do we catch FileIOException?

Gobberwart:
@mjmadsen There's a catcher for it in each file where it occurs. I dislike that approach but it's supposed to be a bandaid so...

mjmadsen:
class FileIOException(Exception):
     pass

This just ignores the exception, right? I suppose the goal was not crashing anyway (like you said). 

Anyway, we have double excepts in here. Is the purpose in case the exception is not IOError or ValueError?

Gobberwart:
@mjmadsen Pretty much. It will allow the bot to continue if there's an error, although I suspect there might be more after... which will give us a trace to play with.

Yes, the intent is to catch either IOerror or ValueError.. they're the two I've seen happen here. Either it can't open the file or if it can, it can't handle the contents.

mjmadsen:
Let's pull the trigger.

Jasperrr91:
Sorry guys, only just woke up. Looks good, should catch the errors just fine 👍 

Now I was thinking, this is caused by invalid JSON right? Why not put the writing of the `inventory.json` in a loop in which we try to decode it and as long as that gives error; restart the loop to try to write it away again.

This will work if the error is somehow caused by a bad write. However, this will put your bot on a hold if it fails. In that case, after X retries; we can display a message telling people to upload the invalid JSON file; so we can analyse it and track down the error.

Gobberwart:
@Jasperrr91 Agree it would be nice to know what the problem actually is, other than "something went wrong".

PR #4881, opened 2016-08-29
Incubate eggs fix

DeXtroTip:
## Short Description:
Fixed a bug when hatching an egg and displaying the eggs at the same time would print wrong output (something like an egg hatching '2.26/2.0 km').
Fixed a bug where breakable incubators weren't being used.
Updated cached inventory after an egg hatched, removing the egg from the inventory and adding it as pokemon.

### DIFF
--- a/pokemongo_bot/cell_workers/incubate_eggs.py
+++ b/pokemongo_bot/cell_workers/incubate_eggs.py
@@ -35,15 +35,18 @@ class IncubateEggs(BaseTask):
         except:
             return
 
+        should_print = self._should_print()
+
         if self.used_incubators and IncubateEggs.last_km_walked != self.km_walked:
             self.used_incubators.sort(key=lambda x: x.get("km"))
             km_left = self.used_incubators[0]['km']-self.km_walked
             if km_left <= 0:
                 self._hatch_eggs()
+                should_print = False
             else:
                 self.bot.metrics.next_hatching_km(km_left)
 
-        if self._should_print():
+        if should_print:
             self._print_eggs()
             self._compute_next_update()
 
--- a/pokemongo_bot/cell_workers/incubate_eggs.py
+++ b/pokemongo_bot/cell_workers/incubate_eggs.py
@@ -180,9 +180,6 @@ class IncubateEggs(BaseTask):
                             pokemon.get('individual_stamina', 0)
                         ]})
                     matched_pokemon.append(pokemon)
-                    #remove as egg and add as pokemon
-                    inventory.pokemons().remove(pokemon['id'])
-                    inventory.pokemons().add(inventory.Pokemon(pokemon))
                 continue
             if "player_stats" in inv_data:
                 self.km_walked = inv_data.get("player_stats", {}).get("km_walked", 0)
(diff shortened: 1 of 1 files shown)


pokemongo_bot/cell_workers/incubate_eggs.py
```diff
@@ -176,6 +180,9 @@ def _check_inventory(self, lookup_ids=[]):
                             pokemon.get('individual_stamina', 0)
                         ]})
                     matched_pokemon.append(pokemon)
+                    #remove as egg and add as pokemon
+                    inventory.pokemons().remove(pokemon['id'])
+                    inventory.pokemons().add(inventory.Pokemon(pokemon))
```
rawgni: @DeXtroTip it's better for 184 and 185 to go inside _apply_incubators (line 99), ie. once you get a reply that you successfully use an incubator on an egg, you remove it from the inventory.

DeXtroTip: Not sure about that, because even when incubating, it is still an egg and is in the inventory. It never leaves it, it just switches from egg to actual pokemon.

rawgni: @Anakin5

mjmadsen:
Going to wait for @Anakin5 to give his input. But I like it!

pokemongo_bot/cell_workers/incubate_eggs.py
Jasperrr91: I Agree with @DeXtroTip. Removing it while being incubated just causes an inventory size mismatch.

Jasperrr91:
```
        if self.ready_breakable_incubators:
            # get available eggs
            eggs = self._filter_sort_eggs(self.infinite_incubator,
                    self.infinite_longer_eggs_first)
            self._apply_incubators(eggs, self.ready_infinite_incubators)
```

Weird that this was ever merged 😛

pokemongo_bot/cell_workers/incubate_eggs.py
Anakin5: It seems like it is going to work like that. It is hard to be entirely sure since this file is a maze.
The `_check_inventory` should not call the api (at every loop !) so, removing the egg and adding the Pokemon will have to be moved in the `_hatch_eggs` function one day or another.

mjmadsen: "It seems like it is going to work like that. It is hard to be entirely sure since this file is a maze." I know right?

DeXtroTip: Removing the egg and adding the Pokemon can easily be done in the `_hatch_eggs`.
If we remove the api call, the we will need to update the cached inventory with incubators when no uses are left and use distance walked in metrics.

Anakin5: I don't think so because this function also fetch the km walked on each eggs to know if they have to be hatched. So right now the egg incubator does not work if it does not query the api at every loop.
Ideally, walkers would update km walked somewhere and egg incubator would use that to know when to hatch eggs, or at least when to query api to check. That would require complete rewrite of this task.
What is the phone app doing anyway ? Is it querying the inventory regularly ?

mjmadsen:
While we may not have all of our wants in this merge, we fixed one of the problems. Can we settle on that and make a new issue on improvements to this task?

Anakin5:
ok for me

mjmadsen:
:+1:

mjmadsen:
@DeXtroTip Do you agree?

DeXtroTip:
Yes, it's fine for me.

mjmadsen:
Thanks for the PR :)

mjmadsen:
Who'd like to sum up our thoughts for improvement in an issue? I'm sure you guys know what's going on here far better than myself.

pokemongo_bot/cell_workers/incubate_eggs.py
Jasperrr91: It all depends on what the phone app is doing. If it only queries the inventory once, so should we. Otherwise it would be a big red flag for Niantic.

Issue #4760, opened 2016-08-26
Noise / Actual location and pathing


mjmadsen:
Currently our pathing is effected by our noise. While it should be effected by human variance (walking side to side), noise should not effect it as it is a device issue. After we apply noise, we're pathing from the noise location rather than the location we're pretending to be at. Basically, noise should only effect API requests and not our internal, "where I want to go from here" bits of goodness.

kanemasa1987:
Aren't the positions changed through api.set_position and retrieved from api.get_position/bot.position (which manipulates actual position)?
The only strange code code I found was in follow_path
https://github.com/PokemonGoF/PokemonGo-Bot/blob/ae314b666a1eb71de8872289ef3a298af09e4637/pokemongo_bot/cell_workers/follow_path.py#L145-L146
which is using noised position for logging. Other parts seem to be using set/get, which should be the actual position.

kanemasa1987:
Oh no! you were right
https://github.com/PokemonGoF/PokemonGo-Bot/blob/master/pokemongo_bot/walkers/polyline_walker.py#L29

kanemasa1987:
This is totally unintended, but cached polyline is returning the position you are supposed to be, regardless of this bug, so the behavior itself is correct after all (we are not affected by noise.)

That line in polyline is not used, if we look at it carefully.
I will do some cleanups and add the long waited altitude thingy, after taking some sleep...

mjmadsen:
https://github.com/PokemonGoF/PokemonGo-Bot/blob/master/pokemongo_bot/cell_workers/follow_spiral.py#L70
https://github.com/PokemonGoF/PokemonGo-Bot/blob/master/pokemongo_bot/cell_workers/follow_path.py#L102
https://github.com/PokemonGoF/PokemonGo-Bot/blob/master/pokemongo_bot/cell_workers/random_alive_pause.py#L141 (this code should be called from somewhere else)
https://github.com/PokemonGoF/PokemonGo-Bot/blob/master/pokemongo_bot/cell_workers/pokemon_catch_worker.py#L49
https://github.com/PokemonGoF/PokemonGo-Bot/blob/master/pokemongo_bot/api_wrapper.py#L29
https://github.com/PokemonGoF/PokemonGo-Bot/blob/master/pokemongo_bot/walkers/polyline_walker.py#L18

th3w4y:
[...]

PR #4883, opened 2016-08-29
Use real position for spins

mjmadsen:
- Removed use of noised position for fort spins

I'll leave this up for discussion. In my opinion, we use noise to better mimic phone GPS data. With that, however, we should be using our "actual" position for interacting in the world. Thoughts?

Closes last bit of #4760

### DIFF
--- a/pokemongo_bot/cell_workers/spin_fort.py
+++ b/pokemongo_bot/cell_workers/spin_fort.py
@@ -180,20 +180,13 @@ class SpinFort(Datastore, BaseTask):
     def get_forts_in_range(self):
         forts = self.bot.get_forts(order_by_distance=True)
         forts = filter(lambda fort: fort["id"] not in self.bot.fort_timeouts, forts)
-        if self.bot.config.replicate_gps_xy_noise:
-            forts = filter(lambda fort: distance(
-                self.bot.noised_position[0],
-                self.bot.noised_position[1],
-                fort['latitude'],
-                fort['longitude']
-            ) <= Constants.MAX_DISTANCE_FORT_IS_REACHABLE, forts)
-        else:
-            forts = filter(lambda fort: distance(
-                self.bot.position[0],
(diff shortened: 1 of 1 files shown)


mjmadsen:
@kanemasa1987

cmezh:
Removing this part of code will cause "Fort out of range" errors. We need better solution.

cmezh:
Also, there are another part of "noised" code in move_to_fort.

mjmadsen:
Good point about out of range. Where is the other portion using noised?

Maybe if we're using noised, we check that both are in range?

cmezh:
I know it because I made this changes 8)
I think, move_to_fort is ok, please take a look at it.
Checking both position is a good idea, I think it will solve a problem. Also it should throw warning/error (and maybe use default value in this case) if noise range is greater than distance fort is reachable.

mjmadsen:
Out of curiosity, how much variance does our default values for noise have in meters? Changed the max range to 38 for forts a while back.

cmezh:
Default value of noise range is 0.000125. Value in meters varies depending on location and should be calculated using utils.distance.
For example, if we are at (0,0) it is 19.65 m, but if we are at (50,50) it is 16.52 m.

kanemasa1987:
The app itself is using the noised data for calculating the range of spinning/encountering pokemon, thus my idea is using noised position for those 2 are ok(as there are cases gps noise let us spin a fort further than our actual position in the actual app.)
Besides, implementation of taking noise into the account would be dirty.

mjmadsen:
@kanemasa1987 @cmezh Ok, screw this PR then :+1:

kanemasa1987:
Slightly off the topic, but it might be good idea to spin the fort when we are slightly away from it.
It is much human like.

mjmadsen:
Interesting idea @kanemasa1987. Making requests we know will fail some of the time.

Issue #4873, opened 2016-08-28
Crash when evolving pokemon


r3pek:
### Expected Behavior

Bot to evolve pokemon

### Actual Behavior
Crash with the following call trace while trying to evolve a pokemon:
http://pastebin.com/z3Dj932k

### Your FULL config.json (remove your username, password, gmapkey and any other private info)

Pretty much the default config, just enable evolve pokemons

### Output when issue occurred

http://pastebin.com/z3Dj932k

### Steps to Reproduce

start bot with a pokemon available to evolve

### Other Information
OS: linux
 
Branch: dev
 
Git Commit:  007f02a
 
Python Version: Python 2.7.12
 
Any other relevant files/configs (eg: path files)

Gobberwart:
Did this happen just once or is it a regular/recurring thing?

r3pek:
Since it doesn't even evolve the pokemon, everytime i start the bot, it triggers the evolution, and so it crashes.

Gobberwart:
I changed the title of this issue because "Yet another UTF8 crash" means nothing.

mjmadsen:
Looks related to #4798. @rawgni 

Is this because we have the default value = none. So we try getting the first index and it's none (null).

rawgni:
@r3pek can you post your config file ? just the EvolvePokemon part

r3pek:
@rawgni yes. my bad then? :(

mjmadsen:
Not necessarily, whatever you have evolve_all is the key to figuring this out.

mjmadsen:
@rawgni I looked over what is there, realize it wouldn't get to that line if it was set to none. Side note, we need to change it to something other then evolve_all (it's purposes go well beyond it's initial purpose).

r3pek:
i have `"evolve_all": ""`

mjmadsen:
Ah, set to none if you don't want to have any evolved.

r3pek:
I want the exact opposite... evolve all ;)

rawgni:
`"evolve_all": "all"`

r3pek:
works ;) thx and sorry for the misleading error

mjmadsen:
Maybe change it to evolve_list. Do we indicate the values in the md?

@r3pek No problem, the variable is a bit misleading. Plus helped us realize a few places to make sure we're getting correct data.

mjmadsen:
Several solutions merged to dev. If problems persist, please reopen with new logs.

PR #4891, opened 2016-08-29
add some sanitycheck

rawgni:
## Short Description:
Evolve Pokemon crashes the bot if `evolve_all : ''`

- #4873 
-
-

### DIFF
--- a/pokemongo_bot/cell_workers/evolve_pokemon.py
+++ b/pokemongo_bot/cell_workers/evolve_pokemon.py
@@ -40,7 +40,7 @@ class EvolvePokemon(Datastore, BaseTask):
 
         if self.evolve_all[0] != 'all':
             # check for negation
-            negate = filter(lambda x: x[0] == '-', self.evolve_all)
+            negate = filter(lambda x: len(x) > 0 and x[0] == '-', self.evolve_all)
 
             # if there are things to negate
             if len(negate) > 0:


rawgni:
@mjmadsen

pokemongo_bot/cell_workers/evolve_pokemon.py
```diff
@@ -40,7 +40,7 @@ def work(self):
 
         if self.evolve_all[0] != 'all':
             # check for negation
-            negate = filter(lambda x: x[0] == '-', self.evolve_all)
+            negate = filter(lambda x: len(x) > 0 and x[0] == '-', self.evolve_all)
```
mjmadsen: Could we have a check when we check for it being 'none' as well (though this might cover both).

mjmadsen:
:+1:

pokemongo_bot/cell_workers/evolve_pokemon.py
rawgni: you mean if `"evolve_all": "none"` or `"evolve_all": "Pidgey, none"` ?

The first will not get to this line as it will be return at 37.
The second one will not be in negate as it doesn't have -. Even if someone try to do "Pidgey, -none", it won't match any pokemon.

mjmadsen: Actually, since we had an empty string example come through, this will suffice.

Issue #4754, opened 2016-08-26
Decoupling and refactoring


mjmadsen:
Our task order is currently static. Some of our tasks need to be performed on a state change. My thoughts that turn into ramblings:

- We should always try to collect level up rewards on exp change
- We should always check for badge responses (on each heartbeat?)
- Our API requests should be detached from bot decisions (store what we want locally, but only send recent when a response is required)
- Move emit calls to centralized place (store emit requests from each task). Allows for easier configuration, decouples task config from log config. 
- Move delay times and sleep features to centralized place. Same as above, we repeat too many (similar kinds of) waits and make our config entirely too long. 
- Too many parts of code that are trying to accomplish the same thing. Between catch_pokemon, pokemon_catch_worker, map_move_to_pokemon we repeat much of the code. Plus map_move_to_pokemon changes position and initiates capture attempts (and checks balls). All capture attempts should be force through our catch_pokemon task first (with refactor to implement order logic of map_move_to_pokemon). pokemon_catch_work should be the only one checking balls of them. 

I'll stop for now. My idea is, some files do too much that other files should and some files just too much in general.

Gobberwart:
Re api calls.. In an ideal universe the bot would only call the api when absolutely necessary, ie. To perform an action for the most part. As far as  things like player stats and inventory,  the bot should grab that info at start, store it in a database, then only refer to the database for future info. Sanity check every so often just to make sure nothing is wrong. 

Just my ramble

duttonw:
> We should always check for badge responses (on each heartbeat?)
check awarded badges is currently in the heartbeat.

My PR #4746 stored the awarded badge in the bot, and also updated the player_data on every heartbeat. so we could remove the get player_data api calls from some tasks if last 10 second (give or take delays) is good enough.

mjmadsen:
@Gobberwart @duttonw I love it.

sohje:
I make some debug over api calls. There is a lot of calls like _get_player/get_inventory_ one after another without delay or delay <= 1s.
Looks like:
```
bot_task()
-> function get_inventory at 0x107a1bc80
-> function get_inventory at 0x107a1bc80
-> function get_player at 0x107a1bc08
-> function get_inventory at 0x107a1bc80
bot_task()
-> function get_player at 0x107a1bc08
-> function get_inventory at 0x107a1bc80
-> function get_inventory at 0x107a1bc80
-> function get_inventory at 0x107a1bc80
```
[...]

PR #4896, opened 2016-08-29
Call level_up_rewards on exp changes/Some pep-8

sohje:
## Short Description:
Call level_up_reward on exp changes
- #4754 
- Some pep-8
- Need refactor other tasks, wrap refresh inventory with cached object -> reduce get_inventory api calls


mjmadsen:
:+1:

PR #4909, opened 2016-08-29
Fix for run.sh

alexyaoyang:
## Fixes
- Currently `run.sh` requires an auth file as the first parameter for `pokecli.py` to run, which forces everyone to use auth files.
- Swapped config and auth position so old command of `./run.sh ./configs/myconfig.json` works again.

Note: Those running `./run.sh ./configs/myauth.json ./configs/configalexnew.json` need to swap the parameters back.


mjmadsen:
:+1:

PR #4921, opened 2016-08-30
Pokemon Optimizer round 4

Anakin5:
- Add full documentation of Pokemon Optimizer (need to add a few things but mostly done)
- Add support for different mode of selection "by_pokemon", "by_family", "overall"
- Add parameter to create and reference groups of Pokemon
- Add possibility to exclude Pokemon or groups of Pokemon
- Add possibility to configure evolution threshold when selecting Pokemon to keep
- Add possibility to keep Pokemon whose stats are no further that x% of the best
- Auto correct misspelled Pokemon names in configuration

### DIFF
--- a/pokemongo_bot/cell_workers/pokemon_optimizer.py
+++ b/pokemongo_bot/cell_workers/pokemon_optimizer.py
@@ -1,10 +1,8 @@
-import copy
-import json
+import difflib
+import itertools
 import math
-import os
 
 from pokemongo_bot import inventory
-from pokemongo_bot.base_dir import _base_dir
 from pokemongo_bot.base_task import BaseTask
 from pokemongo_bot.datastore import Datastore
 from pokemongo_bot.human_behaviour import sleep, action_delay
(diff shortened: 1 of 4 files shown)


mjmadsen:
:+1:

Cptspal:
I wonder if lines 261 to 263 of pokemon_optimizer.py are not a duplicate of lines above ? I've a TypeError 'Pokemon' type not iterable when I reach this step.

Anakin5:
Which lines are you referring to ?

mjmadsen:
I think he's referring to this:
            for _, pokemon_list in self.group_by_pokemon_id(senior_pokemon_list):
                try_evolve, keep = self.get_best_pokemon_for_rule(pokemon_list, rule)
                try_evolve_all += try_evolve
                keep_all += keep

            try_evolve, keep = self.sort_pokemon_list(other_family_list, rule)
            try_evolve_all += try_evolve
            keep_all += keep

pokemongo_bot/cell_workers/pokemon_optimizer.py
```diff
@@ -62,211 +60,318 @@ def work(self):
...
+                keep_all += keep
+        elif len(senior_pids) < nb_branch:
+            # We did not get every combination yet = All other Pokemon are potentially good to keep
+            for _, pokemon_list in self.group_by_pokemon_id(senior_pokemon_list):
+                try_evolve, keep = self.get_best_pokemon_for_rule(pokemon_list, rule)
+                try_evolve_all += try_evolve
+                keep_all += keep
+
+            try_evolve, keep = self.sort_pokemon_list(other_family_list, rule)
+            try_evolve_all += try_evolve
```
mjmadsen: Should have just commented here instead.

Anakin5:
Sorry, can you point me on which line is reported in the error message ? I do not see where the error is.

Cptspal:
I don't have the error on my screen anymore, but it was pointing to line 262 of pokemon_optimizer.py.
I've just set the evolve=false on config.json to get rid of the error, but now I transfert most of my pokemon so even if I set it back to true to check the error message, I've to wait for more pokemon.

Bit if you check the file pokemon_optimizer.py, you can see that line 261, 262 and 263 are just a duplicating of above lines 257, 258, 259 but outside of the "for" loop. I don't catch all the logic behind the code, so I may be wrong.

Anakin5:
Ok I see the error, I will fix.

For the lines you are talking about, they are not exactly duplicated, there is a small difference ;)

Anakin5:
I pushed the fix in the other pull request.

mjmadsen:
I'm not seeing the PR

PR #4922, opened 2016-08-30
documented docker for the auth.json use case

th3w4y:
Fixes:
- add instructions for the docker run command for the multiple config files use case.


mjmadsen:
:+1:

Issue #4966, opened 2016-08-30
Errors logged in output when Eggs hatching.


JimmyMultani:
### Expected Behavior

When egg is hatching, details of Pokemon should appear.

### Actual Behavior

Details appear as:
```Shell
[IncubateEggs] [INFO] [egg_hatched] {'candy': 'error', 'iv': 'error', 'stardust': 'error', 'exp': 'error', 'cp': 'error', 'pokemon': 'error'}
```

### Your FULL config.json (remove your username, password, gmapkey and any other private info)

http://pastebin.com/JbAE8rGC

### Output when issue occurred

```Shell
[IncubateEggs] [INFO] [egg_hatched] {'candy': 'error', 'iv': 'error', 'stardust': 'error', 'exp': 'error', 'cp': 'error', 'pokemon': 'error'}
```

### Steps to Reproduce

Set up Egg Incubation settings, let eggs hatch. Happens inconsistently.

### Other Information
OS: OSX
 
Branch: Master
 
Git Commit: 56339fba1ba7f77a0275078ea0cf7a73e60c2ed0
 
Python Version: 2.7.10
 
Any other relevant files/configs (eg: path files) 

N/A

Gobberwart:
Looking at this now

leadboots5:
Mine actually crashed when it hatched on the latest dev:

```
Traceback (most recent call last):
  File "pokecli.py", line 781, in <module>
    main()
  File "pokecli.py", line 139, in main
    bot.tick()
  File "C:\Users\Josh\Downloads\PokemonGo-Bot\pokemongo_bot\__init__.py", line 658, in tick
    if worker.work() == WorkerResult.RUNNING:
  File "C:\Users\Josh\Downloads\PokemonGo-Bot\pokemongo_bot\cell_workers\incubate_eggs.py", line 45, in work
    self._hatch_eggs()
  File "C:\Users\Josh\Downloads\PokemonGo-Bot\pokemongo_bot\cell_workers\incubate_eggs.py", line 215, in _hatch_eggs
    if not pokemon_ids or pokemon_data[0]['name'] == "error":
IndexError: list index out of range
```

mmixx:
Encountered same errors with mine...

Asafffff:
mine crashed too, but i have another problem - it says it reads from C:\Python while my Python installation is on D:\
http://pastebin.com/dXQTTiBf

Jasperrr91:
[...]

PR #4974, opened 2016-08-31
Inventory refresh before hatching eggs

Gobberwart:
## Short Description:

Added inventory.refresh() to egg hatching, otherwise the inventory can't provide information needed for correct display.

- #4966 

Egg hatching needs a current copy of inventory (post egg hatching) to be
able to display details. Api call to refresh inventory occurs if there
are eggs to hatch.

### DIFF
--- a/pokemongo_bot/cell_workers/incubate_eggs.py
+++ b/pokemongo_bot/cell_workers/incubate_eggs.py
@@ -130,6 +130,11 @@ class IncubateEggs(BaseTask):
         temp_used_incubators = []
         temp_ready_breakable_incubators = []
         temp_ready_infinite_incubators = []
+        
+        # Force inventory refresh to ensure we have correct data to compare against
+        if lookup_ids:
+            inventory.refresh()
+            
         inv = inventory.jsonify_inventory()
         for inv_data in inv:
             inv_data = inv_data.get("inventory_item_data", {})


Gobberwart:
I *think* this should take care of the issue people are seeing where egg hatching reports "[IncubateEggs] [INFO] [egg_hatched] {'candy': 'error', 'iv': 'error', 'stardust': 'error', 'exp': 'error', 'cp': 'error', 'pokemon': 'error'}"

I've spent about 3 hours debugging and testing this... seems that the bot compares hatched egg ids (from api call) against inventory to get complete details, but if the inventory isn't 100% up-to-date it fails. 

This fix adds a new api call, but ONLY when the api says there are newly-hatched eggs, so should be very infrequent.

Gobberwart:
For those playing at home, it works like this:

- If km left to walk <= 0...
- Send api request to server to get the ids of the newly-hatched pokemons, then...
- Look in the cached inventory for matching ids (that aren't still in eggs) and retrieve extended data

If the inventory is not 100% up-to-date, there won't be any matches, so the bot displays errors everywhere.

Does this make sense? I hope so, I'm tired.

solderzzc:
👍

Gobberwart:
Nooooooooooo!!! Looks like @DeXtroTip already did the same thing while I was faffing about. Now there's duplicate refreshes.

Gobberwart:
Closing this as it's duplicated by https://github.com/PokemonGoF/PokemonGo-Bot/commit/fdcbd7644543c9b2b719fd1ab115f3225313d0d3

PR #4969, opened 2016-08-30
Updated docs for MoveToMap

DBa2016:
As mentioned in PR#4937, udpated configuration_files.md

PR #4979, opened 2016-08-31
Minor telegram tweak (km_walked truncated to 2 decimals)

Gobberwart:
## Short Description:

Minor tweaks to telegram

- Truncate km_walked to 2 decimals
- Issue with recognising "master"

### DIFF
--- a/pokemongo_bot/event_handlers/telegram_handler.py
+++ b/pokemongo_bot/event_handlers/telegram_handler.py
@@ -56,7 +56,7 @@ class TelegramClass:
                     self.update_id = update.update_id+1
                     if update.message:
                         self.bot.logger.info("message from {} ({}): {}".format(update.message.from_user.username, update.message.from_user.id, update.message.text))
-                        if self.master and str(self.master) != str(update.message.from_user.id):
+                        if self.master and self.master not in [update.message.from_user.id, "@{}".format(update.message.from_user.username)]:
                             continue
                         if update.message.text == "/info":
                             stats = self._get_player_stats()
--- a/pokemongo_bot/event_handlers/telegram_handler.py
+++ b/pokemongo_bot/event_handlers/telegram_handler.py
@@ -7,11 +7,9 @@ import time
 import telegram
 import thread
 import re
+from pokemongo_bot import inventory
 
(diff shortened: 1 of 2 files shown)


pokemongo_bot/event_handlers/telegram_handler.py
```diff
@@ -56,7 +56,7 @@ def run(self):
                     self.update_id = update.update_id+1
                     if update.message:
                         self.bot.logger.info("message from {} ({}): {}".format(update.message.from_user.username, update.message.from_user.id, update.message.text))
-                        if self.master and self.master not in [update.message.from_user.id, "@{}".format(update.message.from_user.username)]:
```
solderzzc: Original code has two way to config the master:
username or chat_id, you changed to only chat_id.

DBa2016:
Agreed with @solderzzc here, no need to remove username recognition. The piece which automatically recognized chat_id was removed in the PR which moved Telegram into thread, but will be re-added once #4969 is merged.

Gobberwart:
I removed it because the original didn't work.

pokemongo_bot/event_handlers/telegram_handler.py
Gobberwart: I see what you're saying, but it wasn't responding to my /info at all with the way it was.. Will keep looking.

Gobberwart:
Looks like I had a telegram api issue while testing this. Reverted to original because it's working fine.

The only change here is truncating km to 2 decimals. Should be a no-brainer.

solderzzc:
yep, easy to merge this one:)

solderzzc:
👍

pokemongo_bot/event_handlers/telegram_handler.py
askovpen: Username not work. Master can be userid or @channelname.

Gobberwart: I just put this back the way it was. If the original isn't working, fix it and create a new PR

DBa2016:
The recognition of non-numeric master is not working right now because #4951 has removed the relevant piece of code without replacement.

PR #4982, opened 2016-08-31
Upgrade improvements

Anakin5:
- Add upgrade_level parameter
- Fix a mistake in upgrade cost
- Some refactoring


pokemongo_bot/cell_workers/pokemon_optimizer.py
```diff
@@ -154,14 +155,15 @@ def work(self):
             xp_all += xp
 
         self.apply_optimization(transfer_all, evolve_all, upgrade_all, xp_all)
-        inventory.update_web_inventory()
```
solderzzc: Is this deleted by merge issue?

Anakin5: No it is just not needed. Web inventory is saved at next loop.

solderzzc:
👍

z-jun:
Hi, very useful should be extension of evolve_for_xp parametr with an white/black list of pokemons/groups. Idea is to save candies for upgrades  while evolving for xp on trash mons. For example:

evolve_for_xp: [ "all", "!Eevee", "!Gym"],

PR #4984, opened 2016-08-31
Fixing WebUI Logging

oralunal:
I added websocket start_embedded_server and address to config files. With this I'm opening websocket.

Web UI will push log toasts.


solderzzc:
👍

PR #5005, opened 2016-08-31
recent_forts sanity check

rawgni:
## Short Description:

recent_forts might be an empty list

## Fixes

  File "pokecli.py", line 782, in <module>
    main()
  File "pokecli.py", line 224, in main
    if bot.recent_forts[-1] is not None and bot.config.forts_cache_recent_forts:
IndexError: list index out of range


solderzzc:
👍

PR #5013, opened 2016-08-31
fix inventory, only call level_up_reward api when needed, up-to-date …

rawgni:
## Short Description:
- get rid of redundant level_up_rewards api call.
- update player exp on spin fort, pokemon caught, evolve pokemon

-
-
-

### DIFF
--- a/pokemongo_bot/inventory.py
+++ b/pokemongo_bot/inventory.py
@@ -113,7 +113,7 @@ class Player(_BaseInventoryComponent):
     def exp(self, value):
         # if new exp is larger than or equal to next_level_xp
         if value >= self.next_level_xp:
-            self.level = self.level + 1
+            self.level = self._level + 1
             # increase next_level_xp to a big amount
             # will be fix on the next heartbeat
             self.next_level_xp += 10000000
--- a/pokemongo_bot/inventory.py
+++ b/pokemongo_bot/inventory.py
@@ -94,7 +94,6 @@ class Player(_BaseInventoryComponent):
         self.next_level_xp = None
         self.pokemons_captured = None
         self.poke_stop_visits = None
-        self.last_lvl_up_reward = time.time()  # ts of last lvl_up_reward api call
         self.player_stats = None
         super(_BaseInventoryComponent, self).__init__()
 
@@ -104,11 +103,6 @@ class Player(_BaseInventoryComponent):
 
     @level.setter
     def level(self, value):
-        if self._level != value:
-            now = time.time()
-            if now - self.last_lvl_up_reward > self.ttl:
-                self.bot.api.level_up_rewards(level=self.level)
-
         self._level = value
 
     @property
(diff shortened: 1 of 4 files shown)


pokemongo_bot/cell_workers/collect_level_up_reward.py
```diff
@@ -70,3 +70,5 @@ def _collect_level_reward(self):
                     'items': data
                 }
             )
+            # if successfully collect reward, update previous level
+            self.previous_level = self.current_level
```
alexyaoyang: We already have this line in `work()`. Do we need this line again in `_collect_level_reward()` which we call in `work()`?

rawgni: oops my bad. definitely not needed

pokemongo_bot/inventory.py
```diff
@@ -1187,6 +1183,7 @@ def refresh(self, inventory=None):
         if inventory is None:
             inventory = self.bot.api.get_inventory()
 
+        print "Inventory refresh"
```
alexyaoyang: A `print` here?

pokemongo_bot/inventory.py
```diff
@@ -117,10 +111,12 @@ def exp(self):
 
     @exp.setter
     def exp(self, value):
-        if self._exp != value:
-            now = time.time()
-            if now - self.last_lvl_up_reward > self.ttl:
-                self.bot.api.level_up_rewards(level=self.level)
+        # if new exp is larger than or equal to next_level_xp
+        if value >= self.next_level_xp:
+            self.level = self.level + 1
```
alexyaoyang: I think they should be `self._level`.

pokemongo_bot/inventory.py
```diff
@@ -117,10 +111,12 @@ def exp(self):
 
     @exp.setter
     def exp(self, value):
-        if self._exp != value:
-            now = time.time()
-            if now - self.last_lvl_up_reward > self.ttl:
-                self.bot.api.level_up_rewards(level=self.level)
+        # if new exp is larger than or equal to next_level_xp
+        if value >= self.next_level_xp:
+            self.level = self._level + 1
```
alexyaoyang: I think it should be `self._level = self._level + 1`?

rawgni: oh god

sohje: self._exp += 1 mb?

alexyaoyang: Oh right 👍 the setter should take care of that, my bad on this.

solderzzc:
👍

pokemongo_bot/inventory.py
```diff
@@ -104,11 +103,6 @@ def level(self):
 
     @level.setter
     def level(self, value):
-        if self._level != value:
-            now = time.time()
-            if now - self.last_lvl_up_reward > self.ttl:
```
alexyaoyang: Just one last thing, since we don't use `self.ttl` anymore, we can remove it.

pokemongo_bot/inventory.py
sohje: So, we will no longer call level_up_rewards?

PR #5015, opened 2016-08-31
Update config.json.map.example

sohje:
## Short Description:
Missing '_' added
- Prevent issues from newly created configs via copy_paste

max_sniping_distance


alexyaoyang:
👍

Issue #5020, opened 2016-08-31
poly line walker cause a problem, just started, then error.


avexus:
In poly line walker, it has a problem, won't start.

{
    "websocket_server": false,
    "heartbeat_threshold": 15,
    "enable_social": false,
    "live_config_update": {"enabled": true, "tasks_only": false},
    "tasks": [
      { "type": "RandomAlivePause",  "config": {"enabled": true, "min_duration": "00:00:10", "max_duration": "00:10:00", "min_interval": "00:05:00", "max_interval": "01:30:00"}},
      { "type": "HandleSoftBan"},
      { "type": "RandomPause",      "config": {"enabled": true, "min_duration": "00:01:00", "max_duration": "00:05:00","min_interval": "00:30:00", "max_interval": "02:00:00"}},
      { "type": "CompleteTutorial", "config": {"enabled": false, "nickname": ""}},
      { "type": "CollectLevelUpReward", "config": {"collect_reward": true, "level_limit": -1}},
      { "type": "IncubateEggs",     "config": {"enabled": true, "infinite_longer_eggs_first": false, "breakable_longer_eggs_first": true, "min_interval": 120, "infinite": [2,5,10], "breakable": [5,10]}},
      { "type": "UpdateLiveStats",  "config": {"enabled": true, "min_interval": 500,
        	"stats": ["username", "uptime", "stardust_earned", "xp_earned", "xp_per_hour", "stops_visited"],
          "terminal_log": true, "terminal_title": true}},
      { "type": "UpdateLiveInventory",  "config": {"enabled": true,  "min_interval": 300, "show_all_multiple_lines": false, "items": ["pokemon_bag", "space_info", "pokeballs", "greatballs", "ultraballs", "razzberries", "luckyegg"]}},
      { "type": "EvolvePokemon",        "config": {"enabled": true,
        	"evolve_list": "Pidgey, Drowzee, Weedle, Caterpie, Spearow, Rattata",
          "min_evolve_speed": 25, "max_evolve_speed": 30, "use_lucky_egg": false}},
      { "type": "TransferPokemon",      "config": {"enabled": true,  "min_free_slot": 20, "transfer_wait_min": 7, "transfer_wait_max": 12}},
      { "type": "RecycleItems",         "config": {"enabled": true,
          "min_empty_space":    5,
          "max_potions_keep":   0,
          "max_revives_keep":   0,
          "max_balls_keep":   210,
          "max_berries_keep": 100,
[...]

PR #5026, opened 2016-09-01
Polyline handle  Google Directions API request  exception  without raise

th3w4y:
## Short Description:
handle IndexError in request response from the Google Directions API 

Two cases are noticed:
a)
 In case of API limit reached we get back we get a status 200 code with an empty routes []

{u'error_message': u'You have exceeded your rate-limit for this API.',
 u'routes': [],
  u'status': u'OVER_QUERY_LIMIT'
}

 b) ZERO_RESULTS {
  "geocoded_waypoints" : [ {}, {} ],
 "routes" : [],
 "status" : "ZERO_RESULTS"
 }

- #5020 
- handle exceptions inside the  Polyline class 
- does not raise to walker_factory or PolylineWalker
- has internal fallback behaviour similar with StepWalker, walking a geometrical walk in  both a) and b) exception cases


solderzzc:
👍

PR #5027, opened 2016-09-01
Fix distance parameters (introduced defaults)

DBa2016:
## Short Description:
Fix distance parameters (introduced defaults)


solderzzc:
Yes, to save from issue reporting wave.

solderzzc:
👍

PR #5031, opened 2016-09-01
Fix nickname

DBa2016:
## Short Description:
Make sure nickname_pokemon only calls action_delay if any calls were made to the api


alexyaoyang:
👍

PR #5034, opened 2016-09-01
hatch egg update player exp

rawgni:
## Short Description:
update player's exp when egg hatches

-
-
-


solderzzc:
👍

PR #5038, opened 2016-09-01
follow_* and walkers set_position through StepWalker

mjmadsen:
I do not feel this is ready to merge. But I want to get testers.

Goals it to drive all set position api calls through StepWalker. That includes centralizing noise and humanizing to position to one location. Outputting position updates from one location.

### Repeat, I'm not done. Do not merge.


mjmadsen:
Not entirely sure why the CI is crapping out. But I'm still not finished :)

th3w4y:
@mjmadsen 

is it .get_position() or .position() ?

I see some inconsistencies between:


https://github.com/mjmadsen/PokemonGo-Bot/blob/8a57a2656d8d8636901512fc07258d2b4d1efd6e/pokemongo_bot/cell_workers/follow_path.py#L146 

and 


https://github.com/mjmadsen/PokemonGo-Bot/blob/8a57a2656d8d8636901512fc07258d2b4d1efd6e/pokemongo_bot/cell_workers/follow_spiral.py#L70

pokemongo_bot/walkers/polyline_walker.py
```diff
@@ -34,7 +34,14 @@ def step(self):
...
         cAlt = self.polyline_walker.get_alt()
-        self.api.set_position(cLat, cLng, cAlt)
+    
+        step_walker = StepWalker(
+            self.bot,
+            cLat,
+            cLng,
+            cAlt
+        )
+
```
th3w4y: This i not going to change the position..

since you replaced .set_position() with just an object initiation but there is nothing calling .step() on the new object

th3w4y:
@mjmadsen see https://github.com/mjmadsen/PokemonGo-Bot/pull/1

th3w4y:
@mjmadsen  fixed some more issues that would arrive from this regarding the PolylineObjecthandler cache 
https://github.com/mjmadsen/PokemonGo-Bot/pull/1/commits/399c65208215a211476eb056c42dd5210ad84e93

mjmadsen:
@th3w4y Thank you!

pokemongo_bot/cell_workers/follow_spiral.py
```diff
@@ -79,48 +78,13 @@ def work(self):
...
-                    'last_position': (last_lat, last_lng, last_alt),
-                    'current_position': (point['lat'], point['lng'], alt),
-                    'distance': dist,
-                    'distance_unit': 'm'
-                }
-            )
-
-        if dist <= 1 or (self.bot.config.walk_min > 0 and step_walker == None):
-            if self.ptr + self.direction >= len(self.points) or self.ptr + self.direction <= -1:
-                self.direction *= -1
```
mjmadsen: Am I breaking anything by removing this? It didn't seem to be applied last night when I was looking.

th3w4y: it is breaking indeed...

you still need something from the if dist <=1

```
        if dist <= 1 or step_walker == None:
            if self.ptr + self.direction >= len(self.points) or self.ptr + self.direction <= -1:
                self.direction *= -1
            if len(self.points) != 1:
                self.ptr += self.direction
            else:
                self.ptr = 0

```

mjmadsen: Even though we check if dist < speed in StepWalker?

th3w4y: This would be my simplified version of the whole  .work()

```
    def work(self):
        last_lat, last_lng, last_alt = self.bot.position()
        point = self.points[self.ptr]

        step_walker = StepWalker(
            self.bot,
            point['lat'],
            point['lng']
        )
        
        if step_walker.step():
            if len(self.points) > 1:
                self.points.pop(self.ptr)

        return [point['lat'], point['lng']]
```

mjmadsen: Bah, I was too sleepy when I change all of this! I'll try to put the corrections in this afternoon.

th3w4y:
@mjmadsen  what do we expect to happen when we walked the whole spiral???

mjmadsen:
@th3w4y I reset the whole file. I'll have to take a closer look later. Have you tested the polyline portion? I have yet to have a chance to - either way, I don't expect this to be mergable for a while yet :)

th3w4y:
@mjmadsen I have a proposal for a simplified follow_spiral https://github.com/mjmadsen/PokemonGo-Bot/pull/3
This replicates the current behaviour in which the Spiral currently walks like this:

imagine the self.points are [p1,p2,p3,p4]
the current follow_spiral will walk p1,p2,p3,p4,p3,p2,p1,p2,p3,p4,p3....

mjmadsen:
@th3w4y nice! That is much easier to follow.

th3w4y:
@mjmadsen follow_path is still a rogue worker setting position as it sees fit..

th3w4y:
@mjmadsen I looked at the code and concluded move_to_map_pokemon and random_alive_pause are both teleporting related tasks... (I will not touch)

mjmadsen:
As far as I know, we should be able to pass them to our stepper. Except maybe the teleporting aspects of move_to_map_pokemon, or we just change the speed in those instances.

mjmadsen:
Do you think we could move loitering functionality to our main stepper?

th3w4y:
@mjmadsen we could instead the fixed_speed=False in StepWalker... we could say.. speed=None

they if not specified take the uniform(self.bot.config.walk_min, self.bot.config.walk_max)
if specified we take whatever is passed through (by the polyline a fixed value) by the teleport tasks we pass a humongous value to instantly arrive at destination...

th3w4y:
@mjmadsen looking at the loitering.. i never have read that code before.. will take a bit...

th3w4y:
@mjmadsen does not look like loitering will serve a purpose in StepWalker... it's definition is strict to the follow_path... (Otherwise to much code to refactor...)

mjmadsen:
Yeah, maybe for another day

th3w4y:
@mjmadsen I/we made a mistake bot.position is a property not a method
please merge https://github.com/mjmadsen/PokemonGo-Bot/pull/5

mjmadsen:
Doh, nice catch.

th3w4y:
@mjmadsen I kinda ended up refactoring StepWalker https://github.com/mjmadsen/PokemonGo-Bot/pull/6 

We need some additional reviewers....

mjmadsen:
@th3w4y I think I'm going to revert some changes and get the basics merged - we can extend StepWalker functionality in a different PR to accomplish what we are trying. I fear we'll get too bogged down and have this PR up forever.

mjmadsen:
@solderzzc @Gobberwart @th3w4y 

This is the stripped down version of what we wanted to attempt. I'm working through tests, but some of them only happen if no targets are found. We're passing CI without editing tests. 

@th3w4y We'll make a new branch to accomplish our extended goals.

solderzzc:
@mjmadsen I have an exception. I'm using move_to_map and Move_to_fort:
`  File "pokecli.py", line 795, in <module>
    main()
  File "pokecli.py", line 153, in main
    bot.tick()
  File "/Users/simba/Projects/pokemon/PokemonGo-Bot/pokemongo_bot/__init__.py", line 662, in tick
  File "/Users/simba/Projects/pokemon/PokemonGo-Bot/pokemongo_bot/cell_workers/move_to_fort.py", line 89, in work
    lng
  File "/Users/simba/Projects/pokemon/PokemonGo-Bot/pokemongo_bot/walkers/walker_factory.py", line 11, in walker_factory
    ret = PolylineWalker(bot, dest_lat, dest_lng)
  File "/Users/simba/Projects/pokemon/PokemonGo-Bot/pokemongo_bot/walkers/polyline_walker.py", line 14, in __init__
    self.speed)
  File "/Users/simba/Projects/pokemon/PokemonGo-Bot/pokemongo_bot/walkers/polyline_generator.py", line 25, in cached_polyline
    abs_offset = abs(haversine.haversine(tuple(origin), PolylineObjectHandler._cache.get_last_pos())*1000)
AttributeError: 'NoneType' object has no attribute 'get_last_pos'`

mjmadsen:
I'm trying to do some debugging, but my mind isn't in the game tonight. I'm working on it @solderzzc

mjmadsen:
@solderzzc While I have you, shouldn't our walker return WorkerResult.RUNNING when we plan on keeping the process running until dest, then WorkerResult.SUCCESS when we hit our dest?

Again, the goal is to make more homogeneous code.

th3w4y:
@mjmadsen @solderzzc i identified the AttributError issue, Fix in: https://github.com/mjmadsen/PokemonGo-Bot/pull/7

solderzzc:
@mjmadsen  The design should be as you describe.  But the code is not like that.
When there's worker return RUNNING status, next tick need keep running on that worker.
`
        self.health_record.heartbeat()
        self.cell = self.get_meta_cell()

        if self.sleep_schedule:
            self.sleep_schedule.work()

        now = time.time() * 1000

        for fort in self.cell["forts"]:
            timeout = fort.get("cooldown_complete_timestamp_ms", 0)

            if timeout >= now:
                self.fort_timeouts[fort["id"]] = timeout

        self.tick_count += 1

        self.check_session(self.position)

        for worker in self.workers:
                return
`

mjmadsen:
@solderzzc But doesn't some of the walking units operate like that? That was the big huff and puff when we wanted to allow 0.0 walking speeds. I was told the step function would be called forever.

solderzzc:
@th3w4y  I have it merged locally, did fix the exception.

mjmadsen:
But aside from that, I want our tests to go through with the code as is - no more functionality. Just bug fixes. We will save more for round two of improvements.

solderzzc:
Then another exception....
Error in sys.excepthook:
  File "/Users/simba/Projects/pokemon/PokemonGo-Bot/venv/lib/python2.7/site-packages/raven/base.py", line 238, in handle_exception
  File "/Users/simba/Projects/pokemon/PokemonGo-Bot/venv/lib/python2.7/site-packages/raven/base.py", line 768, in captureException
  File "/Users/simba/Projects/pokemon/PokemonGo-Bot/venv/lib/python2.7/site-packages/raven/base.py", line 596, in capture
  File "/Users/simba/Projects/pokemon/PokemonGo-Bot/venv/lib/python2.7/site-packages/raven/base.py", line 440, in build_msg
  File "/Users/simba/Projects/pokemon/PokemonGo-Bot/venv/lib/python2.7/site-packages/raven/base.py", line 256, in get_processors
  File "/Users/simba/Projects/pokemon/PokemonGo-Bot/venv/lib/python2.7/site-packages/raven/base.py", line 67, in __missing__
ImportError: No module named processors

Original exception was:
  File "pokecli.py", line 795, in <module>
  File "pokecli.py", line 153, in main
  File "/Users/simba/Projects/pokemon/PokemonGo-Bot/pokemongo_bot/__init__.py", line 662, in tick
  File "/Users/simba/Projects/pokemon/PokemonGo-Bot/pokemongo_bot/cell_workers/spin_fort.py", line 101, in work
sqlite3.OperationalError: unable to open database file

th3w4y:
@solderzzc @mjmadsen ok we continue the 0.0 and inf in the other branch you were mentioning.. lets open an issue for that to comment extras...

mjmadsen:
@solderzzc Hm. Don't know the db bit well enough

solderzzc:
@mjmadsen  agreed about that we need focus on one issue per PR.

mjmadsen:
I'm going to have to call it quits for tonight. Sorry guys. I'll catch up in the morning.

solderzzc:
I'll do pip install, then have a test.

solderzzc:
@mjmadsen good night.

solderzzc:
**Still  have issue:**

  File "pokecli.py", line 795, in <module>
    main()
  File "pokecli.py", line 153, in main
    bot.tick()
  File "/Users/simba/Projects/pokemon/PokemonGo-Bot/pokemongo_bot/__init__.py", line 662, in tick
  File "/Users/simba/Projects/pokemon/PokemonGo-Bot/pokemongo_bot/cell_workers/move_to_fort.py", line 89, in work
    lng
  File "/Users/simba/Projects/pokemon/PokemonGo-Bot/pokemongo_bot/walkers/walker_factory.py", line 11, in walker_factory
    ret = PolylineWalker(bot, dest_lat, dest_lng)
  File "/Users/simba/Projects/pokemon/PokemonGo-Bot/pokemongo_bot/walkers/polyline_walker.py", line 14, in __init__
    self.speed)
  File "/Users/simba/Projects/pokemon/PokemonGo-Bot/pokemongo_bot/walkers/polyline_generator.py", line 25, in cached_polyline
    if PolylineObjectHandler._cache.get_last_pos():
AttributeError: 'NoneType' object has no attribute 'get_last_pos'

**Local Branch: **

commit aadc1f84728528c2d5e8273931e8e48c50550a62
Author: Ovidiu Ghinet <ovidiu@ghinet.ch>
Date:   Fri Sep 2 08:54:57 2016 +0200
    Fixed cache validation when PolylineObjecthander cache is empty

commit 03610a62d38e70b78e4587ce4217519dc101db9f
Merge: 669446e e83c7e6
Author: Matt J Madsen <mattjmadsen@gmail.com>
Date:   Fri Sep 2 01:49:59 2016 -0500

    Merge branch 'dev' into mjmadsen-walker-test

th3w4y:
@solderzzc Fixe at: https://github.com/mjmadsen/PokemonGo-Bot/pull/8

Gobberwart:
Too many changes + too much beer = useless Gobberwart. I'll come back to this in the morning :)

th3w4y:
@Gobberwart it is morning :)

Gobberwart:
Out of interest, who's our db expert? I'd be keen to help make that more globally useful, but don't want to step on anyone's toes.

solderzzc:
@th3w4y  need another protection:
`+++ b/pokemongo_bot/walkers/polyline_generator.py
@@ -22,7 +22,7 @@ class PolylineObjectHandler:
 -        if PolylineObjectHandler._cache.get_last_pos() != (None, None):
 +        if PolylineObjectHandler._cache and PolylineObjectHandler._cache.get_last_pos() != (None, None):
             abs_offset = haversine.haversine(tuple(origin), PolylineObjectHandler._cache.get_last_pos())*1000
         else:
             abs_offset = float("inf")
`

th3w4y:
@solderzzc i agree

mjmadsen:
Alright, I'm awake and able to do some testing now :)

solderzzc:
It need fall back when google api out of quota. 
```
  File "pokecli.py", line 795, in <module>
    main()
  File "pokecli.py", line 153, in main
    bot.tick()
  File "/Users/simba/Projects/pokemon/PokemonGo-Bot/pokemongo_bot/__init__.py", line 662, in tick
  File "/Users/simba/Projects/pokemon/PokemonGo-Bot/pokemongo_bot/cell_workers/move_to_fort.py", line 89, in work
    lng
  File "/Users/simba/Projects/pokemon/PokemonGo-Bot/pokemongo_bot/walkers/walker_factory.py", line 11, in walker_factory
    ret = PolylineWalker(bot, dest_lat, dest_lng)
  File "/Users/simba/Projects/pokemon/PokemonGo-Bot/pokemongo_bot/walkers/polyline_walker.py", line 16, in __init__
    self.pol_alt = self.polyline.get_alt()
  File "/Users/simba/Projects/pokemon/PokemonGo-Bot/pokemongo_bot/walkers/polyline_generator.py", line 151, in get_alt
    return self.polyline_elevations[-1]
```

th3w4y:
@solderzzc i think i did not catch exceptions for the  Elevation part yet...

mjmadsen:
The -1 index was in there originally. Not sure why we were sending a negative index when we had a bad index?

@th3w4y it looks like it generates the index from time passed, so we much have to call the pause and unpause bits for it to work properly? The more I look at this file, the more I'm unsure as to what is happening (my python skills apparently suck).

th3w4y:
@mjmadsen & @solderzzc the issue with the API limits in Elevation that @solderzzc  mentioned is unrelated with current changes...

solderzzc:
@th3w4y  agreed, it need another PR for fixing.

mjmadsen:
@th3w4y @solderzzc So, as is, the polywalker is functioning in this PR (on a level consistent with before the PR)?

th3w4y:
@mjmadsen yes... bugs have been taken over :)

@solderzzc if you are still in breach of API quota and can reproduce the failure.. 
add a print/log of the content of self.polyline as it is after line 85 please...

https://github.com/mjmadsen/PokemonGo-Bot/blob/3892fc54f4fb4b20925fd4fccd7d31358feb0dda/pokemongo_bot/walkers/polyline_generator.py#L85

mjmadsen:
Related to API quota problems. We should allow folks to switch keys when the limit is hit. I'm sure most of us have more than one google account. :)

th3w4y:
@mjmadsen currently i did not add the key... is using the 2500/day per IP limit...  the limit for the key is also 2500/day.

My thinking was that we get 2500/day non identified / API requests for polyline 
and 2500/day identified by key for the map

mjmadsen:
I was under the impression it is total requests per key a day. But I haven't found information to verify that or test myself.

th3w4y:
@solderzzc @mjmadsen  I identified the issue with the Elevation API limit

https://github.com/mjmadsen/PokemonGo-Bot/pull/9

mjmadsen:
@th3w4y I put in a PR to alter that submission. Take a gander. I will probably retool the random alt bit down the road. The default values are a bit insane.

mjmadsen:
Time to merge?

solderzzc:
Running well here.

solderzzc:
👍

mjmadsen:
Give me a few minutes. Doing a few extra tests.

th3w4y:
@mjmadsen I believe we removed the emit from follow spiral... do we want them back?

mjmadsen:
It occurs in StepWalker. But I don't love the format of the output.

th3w4y:
@mjmadsen  you might see if you use it with the PolylineWlaker that will say for A-Z
will most probably emit:
A to B
B to C
C to D
...
Y to Z

i am thinking in the refactoring of the stepWalker to add support for something like... displayed_destination=Z for display purpose different then the passed through destination for walking calculations...

same for distances... since the walking distance is different then the haversine (A-Z straight line) distance..

solderzzc:
My Move_to_Map is not sniping, have you seen the same issue on this branch?

mjmadsen:
Just finished my last test on the follow_* bits. Still haven't tried poly or move_to_

solderzzc:
oh, should be poke ball limit, not the code caused.

solderzzc:
You can use like this:
```
      {
        "type": "MoveToFort",
        "config":{
          "lure_attraction": true,
          "lure_max_distance": 20,
          "walker": "PolylineWalker",
          "log_interval": 5
        }
      },
```

```
        "type": "MoveToMapPokemon",
        "config": {
          "walker": "StepWalker",
```

solderzzc:
This is ready to merge. @mjmadsen

mjmadsen:
Lets do it. :+1:

mjmadsen:
@solderzzc PolyWalker seems to not add in the human variance changes still.

th3w4y:
@solderzzc  in this commit it does...

mjmadsen:
Should we make polywalker the default for movetofort? This honestly seems to be superior.

th3w4y:
@mjmadsen you don't see any jitter to it from the StepWalker?

mjmadsen:
Very little compared to the other walkers (follow_*)

th3w4y:
@mjmadsen i though the random_lat_long_delta() would be enough... 
(that is why i removed the rounding from the .get_pos() hmmm...)

mjmadsen:
Those random deltas are the only thing added if our noise is disabled right? The variance seemed quiet high with the followers and the polylinewalker seems to be very minimal.

th3w4y:
@mjmadsen wow i think i know why...  it has to do with the magnitude self.magnitude way of calculation.. the higher the distance to walk the higher the jitter maybe...

I have to do the math...

mjmadsen:
Lol. I'll probably work on a few other tasks before returning to this. It's a bit of a minefield.

th3w4y:
the problem is with the calculations in below code:

```
        totalDLat = (self.destLat - self.initLat)
        totalDLng = (self.destLng - self.initLng)
        magnitude = self._pythagorean(totalDLat, totalling)
        unitLat = totalDLat / magnitude
        unitLng = totalDLng / magnitude

        scaledDLat = unitLat * self.magnitude
        scaledDLng = unitLng * self.magnitude

        cLat = self.initLat + scaledDLat + random_lat_long_delta()
        cLng = self.initLng + scaledDLng + random_lat_long_delta()
        cAlt = self.initAlt + self.unitAlt + random_alt_delta()
```

the more steps you have the higher the scaledDLat thus higher jitter

for an example of a distance of 125m with speed 3m/s the above calculations are walking only 2.68m due to jitter before the random jitter of extra +/-2m

th3w4y:
@mjmadsen since the Polyline walker is telling the StepWalker to walk in baby steps.. A-B B-C.. then the less affected it is by the magnitude*self.magnitude factor.

mjmadsen:
follow_path is also affected.

th3w4y:
This looks odd to me the method we calculate now....


made a small test with speed 3 trying to reproduce what step walker does now...

```
In [105]: def current_calculation(orig, dest):
     ...:     distance = haversine.haversine(orig, dest)*1000
     ...:     speed = 3
     ...:     steps = int(distance/speed)
     ...:     if steps == 0:  # set steps to 1 to prevent division by zero
     ...:         steps = 1    
     ...:     dLat = (dest[0] - orig[0])/steps
     ...:     dLon = (dest[1] - orig[1])/steps
     ...:     self_magnitude = sqrt(dLat**2+dLon**2)
     ...:     totalDLat = (dest[0] - orig[0])
     ...:     totalDLon = (dest[1] - orig[1])
     ...:     magnitude = sqrt(totalDLat**2+totalDLon**2)
     ...:     unitLat = totalDLat / magnitude
     ...:     unitLon = totalDLon / magnitude
     ...:     scaledDLat = unitLat * self_magnitude
     ...:     scaledDLon = unitLon * self_magnitude
     ...:     cLat = orig[0] + scaledDLat
     ...:     cLng = orig[1] + scaledDLng
     ...:     return haversine.haversine(orig, dest)*1000, steps, haversine.haversine(orig,(cLat, cLng))*1000
     ...:     
     ...: 

In [106]: 
```
15 cm 1 step 2.6m walked
```
In [106]: current_calculation((8.000001, 8.000001),(8.000002, 8.000002))
Out[106]: (0.15649005443906103, 1, 2.653969095391703)
```

1.56 m 1 step 2.87m walked  in 1sec
```
In [107]: current_calculation((8.00001, 8.00001),(8.00002, 8.00002))
Out[107]: (1.5649005174894743, 1, 2.8753466491876645)
```
15.65m 5 steps 3.46m walked  in 1sec
```
In [108]: current_calculation((8.0001, 8.0001),(8.0002, 8.0002))
Out[108]: (15.649002609436945, 5, 3.4607670521667617)
```
156m 52 steps 3.40m walked  in 1sec
```
In [109]: current_calculation((8.001, 8.001),(8.002, 8.002))
Out[109]: (156.4897695005808, 52, 3.406428240494767)
```
1.5km 521 steps 3.4m walked in 1sec
```
In [110]: current_calculation((8.01, 8.01),(8.02, 8.02))
Out[110]: (1564.8720111329158, 521, 3.40380763031615)

In [111]: 

```

Anakin5:
I have some big problems with this merge:
- Polyline worker totally broken:
    - The polyline step must be paused before step and unpaused after. Otherwise the polyline worker is jumping directly to the destination when calling step.
    - Polyline pol_alt now default to a random value when altitude is zero. I would prefer it stays zero if that is the altitude returned. zero is not an error case.
    - There is way too much variance in the polyline step. gps does not jump 5 meters in every direction every seconds.

th3w4y:
>Polyline worker totally broken:
-The polyline step must be paused before step and unpaused after. Otherwise the polyline worker is jumping directly to the destination when calling step.

yes it jumps directly to destination.. but the destination is the .get_pos() of the cached polyline not final destination so i don't see an issue...

>-Polyline pol_alt now default to a random value when altitude is zero. I would prefer it stays zero if that is the altitude returned. zero is not an error case.

~~that should be only if there is no cached polyline...~~
in the comparison .get_alt() or uniform(self.bot.config.alt_min, self.bot.config.alt_max)
That is a valid issue!
will fix as @mjmadsen  suggested with self.api.actual_alt https://github.com/th3w4y/PokemonGo-Bot/pull/1

>-There is way too much variance in the polyline step. gps does not jump 5 meters in every direction every seconds.

I agree that gps does not jump 5m every second but you kinda have exact opposite behaviour then @mjmadsen  who is complaining of too little variance..
The variance comes now only from the random_lat_long_delta() in the StepWalker

mjmadsen:
#5132 I have a few alterations to this. The first portion is not implemented yet.

Anakin5:
get_pos() return the final destination is Walker has never been paused.

random_lat_long_delta is too big. When I look at my GPS position reported by my phone versus what the bot set in StepWalker, there is a very big difference.
After some tests, I settled on dividing by 5 whatever number is returned by random_lat_long_delta. That is what gave me the results closest to reality.

th3w4y:
@mjmadsen @Anakin5  https://github.com/PokemonGoF/PokemonGo-Bot/pull/5135 will fallback to previous bot position...

th3w4y:
@Anakin5 

how big is the distance till final destination? it should return final destination only if enough time passes...
>get_pos() return the final destination if Walker has never been paused.

th3w4y:
@Anakin5 You are actually right... i think i found the issue...

Anakin5:
I think the problem comes from the fact that step_walker.step return True at the first step of the PolylineWalker. That is because the first point of the Polyline is the current position, so the distance to step is zero, which translate to "we are arrived at destination".
The StepWalker.step should not return True until final destination is reached, as it was the case before.

Anakin5:
But my opinion is that the real evil there is in the fact that we do not reuse walkers. There is no concept of final destination and next point in the walker, we are expected to create a new object at each step.
Without these concepts, there is no way for the walker to know that we are arrived. It would be nice to keep the same walker object during the whole trip.

th3w4y:
what are you combining the PolylineWalker with?
>I think the problem comes from the fact that step_walker.step return True at the first step of the PolylineWalker. That is because the first point of the Polyline is the current position, so the distance to step is zero, which translate to "we are arrived at destination".
The StepWalker.step should not return True until final destination is reached, as it was the case before.

Anakin5:
It return True whatever is the task I am using it with. Be it MoveToFort or CampFort. Both of them relying on the fact that it should return True only when arrived.

mjmadsen:
Looks like I was removed as an admin over this :(

I wouldn't have merged if I was told by @solderzzc it was ready to merge.

th3w4y:
@Anakin5  i had to override step() again... i didn't want to but hey... https://github.com/PokemonGoF/PokemonGo-Bot/pull/5135/commits/130d630fb94af5e79cc9cdd4a19fa1bf33c41185

th3w4y:
The change was bit... some issues slipped through...

th3w4y:
@Anakin5 

What do you consider a normal range for variation?

>get_pos() return the final destination if Walker has never been paused.

random_lat_long_delta is too big. When I look at my GPS position reported by my phone versus what the bot set in StepWalker, there is a very big difference.
After some tests, I settled on dividing by 5 whatever number is returned by random_lat_long_delta. That is what gave me the results closest to reality.

mjmadsen:
Oh I see. random_lat_long_delta is supposed to make us look more human with wandering. What would be consider as a better range? Currently supposed to be -9.1 to 9.1m variance. Even assuming those comments are correct, we could move 9.1 x 9.1 in a single step.

The gps noise bit is the one that is supposed to alter our coords to match phone's inaccuracies. @Anakin5 I assume you have this bit disabled too?

th3w4y:
@mjmadsen @Anakin5 

this was the rookie mistache causing everything....

https://github.com/PokemonGoF/PokemonGo-Bot/pull/5135/commits/8f7f22c90e92ccafafa4956ba006624c3f90ab5e

mjmadsen:
Doh! At least you found it :+1:

PR #5039, opened 2016-09-01
Don't run summary on API exception

mjmadsen:
We try to run the bot summary after API exceptions.
Need help to test. 

### This is also not ready to merge. 

From looking at a lot of the issues, we don't handle our exceptions well, it seems metrics trys to output our data after an exception crash (which we can only do it we have populated the data). Instead, I want to suppress the metrics and attempt a reconnect. Will also limit output folks who have problems see (resulting in better issue titles, etc).


mjmadsen:
At the least this won't break anything. We were checking an undefined variables for a value of None which always failed because it was never initialized.

mjmadsen:
:+1:

mjmadsen:
Bah, I'll try later. Coders block.

PR #5048, opened 2016-09-01
Add xp earned to player xp + doc fix

Anakin5:
- Add xp earned to player xp
- Clarify doc about evolution priority over upgrade
- Fix doc: ncp was a string instead of numeric value


solderzzc:
👍

PR #5063, opened 2016-09-01
Remove spaces in custom keep best

alexyaoyang:
- Remove spaces in `keep_best_custom` or else it will not be parsed correctly


solderzzc:
👍

Issue #5042, opened 2016-09-01
[DEV] Error on startup of fresh install


Bashin:
### Expected Behavior

No errors on startup.

### Actual Behavior

TypeError on fresh install (error occurs ONLY during the first startup!)

### Your FULL config.json (remove your username, password, gmapkey and any other private info)

Default config.example.json

### Output when issue occurred

[2016-09-01 10:23:35] Found encrypt.dll! Platform: win32 encrypt.dll directory: [...]
[2016-09-01 10:23:35]
Traceback (most recent call last):
  File "[...]Python\lib\logging\__init__.py", line 861, in emit
    msg = self.format(record)
  File "[...]Python\lib\logging\__init__.py", line 734, in format
    return fmt.format(record)
  File "[...]Python\lib\logging\__init__.py", line 465, in format
    record.message = record.getMessage()
  File "[...]Python\lib\logging\__init__.py", line 329, in getMessage
    msg = msg % self.args
TypeError: not all arguments converted during string formatting
Logged from file __init__.py, line 1354
[2016-09-01 10:23:36] No inventory file [...]pokemongo_bot\..\web\inventory-[...].json found. Creating a new one

### Steps to Reproduce

Use fresh install of dev version and start bot

### Other Information
OS: Win10 64 bit
Branch: dev
Git Commit: d5059ed
Python Version: 2.7.12

Gobberwart:
PR #5072 should fix

Gobberwart:
Merged #5072

PR #5071, opened 2016-09-01
'red' not supported by logger

Gobberwart:
## Short Description:

When trying to read cached forts on first run (or if file has been deleted), tries to generate a log message, but option 'red' is not supported. Removed.

- #5042


Gobberwart:
Wrong branch, sorry

PR #5074, opened 2016-09-02
handling non-numeric "master" in telegram config

DBa2016:
## Short Description:
"master" in telegram bot config can be set to a numeric value or to a "@username". If set to "@username", the correct "master" value will be calculated on first "/info" (or similar) call coming from the correct username.


solderzzc:
👍

PR #5077, opened 2016-09-02
make min free slots for optimizer configurable

DBa2016:
## Short Description:
make min free slots for optimizer configurable through the "min_slots_left" parameter.


solderzzc:
👍

PR #5079, opened 2016-09-02
Reduce move to fort spam

DBa2016:
## Short Description:
introduced a new config parameter for MoveToFort, log_interval_meters - this reduces the frequency of "moving to pokestop..." messages to appear only every X meters


solderzzc:
👍

Gobberwart:
What happened to the previous fix for this issue? Recently there was an addition to the bot to overwrite the previous line if it was just another "moving towards BLAH BLAH" but that seems to have stopped working.

Gobberwart:
Hmmm ok that's still working, sort of. Why do we need this change as well?

mjmadsen:
Yeah the previous addition made it overwrite the output for position_update, movetofort and maybe one or two more. I suppose his PR is still valid in that it will make THAT part update even less (though 1 line that keeps being rewritten limits the log pretty well..)

Gobberwart:
I don't really see a need for this. Overwriting the previous line should be sufficient. I don't really want the output suppressed like this.

DBa2016:
For some reason, overwriting last line does not work for me. Also, if you are redirecting (or tee'ing) the log to a file (in order to be able to inspect it later), every "overwrite last line" still causes an additional line in the log, just with some control characters.

sohje:
Why not to move all this delay logic to event manager?

DBa2016:
Because the event manager does not know whether the position has changed or not and therefore cannot decide whether a new line would be output.

sohje:
@DBa2016 Yeap, i know. I've talked about dont mess emit logic in tasks modules. If you dont want to seeing logs - adjust walk speed, logs_delay, etc. Instead, we added some logic to task, new config options, etc. Our config files already redundant, contains a lot of unnecessary stuff. Lot of people dont even know wtf is this option or another.

DBa2016:
@sohje : no, that's something different. The event manager does not [need to] know anything about the task logic. This is something which needs to be known by the tasks.

solderzzc:
Good discussion, I'm waiting for your guys' decision.

DBa2016:
Also, I don't see how moving this logic into event manager would reduce the number of options in the logfile. Oh, and - the option has a default, therefore even if the user does not set it, the bot does not error out or something.

DBa2016:
Yeah, that's great - just let the PR sit around until there are conflicts which prevent automatic merging.

BreezeRo:
I agree with @Gobberwart, I think it's sufficient as is, though perhaps exception could be made if this were aimed at the master branch instead? Anyone else?

pokemongo_bot/cell_workers/move_to_fort.py
```diff
@@ -71,17 +74,21 @@ def work(self):
 
             if self.is_attracted() > 0:
                 fort_event_data.update(lure_distance=format_dist(self.lure_distance, unit))
-                self.emit_event(
-                    'moving_to_lured_fort',
-                    formatted="Moving towards pokestop {fort_name} - {distance} (attraction of lure {lure_distance})",
-                    data=fort_event_data
-                )
+                if int(self.prev_dist/self.config.get("log_interval_meters", 10)) != int(dist/self.config.get("log_interval_meters", 10)):
```
mjmadsen: If we're going to go ahead with this PR, can we fetch "log_interval_meters" once? We use it twice in this `if` alone.

DBa2016:
Looks like this one is not going to be merged, I am closing this.

PR #5082, opened 2016-09-02
Add poke lvl

rawgni:
## Short Description:

add poke level information to ._data so that it will show in the inventory-* dump.
This can then be used to display pokemon level in the web ui.

-
-
-


solderzzc:
👍

PR #5101, opened 2016-09-02
utilize inventory instead of api

rawgni:
## Short Description:

uses what we have in the inventory instead of using API


solderzzc:
👍

PR #5104, opened 2016-09-02
Pokemon Optimizer limit on Eevees

Anakin5:
Put a limit (12)  to the number of Eevee that can be kept when player is missing one evolution.
This will avoid inventory being full without any solution to empty it.


solderzzc:
:

PR #5105, opened 2016-09-02
Add CampFort task

Anakin5:
New task that walk between fort clusters and stay there for the time configured.

`max_distance` is radius in meters of the circle around starting position to limit movements of the bot.
`min_forts_count` is the minimum number of forts to be considered a cluster.
`min_lured_forts_count` is the minimum number of lured forts to be considered a cluster.
`camping_time` is the minimum time to stay at a cluster. A cluster can be selected again after if it is still the best place to be.

To be place before MoveToFort
```
{
    "type": "CampFort",
    "config": {
        "max_distance": 2000,
        "min_forts_count": 2,
        "min_lured_forts_count": 1,
        "camping_time": 1800,
        "moving_time": 600
    }
},
```

### DIFF
--- a/pokemongo_bot/cell_workers/camp_fort.py
+++ b/pokemongo_bot/cell_workers/camp_fort.py
@@ -1,0 +1,193 @@
+import math
+import time
+
+from pokemongo_bot.base_task import BaseTask
+from pokemongo_bot.cell_workers.utils import distance, coord2merc, merc2coord
+from pokemongo_bot.constants import Constants
+from pokemongo_bot.datastore import Datastore
+from pokemongo_bot.human_behaviour import random_lat_long_delta, random_alt_delta
+from pokemongo_bot.walkers.polyline_walker import PolylineWalker
+from pokemongo_bot.worker_result import WorkerResult
+
+
+class CampFort(Datastore, BaseTask):
+    SUPPORTED_TASK_API_VERSION = 1
+
+    def __init__(self, bot, config):
+        super(CampFort, self).__init__(bot, config)
+
+    def initialize(self):
+        self.destination = None
+        self.stay_until = 0
+        self.move_until = 0
+        self.last_position_update = 0
+        self.walker = None
+
+        self.config_max_distance = self.config.get("max_distance", 2000)
+        self.config_min_forts_count = self.config.get("min_forts_count", 2)
+        self.config_min_lured_forts_count = self.config.get("min_lured_forts_count", 1)
+        self.config_camping_time = self.config.get("camping_time", 1800)
+        self.config_moving_time = self.config.get("moving_time", 600)
+
+    def work(self):
+        if not self.enabled:
+            return WorkerResult.SUCCESS
+
+        now = time.time()
+
+        if now < self.move_until:
+            return WorkerResult.SUCCESS
+
+        if 0 < self.stay_until < now:
+            self.destination = None
+            self.stay_until = 0
+            self.move_until = now + self.config_moving_time
+
+            if self.config_moving_time > 0:
+                return WorkerResult.SUCCESS
+
+        if self.destination is None:
+            forts = self.get_forts()
+            forts_clusters = self.get_forts_clusters(forts)
+
+            if len(forts_clusters) > 0:
+                self.destination = forts_clusters[0]
+                self.logger.info("New destination at %s meters: %s forts, %s lured.", int(self.destination[4]), self.destination[3], self.destination[2])
+            else:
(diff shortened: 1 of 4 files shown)


pokemongo_bot/cell_workers/camp_fort.py
```diff
@@ -0,0 +1,174 @@
...
+    def initialize(self):
+        self.destination = None
+        self.stay_until = 0
+        self.last_position_update = 0
+        self.polyline = None
+
+        self.config_max_distance = self.config.get("max_distance", 2000)
+        self.config_min_forts_count = self.config.get("min_forts_count", 2)
+        self.config_min_lured_forts_count = self.config.get("radius", 1)
+        self.config_camping_time = self.config.get("radius", 1800)
```
davidakachaos: typo? Shouldn't this be "camping_time" instead of "radius"?

solderzzc:
This task includes many logic task here. Is there better way to struct it?

pokemongo_bot/cell_workers/camp_fort.py
Anakin5: yes typo. thanks i will fix

Anakin5:
What logic are you referring to ?

pokemongo_bot/cell_workers/camp_fort.py
```diff
@@ -0,0 +1,162 @@
...
+        now = time.time()
+
+        if self.stay_until >= now:
+            if now - self.last_position_update >= 1:
+                lat = self.destination[0] + random_lat_long_delta() / 5
+                lon = self.destination[1] + random_lat_long_delta() / 5
+                alt = self.walker.pol_alt + random_alt_delta() / 2
+                self.bot.api.set_position(lat, lon, alt)
+                self.last_position_update = now
+                self.bot.heartbeat()
```
solderzzc: heartbeat should be call in tick.

pokemongo_bot/cell_workers/camp_fort.py
```diff
@@ -0,0 +1,162 @@
...
+        self.walker = PolylineWalker(self.bot, self.destination[0], self.destination[1])
+        self.walker.polyline.pause()
+
+        if self.walker.step():
+            self.logger.info("Arrived at destination: %s forts, %s lured.", self.destination[3], self.destination[2])
+            self.stay_until = now + self.config_camping_time
+
+        self.walker.polyline.unpause()
+
+        return WorkerResult.RUNNING
```
solderzzc: The RUNNING will be unpredictable, the task depending on its order in task list.

solderzzc:
@Anakin5 I have two inline comments

pokemongo_bot/cell_workers/camp_fort.py
Anakin5: I am not sure I understand. What do you suggest to return ?

solderzzc: Hmm, I don't know, RUNNING handle is not good in tick, but that's not related in this PR imo

pokemongo_bot/cell_workers/camp_fort.py
Anakin5: I will remove it. It is not needed anymore since commit 5b3cfbb.

Otherwise, if heartbeat should be called in tick, why isn't it ?

pokemongo_bot/cell_workers/camp_fort.py
Anakin5: What is wrong with it ?

Anakin5:
I added moving_time parameter to forward movement to next task during that time.

BreezeRo:
I believe @solderzzc is saying that it is not advised to use info from workers in this way as there are cases where "running" information would not be ready to read depending on which order it were in in the task list - but as he also said

> but that's not related in this PR imo

this look good to me :dancer:

solderzzc:
Lol @BreezeRo has much better English language ability.

solderzzc:
👍

th3w4y:
@Anakin5 I believe you are invalidating the polyline indirectly here

```
        if self.stay_until >= now:
            lat = self.destination[0] + random_lat_long_delta() / 5
            lon = self.destination[1] + random_lat_long_delta() / 5
            alt = self.walker.pol_alt + random_alt_delta() / 2
            self.bot.api.set_position(lat, lon, alt)
        else:
            self.walker = PolylineWalker(self.bot, self.destination[0], self.destination[1])
            self.walker.step()
```

th3w4y:
What tasks can this be combined with? What is you config for tasks https://github.com/PokemonGoF/PokemonGo-Bot/pull/5169

PR #5123, opened 2016-09-03
Mqtt web ui

solderzzc:
# MQTT Web UI   -- Working branch, don't merge it.
## Why need it, isn't Socket.IO good enough?

well, it's complex for multiple instance
## is that safe to send data to public mqtt broker

E2E encryption will keep your data safe


rawgni:
@solderzzc isn't mqtt too overkill for the web ui?

solderzzc:
@rawgni  I don't have better idea than MQTT for multiple instance running locally. Socket.io is complex to handle. And a MQTT broker is very lightweight.

rawgni:
@solderzzc hm.. but, is it safe to run multiple bot locally ? what happened if the public broker "broke" ? would it be possible to have private broker?

BreezeRo:
That is actually the goal at it's core @rawgni, to have the public create their own broker, and establish "trusted brokers" within the botting community that people could add into a config list to pull data from.

solderzzc:
@BreezeRo I have something to do before merge. The channel need to be configurable, E2E encrypt is best to have to protect private data if there's any.
And good morning :)

douglascamata:
SocketIO-capable web server is very lightweight too, because it runs on Flask. If someone thinks it's not lightweight, he can implement a server with the same features in any other language he wants that has a SocketIO library for web servers, like C or Golang. SocketIO is a very common library used for this type of communication. I don't see any reason to use MQTT over SocketIO. In fact, from what I see MQTT seems like a layer on top of websocket, while SocketIO just provides a compatible interface between servers/clients (correctly choosing websocket or long polling based on client/server). 

It's not complex for multiple instance of the bot. It routes different account stuff to websocket events with the account email in it. It's easy to support and very leightweight on the server too, because it's not broadcasting events from all accounts to all clients, which is totally uncessary and very costy.

solderzzc:
@douglascamata Route is exactly what MQTT broker does, so that's easier?

Gobberwart:
Is this still required? I removed the "Approved" label, since it's not actually approved. Wondering if it should just be closed as abandoned?

PR #5135, opened 2016-09-03
Fixes in PolylineWalker

th3w4y:
Scenario:
- when there is no cached Polyline we get a None back from the .get_alt()
- when altitude is 0  we get 0 from .get_alt

Issue:
- In polylineWalker I was checking if .get_alt() or uniform(min_alt, max_lat) that clearly fails when altitude is zero.

Fixes:
- will fallback to read the previous position of the bot in the alt 0 or None case
- set the speed in PolylineWalker and passthrough to StepWalker

Additional very Important Fixes:
- Polyline Walker should return False if final destination not reached (might have teleported in some follow workers)
- fixed cache validation.. (this was causing API quota reached)

### DIFF
--- a/pokemongo_bot/walkers/polyline_walker.py
+++ b/pokemongo_bot/walkers/polyline_walker.py
@@ -1,5 +1,6 @@
 from random import uniform
 
+from pokemongo_bot.cell_workers.utils import distance
 from pokemongo_bot.walkers.step_walker import StepWalker
 from polyline_generator import PolylineObjectHandler
 
@@ -19,3 +20,10 @@ class PolylineWalker(StepWalker):
         self.pol_alt = self.polyline.get_alt() or self.actual_alt
         super(PolylineWalker, self).__init__(self.bot, self.pol_lat, self.pol_lon,
                                              self.pol_alt, fixed_speed=self.speed)
+
+    def step(self):
+        step = super(PolylineWalker, self).step()
+        if not (distance(self.pol_lat, self.pol_lon, self.dest_lat, self.dest_lng) > 10 and step):
+            return False
+        else:
+            return True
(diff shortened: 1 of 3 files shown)


mjmadsen:
@th3w4y Thank you.

pokemongo_bot/walkers/polyline_walker.py
```diff
@@ -8,13 +9,20 @@ class PolylineWalker(StepWalker):
...
+        self.pol_alt = self.polyline.get_alt() or self.actual_alt
         super(PolylineWalker, self).__init__(self.bot, self.pol_lat, self.pol_lon,
-                                             self.pol_alt, fixed_speed=True)
+                                             self.pol_alt, fixed_speed=self.speed)
+
+    def step(self):
+        step = super(PolylineWalker, self).step()
+        if not (distance(self.pol_lat, self.pol_lon, self.dest_lat, self.dest_lng) > 10 and step):
+            return False
+
```
Anakin5: There is branch without any return statement

th3w4y: Which one?

mjmadsen:
@th3w4y can you rebase? @solderzzc put some changes in to polyline_walker

th3w4y:
@mjmadsen i rebased...

mjmadsen:
:+1:

Anakin5:
With these changes, now the problem is that movements are chaotics just like the screenshot of @solderzzc .
So there is still something to fix :P.

th3w4y:
@Anakin5 check last commit I believe this was causing the whole thing... https://github.com/PokemonGoF/PokemonGo-Bot/pull/5135/commits/8f7f22c90e92ccafafa4956ba006624c3f90ab5e

Anakin5:
Yes looks better :)

mjmadsen:
:+1: 

We agree this is in a working state now? :)

Anakin5:
Yes, it is working fine

PR #5140, opened 2016-09-03
Reduce lat_long_delta variance range

mjmadsen:
### Waiting on @Anakin5 s data before merging.

Reduced lat_long_delta per conversation in #5038

### DIFF
--- a/pokemongo_bot/human_behaviour.py
+++ b/pokemongo_bot/human_behaviour.py
@@ -23,7 +23,9 @@ def action_delay(low, high):
 def random_lat_long_delta():
     # Return random value from [-.000025, .000025]. Since 364,000 feet is equivalent to one degree of latitude, this
     # should be 364,000 * .000025 = 9.1. So it returns between [-9.1, 9.1]
-    return ((random() * 0.00001) - 0.000005) * 5
+    # return ((random() * 0.00001) - 0.000005) * 5
+    # reduced per conversation in #5038
+    return ((random() * 0.00001) - 0.000005)
 
 def random_alt_delta():
     # Return random value from [-0.2, 0.2]. Altitude is measured in meters. A delta of 0.4 could for example 


mjmadsen:
@Anakin5 Think this will better match your GPS tests?

th3w4y:
@mjmadsen i did the math... by reducing it 5 fold then the random_lat_long adds or removes anything between [-15cm , 15cm] is that enough?

mjmadsen:
Whoa. Yeah, I think 30cm variance isn't quite enough, haha - didn't realize the comments were wrong. Should have done the math myself!

th3w4y:
I think the problem was coming from something else not even sure.. what was the problem?

mjmadsen:
He said the gps variance we were adding in was way too much and from his data thought it was about 5 times too big. My immediate thought was that this function was off. Maybe there is something else..

th3w4y:
@mjmadsen @Anakin5 I see.. i believe is hard to say what is a correct range...
maybe the real gps probably has [-20, +20] cm variance

but this will f*uck up completely the Walkers that don't use the Polyline...

mjmadsen:
Check out my comment in the other section.

Anakin5:
The gps accuracy might be 10 meters, that is not what we are representing here. We are dealing with the jitter, which is not 10 meters when coming from the same source.

If we want to represent a sudden loss of accuracy (switch from gps to network positioning for example), then we have to make the gps jump a few meters, but for not more than a few points, then come back to normal position. I don't think we need to do that at all !

If we want a bit a jitter, that is going to be a few centimeters Turn on your phone and open google map, big blue circle is accuracy, but your position is stable at few centimeters.

mjmadsen:
This bit of code is to simulate a person wandering a bit while walking. The noise settings are for gps inaccuracy.

Anakin5:
So we need a better pattern.
Let's say we have have a good drunken man pattern, who is constantly stepping in every direction.
Now we need the shopping lady pattern, who might be stopping, going a bit on the left or the right, but who is always getting closer to her destination.

mjmadsen:
@Anakin5 I hear you. I've been walking around my office to figure out better distances. If @th3w4y did the math right, the current variance is around a -75cm to 75cm. That isn't horrible (the -9.1 to 9.1 was worse than a drunken freshman stumbling home). But I'll agree we need to make this a bit smarter.

th3w4y:
@Anakin5 we could also +/- altitude to have a happy little girl walking and jumping around :)

mjmadsen:
lol.

th3w4y:
@mjmadsen @Anakin5 oops we actually do that right? [+/- altitude] :)

th3w4y:
Thus we currently have a happy little drunken girl :)

mjmadsen:
The range of the alt humanizer is 40cm. She just has spastic arms.

th3w4y:
now back to serious note...

not sure  how rhe apu uses the accuracy internally (is it fixed value, random value in a range) but that value is sent in an network envelope...

now lets think a bit... 

the whole random_lat_long we add to be more humman and prevent bit detection right?

but if the position is tempered with and the accuracy is tempered with in same time we are actually sending crap anyway? we sendind something that can infer we were botting... because of non-correlation between position (inferred speed) and accuracy

mjmadsen:
I think we need to account for our intended direction. We don't really want to suddenly leap towards or away from our target. Instead, only add in side to side wandering, but not much (to keep it simple). In that case, we'd need to look at the angle we're walking at and adjust the random values assigned to each long/lat accordingly. 

Side note, I think the default noise values are probably too high. 
    "gps_xy_noise_range": 0.000125,
    "gps_z_noise_range": 12.5,

mjmadsen:
If either of you have any ideas on how to fetch our bearing, then use that value to scale how much both long/lat should vary along that direction. I'm more than open to suggestions - the math looks a bit confusing.

PR #5141, opened 2016-09-03
decouple datastore from the bot

rawgni:
## Short Description:

The Datastore is being initialized multiple times, resulting in multiple duplicate migrations.

- Initialize the Datastore once, migrate once, and done with it.
- Pass the datastore connection to the bot.
- Easier to mock datastore since it's no longer initialized inside PokemonGoBot()

### DIFF
--- a/pokecli.py
+++ b/pokecli.py
@@ -94,7 +94,14 @@ def main():
         bot.workers = tree
 
     def initialize(config):
-        bot = PokemonGoBot(config)
+        from pokemongo_bot.datastore import Datastore
+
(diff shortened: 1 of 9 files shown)


BreezeRo:
Excellent work right here brother, please update this branch with the latest changes from dev so we can merge

rawgni:
@BreezeRo should be updated now

solderzzc:
Since the DB expert approval, we can have this merged.

MZorzy:
alt_min default 500
alt_max default 1000
alt default 8 ?

please set null and force chek exist on a config file

PR #5147, opened 2016-09-03
Improve humanize walking variance

mjmadsen:
- Account for bearing when adding in walking "sway"
- We now only sway "left" and "right" as we head towards our destination

### DIFF
--- a/pokemongo_bot/human_behaviour.py
+++ b/pokemongo_bot/human_behaviour.py
@@ -24,9 +24,6 @@ def random_lat_long_delta():
     # Return random value from [-.000025, .000025]. Since 364,000 feet is equivalent to one degree of latitude, this
     # should be 364,000 * .000025 = 9.1. So it returns between [-9.1, 9.1]
     return ((random() * 0.00001) - 0.000005) * 5
-    
-def random_lat_long_delta2(bearing):
-    return ((random() * 0.00001) - 0.000005) * 5, ((random() * 0.00001) - 0.000005) * 5
 
 def random_alt_delta():
     # Return random value from [-0.2, 0.2]. Altitude is measured in meters. A delta of 0.4 could for example 
--- a/pokemongo_bot/human_behaviour.py
+++ b/pokemongo_bot/human_behaviour.py
@@ -20,10 +20,13 @@ def action_delay(low, high):
     time.sleep(shortNum)
 
 
-def random_lat_long_delta(bearing):
+def random_lat_long_delta():
     # Return random value from [-.000025, .000025]. Since 364,000 feet is equivalent to one degree of latitude, this
     # should be 364,000 * .000025 = 9.1. So it returns between [-9.1, 9.1]
     return ((random() * 0.00001) - 0.000005) * 5
+    
+def random_lat_long_delta2(bearing):
+    return ((random() * 0.00001) - 0.000005) * 5, ((random() * 0.00001) - 0.000005) * 5
 
 def random_alt_delta():
     # Return random value from [-0.2, 0.2]. Altitude is measured in meters. A delta of 0.4 could for example 
--- a/pokemongo_bot/human_behaviour.py
+++ b/pokemongo_bot/human_behaviour.py
@@ -20,7 +20,7 @@ def action_delay(low, high):
     time.sleep(shortNum)
 
 
-def random_lat_long_delta():
+def random_lat_long_delta(bearing):
     # Return random value from [-.000025, .000025]. Since 364,000 feet is equivalent to one degree of latitude, this
     # should be 364,000 * .000025 = 9.1. So it returns between [-9.1, 9.1]
     return ((random() * 0.00001) - 0.000005) * 5
--- a/pokemongo_bot/walkers/step_walker.py
+++ b/pokemongo_bot/walkers/step_walker.py
@@ -13,6 +13,7 @@ class StepWalker(object):
     def __init__(self, bot, dest_lat, dest_lng, dest_alt=None, fixed_speed=None):
         self.bot = bot
         self.api = bot.api
+        self.offset_angle=5
 
         self.initLat, self.initLng = self.bot.position[0:2]
 
@@ -73,7 +74,7 @@ class StepWalker(object):
             self.bot.heartbeat()
             return True
 
-        self._new_position = self._get_next_pos(self.initLat, self.initLng, self.bearing, self.speed, 1)
+        self._new_position = self._get_next_pos(self.initLat, self.initLng, self.bearing, self.speed, self.offset_angle)
         cAlt = self.initAlt + random_alt_delta()
 
         self.api.set_position(self._new_position[0], self._new_position[1], cAlt)
@@ -135,7 +136,7 @@ class StepWalker(object):
     
         return compass_bearing
         
-    def _get_next_pos(lat, lon, bearing, speed, offset_angle):
+    def _get_next_pos(self, lat, lon, bearing, speed, offset_angle):
         origin = Point(lat, lon)
         lat, lon, _ = VincentyDistance(kilometers=speed*1e-3).destination(origin, bearing+random.randrange(-offset_angle, offset_angle))
         return lat, lon
--- a/pokemongo_bot/walkers/step_walker.py
+++ b/pokemongo_bot/walkers/step_walker.py
@@ -2,7 +2,7 @@
 
 import math
 
-from random import uniform, randrange
+from random import uniform
 from pokemongo_bot.cell_workers.utils import distance
 from pokemongo_bot.human_behaviour import sleep, random_alt_delta
 from geopy.distance import VincentyDistance
@@ -34,7 +34,7 @@ class StepWalker(object):
         else:
             self.speed = uniform(self.bot.config.walk_min, self.bot.config.walk_max)
             
-        self.offset_angle=int((1/self.speed)*(precision/1.74))
+        self.offset_angle=(1/self.speed)*(precision/1.74)
 
         if len(self.bot.position) == 3:
             self.initAlt = self.bot.position[2]
@@ -139,5 +139,5 @@ class StepWalker(object):
         
     def _get_next_pos(self, lat, lon, bearing, speed, offset_angle):
         origin = Point(lat, lon)
-        lat, lon, _ = VincentyDistance(kilometers=speed*1e-3).destination(origin, bearing+randrange(-offset_angle, offset_angle))
+        lat, lon, _ = VincentyDistance(kilometers=speed*1e-3).destination(origin, bearing+uniform(-offset_angle, offset_angle))
         return lat, lon
--- a/pokemongo_bot/walkers/step_walker.py
+++ b/pokemongo_bot/walkers/step_walker.py
@@ -80,7 +80,7 @@ class StepWalker(object):
         self._new_position = get_next_pos(self.initLat, self.initLng, self.bearing, self.speed, 1)
         cAlt = self.initAlt + random_alt_delta()
 
-        self.api.set_position(cLat, cLng, cAlt)
+        self.api.set_position(new_position[0], new_position[1], cAlt)
         self.bot.event_manager.emit(
             'position_update',
             sender=self,
--- a/pokemongo_bot/walkers/step_walker.py
+++ b/pokemongo_bot/walkers/step_walker.py
@@ -138,6 +138,11 @@ class StepWalker(object):
         
     def _get_next_pos(self, lat, lon, bearing, speed, precision):
         origin = Point(lat, lon)
-        offset_angle = (1/self.speed)*(precision/1.74)
-        lat, lon, _ = VincentyDistance(kilometers=speed*1e-3).destination(origin, bearing+uniform(-offset_angle, offset_angle))
-        return lat, lon
+        if speed == 0.0:
+            return lat, lon
+        if speed == float("inf"):
+            return self.destLat, self.destLng
+        else:
+            offset_angle = (1/self.speed)*(precision/1.74)
+            lat, lon, _ = VincentyDistance(kilometers=speed*1e-3).destination(origin, bearing + uniform(-offset_angle, offset_angle))
+            return lat, lon
--- a/pokemongo_bot/walkers/step_walker.py
+++ b/pokemongo_bot/walkers/step_walker.py
@@ -10,10 +10,10 @@ from geopy import Point
 
 class StepWalker(object):
 
-    def __init__(self, bot, dest_lat, dest_lng, dest_alt=None, fixed_speed=None):
+    def __init__(self, bot, dest_lat, dest_lng, dest_alt=None, fixed_speed=None, offset_angle=5):
         self.bot = bot
         self.api = bot.api
-        self.offset_angle=5
+        self.offset_angle=offset_angle
 
         self.initLat, self.initLng = self.bot.position[0:2]
 
@@ -138,5 +138,5 @@ class StepWalker(object):
         
     def _get_next_pos(self, lat, lon, bearing, speed, offset_angle):
         origin = Point(lat, lon)
-        lat, lon, _ = VincentyDistance(kilometers=speed*1e-3).destination(origin, bearing+random.randrange(-offset_angle, offset_angle))
+        lat, lon, _ = VincentyDistance(kilometers=speed*1e-3).destination(origin, bearing+randrange(-offset_angle, offset_angle))
         return lat, lon
--- a/pokemongo_bot/walkers/step_walker.py
+++ b/pokemongo_bot/walkers/step_walker.py
@@ -7,7 +7,7 @@ from pokemongo_bot.human_behaviour import random_lat_long_delta, sleep, random_a
 
 class StepWalker(object):
 
-    def __init__(self, bot, dest_lat, dest_lng, dest_alt=None, fixed_speed=False):
+    def __init__(self, bot, dest_lat, dest_lng, dest_alt=None, fixed_speed=None):
         self.bot = bot
         self.api = bot.api
 
@@ -25,9 +25,9 @@ class StepWalker(object):
         else:
             self.alt = dest_alt
             
-        if fixed_speed:
+        if fixed_speed != None:
             # PolylineWalker uses a fixed speed!
-            self.speed = self.bot.config.walk_min
+            self.speed = fixed_speed
         else:
             self.speed = uniform(self.bot.config.walk_min, self.bot.config.walk_max)
 
--- a/pokemongo_bot/walkers/step_walker.py
+++ b/pokemongo_bot/walkers/step_walker.py
@@ -1,3 +1,5 @@
+# -*- coding: utf-8 -*-
+
 import math
 
 from random import uniform
@@ -114,6 +116,10 @@ class StepWalker(object):
     def _calc_bearing(start_lat, start_lng, dest_lat, dest_lng):
         """
         Calculates the bearing between two points.
+    
+        The formulae used is the following:
+            θ = atan2(sin(Δlong).cos(lat2),
+                      cos(lat1).sin(lat2) − sin(lat1).cos(lat2).cos(Δlong))
     
         :Parameters:
           - `start_lat in decimal degrees
--- a/pokemongo_bot/walkers/step_walker.py
+++ b/pokemongo_bot/walkers/step_walker.py
@@ -55,7 +55,7 @@ class StepWalker(object):
             self.dLat = (dest_lat - self.initLat) / int(self.steps)
             self.dLng = (dest_lng - self.initLng) / int(self.steps)
             
-        self.bearing = self._calc_bearing(self.initLat, self.initLng, self.dLat, self.dLng)
+        self.bearing = self._calc_bearing(self.initLat, self.initLng, self.destLat, self.destLng)
 
     def step(self):
         if (self.dLat == 0 and self.dLng == 0) or self.dist < self.speed:
--- a/pokemongo_bot/walkers/step_walker.py
+++ b/pokemongo_bot/walkers/step_walker.py
@@ -10,7 +10,7 @@ from geopy import Point
 
 class StepWalker(object):
 
-    def __init__(self, bot, dest_lat, dest_lng, dest_alt=None, fixed_speed=None, precision=15):
+    def __init__(self, bot, dest_lat, dest_lng, dest_alt=None, fixed_speed=None, precision=50):
         self.bot = bot
         self.api = bot.api
         self.precision = precision
--- a/pokemongo_bot/walkers/step_walker.py
+++ b/pokemongo_bot/walkers/step_walker.py
@@ -10,10 +10,9 @@ from geopy import Point
 
 class StepWalker(object):
 
-    def __init__(self, bot, dest_lat, dest_lng, dest_alt=None, fixed_speed=None, offset_angle=5):
+    def __init__(self, bot, dest_lat, dest_lng, dest_alt=None, fixed_speed=None, precision=50):
         self.bot = bot
         self.api = bot.api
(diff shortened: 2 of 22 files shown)


pokemongo_bot/human_behaviour.py
```diff
@@ -24,6 +24,9 @@ def random_lat_long_delta():
     # Return random value from [-.000025, .000025]. Since 364,000 feet is equivalent to one degree of latitude, this
     # should be 364,000 * .000025 = 9.1. So it returns between [-9.1, 9.1]
     return ((random() * 0.00001) - 0.000005) * 5
+    
+def random_lat_long_delta2(bearing):
+    return ((random() * 0.00001) - 0.000005) * 5, ((random() * 0.00001) - 0.000005) * 5
```
th3w4y: How do you plan to use bearing in the calculation...?

pokemongo_bot/walkers/step_walker.py
```diff
@@ -106,3 +110,43 @@ def step(self):
 
     def _pythagorean(self, lat, lng):
         return sqrt((lat ** 2) + (lng ** 2))
```
th3w4y: If you import math 

then this should become

`math.sqrt`

pokemongo_bot/human_behaviour.py
mjmadsen: If we're heading at 0/360, we only add variance to latitude. Do some more math between there to scale each respective to that. I'm not sure

mjmadsen: Wait I got that backwards. If we're headed north (0/360), we'd change our longitude.

Anakin5: The Delta applies to the destination. So as long as the Delta is smaller than the (destination - origin), you will never walk backward.
And as it is now, it is small enough for that. So not sure you need to change anything.

pokemongo_bot/walkers/step_walker.py
```diff
@@ -109,7 +109,7 @@ def step(self):
         #     alt, False)
 
     def _pythagorean(self, lat, lng):
-        return sqrt((lat ** 2) + (lng ** 2))
+        return math.sqrt((lat ** 2) + (lng ** 2))
 
     def calc_bearing(start_lat, start_lng, dest_lat, dest_lng):
```
sohje: Should it be static method? calc_bearing(self...) ? Or you forget @static decorator?

th3w4y:
@mjmadsen  I am afraid that the travis build is failing due to non-uff carracters 

`  θ = atan2(sin(Δlong).cos(lat2),`

pokemongo_bot/walkers/step_walker.py
```diff
@@ -54,10 +54,14 @@ def __init__(self, bot, dest_lat, dest_lng, dest_alt=None, fixed_speed=None):
             self.dLng = (dest_lng - self.initLng) / int(self.steps)
             self.magnitude = self._pythagorean(self.dLat, self.dLng)
             self.unitAlt = (self.alt - self.initAlt) / int(self.steps)
+            
+        self.bearing = calc_bearing(self.initLat, self.initLng, self.dLat, self.dLng)
 
     def step(self):
+        walk_sway = random_lat_long_delta2(self.bearing)
```
sohje: random_lat_long_delta2 import missed?

pokemongo_bot/walkers/step_walker.py
```diff
@@ -2,7 +2,7 @@
```
sohje: Add encoding declaration
# coding=utf-8

th3w4y:
I don't think that adding a value to only latitude or only longitude based on bearing adds much extra...

mjmadsen:
@th3w4y this is for our discussion in #5140

We're adding in a bit of variance to simulate human walking. We want it to only move us "left" or "right".

pokemongo_bot/walkers/step_walker.py
```diff
@@ -105,4 +111,44 @@ def step(self):
         #     alt, False)
 
     def _pythagorean(self, lat, lng):
-        return sqrt((lat ** 2) + (lng ** 2))
+        return math.sqrt((lat ** 2) + (lng ** 2))
+
+    def _calc_bearing(start_lat, start_lng, dest_lat, dest_lng):
```
sohje: make def _calc_bearing(self, start_lat, start_lng, dest_lat, dest_lng)

sohje:
Works for me.

th3w4y:
@sohje @mjmadsen I agreed with the idea... but the calculation doesn't seem to do that yet...

pokemongo_bot/walkers/step_walker.py
```diff
@@ -4,7 +4,7 @@
 
 from random import uniform
 from pokemongo_bot.cell_workers.utils import distance
-from pokemongo_bot.human_behaviour import random_lat_long_delta2, sleep, random_alt_delta
+from pokemongo_bot.human_behaviour import random_lat_long_delta, random_lat_long_delta2, sleep, random_alt_delta
```
sohje: Any usage of random_lat_long_delta?

mjmadsen:
@th3w4y Yeah, I still need to figure out the important part :)

pokemongo_bot/walkers/step_walker.py
mjmadsen: The test files use it still. I'll update those too if I can get this working :)

sohje:
@th3w4y @mjmadsen Just to make sure bot not crashing ;]

pokemongo_bot/walkers/step_walker.py
mjmadsen: Actually the CI still fails because of how the test files are setup.

BreezeRo:
looks good :dancer: might have to edit pokemongo_bot/test/step_walker_test.py to pass CI

th3w4y:
My proposal:

```
IIn [209]: from geopy.distance import VincentyDistance

In [210]: 

In [210]: from geopy import Point

In [211]: def get_next_pos(lat, lon, bearing, speed, offset_angle):
     ...:     origin = Point(lat, lon)
     ...:     lat, lon, _ = VincentyDistance(kilometers=.003).destination(origin, bearing+random.randint(-offset_angle, offset_angle))
     ...:     return lat, lon
     ...:     
     ...: 

In [212]: ex_orig
Out[212]: (47.1706378, 8.5167405)

In [213]: get_next_pos(ex_orig[0], ex_orig[1], 180, 3, 5)
Out[213]: (47.170610819427246, 8.516741190604915)

In [214]: get_next_pos(ex_orig[0], ex_orig[1], 180, 3, 5)
Out[214]: (47.17061085229894, 8.516738429026649)

In [215]: get_next_pos(ex_orig[0], ex_orig[1], 180, 3, 5)
Out[215]: (47.1706109180023, 8.516743948818615)

In [216]: 
```

th3w4y:
This above obviously can replace the whole pythagorean.. magnitude.. the big convoluted part in StepWalker

mjmadsen:
@th3w4y oh that's a lot sexier than what I was getting into. :+1: 

Do you want to put a PR with that and I'll close this? Seems like you have a firmer grasp on this :smile:

th3w4y:
@mjmadsen @BreezeRo @sohje what would be an good max offset_angle value?

mjmadsen:
VincentyDistance(kilometers=speed*1e-3).

Our speed currently doesn't have anything to do with a real life unit. Will that cause problems with this? Speaking of which, it would be nice if it did - we could even adjust it based on our long/lat to we're always in a physical unit.

th3w4y:
@mjmadsen Yes this assumes you only walk for one second.
>VincentyDistance(kilometers=speed*1e-3).
>
>Our speed currently doesn't have anything to do with a real life unit. Will that cause problems with this? Speaking of which, it would be nice if it did - we could even adjust it based on our long/lat to we're always in a physical unit.

mjmadsen:
Applied some changes. Will need to look over it a bit more. I removed the randomness from the portion at the start of step (when we're very close to dest) for the time being.

pokemongo_bot/walkers/step_walker.py
```diff
@@ -72,26 +77,16 @@ def step(self):
...
+        cAlt = self.initAlt + random_alt_delta()
 
         self.api.set_position(cLat, cLng, cAlt)
         self.bot.event_manager.emit(
             'position_update',
             sender=self,
             level='debug',
             data={
-                'current_position': (cLat, cLng, cAlt),
+                'current_position': (new_position[0], new_position[1], cAlt),
```
th3w4y: self.newposition !

pokemongo_bot/walkers/step_walker.py
```diff
@@ -72,26 +77,16 @@ def step(self):
...
 
-        scaledDLat = unitLat * self.magnitude
-        scaledDLng = unitLng * self.magnitude
-
-        cLat = self.initLat + scaledDLat + random_lat_long_delta()
-        cLng = self.initLng + scaledDLng + random_lat_long_delta()
-        cAlt = self.initAlt + self.unitAlt + random_alt_delta()
-
-        self.api.set_position(cLat, cLng, cAlt)
+        self.api.set_position(new_position[0], new_position[1], cAlt)
```
th3w4y: self._new_position

mjmadsen:
Oh I realized I have a decent more to cleanup in init. Bah.

pokemongo_bot/walkers/step_walker.py
```diff
@@ -104,5 +99,47 @@ def step(self):
...
+    
+        # Now we have the initial bearing but math.atan2 return values
+        # from -180° to + 180° which is not what we want for a compass bearing
+        # The solution is to normalize the initial bearing as shown below
+        initial_bearing = math.degrees(initial_bearing)
+        compass_bearing = (initial_bearing + 360) % 360
+    
+        return compass_bearing
+        
+    def _get_next_pos(lat, lon, bearing, speed, offset_angle):
```
th3w4y: self or staticmethod

pokemongo_bot/walkers/step_walker.py
```diff
@@ -54,10 +57,12 @@ def __init__(self, bot, dest_lat, dest_lng, dest_alt=None, fixed_speed=None):
             self.dLng = (dest_lng - self.initLng) / int(self.steps)
             self.magnitude = self._pythagorean(self.dLat, self.dLng)
```
th3w4y: method was removed...

th3w4y:
@mjmadsen Hint: 

test cases will have to patch random.randrange to return 0.. i think..

pokemongo_bot/walkers/step_walker.py
```diff
@@ -104,5 +95,47 @@ def step(self):
...
+        # from -180° to + 180° which is not what we want for a compass bearing
+        # The solution is to normalize the initial bearing as shown below
+        initial_bearing = math.degrees(initial_bearing)
+        compass_bearing = (initial_bearing + 360) % 360
+    
+        return compass_bearing
+        
+    def _get_next_pos(lat, lon, bearing, speed, offset_angle):
+        origin = Point(lat, lon)
+        lat, lon, _ = VincentyDistance(kilometers=speed*1e-3).destination(origin, bearing+random.randrange(-offset_angle, offset_angle))
```
sohje: add
```python
from random import randrange
```

pokemongo_bot/walkers/step_walker.py
```diff
@@ -72,26 +73,16 @@ def step(self):
             self.bot.heartbeat()
             return True
 
-        totalDLat = (self.destLat - self.initLat)
-        totalDLng = (self.destLng - self.initLng)
-        magnitude = self._pythagorean(totalDLat, totalDLng)
-        unitLat = totalDLat / magnitude
-        unitLng = totalDLng / magnitude
+        self._new_position = self._get_next_pos(self.initLat, self.initLng, self.bearing, self.speed, 1)
```
th3w4y: maybe don't hardcode this to 1... add some offset_angle=5  in __init__

th3w4y: i i think about in for small distances like our speed range... even 10 should do...
we can even make it configurable...

pokemongo_bot/walkers/step_walker.py
sohje: Also
```python
randrange(-offset_angle, offset_angle)
#random.randrange(...)
```

pokemongo_bot/walkers/step_walker.py
```diff
@@ -1,15 +1,19 @@
...
 
+from random import uniform, randrange
+from pokemongo_bot.cell_workers.utils import distance
+from pokemongo_bot.human_behaviour import sleep, random_alt_delta
+from geopy.distance import VincentyDistance
+from geopy import Point
 
 class StepWalker(object):
 
     def __init__(self, bot, dest_lat, dest_lng, dest_alt=None, fixed_speed=None):
```
th3w4y: What i meant was:

```
def __init__(self, bot, dest_lat, dest_lng, dest_alt=None, fixed_speed=None, offset_angle=5):
    self.offset_angle = offset_angle
```

This way different follow can tweak it :)

pokemongo_bot/walkers/step_walker.py
```diff
@@ -104,5 +96,47 @@ def step(self):
...
+        # from -180° to + 180° which is not what we want for a compass bearing
+        # The solution is to normalize the initial bearing as shown below
+        initial_bearing = math.degrees(initial_bearing)
+        compass_bearing = (initial_bearing + 360) % 360
+    
+        return compass_bearing
+        
+    def _get_next_pos(self, lat, lon, bearing, speed, offset_angle):
+        origin = Point(lat, lon)
+        lat, lon, _ = VincentyDistance(kilometers=speed*1e-3).destination(origin, bearing+random.randrange(-offset_angle, offset_angle))
```
th3w4y: This is now bearing+randrange(-offset_angle, offset_angle)

instead of bearing+random.randrange(-offset_angle, offset_angle)

due to from random import randrange..

mjmadsen:
Lol @th3w4y again, I am sorry.

th3w4y:
offset_angle = 5 gives and distance of [-26cm, +26cm] from original distance
offset_angle = 10 gives and distance of [-52cm, +52cm] from original distance
offset_angle = 15 gives and distance of [-78cm, +78cm] from original distance

all of them in the direction of bearing...

above values calculated with speed 3m/s

th3w4y:
@mjmadsen what speeds do we usually advise the user to set?

mjmadsen:
2.16 to 4.16

th3w4y:
Do we want to have constant ranges here?
with offset = 5

with walk_max = 4.16 it can jump  36cm
with walk_min = 2.16 it can jump 18cm

mjmadsen:
I suppose I'd prefer it on the config. That way @Anakin5 can remove all variance :smile:

th3w4y:
One way of making the range not speed dependant... is to say that:

offset_angle = 1/speed*20 # for 36cm range offset
offset_angle = 1/speed*30 # for 52cm range offset
offset_angle = 1/speed*40 # for 70cm range offset

th3w4y:
This actually means...

that offset angle formula can be:


`offset_angle = 1/speed*precission_in_cm/2`

and we let the user specify precission_in_cm

mjmadsen:
Interesting :+1:

th3w4y:
magic formula is 1/speed*precission_in_cm/1.7

mjmadsen:
What values should we use for precision? Should that be configurable?

Anakin5:
Please no more configuration parameter to make the bot works. Or at least with good default value so that we so not have to touch them if we want the bot to act normally.

th3w4y:
@Anakin5  it will have a default value... in init

th3w4y:
math.sqrt((math.pi*math.e)/2)*math.erf(1) ~= 1.74

pokemongo_bot/walkers/step_walker.py
```diff
@@ -104,5 +97,47 @@ def step(self):
...
+        # from -180° to + 180° which is not what we want for a compass bearing
+        # The solution is to normalize the initial bearing as shown below
+        initial_bearing = math.degrees(initial_bearing)
+        compass_bearing = (initial_bearing + 360) % 360
+    
+        return compass_bearing
+        
+    def _get_next_pos(self, lat, lon, bearing, speed, offset_angle):
+        origin = Point(lat, lon)
+        lat, lon, _ = VincentyDistance(kilometers=speed*1e-3).destination(origin, bearing+randrange(-offset_angle, offset_angle))
```
th3w4y: lets use precision then...

mjmadsen:
Not sure we've had a PR get 10+ comments an hour for 5 hours and 31 :cry: commits over the same period.

mjmadsen:
@sohje @solderzzc @Anakin5 We were discussing the different walkers and think that spiral could probably be removed. It only really seems to be used if we're banned and can't see anything to move towards. Your thoughts?

Also, I think we should possibly adjust how our walker **task** to account for all forms of guidance (we don't need three follow_* tasks). With this, the move_to_map_pokemon bit should be decoupled from walking and instead, if we want to use this portion, it is in catch_pokemon, which is our "find available pokemon" list builder. In addition, we have small edge errors if close to a pokemon and close to a fort where we keep moving back and forth. To solve this might be to continue this work and set a "next target" - also related to comments about keeping an instance of our walker open over iterations so we don't keep reinitializing. Ramblings... I know :smile:

sohje:
Spiral Task causes a lot of issues on GH. Dunno if any1 still using it.
What about switching between walkers dynamically? Can this be usefull?

th3w4y:
@sohje good idea...

mjmadsen:
I think we could do a lot of refactoring still.

Is there any reason NOT to use polyline_walker? If we can make it work better with cluster and path, it's superior. I just don't see any advantage of running things to stepwalker directly.

As far as cluster, path, move_to_map_pokemon and move_to_fort - we need a generalized task to find our next target. This would kill off about 4 example configs and reduce code in places that walking shouldn't be involved with. If we could combine the logic of these to switch on the fly for max efficiency - then we're freaking talking.

mjmadsen:
I think we could do a lot of refactoring still.

Is there any reason NOT to use polyline_walker? If we can make it work better with cluster and path, it's superior. I just don't see any advantage of running things to stepwalker directly. Maybe they should be closer friends. :smile:

As far as cluster, path, move_to_map_pokemon and move_to_fort - we need a generalized task to find our next target. This would kill off about 4 example configs and reduce code in places that walking shouldn't be involved with. If we could combine the logic of these to switch on the fly for max efficiency - then we're freaking talking.

Any thoughts on how we can clean up the position_update goodness? Maybe check if we're aiming at something in particular and print that? I'd say round but then we lose all the important numbers. Maybe print long/lat difference in meters (moving X meters east). 

CI passed! Please review :smile:

sohje:
Code looks good, didnt test it yet.

mjmadsen:
Since this is mostly a nuanced bit, I want to wait a bit. Everything, more or less, works without this being merged. I don't want to be tagged on slack again for breaking stuff :smile:

th3w4y:
you gus can blame me also not only @mjmadsen :)

sohje:
Hmm... guys, any1 else experiencing something like this.
Imgur

th3w4y:
@sohje this checkout right? what are the coordinates?

mjmadsen:
{
       "type": "MoveToFort",
       "config": {
         "enabled": true,
         "lure_attraction": true,
         "lure_max_distance": 2000,
         "walker": "PolylineWalker",
         "log_interval": 5
       }

https://s17.postimg.io/589c0b933/147291834694374.png
https://s18.postimg.io/u2b8fomjt/test1.jpg
https://s14.postimg.io/76ttrkq29/test2.jpg
https://s16.postimg.io/oj5dvraid/test3.jpg

Latter half is not catching pokemon and removing type: "FollowSpiral"

sohje:
@th3w4y yeap, checkout from @mjmadsen branch. Coords 55.639345, 37.360947

th3w4y:
@sohje one explanation would be:


pokestops are to close... google maps does not have walking directions in between them


but your picture shows that you were automatically fallback to classic behaviour...

mjmadsen:
I'm running through all the walkers. Might not finish before I go to bed. 

@Anakin5 does this seem to give your poly the results you're looking for? If we seal this portion down, we don't have to mess with it for a long ass time!

sohje:
@th3w4y according to google maps there are 1-3 routes between pokestops on that coords.

th3w4y:
@mjmadsen  one more commit:

https://github.com/mjmadsen/PokemonGo-Bot/pull/2
To make the function complete...(i don't want to see divisions by zero of walking of planet)

th3w4y:
@mjmadsen i know that we'll not see 0.0 or infinite but it make the function complete and protected from ZeroDivizion

th3w4y:
@mjmadsen I found a bug...

https://github.com/mjmadsen/PokemonGo-Bot/blob/13819891d3d0df4ec91fbe4bfeefcb7d64a1fbab/pokemongo_bot/walkers/step_walker.py#L58

th3w4y:
This is incorrect

`self.bearing = self._calc_bearing(self.initLat, self.initLng, self.dLat, self.dLng)`

sohje:
still moving straight ;]
```shell
[2016-09-03 20:31:12] [MoveToFort] Moving towards pokestop XX - 0.13km
[2016-09-03 20:31:17] [MoveToFort] Moving towards pokestop  XX - 0.11km
[2016-09-03 20:31:23] [MoveToFort] Moving towards pokestop XX - 0.10km
[2016-09-03 20:31:29] [MoveToFort] Moving towards pokestop  XX - 0.10km
[2016-09-03 20:31:35] [MoveToFort] Moving towards pokestop  XX - 0.11km
[2016-09-03 20:31:42] [MoveToFort] towards pokestop  XX - 0.12km
```

th3w4y:
Should be:

`self.bearing = self._calc_bearing(self.initLat, self.initLng, self.destLat, self.destLng)`

sohje:
yeap, with destLat and destLng works fine.

th3w4y:
@mjmadsen  bearing calculation should use final destination https://github.com/mjmadsen/PokemonGo-Bot/pull/3

mjmadsen:
@sohje This should be a very unnoticeable change to look like strafing while walking.

th3w4y:
@sohje so is working now?

screenshot please :)

sohje:
@th3w4y yeap, with bearing fix works fine. Wait a sec, i'll make pull - edited manually.

sohje:
hmm, i dont see on the map any strafing ;[

mjmadsen:
https://s15.postimg.org/4extwtzjf/test4.jpg

      {
        "type": "SpinFort",
        "config": {
          "enabled": true,
          "spin_wait_min": 3,
          "spin_wait_max": 5,
          "daily_spin_limit": 1900
        }
      },
      {
        "type": "MoveToFort",
        "config": {
          "enabled": true,
          "lure_attraction": true,
          "lure_max_distance": 2000,
          "walker": "PolylineWalker",
          "log_interval": 5
        }
      }
    ],
    "map_object_cache_time": 5,
    "forts": {
      "avoid_circles": false,
      "max_circle_size": 6,
      "cache_recent_forts": true
    },

sohje:
Seems we have to implement some sort of speed acceleration and deceleration.

th3w4y:
@sohje for the polyline... that would be hard...

sohje:
@th3w4y yeap...

th3w4y:
@sohje i take the challenge then !

th3w4y:
but this does not need to wait for that refactoring...

BreezeRo:
Tried, testing and working on my end - can confirm speed variance issue mentioned by @sohje 

I'm approving this for now, will check in a moment before merging

Issue #5150, opened 2016-09-03
[guide] setup telegram push notification for bot


lntn:
It took me a while to Google this out so I am going to put it here, hoping it may benefit other people.

If you have not known, currently the bot has already implemented Telegram notification. The reason why you are still not able to receive anything from the `alert_catch` option is you have not told the bot which Telegram account it should send the message to. This is where the "master" option come in.

As the instruction says, you need to enter your chat ID into that `master` option.

In order to find out your chat ID, open Telegram app, find the **@get_id_bot** bot. Once found you can ask `\my_id` and you will get it.

That's it, put the ID into `master` option and cross your finger.

P.S. the bot does inform you on newly hatched pokemons, not only the new catches.

avexus:
Thank you. Now I can talk both ways.

Gobberwart:
Note also that the bot will display a log message like:

`[2016-09-04 06:55:31] [MainThread] [PokemonGoBot] [INFO] message from USERNAME (USERID): MESSAGETEXT`

whenever a message is received. You can just grab your ID from there.

oralunal:
Please make a PR and update configuration file ;)

PR #5153, opened 2016-09-03
add getmaster command

askovpen:
#5150


Gobberwart:
Forgive me if I'm being blatantly stupid here, but if master isn't already correctly set, then surely the bot will not respond to any commands, including "/master". Isn't that sort-of the point?

Gobberwart:
If the bot WILL respond to a command without it coming from the correct master, I'd suggest that's a security issue.

askovpen:
bot answer to anyone if `"master": null`

Gobberwart:
Then I'd say that's a security issue, especially given that the default is null.

askovpen:
default telegram disabled

Gobberwart:
Still risk of user changing to enabled: "true", testing the "info" command, finding that it works and then leaving it without ever setting master. I honestly think anyone capable of being able to create a bot on telegram should be capable of finding their master id without this functionality.

Better yet, bot should be able to accept telegram username as master and not have to find the id at all.

Gobberwart:
Also note that, regardless of master setting, the bot will log a received message as follows:

message from USERNAME (USERID): MESSAGETEXT

Is it not easy enough to just get the master userid from there?

DBa2016:
I agree with @Gobberwart (there is no need to accept anything except "/info" if master is not explicitly set) , though need to add a correction: the bot DOES accept a telegram username (in form of "@username"), however in such cases, it will not be able to send any notifications until a "/info" command has been issued. I will gladly create a PR to automatically find out the numeric ID if somebody knows a way how to do this (find chat ID).

Gobberwart:
@DBa2016 That only sort-of works, provided that you've specified a username in Telegram, which isn't required or very clear how to do. 

TBH I'm a bit over the whole Telegram thing. The performance hit it causes alone is enough to make it more of a hindrance than a help.

Gobberwart:
How about instead of responding to /getmaster command, you could trap "master": null and respond with a message like below. Message could be output to both bot log and sent to telegram. 

That said, I'm still not really a fan of the bot responding to an unknown sender of any kind. It confirms the bot's existence and offers potential for abuse.

```
                        if not self.master:
                            outMessage = "Telegram bot setup not yet complete (master = null). Please enter your userid {} into bot configuration file.".format(update.message.from_user.id)
                            self.bot.logger.info(outMessage)
                            self._tbot.sendMessage(chat_id=update.message.from_user.id, parse_mode='Markdown', text=outMessage)
                            continue
```

BreezeRo:
I approve, however I didn't merge as I figured there would be concerns (as mentioned above), though not entirely relative to this PR.

I don't see this as a security issue more than what isn't already present, but I'm not sure it's necessary. There are more riskier triggers finding bot use

Gobberwart:
I don't see any need for this at all. Existing security issue should be fixed. Bot should ideally not respond to ANY unknown sender under any circumstance.

Gobberwart:
I would like to propose this instead:

1. Does not respond to an unknown sender, ever.
2. Prints detailed information to logger to assist in completing setup in case of invalid(null) master or unknown sender.

```
    def run(self):
        time.sleep(1)
        while True:
            try:
                for update in self._tbot.getUpdates(offset=self.update_id, timeout=10):
                    self.update_id = update.update_id+1
                    if update.message:
                        self.bot.logger.info("message from {} ({}): {}".format(update.message.from_user.username, update.message.from_user.id, update.message.text))
                        if not self.master:
                            outMessage = "Telegram bot setup not yet complete (master = null). Please enter your userid {} into bot configuration file.".format(update.message.from_user.id)
                            self.bot.logger.warn(outMessage)
                            # Bot could reply to Telegram, but ideally the bot should not respond to ANY unknown sender, ever.
                            # self._tbot.sendMessage(chat_id=update.message.from_user.id, parse_mode='Markdown', text=outMessage)
                            continue
                        if self.master not in [update.message.from_user.id, "@{}".format(update.message.from_user.username)]:
                            outMessage = "Telegram message received from unknown sender. If this was you, please enter your userid {} as master in bot configuration file.".format(update.message.from_user.id)
                            self.bot.logger.warn(outMessage)
                            continue
                        if self.master and not re.match(r'^[0-9]+$', str(self.master)):
                            # the "master" is not numeric, set self.master to update.message.chat_id and re-instantiate the handler
                            newconfig = self.config
                            newconfig['master'] = update.message.chat_id
                            # remove old handler
                            self.bot.event_manager._handlers = filter(lambda x: not isinstance(x, TelegramHandler), self.bot.event_manager._handlers)
                            # add new handler (passing newconfig as parameter)
                            self.bot.event_manager.add_handler(TelegramHandler(self.bot, newconfig))
                        if update.message.text == "/info":
                            self.send_player_stats_to_chat(update.message.chat_id)
                        elif update.message.text == "/start" or update.message.text == "/help":
                            res = (
                                "Commands: ",
                                "/info - info about bot"
                            )
                            self._tbot.sendMessage(chat_id=update.message.chat_id, parse_mode='Markdown', text="\n".join(res))
                        else:
                            self._tbot.sendMessage(chat_id=update.message.chat_id, parse_mode='Markdown', text="Unrecognized command: {}".format(update.message.text))
```

Gobberwart:
New PR submitted to replace this.

DBa2016:
@Gobberwart : I do not remember exactly how to specify the user name in Telegram (did it long ago) but I think you can easily do it in the Telegram app. This is not a bot issue though :)

Still, I agree with your main concern (not the "I don't like Telegram" one...if you don't like Telegram, just disable it... ;)): the bot should not respond to unknown senders.

For the "we have multiple users observing the same bot" issue - "master" can be easily converted to an array, thus allowing multiple users.

Gobberwart:
@DBa2016 I didn't say I don't like Telegram, just that I'm a bit over it. And yes, I do disable it generally, but still happy to help work on it :)

DBa2016:
@Gobberwart : I don't think that a user being "a bit over" a certain bot's feature is an issue :) - certainly it is not the bot's issue nor an issue of the contributor.

Gobberwart:
LOL retracted then, it's not really relevant :)

PR #5156, opened 2016-09-03
Telegram - Alternate method of finding master id

Gobberwart:
Will not respond to unknown sender, but will instead generate helpful log messages.

This is in response to people struggling to find master (#5150), and is an alternative solution to PR #5153 

- Eliminates current security issue of bot responding to unknown sender
- Prints detailed log information to assist users in entering user id in config file as master

### DIFF
--- a/docs/configuration_files.md
+++ b/docs/configuration_files.md
@@ -1000,8 +1000,10 @@ Bot answer on command '/info' self stats.
 ### Options
 
 * `telegram_token` : bot token (getting [there](https://core.telegram.org/bots#6-botfather) - one token per bot)
-* `master` : id (without quotes) of bot owner, who will gett announces.
+* `master` : id (without quotes) of bot owner, who will get alerts and may issue commands.
 * `alert_catch` : dict of rules pokemons catch.
+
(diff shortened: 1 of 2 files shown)


mjmadsen:
I'd say update the configuration_files.md and I prefer over #5153 

@askovpen Do you find this to be a satisfactory compromise to the initial goal?

mjmadsen:
:+1:

DBa2016:
From what I see, this PR breaks existing functionality of providing a username as master. Can we have it reverted and implemented correctly please?

askovpen:
with telegram bot api can't use correct usernames (only userids)

Gobberwart:
@DBa2016 This change has no effect on the use of "@username" as master. It does, however, strongly encourage using the id instead if no id is supplied (or an incorrect id/username). No need to revert.

DBa2016:
@Gobberwart : sorry, I misread the dff (it looked like you removed the whole non-numeric comparison section, while it is modified and not removed completely). Never read diffs on a cellphone...

Gobberwart:
@DBa2016 That's exactly the same mistake I made reading #5166

Issue #5142, opened 2016-09-03
[??BUG??] Possibly searching too often, take a rest.


davidspeijer:
### Expected Behavior

Bot walking around without any warnings

### Actual Behavior

Getting the warning: Possibly searching too often, take a rest.
Strange thing is that I thought that this warning should only happen when spinning forts and not while walking between forts. Can anybody confirm this?

### Your FULL config.json (remove your username, password, gmapkey and any other private info)

http://pastebin.com/Ekmrdh65

### Output when issue occurred

[2016-09-03 11:29:23] [  SpinFort] [spun_pokestop] Spun pokestop xxx. Experience awarded: 50. Items awarded: 4x Pokeball
[2016-09-03 11:29:26] [MoveToFort] [moving_to_fort] Moving towards pokestop yyy - 0.18km
[2016-09-03 11:29:28] [  SpinFort] [pokestop_searching_too_often] Possibly searching too often, take a rest.
[2016-09-03 11:29:30] [  SpinFort] [pokestop_searching_too_often] Possibly searching too often, take a rest.
[2016-09-03 11:29:31] [  SpinFort] [pokestop_searching_too_often] Possibly searching too often, take a rest.
[2016-09-03 11:29:32] [  SpinFort] [pokestop_searching_too_often] Possibly searching too often, take a rest.
[2016-09-03 11:29:32] [MoveToFort] [moving_to_fort] Moving towards pokestop yyy - 0.17km
[2016-09-03 11:29:34] [  SpinFort] [pokestop_searching_too_often] Possibly searching too often, take a rest.
[2016-09-03 11:29:35] [  SpinFort] [pokestop_searching_too_often] Possibly searching too often, take a rest.
[2016-09-03 11:29:36] [  SpinFort] [pokestop_searching_too_often] Possibly searching too often, take a rest.
[2016-09-03 11:29:38] [  SpinFort] [pokestop_searching_too_often] Possibly searching too often, take a rest.
[2016-09-03 11:29:43] [MoveToFort] [moving_to_fort] Moving towards pokestop yyy - 0.13km

### Steps to Reproduce

Running the bot

### Other Information
OS: CentOS
 
Branch: dev
 
Git Commit: 83c154645da127e0d061f93a2d073ee1423e7e3d
 
Python Version: 2.7
 
Any other relevant files/configs (eg: path files)

Anakin5:
You can pass by some forts while walking between two forts.
[...]

PR #5160, opened 2016-09-04
Fix spin fort task

Anakin5:
This is a fix for the spin fort task.
The task might need to be reviewed overall but for now this change is enough to fix issue https://github.com/PokemonGoF/PokemonGo-Bot/issues/5142

I fixed some pep8 issues as well


mjmadsen:
:+1:

PR #5166, opened 2016-09-04
telegram - remove unneeded

askovpen:
Removing support telegram usernames in config. Only userids


askovpen:
@Gobberwart , @DBa2016 
I understand correctly, that it's too much?

Gobberwart:
How does this improve anything?

askovpen:
It removes code that is not being used

Gobberwart:
PR #5156 addressed all the issues and has been merged. This pr is unnecessary.

askovpen:
this does not apply #5156

Gobberwart:
What does it apply to? The code you've removed in this commit is exactly the code I rewrote in 5156 and is definitely being used.

askovpen:
why you need unsernames in config? (not userid)

Gobberwart:
If it can work with user names, why shouldn't it? 

Also you are removing far too much code, eg. Code which checks for master == null, or invalid /unknown user id/name.

askovpen:
with usernames alerts work only after you send message to bot (every bot restarts). why need it?

Gobberwart:
Sorry I misread something there. For a minute it looked like you'd removed additional code, which I now see isn't the case.

I agree that this use of username doesn't work particularly well and until we can find a way to get id from username, or send to username instead of id, it's probably not something we want to encourage/support. Users should enter their id in config, not username.

pokemongo_bot/event_handlers/telegram_handler.py
```diff
@@ -96,19 +96,11 @@ def run(self):
                         outMessage = "Telegram bot setup not yet complete (master = null). Please enter your userid {} into bot configuration file.".format(update.message.from_user.id)
                         self.bot.logger.warn(outMessage)
                         continue
-                    if self.master not in [update.message.from_user.id, "@{}".format(update.message.from_user.username)]:
```
solderzzc: Some users already configure master as username, it's not good to remove it

askovpen: in documentation can use only id. username not work now, if every time after restart not send message to bot

Gobberwart: @solderzzc I agree, but I'd be curious to know how many people actually use the username in their config, since that doesn't actually work properly.

DBa2016: @askovpen after #5182 is merged, uid-username relation will persist bot restarts. Also, I think removing a feature deserves a specific mentioning in the PR head, not just a line note when someone notices it.

Gobberwart:
This isn't going anywhere and I think we're mostly agreed it's moving in the wrong direction. Closing.

askovpen:
Ок. I'll try to make discord support :-)
But I think - use usernames in telegram - wrong way

Gobberwart:
@askovpen Feel free to add your comments to issue #5192 for discussion.

PR #5177, opened 2016-09-04
CampFort - Check pokeballs before camping

davidakachaos:
## Short Description:
I noticed that the bot doesn't move when camping at a lure, but had run
out of pokeballs. This will first check IF we have any pokeballs left.
If we have no pokeballs left, abandon the lure (to move to a fort to get
more pokeballs) when we got some pokeballs again, will move back to the
lure.

### DIFF
--- a/pokemongo_bot/cell_workers/camp_fort.py
+++ b/pokemongo_bot/cell_workers/camp_fort.py
@@ -23,7 +23,6 @@ class CampFort(BaseTask):
         self.move_until = 0
         self.last_position_update = 0
         self.walker = None
-        self.announced_out_of_balls = False
 
         self.config_max_distance = self.config.get("max_distance", 2000)
         self.config_min_forts_count = self.config.get("min_forts_count", 2)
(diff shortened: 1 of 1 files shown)


pokemongo_bot/cell_workers/camp_fort.py
```diff
@@ -29,6 +31,13 @@ def initialize(self):
         self.config_moving_time = self.config.get("moving_time", 600)
 
     def work(self):
+        # Let's make sure we have balls before we sit at a lure!
+        # See also catch_pokemon.py
+        if sum([inventory.items().get(ball.value).count for ball in
+            [Item.ITEM_POKE_BALL, Item.ITEM_GREAT_BALL, Item.ITEM_ULTRA_BALL]]) <= 0:
+            self.logger.info('Not any pokeballs left, refuse to sit at lure!')
+            return WorkerResult.ERROR
```
Anakin5: That part should go after the check on "move_until". There is no point displaying that we would not sit there if the task is deactivated or if we are in the moving phase.
Also, does not seem like an error to me, more a WorkerResult.SUCCESS.

Anakin5:
On top of my comment, I am wondering something. The CampFort is not staying at a specific fort but in the middle of a cluster, which makes it a good location to spin forts, lazily speaking.
I can image a scenario where you are always out of Pokeballs, and so the task would have no effect at all. 

The original goal was to really stop walking, take at rest at a convenient place. That we agree would be less efficient than walking around whole day, but also much more human.
Did you notice that you run out of Pokeballs while activating the CampFort ?

davidakachaos:
I see what you mean. But my line of thought was; if I was outside, playing Pokemon Go, I'm sitting at a few lures, and I run out of Pokeballs... I would look around for other stops nearby (that might be the tweak here!) to spin and collect balls. As I would wander to other stops to get some balls, I would still come back to the lure "hotspot" when I got a few more balls.
What do you think? Is that a reasonable thing to argue here? Maybe with the distant tweak?

Anakin5:
OK. Let's just move down a bit the code then, to avoid message spam.

davidakachaos:
@Anakin5 Okay, so I've move the check below the enabled check. I also added a boolean to check if we've announced that the bot is out of balls. What do you think?

Anakin5:
@davidakachaos Actually it should go down a little bit more, after the check `now < move_until`. That is because we don't want to warn about not staying at the fort if were going to do so in the first place (because the timer moving is still under way).

To limit the output, you could also activate the `move_until` timer otherwise we are going to get 1 ball at a nearby pokestop and then come back to camp with only 1 ball. Would be better to walk for 10 minutes.

davidakachaos:
Thank you @Anakin5 for all the great advise :smile: Now the worker will move away from the lure for a set time after which it will check again for enough balls.

solderzzc:
👍

Anakin5:
I will do a little adjustement latter on to support move_time = 0 but it can wait a bit

PR #5179, opened 2016-09-04
Add sleep message from Telegram

developerllazy:
Add the function to send Telegram message when Bot is going to sleep.


solderzzc:
👍

PR #5181, opened 2016-09-04
PolylineWalker StepWalker better compatibility

th3w4y:
Fixes:
- PolylineWalker serves a destination further to StepWalker if we are not close to destination

### DIFF
--- a/pokemongo_bot/walkers/polyline_walker.py
+++ b/pokemongo_bot/walkers/polyline_walker.py
@@ -1,6 +1,8 @@
+# -*- coding: utf-8 -*-
 from random import uniform
 
-from pokemongo_bot.cell_workers.utils import distance
+import haversine
+
 from pokemongo_bot.walkers.step_walker import StepWalker
 from polyline_generator import PolylineObjectHandler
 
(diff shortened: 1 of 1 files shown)


Anakin5:
hum can you explain ? Why no giving the real destination ?

th3w4y:
real destination is at a different bearing/heading..

th3w4y:
Polyline says from O->D i have to walk O-A-B-C-...D

If i feed .get_pos() to StepWalker() it instantly set the position without any randomness...

Thus i virtually say that the destination is 2* speed thus SpetWalker will walk correct distance and return False because it did not finished...

img_0042

th3w4y:
@Anakin5 It is just a way to tell the StepWalker... walk in this direction but you will not be done...

Anakin5:
It is for choosing whether we should return true or false ? If only that, the polyline has `self.get_last_pos() == self.destination`

th3w4y:
@Anakin5 obviously i know that bit... but that does not suffice... 
I need the generated PolylineWalker(StepWalker) to  also use the offset_angle also in the close vicinity of the point...

Anakin5:
As I suggested earlier, we should just remove this code. It does not makes sense to set_position when we are close enough. Everything should be handled the same way.
We just check for distance to destination to know if we are arrived.

Anakin5:
What do you tihnk of something like that:
https://github.com/Anakin5/PokemonGo-Bot/blob/walker/pokemongo_bot/walkers/step_walker.py

th3w4y:
https://github.com/Anakin5/PokemonGo-Bot/blob/walker/pokemongo_bot/walkers/step_walker.py#L49

I think it will walk past the point... and it the user sets a high speed... it can be quite far...

Anakin5:
That is a problem in _get_next_pos then. We can bound it to the distance between the 2 points.

th3w4y:
Please elaborate not sure i follow to bound it by distance between 2 points..
Like a get closest point between real destination and the calculated destination based on bearing and speed...?
>That is a problem in _get_next_pos then. We can bound it to the distance between the 2 points

Anakin5:
But wait, that was already the case before...

th3w4y:
@Anakin5  you see this is kinda PolylineWalker specific.. not sure we can generalise that use case to the StepWalker...

th3w4y:
@Anakin5 

This will not work also because if destination given by .get_pos() is exactly one step.. the closes point will be the point itself without the offset.

>Yes get closest point between calculated destination and real destination.

Anakin5:
No but that means the StepWalker is not working right now for high speeds. Because the step returned by VincentyDistance is always as big as the distance travelled during one second.
So it is not related to Polyline, we should never go beyond the destination point.

th3w4y:
@Anakin5 

That is why we have and should keep
```
if distance < speed:
       set_position 

```

>No but that means the StepWalker is not working right now for high speeds. Because the step returned by VincentyDistance is always as big as the distance travelled during one second.

Anakin5:
or this:
https://github.com/Anakin5/PokemonGo-Bot/blob/walker/pokemongo_bot/walkers/step_walker.py#L114

th3w4y:
@Anakin5 

Now that is elegant..

>or this:
https://github.com/Anakin5/PokemonGo-Bot/blob/walker/pokemongo_bot/walkers/step_walker.py#L114

th3w4y:
@Anakin5 do you plan to do a PR?

if yes you can also get read of this part is handler in _get_next_pos()


```
         if self.speed == 0:
              raise Exception("Walking speed cannot be 0, change your walking speed higher than 1!")		              raise Exception("Walking speed cannot be 0, change your walking speed higher than 1!")
```

Anakin5:
I opened https://github.com/PokemonGoF/PokemonGo-Bot/pull/5184.

th3w4y:
@Anakin5  and i opens https://github.com/PokemonGoF/PokemonGo-Bot/pull/5185 for the remaining bits that still needs modifying in PolylineWalker

th3w4y:
@Anakin5 https://github.com/PokemonGoF/PokemonGo-Bot/pull/5184 that return still returns always True for the PolylineWalker

`return distance(self._new_position[0], self._new_position[1], self.destLat, self.destLng) < 1`

th3w4y:
This proposal would have manage to remove the .step() override and fix the issue also!

I will be tempted to reopen this PR tomorrow if @Anakin5 does not come with a different idea/solution

PR #5184, opened 2016-09-04
Improve step_walker

Anakin5:
Improve step_walker
- Can instanciate walkers one time and call step() multiple times on the same walker
- Speed randomized at each step
- Speed can be overridden in the step parameter to pass 0 to anything

### DIFF
--- a/pokemongo_bot/test/step_walker_test.py
+++ b/pokemongo_bot/test/step_walker_test.py
@@ -1,11 +1,12 @@
 import unittest
-from mock import MagicMock, patch
+
+from geographiclib.geodesic import Geodesic
+from mock import MagicMock, patch, mock
 
 from pokemongo_bot.walkers.step_walker import StepWalker
-from pokemongo_bot.cell_workers.utils import float_equal
 
-NORMALIZED_LAT_LNG_DISTANCE_STEP = 6.3593e-6
 NORMALIZED_LAT_LNG_DISTANCE = (6.3948578954430175e-06, 6.35204828670955e-06)
+
 
 class TestStepWalker(unittest.TestCase):
     def setUp(self):
@@ -16,14 +17,13 @@ class TestStepWalker(unittest.TestCase):
         self.bot.position = [0, 0, 0]
         self.bot.api = MagicMock()
 
-        self.lat, self.lng, self.alt = 0, 0, 0
-
         # let us get back the position set by the StepWalker
         def api_set_position(lat, lng, alt):
-            self.lat, self.lng, self.alt = lat, lng, alt
+            self.bot.position = [lat, lng, alt]
         self.bot.api.set_position = api_set_position
 
     def tearDown(self):
+        self.bot.position = [0, 0, 0]
         self.patcherSleep.stop()
 
     def test_normalized_distance(self):
@@ -34,14 +34,19 @@ class TestStepWalker(unittest.TestCase):
         self.bot.config.walk_min = 1
 
         sw = StepWalker(self.bot, 0.1, 0.1, precision=0.0)
-        self.assertGreater(sw.dLat, 0)
-        self.assertGreater(sw.dLng, 0)
+        self.assertGreater(sw.dest_lat, 0)
+        self.assertGreater(sw.dest_lng, 0)
 
-        stayInPlace = sw.step()
+        @mock.patch('random.uniform')
+        def run_step(mock_random):
+            mock_random.return_value = 0.0
+            return sw.step()
+
+        stayInPlace = run_step()
         self.assertFalse(stayInPlace)
 
-        self.assertTrue(float_equal(self.lat, NORMALIZED_LAT_LNG_DISTANCE[0]))
-        self.assertTrue(float_equal(self.lng, NORMALIZED_LAT_LNG_DISTANCE[1]))
+        self.assertAlmostEqual(self.bot.position[0], NORMALIZED_LAT_LNG_DISTANCE[0], places=6)
+        self.assertAlmostEqual(self.bot.position[1], NORMALIZED_LAT_LNG_DISTANCE[1], places=6)
 
         self.bot.config.walk_max = walk_max
         self.bot.config.walk_min = walk_min
@@ -54,14 +59,19 @@ class TestStepWalker(unittest.TestCase):
         self.bot.config.walk_min = 2
 
         sw = StepWalker(self.bot, 0.1, 0.1, precision=0.0)
-        self.assertTrue(sw.dLat > 0)
-        self.assertTrue(sw.dLng > 0)
+        self.assertTrue(sw.dest_lat > 0)
+        self.assertTrue(sw.dest_lng > 0)
 
-        stayInPlace = sw.step()
+        @mock.patch('random.uniform')
+        def run_step(mock_random):
+            mock_random.return_value = 0.0
+            return sw.step()
+
+        stayInPlace = run_step()
         self.assertFalse(stayInPlace)
 
-        self.assertTrue(float_equal(self.lat, NORMALIZED_LAT_LNG_DISTANCE[0] * 2))
-        self.assertTrue(float_equal(self.lng, NORMALIZED_LAT_LNG_DISTANCE[1] * 2))
+        self.assertAlmostEqual(self.bot.position[0], NORMALIZED_LAT_LNG_DISTANCE[0] * 2, places=6)
+        self.assertAlmostEqual(self.bot.position[1], NORMALIZED_LAT_LNG_DISTANCE[1] * 2, places=6)
 
         self.bot.config.walk_max = walk_max
         self.bot.config.walk_min = walk_min
@@ -73,13 +83,19 @@ class TestStepWalker(unittest.TestCase):
         self.bot.config.walk_max = 1
         self.bot.config.walk_min = 1
 
-        sw = StepWalker(self.bot, 0, 0)
-        self.assertEqual(sw.dLat, 0, 'dLat should be 0')
-        self.assertEqual(sw.dLng, 0, 'dLng should be 0')
+        sw = StepWalker(self.bot, 0, 0, precision=0.0)
+        self.assertEqual(sw.dest_lat, 0, 'dest_lat should be 0')
+        self.assertEqual(sw.dest_lng, 0, 'dest_lng should be 0')
 
-        self.assertTrue(sw.step(), 'step should return True')
-        self.assertTrue(self.lat == self.bot.position[0])
-        self.assertTrue(self.lng == self.bot.position[1])
+        @mock.patch('random.uniform')
+        def run_step(mock_random):
+            mock_random.return_value = 0.0
+            return sw.step()
+        moveInprecision = run_step()
+
+        self.assertTrue(moveInprecision, 'step should return True')
+        distance = Geodesic.WGS84.Inverse(0.0, 0.0, self.bot.position[0], self.bot.position[1])["s12"]
+        self.assertTrue(0.0 <= distance <= (sw.precision + sw.epsilon))
 
         self.bot.config.walk_max = walk_max
         self.bot.config.walk_min = walk_min
@@ -91,17 +107,95 @@ class TestStepWalker(unittest.TestCase):
         self.bot.config.walk_max = 1
         self.bot.config.walk_min = 1
 
-        sw = StepWalker(self.bot, 1e-5, 1e-5)
-        self.assertEqual(sw.dLat, 0)
-        self.assertEqual(sw.dLng, 0)
+        # distance from origin 0,0 to 1e-6, 1e-6 is 0.157253373328 meters
+        total_distance = Geodesic.WGS84.Inverse(0.0, 0.0, 1e-6, 1e-6)["s12"]
+        # we take a precision bigger then the total value...
+        sw = StepWalker(self.bot, 1e-6, 1e-6, precision=0.2)
+        self.assertEqual(sw.dest_lat, 1e-6)
+        self.assertEqual(sw.dest_lng, 1e-6)
+
+        @mock.patch('random.uniform')
+        def run_step(mock_random):
+            mock_random.return_value = 0.0
+            return sw.step()
+
+        moveInprecistion = run_step()
+        self.assertTrue(moveInprecistion, 'step should return True')
+
+        distance = Geodesic.WGS84.Inverse(0.0, 0.0, self.bot.position[0], self.bot.position[1])["s12"]
+        self.assertTrue(0.0 <= abs(total_distance - distance) <= (sw.precision + sw.epsilon))
 
         self.bot.config.walk_max = walk_max
         self.bot.config.walk_min = walk_min
 
-    @unittest.skip('This behavior is To Be Defined')
     def test_big_distances(self):
-        # FIXME currently the StepWalker acts like it won't move if big distances gives as input
-        # see args below
-        # with self.assertRaises(RuntimeError):
-        sw = StepWalker(self.bot, 10, 10)
-        sw.step() # equals True i.e act like the distance is too short for a step
+        walk_max = self.bot.config.walk_max
+        walk_min = self.bot.config.walk_min
+
+        self.bot.config.walk_max = 1
+        self.bot.config.walk_min = 1
+
+        sw = StepWalker(self.bot, 10, 10, precision=0.0)
+        self.assertEqual(sw.dest_lat, 10)
+        self.assertEqual(sw.dest_lng, 10)
+
+        @mock.patch('random.uniform')
+        def run_step(mock_random):
+            mock_random.return_value = 0.0
+            return sw.step()
+
+        finishedWalking = run_step()
+        self.assertFalse(finishedWalking, 'step should return False')
+        self.assertAlmostEqual(self.bot.position[0], NORMALIZED_LAT_LNG_DISTANCE[0], places=6)
+
+        self.bot.config.walk_max = walk_max
+        self.bot.config.walk_min = walk_min
+
+    def test_stay_put(self):
+        walk_max = self.bot.config.walk_max
+        walk_min = self.bot.config.walk_min
+
+        self.bot.config.walk_max = 4
+        self.bot.config.walk_min = 2
+
+        sw = StepWalker(self.bot, 10, 10, precision=0.0)
+        self.assertEqual(sw.dest_lat, 10)
+        self.assertEqual(sw.dest_lng, 10)
+
+        @mock.patch('random.uniform')
+        def run_step(mock_random):
+            mock_random.return_value = 0.0
+            return sw.step(speed=0)
+
+        finishedWalking = run_step()
+        self.assertFalse(finishedWalking, 'step should return False')
+        distance = Geodesic.WGS84.Inverse(0.0, 0.0, self.bot.position[0], self.bot.position[1])["s12"]
+        self.assertTrue(0.0 <= distance <= (sw.precision + sw.epsilon))
+
+        self.bot.config.walk_max = walk_max
+        self.bot.config.walk_min = walk_min
+
+    def test_teleport(self):
+        walk_max = self.bot.config.walk_max
+        walk_min = self.bot.config.walk_min
+
+        self.bot.config.walk_max = 4
+        self.bot.config.walk_min = 2
+
+        sw = StepWalker(self.bot, 10, 10, precision=0.0)
+        self.assertEqual(sw.dest_lat, 10)
+        self.assertEqual(sw.dest_lng, 10)
+
+        @mock.patch('random.uniform')
+        def run_step(mock_random):
+            mock_random.return_value = 0.0
+            return sw.step(speed=float("inf"))
+
+        finishedWalking = run_step()
+        self.assertTrue(finishedWalking, 'step should return True')
+        total_distance = Geodesic.WGS84.Inverse(0.0, 0.0, 10, 10)["s12"]
+        distance = Geodesic.WGS84.Inverse(0.0, 0.0, self.bot.position[0], self.bot.position[1])["s12"]
+        self.assertTrue(0.0 <= abs(total_distance - distance) <= (sw.precision + sw.epsilon))
+
+        self.bot.config.walk_max = walk_max
+        self.bot.config.walk_min = walk_min
--- a/pokemongo_bot/walkers/polyline_generator.py
+++ b/pokemongo_bot/walkers/polyline_generator.py
@@ -5,8 +5,8 @@ from itertools import chain
 import math
 import polyline
 import requests
-import numpy as np
 from geopy.distance import great_circle
+
 
 def distance(point1, point2):
     return Geodesic.WGS84.Inverse(point1[0], point1[1], point2[0], point2[1])["s12"]  # @UndefinedVariable
--- a/pokemongo_bot/walkers/polyline_generator.py
+++ b/pokemongo_bot/walkers/polyline_generator.py
@@ -1,12 +1,14 @@
 # -*- coding: utf-8 -*-
-from geopy.distance import VincentyDistance
-from geopy import Point
+from geographiclib.geodesic import Geodesic
 from itertools import chain
 
-import haversine
 import math
 import polyline
 import requests
+
+
+def distance(point1, point2):
+    return Geodesic.WGS84.Inverse(point1[0], point1[1], point2[0], point2[1])["s12"]  # @UndefinedVariable
 
 
 class PolylineObjectHandler:
@@ -26,10 +28,10 @@ class PolylineObjectHandler:
 
         # Absolute offset between bot origin and PolyLine get_last_pos() (in meters)
         if PolylineObjectHandler._cache and PolylineObjectHandler._cache.get_last_pos() != (None, None):
-            abs_offset = haversine.haversine(tuple(origin), PolylineObjectHandler._cache.get_last_pos())*1000
+            abs_offset = distance(origin, PolylineObjectHandler._cache.get_last_pos())
         else:
             abs_offset = float("inf")
-        is_old_cache = lambda : abs_offset > 8 # Consider cache old if we identified an offset more then 8 m
+        is_old_cache = lambda : abs_offset > 8  # Consider cache old if we identified an offset more then 8 m
         new_dest_set = lambda : tuple(destination) != PolylineObjectHandler._cache.destination
 
         if PolylineObjectHandler._run and (not is_old_cache()):
@@ -38,20 +40,20 @@ class PolylineObjectHandler:
             if PolylineObjectHandler._instability <= 0:
                 PolylineObjectHandler._instability = 0
                 PolylineObjectHandler._run = False
-            pass # use current cache
+            pass  # use current cache
         elif None == PolylineObjectHandler._cache or is_old_cache() or new_dest_set():
             # no cache, old cache or new destination set by bot, so make new polyline
             PolylineObjectHandler._instability += 2
             if 10 <= PolylineObjectHandler._instability:
                 PolylineObjectHandler._run = True
-                PolylineObjectHandler._instability = 20 # next N moves use same cache
+                PolylineObjectHandler._instability = 20  # next N moves use same cache
 
             PolylineObjectHandler._cache = Polyline(origin, destination, speed, google_map_api_key)
         else:
             # valid cache found
             PolylineObjectHandler._instability -= 1
             PolylineObjectHandler._instability = max(PolylineObjectHandler._instability, 0)
-            pass # use current cache
+            pass  # use current cache
         return PolylineObjectHandler._cache
 
 
@@ -60,7 +62,7 @@ class Polyline(object):
         self.speed = float(speed)
         self.origin = origin
         self.destination = tuple(destination)
-        self.DIRECTIONS_API_URL='https://maps.googleapis.com/maps/api/directions/json?mode=walking'
+        self.DIRECTIONS_API_URL = 'https://maps.googleapis.com/maps/api/directions/json?mode=walking'
         self.DIRECTIONS_URL = '{}&origin={}&destination={}'.format(self.DIRECTIONS_API_URL,
                 '{},{}'.format(*self.origin),
                 '{},{}'.format(*self.destination))
@@ -87,12 +89,12 @@ class Polyline(object):
         self._points = [self.origin] + self._get_directions_points() + [self.destination]
         self._polyline = self._get_encoded_points()
         self._last_pos = self._points[0]
-        self._step_dict =  self._get_steps_dict()
+        self._step_dict = self._get_steps_dict()
         self._step_keys = sorted(self._step_dict.keys())
         self._last_step = 0
 
         self._nr_samples = int(min(self.get_total_distance() / self.speed + 1, 512))
-        self.ELEVATION_API_URL='https://maps.googleapis.com/maps/api/elevation/json?path=enc:'
+        self.ELEVATION_API_URL = 'https://maps.googleapis.com/maps/api/elevation/json?path=enc:'
         self.ELEVATION_URL = '{}{}&samples={}'.format(self.ELEVATION_API_URL,
                                                       self._polyline, self._nr_samples)
         if google_map_api_key:
@@ -106,7 +108,7 @@ class Polyline(object):
         points = []
         for point in self._directions_encoded_points:
             points += polyline.decode(point)
-        return [x for n,x in enumerate(points) if x not in points[:n]]
+        return [x for n, x in enumerate(points) if x not in points[:n]]
 
     def _get_encoded_points(self):
         return polyline.encode(self._points)
@@ -125,7 +127,7 @@ class Polyline(object):
         walked_distance = 0.0
         steps_dict = {}
         for step in self._get_walk_steps():
-            walked_distance += haversine.haversine(*step) * 1000
+            walked_distance += distance(*step)
             steps_dict[walked_distance] = step
         return steps_dict
 
@@ -133,7 +135,7 @@ class Polyline(object):
         closest_sample = None
         best_distance = float("inf")
         for point in self._elevation_at_point.keys():
-            local_distance = haversine.haversine(self._last_pos, point)*1000
+            local_distance = distance(self._last_pos, point)
             if local_distance < best_distance:
                 closest_sample = point
                 best_distance = local_distance
@@ -142,91 +144,8 @@ class Polyline(object):
         else:
             return None
 
-    def get_pos(self):
-        if self.speed > self.get_total_distance():
-            self._last_pos = self.destination
-            self._last_step = len(self._step_keys)-1
-        if self.get_last_pos() == self.destination:
-            return self.get_last_pos()
-        distance = self.speed
-        origin = Point(*self._last_pos)
-        ((so_lat, so_lng), (sd_lat, sd_lng)) = self._step_dict[self._step_keys[self._last_step]]
-        bearing = self._calc_bearing(so_lat, so_lng, sd_lat, sd_lng)
-        while haversine.haversine(self._last_pos, (sd_lat, sd_lng))*1000 < distance:
-            distance -= haversine.haversine(self._last_pos, (sd_lat, sd_lng))*1000
-            self._last_pos = (sd_lat, sd_lng)
-            if self._last_step < len(self._step_keys)-1:
-                self._last_step += 1
-                ((so_lat, so_lng), (sd_lat, sd_lng)) = self._step_dict[self._step_keys[self._last_step]]
-                bearing = self._calc_bearing(so_lat, so_lng, sd_lat, sd_lng)
-                origin = Point(so_lat, so_lng)
-                lat, lng = self._calc_next_pos(origin, distance, bearing)
-                if haversine.haversine(self._last_pos, (lat, lng))*1000 < distance:
-                    distance -= haversine.haversine(self._last_pos, (lat, lng))*1000
-                    self._last_pos = (lat, lng)
-            else:
-                return self.get_last_pos()
-        else:
-            lat, lng = self._calc_next_pos(origin, distance, bearing)
-            self._last_pos = (lat, lng)
-            return self.get_last_pos()
-
-
     def get_total_distance(self):
-        return math.ceil(sum([haversine.haversine(*x) * 1000 for x in self._get_walk_steps()]))
-
+        return math.ceil(sum([distance(*x) for x in self._get_walk_steps()]))
 
     def get_last_pos(self):
         return self._last_pos
-
-    def set_speed(self, speed):
-        self.speed = speed
-
-    def _calc_next_pos(self, origin, distance, bearing):
-        lat, lng, _ = origin
-        if distance == 0.0:
-            return lat, lng
-        if distance == float("inf"):
-            return self.destination
-        lat, lng, _ = VincentyDistance(kilometers=distance * 1e-3).destination(origin, bearing)
-        return (lat, lng)
-
-    def _calc_bearing(self, start_lat, start_lng, dest_lat, dest_lng):
-        """
-        Calculates the bearing between two points.
-
-        The formulae used is the following:
-            θ = atan2(sin(Δlong).cos(lat2),
-                      cos(lat1).sin(lat2) − sin(lat1).cos(lat2).cos(Δlong))
-
-        :Parameters:
-          - `start_lat in decimal degrees
-          - `start_lng in decimal degrees
-          - `dest_lat in decimal degrees
-          - `dest_lng in decimal degrees
-
-        :Returns:
-          The bearing in degrees
-
-        :Returns Type:
-          float
-        """
-
-        lat1 = math.radians(start_lat)
-        lat2 = math.radians(dest_lat)
-
-        diffLong = math.radians(dest_lng - start_lng)
-
-        x = math.sin(diffLong) * math.cos(lat2)
-        y = math.cos(lat1) * math.sin(lat2) - (math.sin(lat1)
-                                               * math.cos(lat2) * math.cos(diffLong))
-
-        initial_bearing = math.atan2(x, y)
-
-        # Now we have the initial bearing but math.atan2 return values
-        # from -180° to + 180° which is not what we want for a compass bearing
-        # The solution is to normalize the initial bearing as shown below
-        initial_bearing = math.degrees(initial_bearing)
-        compass_bearing = (initial_bearing + 360) % 360
-
-        return compass_bearing
--- a/pokemongo_bot/walkers/polyline_generator.py
+++ b/pokemongo_bot/walkers/polyline_generator.py
@@ -6,7 +6,7 @@ import math
 import polyline
 import requests
 import numpy as np
-from scipy.spatial import KDTree
+from geopy.distance import great_circle
 
 def distance(point1, point2):
     return Geodesic.WGS84.Inverse(point1[0], point1[1], point2[0], point2[1])["s12"]  # @UndefinedVariable
@@ -104,10 +104,6 @@ class Polyline(object):
         self._elevation_at_point = dict((tuple(x['location'].values()),
                                          x['elevation']) for x in
                                         self._elevation_response['results'])
-        if self._elevation_at_point:
-            self._elevation_kdtree = KDTree(self._elevation_at_point.keys())
-        else:
-            self._elevation_kdtree = None
 
     def _get_directions_points(self):
         points = []
@@ -139,22 +135,8 @@ class Polyline(object):
     def get_alt(self, at_point=None):
         if at_point is None:
             at_point = self._last_pos
-        if self._elevation_kdtree:
-            p1, p2 = np.array(self._elevation_at_point.keys())[self._elevation_kdtree.query(at_point, k=2)[1]]
-            return self._get_relative_hight(at_point, tuple(p1), tuple(p2))
-        else:
-            return None
-
-
-    def _get_relative_hight(self, point, p1, p2):
-        ep1 = self._elevation_at_point[p1]
-        ep2 = self._elevation_at_point[p2]
-        hdelta = ep2 - ep1
-        distance_p1_p2 = distance(p1, p2)
-        distance_to_p1 = distance(p1, point)
-        distance_to_p2 = distance(p2, point)
-        elevation = ((math.pow(distance_p1_p2,2) + math.pow(distance_to_p1,2) - math.pow(distance_to_p2,2)) * hdelta)/ (3 * distance_p1_p2) + ep1
-        return elevation
+        
+        return sorted([(great_circle(at_point, k).meters, v) for k, v in self._elevation_at_point.items()])[0][1]
 
     def get_total_distance(self):
         return math.ceil(sum([distance(*x) for x in self._get_walk_steps()]))
--- a/pokemongo_bot/walkers/polyline_generator.py
+++ b/pokemongo_bot/walkers/polyline_generator.py
@@ -5,7 +5,8 @@ from itertools import chain
 import math
 import polyline
 import requests
-
+import numpy as np
+from scipy.spatial import KDTree
 
 def distance(point1, point2):
     return Geodesic.WGS84.Inverse(point1[0], point1[1], point2[0], point2[1])["s12"]  # @UndefinedVariable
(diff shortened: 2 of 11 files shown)


th3w4y:
@Anakin5 you'll need to adopt some test cases

Anakin5:
We need to modify offset angle as well to base it on km rather than speed.

th3w4y:
the offset angle formula ensures precision in centimetres

pokemongo_bot/walkers/polyline_walker.py
```diff
@@ -20,10 +19,3 @@ def __init__(self, bot, dest_lat, dest_lng):
         self.pol_alt = self.polyline.get_alt() or self.actual_alt
         super(PolylineWalker, self).__init__(self.bot, self.pol_lat, self.pol_lon,
                                              self.pol_alt, fixed_speed=self.speed)
-
```
th3w4y: I tested that.. the StepWalker returns

`return distance(self._new_position[0], self._new_position[1], self.destLat, self.destLng) < 1`

will still always be True for the polyline....

Anakin5:
I will add back the step override in Polyline to check that we walk the last polyline point.

th3w4y:
@Anakin5  I think i have a better solution...
PolylineWalker does not send the dest to StepWalker as the next step but the real destination... 
It will only override the bearing with the local heading.

  https://github.com/Anakin5/PokemonGo-Bot/pull/1/commits/09831cf72081fa81bcfb7138f1b67f7c195d7824   in the PR https://github.com/Anakin5/PokemonGo-Bot/pull/1 towards your branch

Anakin5:
That makes the StepWalker not very consistent. I think bearing should stay an internal variable, not something we can override.
Otherwise, next time we want to introduce some noise or change our calculation, we have to modify all caller of StepWalker.

What is the last point that we are trying to address ? Is it the return value of step() for polyline ?

th3w4y:
@Anakin5  I updated my comment check the display bit also...

> What is the last point that we are trying to address ? Is it the return value of step() for polyline ?

Anakin5:
Can the Polyline override a get_bearing function instead of passing it by parameter ?

th3w4y:
@Anakin5 that is also a solution...
>Can the Polyline override a get_bearing function instead of passing it by parameter ?

Anakin5:
Ok. Can you commit based on last version of the branch and I will merge it ?

th3w4y:
@Anakin5 Done... https://github.com/Anakin5/PokemonGo-Bot/pull/1

th3w4y:
@Anakin5 i have no time to test now... please check if overriding the _calc_bearing like that works... it should right?

Anakin5:
I will do the same for the ``fixed_speed`` and create a ``get_speed``. For the bearing, we could name it `get_bearing`, because I think it makes sense to expose this value to outside world for reading.

Anakin5:
I will test if it works

th3w4y:
@Anakin5  do not remove the fixed_speed.... we can use that to teleport or stay put... that is a nice good to have!

th3w4y:
that will also allow follow task to reduce the speed in certain condition... in locality of pokestops pokemons... whatever...

th3w4y:
@Anakin5 Updated above comment...

Anakin5:
I commited the verison I am testing, I added:
- Can instance walkers one time and call step several time on the same walker
- Speed randomized at each step
- Speed can be overridden in the step parameter to pass 0 to anything

th3w4y:
@Anakin5  i don't like it... your know why?

PolylineWalker generated the pol_lat and pol_lng based on a speed generated at init time... in order to pass it using this approach we have to override step again... in PolylineWalker which would not be the case if we keep it as an init argument....

Anakin5:
One point remaining is computation of the bearing with altitude.

th3w4y:
@Anakin5 what do you mean....?
>One point remaining is computation of the bearing with altitude.

pokemongo_bot/walkers/step_walker.py
```diff
@@ -8,147 +8,105 @@
...
         compass_bearing = (initial_bearing + 360) % 360
-    
+
         return compass_bearing
-        
-    def _get_next_pos(self, lat, lon, bearing, speed, precision):
-        origin = Point(lat, lon)
+
+    def get_next_position(self, origin_lat, origin_lng, origin_alt, dest_lat, dest_lng, dest_alt, speed):
+        origin = Point(origin_lat, origin_lng, origin_alt)
```
th3w4y: Point(origin_lat, origin_lng, origin_alt)

don't you want the point to be (origin_lat, origin_lng, dest_alt)??

pokemongo_bot/walkers/step_walker.py
```diff
@@ -8,147 +8,105 @@
...
-            return lat, lon
+            total_distance = distance(origin_lat, origin_lng, dest_lat, dest_lng)
+
+            if total_distance == 0:
+                return dest_lat, dest_lng, dest_alt
+
+            dist = min(total_distance, speed)
+            bearing = self._calc_bearing(origin_lat, origin_lng, dest_lat, dest_lng)
+            offset_angle = (100 / dist) * (self.precision / 1.74)
+            return VincentyDistance(kilometers=dist * 1e-3).destination(origin, bearing + uniform(-offset_angle, offset_angle))
```
th3w4y: or 

return VincentyDistance(kilometers=dist * 1e-3).destination(origin, bearing + uniform(-offset_angle, offset_angle))[0:2], dest_alt

Anakin5: Yes, it would just be nice to have the altitude progressively reach the destination

th3w4y: well the polyline.get_alt() gives you the alt at the closest point in the elevations_at_point to the .get_pos()

thus it does reach the final elevation progressively...

pokemongo_bot/walkers/step_walker.py
```diff
@@ -96,17 +96,20 @@ def _calc_bearing(self, start_lat, start_lng, dest_lat, dest_lng):
...
         if speed == float("inf"):
-            return dest_lat, dest_lng, dest_alt
+            dist = total_distance
         else:
-            total_distance = distance(origin_lat, origin_lng, dest_lat, dest_lng)
+            dist = min(total_distance, speed)
 
-            if total_distance == 0:
-                return dest_lat, dest_lng, dest_alt
+        if dist == 0:
```
th3w4y: when speed is zero next_alt should be orig_alt

Anakin5:
I meant what I just commited lol

pokemongo_bot/walkers/step_walker.py
```diff
@@ -96,20 +96,23 @@ def _calc_bearing(self, start_lat, start_lng, dest_lat, dest_lng):
...
 
-        if dist == 0:
-            dist = self.precision
-
         bearing = self._calc_bearing(origin_lat, origin_lng, dest_lat, dest_lng)
         offset_angle = (100 / dist) * (self.precision / 1.74)
 
         next_lat, next_lng, _ = VincentyDistance(kilometers=dist * 1e-3).destination(origin, bearing + uniform(-offset_angle, offset_angle))
-        next_alt = dest_alt + random_alt_delta()
+        next_alt = origin_lat + (dist / total_distance) * (dest_alt - origin_lat) + random_alt_delta()
```
th3w4y: when speed is zero next_alt should be orig_alt

th3w4y: if total_distance == 0

you have a DivisionByZero exception in:

dist / total_distance

th3w4y: we need something like...
```
if total_distance = 0:
    next_alt = orig_alt
else:
    next_alt = dest_alt + random_alt_delta()
```

Anakin5: fixed

th3w4y:
if total_distance == 0

you have a DivisionByZero exception in:

dist / total_distance

we need something like...
```
if total_distance = 0:
    next_alt = orig_alt
else:
    next_alt = dest_alt + random_alt_delta()
```

th3w4y:
why did you changed the offset_angle formula to 100/dist instead of 1/dist?
instead of  1/dist???

`offset_angle = (100 / dist) * (precision / 1.74)`

pokemongo_bot/walkers/step_walker.py
```diff
@@ -8,147 +8,111 @@
...
-            return self.destLat, self.destLng
+            dist = total_distance
         else:
-            offset_angle = (1/self.speed)*(precision/1.74)
-            lat, lon, _ = VincentyDistance(kilometers=speed*1e-3).destination(origin, bearing + uniform(-offset_angle, offset_angle))
-            return lat, lon
+            dist = min(total_distance, speed)
+
+        bearing = self._calc_bearing(origin_lat, origin_lng, dest_lat, dest_lng)
+        offset_angle = (100 / dist) * (self.precision / 1.74)
```
th3w4y: why did you changed the offset_angle formula to 100/dist instead of 1/dist?

pokemongo_bot/walkers/step_walker.py
```diff
@@ -8,147 +8,108 @@
...
-            return self.destLat, self.destLng
+            dist = total_distance
         else:
-            offset_angle = (1/self.speed)*(precision/1.74)
-            lat, lon, _ = VincentyDistance(kilometers=speed*1e-3).destination(origin, bearing + uniform(-offset_angle, offset_angle))
-            return lat, lon
+            dist = max(min(total_distance, speed), self.precision)
+
+        bearing = self._calc_bearing(origin_lat, origin_lng, dest_lat, dest_lng)
+        offset_angle = (100 / dist) * (self.precision / 1.74)
```
th3w4y: why did you changed the offset_angle formula to 100/dist instead of 1/dist?

th3w4y:
Please correct the angle formula


Correct dist=2 precision = 50 cm resulting angle 14 degrees
```
In [8]: (1/2.0)*(50/1.74) % 360
Out[8]: 14.367816091954023

```

Incorrect dist=2 precision = 50 cm resulting angle 356 degrees
```
In [9]: (100/2.0)*(50/1.74) % 360
Out[9]: 356.7816091954023

In [10]: 
```

Anakin5:
I put the precision in meters to be consistent with other units. That is not centimeters anymore.

pokemongo_bot/walkers/step_walker.py
```diff
@@ -8,147 +8,108 @@
...
+
+        next_lat, next_lng, _ = VincentyDistance(kilometers=dist * 1e-3).destination(origin, bearing + uniform(-offset_angle, offset_angle))
+        next_alt = origin_lat + (dist / total_distance) * (dest_alt - origin_lat) + random_alt_delta()
```
th3w4y: Still DivisionByZero

th3w4y:
if total_distance == 0

you have a DivisionByZero exception in:

dist / total_distance

we need something like...

if total_distance = 0:
    next_alt = orig_alt
else:
    next_alt = dest_alt + random_alt_delta()

Anakin5:
There is a check for `total_distance` zero few lines above

pokemongo_bot/walkers/step_walker.py
```diff
@@ -8,147 +8,108 @@
...
+            total_distance = self.precision
+
         if speed == float("inf"):
-            return self.destLat, self.destLng
+            dist = total_distance
         else:
-            offset_angle = (1/self.speed)*(precision/1.74)
-            lat, lon, _ = VincentyDistance(kilometers=speed*1e-3).destination(origin, bearing + uniform(-offset_angle, offset_angle))
-            return lat, lon
+            dist = max(min(total_distance, speed), self.precision)
```
th3w4y: offset_angle when total_distance = 0 is  off

In [17]: (100/0.5)/(0.5/1.74) % 360
Out[17]: 336.0

that is about 0.5m away from origin and will start spiralling with a 336 bearing 

```
In [32]: haversine.haversine((0,0), tuple(VincentyDistance(kilometers=0.5 * 1e-3).destination(Point(0,0), 0 + (100/0.5)/(0.5/1.74) % 360)[0:2]))*1000
Out[32]: 0.5022511946715152

In [33]: 
```

th3w4y:
@Anakin5 

offset_angle when total_distance = 0 is  off

In [17]: (100/0.5)/(0.5/1.74) % 360
Out[17]: 336.0

that is about 0.5m away from origin and will start spiralling with a 336 bearing 

```
In [32]: haversine.haversine((0,0), tuple(VincentyDistance(kilometers=0.5 * 1e-3).destination(Point(0,0), 0 + (100/0.5)/(0.5/1.74) % 360)[0:2]))*1000
Out[32]: 0.5022511946715152

In [33]: 
```

Anakin5:
What do you want to do when distance is zero ?
For me, staying in the precision circle around the position is the behavior I am expecting.

th3w4y:
@Anakin5 
I think we have to stay put... that is the safest bet..

if i run for lets say 30min StepWalker.step(speed=0)... we really have to stay put.. otherwise there is no way to predict where we will be...

>What do you want to do when distance is zero ?
For me, staying in the precision circle around the position is the behavior I am expecting.

Anakin5:
It looks like that:

image

Anakin5:
We should not repeatedly send the exact same position.

th3w4y:
@Anakin5 but then for that case... 0.5 m is to much away...

Anakin5:
We reduce the precision then :)

th3w4y:
when you choose the dest = precision...

offset_angle = (100/X)*(X/1.74) = 100/1.74 =  57.47126436781609 degrees always...

in that formula even if you change precision result will be the same

th3w4y:
@Anakin5 

dist should be:

`dist = min(total_distance, speed) `

lets not involve the precision here..

now for total_distance=0

```
if total_distance = 0:
    total_distance = 1e-3 # 1cm
```


that is 1cm away... makes sense?
```

In [75]: haversine.haversine((0,0), tuple(VincentyDistance(kilometers=1e-3 * 1e-3).destination(Point(0,0), 0 + (100/1e-3)*(0.5/1.74) % 360)[0:2]))*1000
Out[75]: 0.0010001441598963957

In [76]: 
```

Anakin5:
I don't understand what you are trying to achieve.
- When distance = 0, we do not care at all about the bearing or offset angle. The formula can return whatever.
- We do not want to send the same position many times.
- What is the purpose of the precision if next to the destination point, you modify it to 1cm ?

th3w4y:
lets stay i want to stay put for 10min
i do StepWalker.step(speed=0) right

600sec*50cm = 300m from original stay put lication....

Anakin5:
No your destination does not change. You always stay in a circle of radius your precision around your destination point.
See my screenshot, that is 30 minutes stay put.

th3w4y:
your destination will change because origin will change origin will be (bot.position[0], bot.position[1]) which just was updated one step above....

Anakin5:
Origin will change, not destination. Bearing will be recalculated.

th3w4y:
That if you pass the same destination always... (Which should be the case...)

>Origin will change, not destination. Bearing will be recalculated.

th3w4y:
@Anakin5 and i have to apologise if you think i commented to much on your PR :)

Anakin5:
Its ok :)

Anakin5:
I see what you mean for the stay put thing. You want to stop the walker in the middle and not at the end, right ?

th3w4y:
@Anakin5 yes... that would be a problem, even if destination does not change...

Let's compromise in total_distance = 1e-2  or 1e-3 what do you think?

if we stay put in the middle... then it will be no problem...

I am thinking of some sleep_schedules that might use that to stay put at time X (in the middle of walk) and still update the position...


>I see what you mean for the stay put thing. You want to stop the walker in the middle and not at the end, right ?

th3w4y:
@Anakin5 I wonder what is the gps precision when stationary...

th3w4y:
my phone returns 9mm

>@Anakin5 I wonder what is the gps precision when stationary...

Anakin5:
Ok we can reduce the value.
I want to say also that calling with speed=0 without giving the proper destination is not a valid use case.
Since we will not always return the same position, we need to know around which point we have to stay. And that is not the origin point, otherwise we could slowly deviate from the position we actually want to stay.
So that is a requirement to give the position at which you want to stay in destination.

th3w4y:
@Anakin5 but if we reduce the value it should be enough we  will still drift towards the final destination but just slowly...

th3w4y:
@Anakin5 actually for the PolylineWalker the speed=0 even if you say total_distance= self.precission 
the polyline.set_speed(0) and .get_pos() will return same intermediary point that it will ensure we drift around same area....


For the main StepWalker... it will drift towards final destination (and I don't care much about the main StepWalker use without the PolylineWalker) thus you decide...

Anakin5:
Yes it is not major issue, but that would still be nicer to remember where we initially stopped and want to stay around.

th3w4y:
@Anakin5 outdoor precision stationary cloudy day is 2m for me... (iPhone)

th3w4y:
@Anakin5 I know...

you fixed it in StepWalker

now we can set it to 0.5 in PolylineWalker something like:

```
PolylineWalker(StepWalker):
     def __init__(self, x,y...)
            super(PolylineWalker, self)._init(x,y,...)
            self.static_precision = 0.5
```

could you do that if you agree and have time...

Anakin5:
I am thinking about saving the position when speed=0 and talking that saved position as center point when subsequent speed=0 calls.
That way, the precision is not involved and things work whatever is the precision given

th3w4y:
@Anakin5 what do you have in mind?
I don't believe you can save it this object are volatile.. new ones are being created on next .tick()

>I am thinking about saving the position when speed=0 and talking that saved position as center point when subsequent speed=0 calls.
That way, the precision is not involved and things work whatever is the precision given

mjmadsen:
I love you guys :smile:

Anakin5:
Well, you cannot wait somewhere without providing the location. Either the destination is correct at initialisation of the StepWalker (so we have it in get_next_step), either the same object is reused.
I don't plan to cover the scenario in which the user create a brand new StepWalker, with origin A, destination B, and intend to stay put at C. Or at least, if he does that, then he reuse that same StepWalker as long as he is waiting. Otherwise I have no object to save the position.

th3w4y:
all tasks... have a .work() with returns WorkerResult.SUCCESS or WorkerResult.RUNNING in their .work() method right?

but the problem is that they usually create the StepWalker object on each .work()

>@Anakin5 Well, you cannot wait somewhere without providing the location. Either the destination is correct at initialisation of the StepWalker (so we have it in get_next_step), either the same object is reused.
I don't plan to cover the scenario in which the user create a brand new StepWalker, with origin A, destination B, and intend to stay put at C. Or at least, if he does that, then he reuse that same StepWalker as long as he is waiting. Otherwise I have no object to save the position.

Anakin5:
To ensure they do not deviate too much from the expected location, tasks that want to wait somewhere will have to create a StepWalker with the destination where they want to wait, or reuse the same StepWalker across calls. Is it acceptable thing ?

th3w4y:
@Anakin5 what along discussion around:

 `total_distance = self.static_precision`

now that we agreed is the Task resposability 

do we want self.static_precision to 0.5m  because the PolylineWalker is safe from that potential problem.. what to you think? :))

th3w4y:
I looked at the failing test cases, after fixing the obvious dLat, dLng which are now dest_lat, dest_lng variable name changes...  and passing precision=0 to StepWalker initialisation, we have 2 out of 4 failing:

- test_small_distance_same_spot 
- test_small_distance_small_step

they are both by the fact the we now do a position changes even on stay put, and on small distance.

I don't think those test cases make sense...
We should make sure that we do change the bot.position even if by little... the android or iOS changes location constantly why wouldn't we do that also...

I vote for removal of those 2 test cases...

Anakin5:
Lol I was looking into them as well.
I choose to keep the static_precision that we have for the PolylineWalker as well. Just so that we have consistent behavior when staying put.

pokemongo_bot/test/step_walker_test.py
```diff
@@ -92,8 +93,8 @@ def test_small_distance_small_step(self):
         self.bot.config.walk_min = 1
 
         sw = StepWalker(self.bot, 1e-5, 1e-5)
```
th3w4y: StepWalker(self.bot, 1e-5, 1e-5,precission=0.0)

to ensure consistency...

Anakin5:
Should be able to handle any precision now, even with speed zero

pokemongo_bot/walkers/step_walker.py
```diff
@@ -8,147 +8,110 @@
...
+
+        if total_distance == 0:
+            total_distance = self.precision
+
+        if speed == 0:
+            if not self.saved_location:
+                self.saved_location = origin_lat, origin_lng, origin_alt
+
+            dest_lat, dest_lng, dest_alt = self.saved_location
+            dist = self.precision
```
th3w4y: except precision 0 in test cases

ZeroDivisionError

offset_angle = (100 / dist) * (self.precision / 1.74)

Anakin5:
I am done once test cases passes. We can figure out what to do with test cases in another PR maybe.

th3w4y:
fbc0282 you will have to patch random.uniform in test cases now since precision is at least 0.01cm...

pokemongo_bot/test/step_walker_test.py
```diff
@@ -91,9 +92,9 @@ def test_small_distance_small_step(self):
         self.bot.config.walk_max = 1
         self.bot.config.walk_min = 1
 
-        sw = StepWalker(self.bot, 1e-5, 1e-5)
-        self.assertEqual(sw.dLat, 0)
-        self.assertEqual(sw.dLng, 0)
+        sw = StepWalker(self.bot, 1e-5, 1e-5, precision=0.0)
+        self.assertEqual(sw.dest_lat, 1e-5)
+        self.assertEqual(sw.dest_lng, 1e-5)
```
th3w4y: @th3w4y     
th3w4y added a line comment 15 minutes ago • edited
BTW:

This assert will always be true

self.assertEqual(sw.dest_lng, 1e-5)
self.lat is the position after update

pokemongo_bot/test/step_walker_test.py
```diff
@@ -73,13 +74,13 @@ def test_small_distance_same_spot(self):
...
-        self.assertEqual(sw.dLat, 0, 'dLat should be 0')
-        self.assertEqual(sw.dLng, 0, 'dLng should be 0')
+        sw = StepWalker(self.bot, 0, 0, precision=0.0)
+        self.assertEqual(sw.dest_lat, 0, 'dest_lat should be 0')
+        self.assertEqual(sw.dest_lng, 0, 'dest_lng should be 0')
 
         self.assertTrue(sw.step(), 'step should return True')
-        self.assertTrue(self.lat == self.bot.position[0])
-        self.assertTrue(self.lng == self.bot.position[1])
+        self.assertTrue(abs(self.lat - self.bot.position[0]) < 0.01)
```
th3w4y: this is 1.5 km

In [2]: haversine.haversine((0.0,0.0), (0.01, 0.01))*1000
Out[2]: 1572.533729286321

In [3]:

Anakin5:
Yes a lot of things to change in these tests cases :s

th3w4y:
@Anakin5 i tried to help https://github.com/Anakin5/PokemonGo-Bot/pull/2 :)

th3w4y:
This is a very good PR, I like your coding style @Anakin5 
now everything is much easy to read...

Anakin5:
thx, I am pain in the ass for my colleagues though ;)
I take a rest and will fix tests later, until you fix them :P

th3w4y:
@Anakin5 i adopted the test cases and opened a PR to your branch

https://github.com/Anakin5/PokemonGo-Bot/pull/3

th3w4y:
@Anakin5 https://github.com/Anakin5/PokemonGo-Bot/pull/3/commits/923e8fd67a280dfae9651bcb082d756eb52638b4 updated few more test cases and added back float("inf") which you have stollen :)



**test_big_distances (pokemongo_bot.test.step_walker_test.TestStepWalker) ... ok**
test_normalized_distance (pokemongo_bot.test.step_walker_test.TestStepWalker) ... ok
test_normalized_distance_times_2 (pokemongo_bot.test.step_walker_test.TestStepWalker) ... ok
test_small_distance_same_spot (pokemongo_bot.test.step_walker_test.TestStepWalker) ... ok
test_small_distance_small_step (pokemongo_bot.test.step_walker_test.TestStepWalker) ... ok
**test_stay_put (pokemongo_bot.test.step_walker_test.TestStepWalker) ... ok**
**test_teleport (pokemongo_bot.test.step_walker_test.TestStepWalker) ... ok**

th3w4y:
I just realised that the test_case don't properly mock the bot.position which is used by StepWalker is_arrived() 

I will look into mocking the bot.position...

pokemongo_bot/cell_workers/utils.py
```diff
@@ -41,17 +46,9 @@ def fort_details(bot, fort_id, latitude, longitude):
     # Just to avoid KeyErrors
     return FORT_CACHE.get(fort_id, {})
 
-def encode(cellid):
```
th3w4y: why did we remove that?

pokemongo_bot/test/step_walker_test.py
```diff
@@ -91,17 +107,95 @@ def test_small_distance_small_step(self):
         self.bot.config.walk_max = 1
         self.bot.config.walk_min = 1
 
-        sw = StepWalker(self.bot, 1e-5, 1e-5)
-        self.assertEqual(sw.dLat, 0)
-        self.assertEqual(sw.dLng, 0)
+        # distance from origin 0,0 to 1e-6, 1e-6 is 0.157253373328 meters
+        total_distance = Geodesic.WGS84.Inverse(0.0, 0.0, 1e-6, 1e-6)["s12"]
```
th3w4y: if you used geodesic in StepWalker... use haversine in tests for distance!

in the end this is how we catch differences

pokemongo_bot/walkers/polyline_generator.py
```diff
@@ -145,36 +147,39 @@ def get_alt(self):
...
+                lat, lng = self._calc_next_pos(origin, dist, bearing)
+                
+                if distance(self._last_pos, (lat, lng)) < dist:
+                    dist -= distance(self._last_pos, (lat, lng))
                     self._last_pos = (lat, lng)
             else:
                 return self.get_last_pos()
         else:
-            lat, lng = self._calc_next_pos(origin, distance, bearing)
+            lat, lng = self._calc_next_pos(origin, dist, bearing)
```
th3w4y: this is distance not dist

th3w4y: my bad

Anakin5:
@th3w4y 
For Polyline, I got the same algo, just written using the get_next_position of the StepWalker.

pokemongo_bot/walkers/polyline_walker.py
```diff
@@ -1,13 +1,31 @@
...
+
+        while True:
+            _, (step_lat, step_lng) = polyline._step_dict[polyline._step_keys[polyline._last_step]]
+            step_alt = polyline.get_alt() or origin_alt
+
+            next_lat, next_lng, next_alt = super(PolylineWalker, self).get_next_position(previous_lat, previous_lng, previous_alt, step_lat, step_lng, step_alt, distance)
+            distance -= Geodesic.WGS84.Inverse(previous_lat, previous_lng, next_lat, next_lng)["s12"]
+            distance = abs(distance)
+
+            if distance > (self.precision + self.epsilon):
```
th3w4y: what if you have a step < self.precision?

th3w4y: What if you previous position is in step but in last 1 cm...

th3w4y: What if total_distance contains 3 steps? what will be your bearing?

Anakin5: - I agree the speed is wrong, I am doing a small change for that
- Bearing is always towards the next step

th3w4y: bearing in your case is in the direction of  nextstep but if you have 3 steps within the total distance the final bearing is off.

pokemongo_bot/walkers/polyline_walker.py
```diff
@@ -20,7 +18,9 @@ def get_next_position(self, origin_lat, origin_lng, origin_alt, dest_lat, dest_l
             distance -= Geodesic.WGS84.Inverse(previous_lat, previous_lng, next_lat, next_lng)["s12"]
             distance = abs(distance)
 
-            if distance > (self.precision + self.epsilon):
+            if polyline._last_step == len(polyline._step_keys) - 1:
+                break
+            elif distance > (self.precision + self.epsilon):
```
th3w4y: bearing in your case is in the direction of nextstep but if you have 3 steps within the total distance the final bearing is off.

th3w4y:
still off with speed in m then the position you should be at!

Anakin5:
I don't understand, can you explain ?

Anakin5:
For now result are good except when precision is greater than the distance to travel. But it is something to adjust in StepWalker rather than Polyline.

Anakin5:
@th3w4y 
I fixed some issues, they might be was you were referring to but I am not sure. I want to be able to support big precision and that is now the case.

Anakin5:
2.5 meters precision:
image

pokemongo_bot/walkers/polyline_walker.py
```diff
@@ -1,29 +1,37 @@
...
+            dest_alt = polyline.get_alt() or origin_alt
+
+            next_lat, next_lng, next_alt = super(PolylineWalker, self).get_next_position(origin_lat, origin_lng, origin_alt, dest_lat, dest_lng, dest_alt, distance)
+
+            if polyline._last_step == len(polyline._step_keys) - 1:
+                break
+            else:
+                travelled = Geodesic.WGS84.Inverse(origin_lat, origin_lng, next_lat, next_lng)["s12"]
+                remaining = Geodesic.WGS84.Inverse(next_lat, next_lng, dest_lat, dest_lng)["s12"]
+                step_distance = Geodesic.WGS84.Inverse(origin_lat, origin_lng, dest_lat, dest_lng)["s12"]
```
th3w4y: BTW, step_distance is the dictionary key the one you are discarding at line 12:

`_, (dest_lat, dest_lng) = polyline._step_dict[polyline._step_keys[polyline._last_step]]`

Anakin5: Great, i will use it :)

th3w4y: polyline._step_keys[polyline._last_step

th3w4y:
i was trying to write some test cases for the polyline_generator in line with a **similar** approach what you have in PolylineWalker now
I am stocked in a loop when speed > distance...

I am not done yet, but the fact that you removed the .get_pos() makes it difficult.. 

~~you could have just rename it _get_pos() and not use it in the PolylineWalker :) then the polyline_generator_test.py would not been easily made to pass :)~~

~~we would have only needed to write new test cases for PolylineWalker....~~

Anakin5:
We can create a function to return the position of the current step and one to increment the step. Maybe that would make it easier to test and I could use them in the code as well in fact.

th3w4y:
lets have a side conversation on https://github.com/Anakin5/PokemonGo-Bot/pull/3

Anakin5:
Speed > distance should return the last point of the Polyline according to what I wrote in get_next_position

solderzzc:
> This branch is out-of-date with the base branch
> The Travis CI build failed
....

mjmadsen:
@solderzzc They're not finished yet :)

th3w4y:
probably the PR of the month with the nr of comments and changes involved

solderzzc:
@mjmadsen  Oh, sure.

th3w4y:
I created test cases for PolylineWalker  https://github.com/Anakin5/PokemonGo-Bot/pull/4 after @Anakin5's  changes and I can tell you all it looks very good except an small altitude issue regarding to which altitude the PolylineWalker would pick from the Polyline object but i am sure @Anakin5  will fix them tommorow.

Anakin5:
We need to declare the polyline ressources files as binary because on windows they get checkout with windows line endings and it messes up the pickle.

th3w4y:
Updated test case - activated the altitude check part...

```
test_one_small_speed (pokemongo_bot.test.polyline_walker_test.TestPolylineWalker) ... ALTITUDE [polyline: 429.589233398 bot: 416.760917594]
FAIL
test_one_small_speed_big_precision (pokemongo_bot.test.polyline_walker_test.TestPolylineWalker) ... ALTITUDE [polyline: 429.589233398 bot: 4.47246046117]
FAIL
test_stay_put (pokemongo_bot.test.polyline_walker_test.TestPolylineWalker) ... ALTITUDE [polyline: 429.589233398 bot: 0.159921135161]
FAIL

```

pokemongo_bot/walkers/step_walker.py
```diff
@@ -1,154 +1,82 @@
...
+        next_lat = direct["lat2"]
+        next_lng = direct["lon2"]
+
+        random_azi = uniform(inverse["azi1"] - 90, inverse["azi1"] + 90)
+        random_dist = uniform(0.0, self.precision)
+        direct = Geodesic.WGS84.Direct(next_lat, next_lng, random_azi, random_dist)
+
+        next_lat = direct["lat2"]
+        next_lng = direct["lon2"]
+        next_alt = origin_alt + (travel / total_distance) * (dest_alt - origin_alt) + random_alt_delta()
```
th3w4y: I think the elevation problem are coming from here...

`dest_alt + random_alt_delta()`

should be enough having in mind that dest_alt is at:
- end of polyline step in PolylineWalker
- random min/max in StepWalker

Anakin5: hum but it is more accurate to take an elevation in between. that should be the result expected.

Anakin5: travel == total_distance at the end of the polyline

Anakin5: Looks like since we don't call with the full remaining distance, we never reach travel == total_distance

th3w4y: A point in between is not good because.. lets say we have below polyline

O...a...b...c...d..e.....z....D
now elevations are:
O- 404m
a - 490m
b - 420m
...
w - 530m
z - 440m
D - 404m

th3w4y:
@Anakin5 I think i will change the .get_alt(at_point=None) to take a point parameter and identify the closest two point and return their average altitude if no point given that would return altitude at self._last_pos

pokemongo_bot/walkers/step_walker.py
Anakin5: Ok so I printed some logs and that is mostly a test case issue that is expecting the same altitude as the Polyline point while we are in between. I prefer to modify the testcase to check that we are between x and x+1 point altitude.

Anakin5:
No average, it won't work as well, but more check that we are in between

th3w4y:
@Anakin5 i meant a weighted "average"  i will actually return the this point
untitled

Anakin5:
ok, you do the projection or you just get the distance travelled somewhere ?

pokemongo_bot/walkers/step_walker.py
Anakin5: I don't get your example of the 800m line. Altitude is computed at proportion of the distance travelled between 2 points, not on the whole line.

th3w4y:
@Anakin5 i updated the photo

th3w4y:
@Anakin5 the logic to calculate the hight will be something like this

at_point = point you want to find elevation
p1, p2 are the closest 2 points on the polyline that we know the elevation

```
 
    def get_relative_hight(self, point, p1, p2):
        ep1 = self._elevation_at_point[p1]
        ep2 = self._elevation_at_point[p2]
        hdelta = ep2 - ep1
        distance_p1_p2 = distance(p1, p2)
        distance_to_p1 = distance(p1, point)
        distance_to_p2 = distance(p2, point)
        elevation = ((math.pow(distance_p1_p2,2) - math.pow(distance_to_p1,2) + math.pow(distance_to_p2,2)) * hdelta)/ (3 * distance_p1_p2) + ep1
        return elevation
```

Can someone check my trigonometry?

Anakin5:
I do not get where the elevation formula comes from, an my trigonometry skills are fine :)

Anakin5:
I get almost the same. I will check my signs
`ep1 + ((distance_p1_p2**2 + distance_to_p1**2 - distance_to_p2**2) / (3 * distance_p1_p2)) * (ep2 - ep1)`

th3w4y:
@Anakin5 I will check the same...

Anakin5:
I have something like that:
```
d1**2 = x**2 + h**2
d2**2 = (d12 - x)**2 + h**2

d1**2 - x**2 = d2**2 - (d12 - x)**2
d1**2 - x**2 = d2**2 - (d12**2 - 2*d12*x + x**2)
d1**2 - x**2 = d2**2 - d12**2 + 2*d12*x - x**2
d1**2 = d2**2 - d12**2 + 2*d12*x
d1**2 - d2**2 + d12**2 = 2*d12*x
(d1**2 - d2**2 + d12**2) / 2 = d12*x
```

th3w4y:
@Anakin5 yes i reverted the signs by mist ache when writing the code on my IDE on paper was correct

```
    def get_relative_hight(self, point, p1, p2):
        ep1 = self._elevation_at_point[p1]
        ep2 = self._elevation_at_point[p2]
        hdelta = ep2 - ep1
        distance_p1_p2 = distance(p1, p2)
        distance_to_p1 = distance(p1, point)
        distance_to_p2 = distance(p2, point)
        elevation = ((math.pow(distance_p1_p2,2) + math.pow(distance_to_p1,2) - math.pow(distance_to_p2,2)) * hdelta)/ (3 * distance_p1_p2) + ep1
        return elevation

```

th3w4y:
the remaining bit would be to identify the closest 2 points... but i have not much time to write that bit now :(

Anakin5:
I was doing some test with geographiclib stuff and just realized it is freaking expensive to compute stuff with it. It takes several seconds to compute the distance to all forts in the cell.

th3w4y:
@Anakin5 would take a second to calculate them all using Google DistanceMatrix API you would even get walking distances not geometrical distances

Anakin5:
I want to add something is for the `sleep(1)`. I want so that we take into account the last time we were called. There is no need to sleep 1 second if time already passed more than that.

pokemongo_bot/walkers/step_walker.py
```diff
@@ -22,9 +23,13 @@ def __init__(self, bot, dest_lat, dest_lng, dest_alt=None, precision=0.5):
             self.dest_alt = dest_alt
 
         self.saved_location = None
+        self.last_update = 0
 
     def step(self, speed=None):
-        sleep(1)
+        now = time.time()
+
+        sleep(1 - min(now - self.last_update, 1))
```
th3w4y: abs(now - self.last_update)

And you do want to take the max not the min!

this should be absolute...  sometimes time can tick back (especially in in while True... loops)

Then you'll get In 
```
[18]: time.sleep(-0.00001)
---------------------------------------------------------------------------
ValueError                                Traceback (most recent call last)
<ipython-input-18-8e671aa2b61c> in <module>()
----> 1 time.sleep(-0.00001)

ValueError: sleep length must be non-negative

In [19]: 
```

Anakin5: The value in sleep cannot end up negative, even if times ticks back.

th3w4y: on tick back:
now- self.last_update = -1.00001
1 - min(-1.00001, 1) = -0.00001 !!!

Anakin5: = 2.00001 :)

th3w4y: by bad... :) i will delete the comments...

Anakin5:
Were you able to get the tests working with the altitude formula ?

th3w4y:
yes

i finished rewrite of get_alt

I use KDTree to calculated closest 2 points in the cluster... something you could reuse for forts...

```
   def get_alt(self, at_point=None):
        if at_point is None:
            at_point = self._last_pos
        if self._elevation_kdtree:
            p1, p2 = np.array(self._elevation_at_point.keys())[self._elevation_kdtree.query(at_point, k=2)[1]]
            return self._get_relative_hight(at_point, tuple(p1), tuple(p2))
        else:
            return None


    def _get_relative_hight(self, point, p1, p2):
        ep1 = self._elevation_at_point[p1]
        ep2 = self._elevation_at_point[p2]
        hdelta = ep2 - ep1
        distance_p1_p2 = distance(p1, p2)
        distance_to_p1 = distance(p1, point)
        distance_to_p2 = distance(p2, point)
        elevation = ((math.pow(distance_p1_p2,2) + math.pow(distance_to_p1,2) - math.pow(distance_to_p2,2)) * hdelta)/ (3 * distance_p1_p2) + ep1
        return elevation
```

th3w4y:
One one test case in polylineWalker is failing

Altitude on stay put... but i guess i should have initiated the bot with some altitude for that test case

```
test_bigger_then_total_speed (pokemongo_bot.test.polyline_walker_test.TestPolylineWalker) ... ok
test_bigger_then_total_speed_big_precision_offset (pokemongo_bot.test.polyline_walker_test.TestPolylineWalker) ... ok
test_intermediary_speed (pokemongo_bot.test.polyline_walker_test.TestPolylineWalker) ... ok
test_intermediary_speed_big_precision (pokemongo_bot.test.polyline_walker_test.TestPolylineWalker) ... ok
test_one_small_speed (pokemongo_bot.test.polyline_walker_test.TestPolylineWalker) ... ok
test_one_small_speed_big_precision (pokemongo_bot.test.polyline_walker_test.TestPolylineWalker) ... ok
test_polyline_fetched (pokemongo_bot.test.polyline_walker_test.TestPolylineWalker) ... ok
test_stay_put (pokemongo_bot.test.polyline_walker_test.TestPolylineWalker) ... ALTITUDE [polyline: 429.548169047 bot: 0.0892242152855]
FAIL
test_teleport (pokemongo_bot.test.polyline_walker_test.TestPolylineWalker) ... ok

======================================================================
FAIL: test_stay_put (pokemongo_bot.test.polyline_walker_test.TestPolylineWalker)
----------------------------------------------------------------------
  File "/Users/oghinet/Checkout/public/PokemonGo-Bot/pokemongo_bot/test/polyline_walker_test.py", line 280, in test_stay_put
    self.assertTrue(abs(self.polyline.get_alt() - self.bot.position[2]) <= 1)
AssertionError: False is not true

----------------------------------------------------------------------
Ran 9 tests in 0.366s

FAILED (failures=1)
```

th3w4y:
@Anakin5 I modified the test case on stay_put it makes sense that altitude does not change much...

```
test_bigger_then_total_speed (pokemongo_bot.test.polyline_walker_test.TestPolylineWalker) ... ok
test_bigger_then_total_speed_big_precision_offset (pokemongo_bot.test.polyline_walker_test.TestPolylineWalker) ... ok
test_intermediary_speed (pokemongo_bot.test.polyline_walker_test.TestPolylineWalker) ... ok
test_intermediary_speed_big_precision (pokemongo_bot.test.polyline_walker_test.TestPolylineWalker) ... ok
test_one_small_speed (pokemongo_bot.test.polyline_walker_test.TestPolylineWalker) ... ok
test_one_small_speed_big_precision (pokemongo_bot.test.polyline_walker_test.TestPolylineWalker) ... ok
test_polyline_fetched (pokemongo_bot.test.polyline_walker_test.TestPolylineWalker) ... ok
test_stay_put (pokemongo_bot.test.polyline_walker_test.TestPolylineWalker) ... ok
test_teleport (pokemongo_bot.test.polyline_walker_test.TestPolylineWalker) ... ok

----------------------------------------------------------------------
Ran 9 tests in 0.335s

OK
```

Anakin5:
Why do we need closest 2 points ? Algo is using previous and next point.

th3w4y:
because actually previous calculations were incorrect...

the elevation dictionary does not contain elevation at the polyline points... but along the polyline thus even for origin we still have to calculate...

Anakin5:
So there is no point doing any projection then. We just gives altitude of the nearest point.
The get_relative_high should help us for the test cases but we should not use it in our Algo otherwise that is like doing 2 times the calculation (1 time in get_next_position an 1 time in get_alt)

th3w4y:
I have to study more how far away are the elevation samples... but you understand... at least now is correct!
>So there is no point doing any projection then. We just gives altitude of the nearest point.
The get_relative_high should help us for the test cases but we should not use it in our Algo otherwise that is like doing 2 times the calculation (1 time in get_next_position an 1 time in get_alt)

th3w4y:
@Anakin5 we don't nee to do it with the geopraphillib... if we just do without looking at the earth curvature... it would not matter because we calculate to get deltas in elevation not distances in meters...
thus i guess i could use scipy.pdist there... to get the distances!

Anakin5:
I guess the polyline is monotone so your computation of nearest 2 points actually return the point before and the point after. That explains why it works. But we could just do that, get the point before and the one after.

th3w4y:
@Anakin5 what you propose will be harder if you look at the picture... 

I have more elevation samples ( i can get up to 512 in one API call ) then polyline points!

th3w4y:
@Anakin5 yes the response from the Elevation API is monotone.. because i request X samples of elevations on the encoded polyline... (the result from Directions...) 

but is not sure that the polyline points coincide with the elevation points... at any point on the line they are probably just equally spread across the polyline...

Anakin5:
OK. What is this kdtree thing doing besides computing distance between points ?

Anakin5:
why not something like `elevation_points.sort(distance)[:2]` ?

th3w4y:
@Anakin5 KDTree is way faster.

Anakin5:
How does kdtree evaluate distances between points ?

th3w4y:
@Anakin5 the KDtree creates a tree on init and then the .query(is very fast...)

th3w4y:
@Anakin5  that is why i was saying that you can also use it for the find_closest_fort...!

th3w4y:
euclidean distance but you can change the method if you like!
>How does kdtree evaluate distances between points ?

th3w4y:
```
In [14]: %timeit polyline.get_alt()
The slowest run took 5.14 times longer than the fastest. This could mean that an intermediate result is being cached.
1000 loops, best of 3: 585 µs per loop

In [15]: 
```

th3w4y:
i brough it down to 
```
In [5]: %timeit p.get_alt()
1000 loops, best of 3: 270 µs per loop

In [6]: 
```

th3w4y:
Changed to guppy on @Anakin5  suggestion since scipy wan not easy to compile on windows...

```
    def get_alt(self, at_point=None):
        if at_point is None:
            at_point = self._last_pos
        if self._elevation_at_point:
            (distance_to_p1, ep1, p1), (distance_to_p2, ep2, p2) = sorted([(great_circle(at_point, k).meters, v, k) for k, v in self._elevation_at_point.items()])[:2]
            distance_p1_p2 = great_circle(p1, p2).meters
            return self._get_relative_hight(ep1, ep2, distance_p1_p2, distance_to_p1, distance_to_p2)
        else:
            return None


    def _get_relative_hight(self, ep1, ep2, distance_p1_p2, distance_to_p1, distance_to_p2):
        hdelta = ep2 - ep1
        elevation = ((math.pow(distance_p1_p2,2) + math.pow(distance_to_p1,2) - math.pow(distance_to_p2,2)) * hdelta)/ (3 * distance_p1_p2) + ep1
        return elevation 
```

Anakin5:
We are done I think. Waiting for confirmation form @th3w4y

solderzzc:
I've tested it locally, seems no issue

th3w4y:
@Anakin5 looks good 👍

solderzzc:
👍

th3w4y:
@Anakin5 you would only need to rebase...

solderzzc:
@th3w4y  will you merge ?

th3w4y:
@solderzzc i'll wait for @Anakin5  to rebase on dev

th3w4y:
@Anakin5 @solderzzc Merged!

PR #5187, opened 2016-09-04
keep stardust count updated

rawgni:
## Short Description:

display stardust gained when capturing a pokemon
keep the stardust count updated

-
-
-


solderzzc:
👍

PR #5189, opened 2016-09-04
Add sleep reminder

cmezh:
Added sleep reminder. Emits ```next_sleep``` event every ```sleep_reminder_interval``` (or more) seconds.
```sleep_reminder_interval``` is 600 seconds by default.

### CODE REVIEW

**Comment by Gobberwart** [comment]
Looks good. Please update configuration_files.md.

**Comment by Gobberwart** [comment]
@cmezh Please update configuration_files.md and update the branch so we can merge.

**Comment by cmezh** [comment]
@Gobberwart Yep, I will do a bit later. Also there will be a little fix for this PR.

**Comment by cmezh** [comment]
@Gobberwart @mjmadsen Done.

**Comment by Gobberwart** [comment]
OK, sorry for not noticing this earlier, but is there any reason this config item is not inside the SleepSchedule task? It only applies if SleepSchedule is enabled so should be kept together with that task rather than as global config.

**Comment by cmezh** [comment]
@Gobberwart Generally it is because all user configs will be broken again by changing the structure of sleep_schedule config. I think too that it's better to put it inside.
If you want me to do it this way, maybe for now I will close this PR and a bit later open another one: currently I'm preparing another PR with some improvements for SleepSchedule and its tests. Also with feature to enable/disable every single SleepSchedule config entry.
In case of changing config structure it will be possible also to add option to enable/disable whole SleepSchedule.

**Comment by cmezh** [comment]
Now config looks like this:
```"sleep_schedule": [
  {
    "time": "",
    "duration": "",
    "time_random_offset": "",
    "duration_random_offset": "",
    "wake_up_at_location": ""
  }
]```

After changes it will look like this:
```"sleep_schedule": {
  "enabled": true,
  "reminder_interval": 600,
  "entries": [
    {
      "enabled": true,
      "time": "",
      "duration": "",
      "time_random_offset": "",
      "duration_random_offset": "",
      "wake_up_at_location": ""
    }
  ]
}```

**Comment by Gobberwart** [comment]
Seems off to me. Should be:

```
"sleep_schedule": {
  "enabled": true,
  "time": "",
  "duration": "",
  "time_random_offset": "",
  "duration_random_offset": "",
  "wake_up_at_location": "",
  "reminder_interval": 600
]
```

Or have I missed some reason for the added child config?

Gobberwart:
Could throw in a "reminder_enabled": true/false in there as well for people who don't want it.

Gobberwart:
Ah OK I missed your comment re changing the sleepschedule to allow multiple sleeps. Makes sense now.

Up to you whether to make change in this PR or close it and incorporate all changes in a new one.

cmezh:
@Gobberwart It allows multiple sleeps for a long time since it was moved from task to main bot instance ;)
I'm going to merge both PRs and change config structure... oh... I hope it's right way.

Issue #5174, opened 2016-09-04
Skip ignored "Pokemon" spamming log.


maxprzemo:
### Actual Behavior
![zrzut ekranu z 2016-09-04 11 13 47](https://cloud.githubusercontent.com/assets/4314537/18230873/f111ba94-72a8-11e6-813b-fb73ec196878.png)

![zrzut ekranu z 2016-09-04 14 56 06](https://cloud.githubusercontent.com/assets/4314537/18231126/2c0985c6-72b0-11e6-90e1-ffd877fa3cb1.png)

Skip ignored "Pokemon" spamming log.
Blocks the movement bots.
Sends too many requests to the server? I am not sure about that.
Softban to pokestop?
![zrzut ekranu z 2016-09-04 14 24 06](https://cloud.githubusercontent.com/assets/4314537/18230958/f48249f8-72aa-11e6-9d6e-c8d2cb5d9a23.png)
![zrzut ekranu z 2016-09-04 14 24 17](https://cloud.githubusercontent.com/assets/4314537/18230959/f7eb0030-72aa-11e6-88cf-893bc57a7977.png)

### Expected Behavior
Ignore pokemon only once.

### config.json
I have very little Pokeballs, That's why I use this restriction:
    },
    "catch": {
      "any": {"catch_above_cp": 0, "catch_above_iv": 0, "logic": "or"},
      "// Example of always catching Rattata:": {},
      "Pidgey": {"catch_above_cp": 300, "catch_above_iv": 0.8, "logic": "or"},
      "Rattata": {"catch_above_cp": 300, "catch_above_iv": 0.8, "logic": "or"},
      "Drowzee": {"catch_above_cp": 300, "catch_above_iv": 0.8, "logic": "or"},
      "Spearow": {"catch_above_cp": 300, "catch_above_iv": 0.8, "logic": "or"},
      "Pidgeotto": {"catch_above_cp": 100, "catch_above_iv": 0.8, "logic": "or"},
      "Zubat": {"catch_above_cp": 300, "catch_above_iv": 0.8, "logic": "or"},
      "Raticate": {"catch_above_cp": 300, "catch_above_iv": 0.8, "logic": "or"},
      "Krabby": {"catch_above_cp": 100, "catch_above_iv": 0.8, "logic": "or"},
      "Gastly": {"catch_above_cp": 100, "catch_above_iv": 0.8, "logic": "or"},
[...]

PR #5197, opened 2016-09-05
Prevent duplicate "skip ignored pokemon" alerts

Gobberwart:
## Short Description:

Persist ignored/skipped pokemon so that alerts are not continually generated.

Important note: An encountered pokemon does not have a useful unique id (at least not that I can see) so I have used pokemon_id, cp_exact and ivcp as comparison values when deciding if an alert has already been generated. That *should* be sufficient but happy to expand if necessary.

- #5174


solderzzc:
👍

PR #5204, opened 2016-09-05
Format output to fit in stdout codepage. Fix #5144.

### LINKED ISSUE #5144: Logging headers repeated on utf8 messages
### Expected Behavior
Message header only appear one time

### Actual Behavior
Headers are repeated on messages containing utf8 characters

```[2016-09-03 18:20:40] [lured_pokemon_found] [2016-09-03 18:20:40] [lured_pokemon_found] Lured pokemon at fort þ¥ÄÕ¡© (f7d3a79d3c874272aa6d7482c74afd2a.16)```

### Your FULL config.json (remove your username, password, gmapkey and any other private info)
Not relevant

```[2016-09-03 18:20:40] [lured_pokemon_found] [2016-09-03 18:20:40] [lured_pokemon_found] Lured pokemon at fort þ¥ÄÕ¡© (f7d3a79d3c874272aa6d7482c74afd2a.16)```

### Steps to Reproduce
Bot in an asia country ^^

### Other Information
OS: Windows
Branch:  dev
Git Commit: HEAD
Python Version: 2.7.12

### PR DESCRIPTION
**Pull request by Anakin5**
+ Do not force default encoding to utf-8 (which causes issues), but adapt the output to the current encoding.
Messages going though the logging and color logging handler will always have an encoding supported by stdout.

+ Support chcp 65001 for displaying utf-8 encoded characters in console window if system is able to display them.

### DIFF
--- a/pokecli.py
+++ b/pokecli.py
@@ -24,6 +24,7 @@ OR OTHER DEALINGS IN THE SOFTWARE.
 
 Author: tjado <https://github.com/tejado>
 """
+from __future__ import unicode_literals
 
 import argparse
 import codecs
@@ -32,8 +33,6 @@ import logging
 import os
 import ssl
 import sys
-reload(sys)
-sys.setdefaultencoding('utf8')
 import time
 import signal
 import string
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

**Comment by mjmadsen** [comment]
:+1:

**Comment by sohje** [comment]
This commit broke event emitter. Default encoding was set to ascii instead of utf-8(sys.reload, setdefaultencoding('utf-8')), when we try to emit events with utf-8 encoded strings(unicode type). Exception is raised:
```python
UnicodeEncodeError: 'ascii' codec can't encode characters in position 0-2: ordinal not in range(128)
```

Anakin5:
You need to give more details. The error message you have is very generic. There is no need to force the default encoding to utf-8 to not have errors with utf-8 encoded strings.
Which event raised the error ? can we have the full stack ?

PR #5214, opened 2016-09-05
metric uses internal stardust counter

rawgni:
## Short Description:

don't need to ask the server for stardust count. our internal counter is always up to date.

-
-
-


mjmadsen:
:+1:

PR #5237, opened 2016-09-06
Fix #5232

Anakin5:
Fix Connection reset by peer issue when evolving

Issue #4605, opened 2016-08-23
Mimic the request workflow of official application


s0crat:
### Short Description
Bot request workflow is not implemented in the same way as the official application do. It's a way for Niantic to detect bots. Below is the request workflow sent by my Android official application, latest version:

**1st step:**
```
[<-] Request for GetPlayer  {}
[->] Response for GetPlayer {
  // some stuff }
```
  
**2d step:**
```
[<-] Request for DownloadRemoteConfigVersion  { platform: 'ANDROID', app_version: 3300, '$unknownFields': [] }
[<-] Request for GetHatchedEggs  {}
[<-] Request for GetInventory  {}
[<-] Request for CheckAwardedBadges  {}
[<-] Request for DownloadSettings  {}
[->] Response for DownloadRemoteConfigVersion  {
  // some stuff }
[->] Response for GetHatchedEggs  {
  // some stuff }
[->] Response for GetInventory  {
  // some stuff }
[->] Response for CheckAwardedBadges  {
  // some stuff }
[->] Response for DownloadSettings  {
    "hash": "1234567890123456789012345678901234567890", // 40bytes hash
    // other settings sent by the server including "google_maps_api_key" }
```
	
**3rd Step**
```
[<-] Request for GetAssetDigest  { platform: 'ANDROID', app_version: 3300, '$unknownFields': [] }
[<-] Request for GetHatchedEggs  {}
[<-] Request for GetInventory  { last_timestamp_ms: '1234567890123', '$unknownFields': [] }
[<-] Request for CheckAwardedBadges  {}
[<-] Request for DownloadSettings  { hash: '1234567890123456789012345678901234567890', // 40bytes hash received previously
  '$unknownFields': [] }
[->] Response for GetAssetDigest  {
    "digest": [
        {
            "asset_id": "12345678-90ab-cdef-1234-567890abcdef/1234567890123456",
            "bundle_name": "pm1234",
            "version": "1234567890123456",
            "checksum": 1234567890,
            "size": 123456,
            "key": {
                "type": "Buffer",
                "data": [
                    12, 34, 567, 89, 012, 345, 678, 901, 234, 56, 78, 901, 23, 45, 678, 901
                ]
            }
        },
		{
		    "asset_id": ...
			// repetead asset_id stuff
	]
}
[->] Response for GetHatchedEggs  {
  // some stuff }
[->] Response for GetInventory  {
  // some stuff }
[->] Response for CheckAwardedBadges  {
[...]

PR #5243, opened 2016-09-06
Heartbeat decoupled from tasks.

sohje:
## Short Description:
Heartbeats decoupled from tasks. Heartbeat coupled to bot main instance, separate main hb logic from any tasks.
- Need some additional tests, works for me (evolved 40 pokemons, heartbeats is ok)
- No need #5237

### DIFF
--- a/pokemongo_bot/__init__.py
+++ b/pokemongo_bot/__init__.py
@@ -112,6 +112,7 @@ class PokemonGoBot(object):
         self.heartbeat_threshold = self.config.heartbeat_threshold
         self.heartbeat_counter = 0
         self.last_heartbeat = time.time()
+        self.hb_locked = False # lock hb on snip
 
         self.capture_locked = False  # lock catching while moving to VIP pokemon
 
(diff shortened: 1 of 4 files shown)


solderzzc:
Need a ignore heartbeat flag, since sniper will change location but should not send heartbeat since the location changed. correct me if I'm wrong.

sohje:
@solderzzc Ops, sure! Add a flag to bot instance:
```python
self.hb_locked = False
```

Anakin5:
Is there any location info in the heartbeat ?

solderzzc:
@Anakin5 every api request include auth info, including device_info/location/sensor data.

Anakin5:
ok. Also the heartbeat function update the timeout_fort list. Do we plan to move that somewhere else ? maybe in the tick ?

rawgni:
any idea what other api is being chained inside the heartbeat call ?
#4605

solderzzc:
👍 
Merge at the time, I believe the get_map_obj is in the heartbeat call as well.

Issue #5216, opened 2016-09-05
Bot very slow with "enable_social:true"


dragonalvaro:
Just clone the bot again since weeks. When I started it takes like hours to start doing something. I ran the script at 14:39 and it is 15:51 and just showed "Starting bot" message.

It's a desktop computer with i7 Skylake, 16 Gb ram and business bandwith. PokeMap runs fine in the same machine.

Take a look at the log time entries

2016-09-05 **14:39:30**,167 [       cli] [INFO] PokemonGO Bot v1.0
2016-09-05 14:39:30,171 [       cli] [INFO] commit: ca94424f
2016-09-05 14:39:30,173 [       cli] [INFO] Configuration initialized
2016-09-05 14:39:30,173 [pokemongo_bot.health_record.bot_event] [INFO] Health check is enabled. For more information:
2016-09-05 14:39:30,173 [pokemongo_bot.health_record.bot_event] [INFO] https://github.com/PokemonGoF/PokemonGo-Bot/tree/dev#analytics
2016-09-05 14:39:30,177 [requests.packages.urllib3.connectionpool] [INFO] Starting new HTTP connection (1): www.google-analytics.com
(3170) wsgi starting up on http://127.0.0.1:4000
2016-09-05 14:39:30,267 [requests.packages.urllib3.connectionpool] [INFO] Starting new HTTP connection (1): 127.0.0.1
[2016-09-05 14:39:30] [SleepSchedule] [INFO] [next_sleep] Next sleep at 18:15:28, for a duration of 2:32:10
[2016-09-05 14:41:37] [PokemonGoBot] [INFO] [set_start_location] Setting start location.
[2016-09-05 14:43:44] [PokemonGoBot] [INFO] [x] Coordinates found in passed in location, not geocoding.
[2016-09-05 14:43:44] [PokemonGoBot] [INFO] [location_found] Location found: 42.578389, -10.616962 (46.558389, -3.616962, 8.0)
[2016-09-05 14:45:52] [PokemonGoBot] [INFO] [position_update] Now at (42.578389, -10.66962, 8.0)
[2016-09-05 **14:52:13**] [PokemonGoBot] [INFO] [login_started] Login procedure started.
[2016-09-05 14:54:22] [PokemonGoBot] [INFO] [login_successful] Login successful.
[...]

PR #5245, opened 2016-09-06
Cleanup social_handler.py

sohje:
## Short Description:
Social_handler cleanup
- pep8
- cleanup
- readability

### DIFF
--- a/pokemongo_bot/event_handlers/social_handler.py
+++ b/pokemongo_bot/event_handlers/social_handler.py
@@ -1,58 +1,66 @@
 # -*- coding: utf-8 -*-
-from pokemongo_bot.event_manager import EventHandler
-import thread
-import paho.mqtt.client as mqtt
-import Geohash
-import errno
 import json
+import thread
 import time
-
 from socket import error as socket_error
+
+import paho.mqtt.client as mqtt
+
+from pokemongo_bot.event_manager import EventHandler
+
 DEBUG_ON = False
+
+
 class MyMQTTClass:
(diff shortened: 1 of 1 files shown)


Gobberwart:
Why removed "need connect" and "connect again" debug messages?

sohje:
Cause it doesnt bring any clue whats happening. This PR is first step to refactor classes and methods.

Gobberwart:
It certainly does bring a clue to what's happening... if the bot is constantly trying to reconnect, it can help to identify a connection issue.

Also, would like to see debug messages use self.bot.config.debug

`if DEBUG_ON or self.bot.config.debug:

And of course use the logger to display messages.

Gobberwart:
Refer to #5216 why these connect messages are useful

sohje:
We can always check file history ;)

sohje:
If we enable printing via self.bot.config.debug there will be a lot of stuff by api requests, response etc. So i left DEBUG_ON module global var.

Gobberwart:
Agree, but both are needed, so that debug messages are printed either if DEBUG_ON == True, OR config.debug == True. Both should work.

In fact, probably easiest to just add:

`if self.bot.config.debug: DEBUG_ON = True`

to init section of each class

Gobberwart:
Anyway, for now please at least add connect messages back because they are actually helpful in some situations.

Gobberwart:
Thanks. Approved.

sohje:
Ok, thanks. Merge then.

PR #5252, opened 2016-09-06
Fix TypeError: object of type 'NoneType' has no len() on telegram event handler initialization

rbusquet:
If the database is instantiated for the first time, the 'res' object may be None.


solderzzc:
👍

DBa2016:
Thanks @rbusquet, I missed to test after the final change with not-existing tables - and according to docs, `fetchone()` returns "empty list", not "None" :(

PR #5253, opened 2016-09-06
Discord support

askovpen:
ready to use

### DIFF
--- a/pokemongo_bot/event_handlers/discord_handler.py
+++ b/pokemongo_bot/event_handlers/discord_handler.py
@@ -1,0 +1,135 @@
+# -*- coding: utf-8 -*-
+from pokemongo_bot.event_manager import EventHandler
+from pokemongo_bot.base_dir import _base_dir
+import json
+import os
+import time
+import discord_simple
+import thread
+import re
+from pokemongo_bot.datastore import Datastore
+import pprint
+
+
+DEBUG_ON = False
+
+class FileIOException(Exception):
+    pass
+
+class DiscordClass:
+
+    update_id = None
+
+
+    def __init__(self, bot, master, pokemons, config):
+        self.bot = bot
+        self.pokemons = pokemons
+        self._dbot = None
+        self.config = config
+
+    def sendMessage(self, to=None, parse_mode='Markdown', text=None):
+        self._dbot.send_message(to, text)
+
+    def connect(self):
+        self._dbot = discord_simple.Bot(self.bot.config.discord_token,on_message=self.on_message)
+
+    def _get_player_stats(self):
+        web_inventory = os.path.join(_base_dir, "web", "inventory-%s.json" % self.bot.config.username)
+        try:
+            with open(web_inventory, "r") as infile:
+                json_inventory = json.load(infile)
+        except ValueError as exception:
+            self.bot.logger.info('[x] Error while opening inventory file for read: %s' % exception)
+            json_inventory = []
+        except:
+            raise FileIOException("Unexpected error reading from {}".format(web_inventory))
+        return next((x["inventory_item_data"]["player_stats"]
+                     for x in json_inventory
+                     if x.get("inventory_item_data", {}).get("player_stats", {})),
+                    None)
+
+    def send_player_stats_to_chat(self, chat_id):
+        stats = self._get_player_stats()
+        if stats:
+            with self.bot.database as conn:
+                cur = conn.cursor()
+                cur.execute("SELECT DISTINCT COUNT(encounter_id) FROM catch_log WHERE dated >= datetime('now','-1 day')")
+                catch_day = cur.fetchone()[0]
+                cur.execute("SELECT DISTINCT COUNT(pokestop) FROM pokestop_log WHERE dated >= datetime('now','-1 day')")
+                ps_day = cur.fetchone()[0]
+                res = (
+                    "*"+self.bot.config.username+"*",
+                    "_Level:_ "+str(stats["level"]),
+                    "_XP:_ "+str(stats["experience"])+"/"+str(stats["next_level_xp"]),
+                    "_Pokemons Captured:_ "+str(stats["pokemons_captured"])+" ("+str(catch_day)+" _last 24h_)",
+                    "_Poke Stop Visits:_ "+str(stats["poke_stop_visits"])+" ("+str(ps_day)+" _last 24h_)",
+                    "_KM Walked:_ "+str("%.2f" % stats["km_walked"])
+                )
+            self.sendMessage(to=chat_id, parse_mode='Markdown', text="\n".join(res))
+        else:
+            self.sendMessage(to=chat_id, parse_mode='Markdown', text="Stats not loaded yet\n")
+
+    def on_message(self,message):
+        if message.content == "/help":
+            res = (
+                "Commands: ",
+                "/info - info about bot"
+            )
+            self.sendMessage(to=str(message.author), parse_mode='Markdown', text="\n".join(res))
+        elif message.content == "/info":
+            self.send_player_stats_to_chat(message.author)
+
+    def run(self):
+      self._dbot.forever_loop()
+
+
+class DiscordHandler(EventHandler):
+    def __init__(self, bot, config):
+        self.bot = bot
+        self.dbot = None
+        self.master = config.get('master', None)
+        if self.master == None:
+            return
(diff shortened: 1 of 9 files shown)


DBa2016:
What does it do? Can you document it in the appropriate file in the docs directory?

askovpen:
While - for testing - not merge :-)

sohje:
https://discordapp.com/ ?

DBa2016:
@sohje - I know roughly what discord is, but what does it do here? Notification only? Controlling the bot?...

askovpen:
@DBa2016 notify and controlling

pokemongo_bot/event_handlers/discord_handler.py
```diff
@@ -0,0 +1,135 @@
...
+        except ValueError as exception:
+            self.bot.logger.info('[x] Error while opening inventory file for read: %s' % exception)
+            json_inventory = []
+        except:
+            raise FileIOException("Unexpected error reading from {}".format(web_inventory))
+        return next((x["inventory_item_data"]["player_stats"]
+                     for x in json_inventory
+                     if x.get("inventory_item_data", {}).get("player_stats", {})),
+                    None)
+
```
Gobberwart: There is no need to reload inventory from json. Just access the appropriate Inventory (inventory.player_stats from memory)

Gobberwart:
You will end up duplicating a lot of code here to give discord the same functionality as telegram.

This is why I have suggested that the code to generate events and respond to commands be separated from the actual communication method. 

So we could have discordbot and telegrambot that just handle the communication aspect of things, while another class passes information to/from these bots.

Gobberwart:
Love the concept though. We should not have to be forced to use Telegram just to get this functionality.

Would like to see handlers for Telegram, Discord, Email, Other common messaging apps, all sharing the same basic alert generation/command response base code.

pokemongo_bot/event_handlers/discord_handler.py
```diff
@@ -0,0 +1,135 @@
...
+                    "*"+self.bot.config.username+"*",
+                    "_Level:_ "+str(stats["level"]),
+                    "_XP:_ "+str(stats["experience"])+"/"+str(stats["next_level_xp"]),
+                    "_Pokemons Captured:_ "+str(stats["pokemons_captured"])+" ("+str(catch_day)+" _last 24h_)",
+                    "_Poke Stop Visits:_ "+str(stats["poke_stop_visits"])+" ("+str(ps_day)+" _last 24h_)",
+                    "_KM Walked:_ "+str("%.2f" % stats["km_walked"])
+                )
+            self.sendMessage(to=chat_id, parse_mode='Markdown', text="\n".join(res))
+        else:
+            self.sendMessage(to=chat_id, parse_mode='Markdown', text="Stats not loaded yet\n")
```
Gobberwart: This is what I mean about duplicate code. Telegram handler does the same thing. Why not separate out and let both Telegram and Discord (and potentially other handlers in future) share?

pokemongo_bot/event_handlers/discord_handler.py
```diff
@@ -0,0 +1,135 @@
...
+            return True
+        try:
+            oper = re.search(r'operator:([^ ]+)', params).group(1)
+            rule_cp = int(re.search(r'cp:([0-9]+)', params).group(1))
+            rule_iv = float(re.search(r'iv:([0-9.]+)', params).group(1))
+            rule_pkmn = re.search(r'pokemon:([^ ]+)', params).group(1)
+            return rule_pkmn == pokemon and (oper == "or" and (cp >= rule_cp or iv >= rule_iv) or cp >= rule_cp and iv >= rule_iv)
+        except:
+            return False
+
```
Gobberwart: More duplicate code from Telegram

pokemongo_bot/event_handlers/discord_handler.py
```diff
@@ -0,0 +1,135 @@
...
+    def handle_event(self, event, sender, level, formatted_msg, data):
+        if self.dbot is None:
+            try:
+                self.bot.logger.info("Discord bot not running, trying to spin it up")
+                self.dbot = DiscordClass(self.bot, self.master, self.pokemons, self.config)
+                self.dbot.connect()
+                thread.start_new_thread(self.dbot.run)
+            except Exception as inst:
+                self.dbot = None
+                self.bot.logger.error("Unable to spin Telegram bot; master: {}, exception: {}".format(self.master, pprint.pformat(inst)))
```
sohje: Telegram?

pokemongo_bot/event_handlers/discord_handler.py
```diff
@@ -0,0 +1,135 @@
...
+
+
+DEBUG_ON = False
+
+class FileIOException(Exception):
+    pass
+
+class DiscordClass:
+
+    update_id = None
```
sohje: any usage?

pokemongo_bot/event_handlers/discord_handler.py
askovpen: In future can rename "TelegramTask" config keys to some like "alerts" and add key: agents: ["telegram", "discord", "email"]

pokemongo_bot/event_handlers/discord_handler.py
Gobberwart: This is still unnecessarily convoluted. Why not just:

```
def _get_player_stats(self):
    return inventory.player().player_stats
```

pokemongo_bot/event_handlers/discord_handler.py
```diff
@@ -78,7 +71,7 @@ def on_message(self,message):
 
     def run(self):
       self._dbot.forever_loop()
-
+      self = None
```
Gobberwart: What?

Gobberwart:
I have significant concerns about this, in that discord support duplicates a lot of the code in telegram_handler.py. Which means any time we want to change it, we have to change both.

OK to go ahead with this approach for now, but ultimately we should make discord and telegram handlers just look after communicating with discord/telegram, and leave the processing of commands and generation of events to a separate class.

solderzzc:
Same as @Gobberwart , is there better way to organize code?

pokemongo_bot/event_handlers/discord_handler.py
askovpen: You view old code. View last

DBa2016:
I think there are two options:
1) move the logic into a base class, from which telegram, discord, and whateverelse handlers can be derived. The handlers should then **only** include code specific to facilitate the communication between the handler and the messenger.
2) move the logic into "control handler" or similar class which processes events generated/emitted by the particular comunication class.

There might be other options which reduce code duplication, I just don't se them right now :)

Both approaches would even allow web interfaces to run over the same logic... As for email - I don't think this would be particularly useful, but it does not hurt either.

If we don't do this, every bug fix or enhancement we do in a control interface will need to be duplicated (including the logic!) to all other control interfaces.

If no objections are xpressed, I can start implementing option 2 - however, a code freeze for Telegram (and discord) code will be needed for this.

pokemongo_bot/event_handlers/discord_handler.py
askovpen: fixed

askovpen:
approve PR and try make joining to your concept :-)

solderzzc:
@DBa2016 Agreed, let's move on.

solderzzc:
👍

MZorzy:
./setup.sh -r

  File "pokecli.py", line 47, in <module>
    from pokemongo_bot import PokemonGoBot, TreeConfigBuilder
  File "/home/dev/Poke/PokemonGo-Bot/pokemongo_bot/__init__.py", line 22, in <module>
    import cell_workers
  File "/home/dev/Poke/PokemonGo-Bot/pokemongo_bot/cell_workers/__init__.py", line 26, in <module>
    from telegram_task import TelegramTask
  File "/home/dev/Poke/PokemonGo-Bot/pokemongo_bot/cell_workers/telegram_task.py", line 10, in <module>
    from pokemongo_bot.event_handlers import TelegramHandler
  File "/home/dev/Poke/PokemonGo-Bot/pokemongo_bot/event_handlers/__init__.py", line 5, in <module>
    from discord_handler import DiscordHandler
  File "/home/dev/Poke/PokemonGo-Bot/pokemongo_bot/event_handlers/discord_handler.py", line 7, in <module>
    import discord_simple
ImportError: No module named discord_simple

Issue #5249, opened 2016-09-06
'PokemonGoBot' object has no attribute '_player'


cbogithub:
```
Traceback (most recent call last):
  File "pokecli.py", line 803, in <module>
    main()
  File "pokecli.py", line 239, in main
    report_summary(bot)
  File "pokecli.py", line 287, in report_summary
    metrics.capture_stats()
  File "P:\pokemon go\PokemonGo-Bot\pokemongo_bot\metrics.py", line 105, in capture_stats
    self.dust['latest'] = self.bot.stardust 
  File "P:\pokemon go\PokemonGo-Bot\pokemongo_bot\__init__.py", line 72, in stardust
    return filter(lambda y: y['name'] == 'STARDUST', self._player['currencies'])[0]['amount']
AttributeError: 'PokemonGoBot' object has no attribute '_player'
```

alexyaoyang:
@rawgni

avexus:
same problem, could not start bot.

alexyaoyang:
Follow that PR and add that line should work for now.

Gobberwart:
#5254 has been merged into dev, but not master. You will need to apply the fix yourself until merged to master (refer to the PR for details)

rawgni:
@alexyaoyang actually, the problem exist because the user was not able to login. Thus, bot having no player information. However, our code still wants to print report_summary on exit.

mjmadsen:
@rawgni I completely forgot about that. Thank you!

mjmadsen:
@cbogithub @avexus Was there an API error printed before that traceback? 

This whole damn file needs to be cleaned up.

Gobberwart:
Yes it does, but closing because there's fix in dev.

PR #5254, opened 2016-09-06
Temp fix for stardust

alexyaoyang:
#5249.

@rawgni Some more testing before a PR would be marvelous! You may remove this temp fix after you figure out a fix. 👍


avexus:
so, can we do a git pull or should we manually change that file?

Gobberwart:
Looks harmless enough. Merging.

avexus:
I did a git pull on master, working right now.
Thank you guys.

PR #5276, opened 2016-09-08
Reduce "waiting near fort..." spam

Gobberwart:
## Short Description:

When waiting at a lured fort, bot generates repeated "Waiting near fort..." messages until start moving again. Set flag so only appears once until moving status changes.


solderzzc:
👍

PR #5277, opened 2016-09-08
Do you want to open a gitter room for development?

solderzzc:
We do have slack channel, but we've lost millions of chat history, how do you think about this one?
https://gitter.im/pikabot-org/Lobby

### DIFF
--- a/README.md
+++ b/README.md
@@ -17,6 +17,9 @@ The project is currently setup in two main branches:
 - `master` also known as `stable` - The bot 'should' be stable on this branch, and is generally well tested.
 
 ## Support
+### Development Channel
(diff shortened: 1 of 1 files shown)


alexyaoyang:
Great decision! 👍

Gobberwart:
love it. git integration is helpful

Gobberwart:
Would be nice to limit it to approved people only (PokemonGoF members, other approved people) instead of "anyone with a git account".

solderzzc:
@Gobberwart  I have not seen this option on gitter...

mjmadsen:
I'd prefer we don't publicly advertise this chat. Instead, invite those who submit PRs and those that are already admins.

PR #5282, opened 2016-09-08
pokecli.py wait_time not defined

Gobberwart:
## Short Description:

NoPlayerPositionSetException tries to sleep for "wait_time" but "wait_time" isn't defined. Causes bot crash.

- #5280

### DIFF
--- a/pokecli.py
+++ b/pokecli.py
@@ -213,6 +213,7 @@ def main():
                 )
                 time.sleep(36000)
             except NoPlayerPositionSetException:
+                wait_time = config.reconnecting_timeout * 60
                 bot.event_manager.emit(
                     'api_error',
                     sender=bot,
--- a/pokecli.py
+++ b/pokecli.py
@@ -145,11 +145,11 @@ def main():
         finished = False
 
         while not finished:
-            try:
+           wait_time = config.reconnecting_timeout * 60
+           try:
                 bot = initialize(config)
                 bot = start_bot(bot, config)
                 config_changed = check_mod(config_file)
-                wait_time = config.reconnecting_timeout * 60
 
                 bot.event_manager.emit(
                     'bot_start',
--- a/pokecli.py
+++ b/pokecli.py
@@ -145,8 +145,8 @@ def main():
         finished = False
 
         while not finished:
-           wait_time = config.reconnecting_timeout * 60
-           try:
+            wait_time = config.reconnecting_timeout * 60
+            try:
                 bot = initialize(config)
                 bot = start_bot(bot, config)
                 config_changed = check_mod(config_file)
--- a/pokecli.py
+++ b/pokecli.py
@@ -149,6 +149,7 @@ def main():
                 bot = initialize(config)
                 bot = start_bot(bot, config)
                 config_changed = check_mod(config_file)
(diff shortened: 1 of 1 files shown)


mjmadsen:
Been working to extend this and others. Could you move it up a level or two? A few of the below exceptions use it

Gobberwart:
@mjmadsen not sure what you mean? wait_time defined above all the exceptions?

mjmadsen:
Yeah, it's declared in an except, but used in a few excepts out of scope. One of the many problems in this file.

Gobberwart:
Yeah I'm just doing bandaids at the moment. How's that latest commit look?

pokecli.py
```diff
@@ -213,6 +213,7 @@ def get_commit_hash():
                 )
                 time.sleep(36000)
             except NoPlayerPositionSetException:
+                wait_time = config.reconnecting_timeout * 60
```
mjmadsen: We need to move the scope of this. We use it further down the line with different excepts.

Gobberwart:
@mjmadsen Fair enough. I'm just doing bandaids right now, but agree no need to define it multiple times. How's the latest commit look?

solderzzc:
@mjmadsen  how do you think about our new gitter channel ?
https://github.com/PokemonGoF/PokemonGo-Bot#development-channel

Gobberwart:
Oi @solderzzc no thread-jacking! :)

solderzzc:
Another breaking :)

mjmadsen:
@Gobberwart I get it, but I realized we were defining the sleep timer in a  place that can't access the declaration. I've been staring at this file all day. I am working on a PR that breaks the functions up. But one of the big things I noticed is that that variable is defined in an exception and used in other exceptions - thus needs to be defined before all of all exceptions that use it.

Gobberwart:
@mjmadsen I completely understand what you're saying. I've defined wait_timeout before the try block starts. Does that work for you?

mjmadsen:
My notes from looking at this file today:
We need to break non essential config stuff to a config handler file (and clean up our general config checks - move checks closer to load).
Break our try statements into smaller chunks - we have absurd try blocks.
Adapt python idioms (we don't apply the best methods on many file handling bits). 
Use exceptions properly to give us the correct traceback. 
Break up certain functions to allow better try/except blocks

Gobberwart:
@mjmadsen Agree 100%. This is just a bandaid fix though. OK to merge?

mjmadsen:
:+1:

mjmadsen:
I'll merge after CI passes (though I know it will). Sorry to be dismal, I've spent too much time looking at pokecli.py being frustrated today. I have many wants from looking at that file.

mjmadsen:
Showing errors:
FILE: /home/travis/build/PokemonGoF/PokemonGo-Bot/pokecli.py
FAILED pylint inspection: E:173, 0: unindent does not match any outer indentation level (syntax-error)

Gobberwart:
@mjmadsen Yep, but that's nothing to do with this PR. Seems config examples are busted.

Gobberwart:
@mjmadsen OK fixed

mjmadsen:
:+1:

PR #5289, opened 2016-09-08
PokemonGo-Bot-Install

LuckyMe4Evers:
## Short Description:
PokemonGo-Bot-Install
- changed path so requirements install from right path
- updated installation manual


solderzzc:
👍

PR #5292, opened 2016-09-08
Decouple hb from step walker

sohje:
## Short Description:
Decouple hb from step walker due to #5243
- Fix

### DIFF
--- a/pokemongo_bot/walkers/step_walker.py
+++ b/pokemongo_bot/walkers/step_walker.py
@@ -47,8 +47,6 @@ class StepWalker(object):
                                           "distance": "",
                                           "distance_unit": ""})
 
-        self.bot.heartbeat()
-
         return self.is_arrived()
 
     def is_arrived(self):


th3w4y:
@sohje what is calling heartbeat now?

sohje:
Heartbeat works in separate thread now, with timer and another checks, like block hb from sniping.

th3w4y:
@sohje  That part i noticed, what is did not noticed yet is when is this called except .login()
>Heartbeat works in separate thread now, with timer and another checks, like block hb from sniping.

sohje:
@th3w4y 
```python
threading.Timer(self.heartbeat_threshold, self.heartbeat).start()
```

th3w4y:
@sohje  so it is actually this change that made the bot walk faster not our StepWalker PolylineWalker! and not the speeds match the reality!

sohje:
@th3w4y How can this be possible? We sending requests as usual.

th3w4y:
@sohje but that is a good thing that we can respect the speed settings...

th3w4y:
@sohje Looks perfect to me

like I was saying previous comments this actually fixes a real bug
 (a new thread would have been spawn on every bot.tick())

solderzzc:
👍

Issue #5275, opened 2016-09-08
Optimizer exchange the only pokemon of a family


liejuntao001:
### Expected Behavior

Using PokemonOptimizer and it should keep the at least one pokemon of a family

### Actual Behavior

It exchange the only one of a family, for a upgraded family.
example: there are many Magnetites and only one Magneton, it exchange the Magneton

### Your FULL config.json (remove your username, password, gmapkey and any other private info)

                {
                        "type": "PokemonOptimizer",
                        "config": {
                                "enabled": true,
                                "transfer": true,
                                "transfer_wait_min": 4,
                                "transfer_wait_max": 6,
                                "evolve": true,
                                "evolve_time": 25,
                                "evolve_for_xp": true,
                                "evolve_only_with_lucky_egg": false,
                                "evolve_count_for_lucky_egg": 72,
                                "may_use_lucky_egg": true,
                                "upgrade": true,
                                "upgrade_level": 60,
                                "min_slots_left": 5,
                                "groups": {
                                        "gym": ["Dragonite", "Snorlax", "Lapras", "Arcanine"],
                                        "my_love": ["Pikachu"],
                                        "vip": ["Exeggutor", "Blastoise", "Vaporeon", "Gyarados", "Flareon", "Venusaur", "Dragonite", "Snorlax", "Lapras", "Arcanine"]
                                },
                                "keep": [
                                        {
                                                "mode": "by_family",
                                                "top": 1,
                                                "sort": [{"iv": 0.9}],
                                                "evolve": true,
                                                "upgrade": false
                                        },
                                        {
                                                "mode": "by_family",
                                                "top": 1,
                                                "sort": [{"ncp": 0.9}],
                                                "evolve": true,
                                                "upgrade": false
                                        },
                                        {
                                                "mode": "by_family",
                                                "top": 1,
                                                "sort": ["cp"],
                                                "evolve": false,
                                                "upgrade": false
                                        },
                                        {
                                                "mode": "by_family",
                                                "names": ["vip"],
                                                "top": 3,
                                                "sort": [{"iv": 0.9}, {"ncp": 0.9}],
                                                "evolve": true,
                                                "upgrade": false
                                        },
                                        {
                                                "mode": "by_family",
                                                "names": ["my_love"]
                                        }
                                ]
                        }
                },
### Output when issue occurred

[...]

Issue #5235, opened 2016-09-06
Limit PokemonOptimizer logs only after catching a Pokemon


hobbes3:
### Short Description

Make the `PokemonOptimizer` display log only after catching a Pokemon.

### Possible solution

`PokemonOptimizer` currently spams the logs every second. Regardless of whether the bot catches a Pokemon or not. For example:

    [2016-09-06 07:48:44] [FollowPath] [INFO] Walking from (37.79268757281985, -122.39339298324255, 8.030725657710692) to (37.7930338, -122.3929975, 8.0), distance left: (0.05 km) ..
    [2016-09-06 07:48:45] [UpdateLiveStats] [INFO] xxx | Uptime : 0:33:37 | Earned 4,801 Stardust | +9,195 XP | 16,412 XP/h | Visited 46 stops
    [2016-09-06 07:48:45] [PokemonOptimizer] [INFO] Transferring 0 Pokemon
    [2016-09-06 07:48:45] [PokemonOptimizer] [INFO] Skipping evolution step. Not enough Pokemon to evolve with lucky egg: 4/80
    [2016-09-06 07:48:45] [PokemonOptimizer] [INFO] Upgrading 0 Pokemon [396113 stardust]
    [2016-09-06 07:48:46] [FollowPath] [INFO] Walking from (37.79271549998604, -122.39336821335196, 8.220596891714258) to (37.7930338, -122.3929975, 8.0), distance left: (0.05 km) ..
    [2016-09-06 07:48:46] [PokemonOptimizer] [INFO] Transferring 0 Pokemon
    [2016-09-06 07:48:46] [PokemonOptimizer] [INFO] Skipping evolution step. Not enough Pokemon to evolve with lucky egg: 4/80
    [2016-09-06 07:48:46] [PokemonOptimizer] [INFO] Upgrading 0 Pokemon [396113 stardust]
    [2016-09-06 07:48:47] [FollowPath] [INFO] Walking from (37.79274049309204, -122.39333704535689, 8.275425897523267) to (37.7930338, -122.3929975, 8.0), distance left: (0.04 km) ..
    [2016-09-06 07:48:48] [PokemonOptimizer] [INFO] Transferring 0 Pokemon
    [2016-09-06 07:48:48] [PokemonOptimizer] [INFO] Skipping evolution step. Not enough Pokemon to evolve with lucky egg: 4/80
[...]

PR #5308, opened 2016-09-09
Improvements Pokemon Optimizer

Anakin5:
Update doc for #5275
Reduce log when nothing to do for #5235

### DIFF
--- a/docs/pokemon_optimizer.md
+++ b/docs/pokemon_optimizer.md
@@ -342,11 +342,12 @@ Every Pokemon not selected is candidate for transfer.
 | `mode`    | `"by_pokemon"`, `"by_family"`, `"overall"` | `"by_family"` |
 
 The `mode` define how the Pokemon Optimizer is going to apply the rule.
-- `"by_pokemon"` will apply the rule for each Pokemon.
-<br>In that mode, each Pokemon will be compared to other Pokemon of the same name.
+- `"by_pokemon"` will apply the rule for each Pokemon. That is for every types of Pokemon individually.
+<br>In that mode, each Pokemon will be compared to other Pokemon of the same type.
 <br>This is the most conservative mode and the one that will result in the most number of Pokemon kept.
 
 - `"by_family"` will apply the rule for each Pokemon family.
+A family is the group of a Pokemon with all its evolutions.
 <br>In that mode, each Pokemon will be compared to other Pokemon of the same family.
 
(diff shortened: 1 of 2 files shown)


docs/pokemon_optimizer.md
```diff
@@ -342,11 +342,12 @@ Every Pokemon not selected is candidate for transfer.
 | `mode`    | `"by_pokemon"`, `"by_family"`, `"overall"` | `"by_family"` |
 
 The `mode` define how the Pokemon Optimizer is going to apply the rule.
-- `"by_pokemon"` will apply the rule for each Pokemon.
-<br>In that mode, each Pokemon will be compared to other Pokemon of the same name.
+- `"by_pokemon"` will apply the rule for each Pokemon. That is for every types of Pokemon individually.
```
Gobberwart: Not sure this is clear. 

Could just be "will apply the rule for each individual pokemon type (eg. "Rattata", "Pidgey", "Clefairy" etc.)"

pokemongo_bot/cell_workers/pokemon_optimizer.py
```diff
@@ -416,48 +416,60 @@ def get_evolution_plan(self, family_id, family_list, try_evolve, try_upgrade, ke
...
+                        skip_evolve = True
+                        self.emit_event("skip_evolve",
+                                        formatted="Waiting for more Pokemon to evolve with lucky egg: %s/%s" % (evolve_xp_count, self.config_evolve_count_for_lucky_egg))
+                else:
+                    self.use_lucky_egg()
 
-        if not skip_evolve:
-            self.logger.info("Evolving %s Pokemon (the best)", len(evolve))
+            if not skip_evolve:
+                if evolve_count > 0:
```
Gobberwart: if evolve_count > 0 and not skip_evolve:

docs/pokemon_optimizer.md
Anakin5: ok, but the example would be misleading since it does not contains 2 pokemon of the same family.

pokemongo_bot/cell_workers/pokemon_optimizer.py
Anakin5: the not skip_evolve include more code

Gobberwart: OK yep

alexyaoyang:
👍

Issue #5266, opened 2016-09-07
Failing to discard a Potion crashes the bot


tiagoAnon:
===============ISSUE SECTION===============

The bot tried to discard a potion and the message "[ItemRecycler] [INFO] Failed to discard Potion" appeared, but the bot crashed right after. It happened quite a few times, it almost seems like Niantic is trying to filter bot behavior...

### Expected Behavior
The bot should not crash.

### Actual Behavior
The bot crashes.

### Output when issue occurred
`Traceback (most recent call last):
  File "pokecli.py", line 803, in <module>
    main()
  File "pokecli.py", line 161, in main
    bot.tick()
  File "/Users/KTachyon/Development/Independent/PoGoBot/PokemonGo-Bot/pokemongo_bot/__init__.py", line 685, in tick
    if worker.work() == WorkerResult.RUNNING:
  File "/Users/KTachyon/Development/Independent/PoGoBot/PokemonGo-Bot/pokemongo_bot/cell_workers/recycle_items.py", line 170, in work
    this_worker_result =  self.recycle_excess_category_max(self.max_potions_keep, [101,102,103,104])
  File "/Users/KTachyon/Development/Independent/PoGoBot/PokemonGo-Bot/pokemongo_bot/cell_workers/recycle_items.py", line 210, in recycle_excess_category_max
    if inventory.items().get(item[0]).recycle(item[1]) == WorkerResult.ERROR:
  File "/Users/KTachyon/Development/Independent/PoGoBot/PokemonGo-Bot/pokemongo_bot/inventory.py", line 219, in recycle
    raise Exception('Tried to remove more {} than you have'.format(self.name))
Exception: Tried to remove more Potion than you have
[2016-09-07 17:56:36] [MainThread] [sentry.errors] [ERROR] Sentry responded with an error: 'ascii' codec can't decode byte 0x9c in position 1: ordinal not in range(128) (url: https://app.getsentry.com/api/90254/store/)
Traceback (most recent call last):
  File "/usr/local/lib/python2.7/site-packages/raven/transport/threaded.py", line 174, in send_sync
    super(ThreadedHTTPTransport, self).send(data, headers)
  File "/usr/local/lib/python2.7/site-packages/raven/transport/http.py", line 47, in send
    ca_certs=self.ca_certs,
  File "/usr/local/lib/python2.7/site-packages/raven/utils/http.py", line 66, in urlopen
    return opener.open(url, data, timeout)
[...]

PR #5312, opened 2016-09-09
UnicodeDecodeError catch

Gobberwart:
## Short Description:

If non-unicode string passed to parse_unicode_str, bot crashes. Previously was catching only UnicodeEncodeError, added UnicodeDecodeError as well.

- #5266


solderzzc:
👍  uh, UTF8 again

PR #5314, opened 2016-09-09
Telegram cleanup/separation

Gobberwart:
## Description:

I am working on separating telegram/discord communication functions from those methods which are duplicated.

For example, "send_player_stats_to_chat" method performs two functions. 
1. Gather stats from inventory and format them
2. Send to telegram/discord

My thinking is that the first item should be handled by a shared class, while only anything telegram/discord specific should be left in the respective handlers.

In this example, telegram sends stats and a location, but discord only sends stats, so both telegram and discord can share the "get_player_stats" function from the new class, but Telegram can add the extra SendLocation call.

I am putting up this PR for comments and feedback so I know the concept is correct. Some code is unused but there for future progress. Very interested in any suggestions.

I hope this makes sense, but if not please ask for clarification.

### DIFF
--- a/docs/manual_installation.md
+++ b/docs/manual_installation.md
@@ -16,7 +16,7 @@ sudo apt-get install build-essential autoconf libtool pkg-config make python2.7-
 ####
 if you are on a different Linux OS you maybe have to adapt things like:
 
-- package mananger (for example yum instead of apt-get)
+- package manager (for example yum instead of apt-get)
 - package names
 
 #####Mac
@@ -25,9 +25,9 @@ brew install --devel protobuf
 brew install  autoconf libtool pkg-config wget git
 ```
 ####Mac + Linux installation
-make shure you installed everything above
+make sure you installed everything above
 
-- get pip for pyton2.7
+- get pip for python2.7
 ```bash
 wget https://bootstrap.pypa.io/get-pip.py
 python2.7 get-pip.py
@@ -47,9 +47,10 @@ cd PokemonGo-Bot
 ```bash
 pip install virtualenv
 virtualenv .
+source bin/activate
 ```
 ####
-- install the requirements and get the needen encryption.so
+- install the requirements and get the needed encryption.so
 
 (we move `encrypt.so` to the root folder of the Bot so no need to edit the config regarding that)
 ```bash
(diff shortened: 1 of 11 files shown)


Gobberwart:
At this stage, I have only replaced the "send_player_stats_to_chat" method. Feedback please.

rawgni:
@Gobberwart one problem might be the formatting. The format that works for telegram to display nicely, might not work for discord or other app, and vice versa. If somehow you are able to figure out a common format, might want to refactor the updatelivestat at the same time to call the same function.

Gobberwart:
@rawgni Definitely going to be an issue, although so far I'm not seeing much evidence of formatting differences. The "SendMessage" method is different (Telegram supports "Markdown", "HTML" etc. while Discord doesn't) but that's fine.

Ultimately, the ChatHandler class will perform two functions:

1. Data collection
2. Event processing (resulting in data collection), eg. "/info" - returns player stat data, "/top n cp" returns list of top n cp pokemon etc.

Data returned must be in a format that can be used by any class.

At this stage, just want to make sure doing something like this is supported as a general concept (and obviously I want someone to tell me if I'm doing it wrong)

Gobberwart:
I've moved the common events out of their respective handlers (discord/telegram) and into chat_handler.py. Tested working well, except pokemon_caught event which I'm debugging.

Gobberwart:
OK, I'm a bit stuck on pokemon_caught event, specifically, this:

```
                if (not "operator" in trigger or trigger["operator"] == "and") and \
                        data["cp"] >= trigger["cp"] and data["iv"] >= trigger["iv"] or \
                        ("operator" in trigger and trigger["operator"] == "or" and \
                        (data["cp"] >= trigger["cp"] or data["iv"] >= trigger["iv"])):
                    msg = "Caught {} CP: {}, IV: {}".format(data["pokemon"], data["cp"], data["iv"])
                else:
                    return
```

What does that condition do and why is it (sometimes) returning false? Anyone have any ideas?

Gobberwart:
Ugh, I'm being thick. Never mind.

Gobberwart:
OK events are working fine in Telegram. Haven't tested Discord yet (I haven't set up a Discord bot yet). 

Got to stop for a while, but feel free to leave any feedback.

Gobberwart:
OK does not work with discord... yet. It'll take a while (2 days ish) for me to finish this due to family commitments. If anyone wants to run with it in the meantime, go ahead.

Gobberwart:
I've had a major computer crash and need to reinstall everything before I can do any more work on anything. Might be another day or two depending on time available.

Gobberwart:
Closing this. No longer needed.

Issue #5318, opened 2016-09-09
Can't start bot after latest DEV update


beltoft:
### Expected Behavior
Bot should start

### Actual Behavior
remote: Counting objects: 5, done.
remote: Total 5 (delta 4), reused 4 (delta 4), pack-reused 0
Unpacking objects: 100% (5/5), done.
From https://github.com/PokemonGoF/PokemonGo-Bot
   732463c..56665a2  dev        -> origin/dev
Some of the required packages are not found / have different version.
Run: ./setup.sh -u

I ran ./setup.sh -u and it finish successfully without any errors but i still get the above error when starting bot  

OS: 
Ubuntu 14.04 LTS

Branch: 
DEV 

Python Version: 
Python 2.7.6
 
Any other relevant files/configs (eg: path files) 
Output from setup.sh -u

```


./setup.sh  -u
Already up-to-date.
fatal: reference is not a tree: 8dcbe285c4df94dc871ea655a8a9ceabf5133881
Unable to checkout '8dcbe285c4df94dc871ea655a8a9ceabf5133881' in submodule path 'web'
Entering 'web'
From https://github.com/OpenPoGo/OpenPoGoWeb
 * branch            master     -> FETCH_HEAD
Already up-to-date.
Requirement already up-to-date: numpy==1.11.0 in ./lib/python2.7/site-packages (from -r requirements.txt (line 1))
Requirement already up-to-date: networkx==1.11 in ./lib/python2.7/site-packages (from -r requirements.txt (line 2))
Obtaining pgoapi from git+https://github.com/joelgreen/pgoapi.git/@fa49ada8570b0817c5a73b9315ba010fa7a47874#egg=pgoapi (from -r requirements.txt (line 3))
  Skipping because already up-to-date.
Requirement already up-to-date: geopy==1.11.0 in ./lib/python2.7/site-packages (from -r requirements.txt (line 4))
Requirement already up-to-date: geographiclib==1.46.3 in ./lib/python2.7/site-packages (from -r requirements.txt (line 5))
Requirement already up-to-date: protobuf==3.0.0b4 in ./lib/python2.7/site-packages (from -r requirements.txt (line 6))
Requirement already up-to-date: requests==2.10.0 in ./lib/python2.7/site-packages (from -r requirements.txt (line 7))
Requirement already up-to-date: requests-mock==1.0.0 in ./lib/python2.7/site-packages (from -r requirements.txt (line 8))
[...]

PR #5323, opened 2016-09-09
run.sh can't handle pip requiremens using ">"

JaapMoolenaar:
## Short Description:
"discord_simple >" works for pip but run.sh parses the requirements file and matches more strictly.
Down the line, changing the parser would be best, imho.

#5318


JaapMoolenaar:
Duplicate, sorry.

PR #5345, opened 2016-09-09
Default tasks

### LINKED ISSUE #5227: [FR] are you seriously not going to collect level up rewards ?
### Short Description

There are "tasks" that shouldn't be a task, ie. CollectLevelUpReward (CLUR), CompleteTutorial.
Currently , if I get rid of the CLUR task from the config file, it will never run. 
However, does it really make sense to not collect level up reward and keep on catching pokemon?

How is it possible to not complete tutorial and be level 10? 

### Possible solution

Make it part of the system and not configurable.

### How it would help others

Less stuff in config file

DeXtroTip:
## Short Description:

Made DefaultTasks where tasks that must not be disabled by user (CollectLevelUpReward and CompleteTutorial) are now. They run every tick as before.
When no nickname is set it it will try to use account username. When it is not available it will generate a new one based on the first 10 chars of the on tried plus 5 random digits. Not sure if I should include this or not. Also, it only tries generated one once, should it keep generating and trying new ones till it accepts?
- Resolves #5227


configs/config.json.cluster.example
```diff
@@ -22,23 +22,6 @@
         "type": "HandleSoftBan"
       },
       {
-        "type": "CompleteTutorial",
```
mjmadsen: We still need to get a nickname and team from somewhere.

configs/config.json.cluster.example
```diff
@@ -22,23 +22,6 @@
...
-          "nickname": "",
-          "// 0 = No Team, 1 = Blue, 2 = Red, 3 = Yellow": "",
-          "team": 0
-        }
-      },
-      {
-        "type": "CollectLevelUpReward",
-        "config": {
-          "collect_reward": true,
-          "level_limit": -1
```
mjmadsen: This can help a bot stop at a certain level. Would need to be readded somewhere.

configs/config.json.cluster.example
```diff
@@ -208,17 +191,14 @@
...
     "pokemon_bag": {
-      "// if 'show_at_start' is true, it will log all the pokemons in the bag (not eggs) at bot start": {},
       "show_at_start": true,
-      "// if 'show_count' is true, it will show the amount of each pokemon (minimum 1)": {},
       "show_count": false,
-      "// if 'show_candies' is true, it will show the amount of candies for each pokemon": {},
       "show_candies": false,
-      "// 'pokemon_info' parameter define which info to show for each pokemon": {},
-      "// the available options are": {},
-      "// ['cp', 'iv_ads', 'iv_pct', 'ivcp', 'ncp', 'level', 'hp', 'moveset', 'dps']": {},
```
mjmadsen: I don't have a problem with in example comments.

pokemongo_bot/cell_workers/complete_tutorial.py
```diff
@@ -1,214 +0,0 @@
-import random
```
mjmadsen: The work should still be in separate files.

configs/config.json.cluster.example
mjmadsen: I see we have it in another place now. That's fine with me. 

Side note, we should probably make sure the nickname is accepted.

DeXtroTip: Make sure how? It generates a new one if the first is not accepted, should it keep generating and trying till one is accepted?

configs/config.json.cluster.example
DeXtroTip: It's in another place now.

configs/config.json.cluster.example
DeXtroTip: Since it was in docs i though it wouldn't be needed, I'll add it back.

pokemongo_bot/cell_workers/complete_tutorial.py
DeXtroTip: So should I keep the files in that folder? Or a new one, since it works a little different.

mjmadsen: IMO, we have "workers" that do specific functions in one place, then our "tasks" that have settings and choose which things we send to the "workers'. Say if a pokemon should be ignored, our task doesn't send it to the worker.

configs/config.json.cluster.example
mjmadsen: Not sure how others feel, but I like in-example.config explanations. Most don't head over to our configuration_files.md and look at what is going on. It's purely a preference.

configs/config.json.cluster.example
Gobberwart: Yes :)

configs/config.json.cluster.example
Gobberwart: I like them, but I absolutely loathe them when people leave them in config.json and then paste the whole file with comments etc :)

configs/config.json.pokemon.example
```diff
@@ -218,17 +201,14 @@
...
-      "// if 'show_candies' is true, it will show the amount of candies for each pokemon": {},
       "show_candies": false,
-      "// 'pokemon_info' parameter define which info to show for each pokemon": {},
-      "// the available options are": {},
-      "// ['cp', 'iv_ads', 'iv_pct', 'ivcp', 'ncp', 'level', 'hp', 'moveset', 'dps']": {},
       "pokemon_info": ["cp", "iv_pct"]
     },
+    "nickname": "",
+    "team": 0,
+    "level_limit": -1,
```
Gobberwart: best to keep related stuff together rather than just putting at base level in config. At a glance, nobody would have any clue what "level_limit" means.

Why not:

```
      "CompleteTutorial": {		
          "// set a name": "",		
          "nickname": "",		
          "// 0 = No Team, 1 = Blue, 2 = Red, 3 = Yellow": "",		
          "team": 0		
      },		
     { "CollectLevelUpReward": {		
          "level_limit": -1		
     },

acurioustale:
Moving nickname and team for CompleteTutorial outside of tasks would be beneficial for multibot setups, because you can move them to auth.config instead of keeping a separate config.json for each bot.

configs/config.json.pokemon.example
DeXtroTip: But having this outside of tasks? We can group it, but not sure how to name it also.

pokemongo_bot/cell_workers/complete_tutorial.py
DeXtroTip: Yeah, I see what you're saying, but I'm not entirely sure how I should do it

configs/config.json.cluster.example
DeXtroTip: Yeah, that's true but I'll add it back :)

configs/config.json.cluster.example
YvesHenri: The team ID could just be a randomized also

DeXtroTip: Not sure about the team ID since it's not required to pick a team, you can be level 30 without a team. And some people may not want to pick any team, I think

configs/config.json.pokemon.example
cmezh: Yep, it would be better to keep the groups of parameters. You can also keep the names of dicts the same as tasks names like @Gobberwart offered.

pokemongo_bot/cell_workers/complete_tutorial.py
cmezh: @DeXtroTip Place different task classes in different files and then import them into the parent class.

Issue #5299, opened 2016-09-08
Major issue with heartbeat, random crashes


Anakin5:
There is a major issue with the heartbeat being moved to a separate thread.
It updates inventory, player data and fort timeout concurrently to other tasks. It causes random crashes.

Simple scenario:
````
Main Thread: Transfer Pokemon
Timeout Thread: Refresh Inventory
Main Thread: inventory.remove(pokemon_id) <- exception
```

I have trigger several scenario like that so far. The update of data structures should be moved out of the heartbeat.

Gobberwart:
What's your proposed solution?

Anakin5:
Must be done in main thread to avoid modifying objects that a tasks is currently using. But does the phone application do that anyway ? (requesting inventory, player data etc.. regularly)

Gobberwart:
Not sure if the app does it, but we need to in order to make sure our information is 100% accurate and avoid other silly errors (like trying to discard stuff we don't have, trying to nickname a pokemon that isn't there etc.)

If every single piece of code that does something inventory-related could be guaranteed to always update inventory/stats, wouldn't be an issue, but... I wouldn't make that guarantee :)

rawgni:
#4605 show which api is being requested for any request

Gobberwart:
```
    def add(self, pokemon):
        if pokemon.unique_id <= 0:
            raise ValueError("Can't add a pokemon without id")
        if pokemon.unique_id in self._data:
            raise ValueError("Pokemon already present in the inventory")
        self._data[pokemon.unique_id] = pokemon
```

This now occurs because of heartbeat being moved to thread, HOWEVER... why on earth are we raising an exception/crashing because of this? Should just print a debug message and move on.

Gobberwart:
Ugh, yes this is a major headache. Why did heartbeat get threaded? It's not as if it was significantly holding up bot execution or anything.

rawgni:
@Gobberwart since heartbeat should be sent periodically, and some of the tasks might prevent this to be done. Say, if you are evolving a whole bunch of pokemon, the heartbeat will not be called until all the tasks are done.

Gobberwart:
@rawgni ok cool thank you. So... why does that matter? Trying to replicate the app behaviour?

rbusquet:
I'm having lots of crashes because of this, I believe. 
Mainly pokemon already in inventory and trying to remove an item I don't have. Is this problem related to this issue?

[...]

PR #5347, opened 2016-09-09
Fix heartbeat inventory add crash

Gobberwart:
## Short Description:

Since heartbeat threaded, inventory refresh may occur asynchronously and may cause crash (ie. trying to add pokemon that's already there). Moved heartbeat to tick instead so should only occur when the bot isn't doing other things.

**Important note** I have included a threshold in this so that inventory_refresh occurs no more frequently than every 10 seconds, same as heartbeat. If anything, this **decreases** api calls, as tick might occur only every few minutes if doing a lot of task work.

- #5299

### DIFF
--- a/pokemongo_bot/__init__.py
+++ b/pokemongo_bot/__init__.py
@@ -677,6 +677,8 @@ class PokemonGoBot(object):
             if timeout >= now:
                 self.fort_timeouts[fort["id"]] = timeout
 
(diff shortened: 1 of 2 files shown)


Gobberwart:
OK, after feedback in chat, I've added a threshold to prevent inventory_refresh occurring under 10 seconds.

cmezh:
:+1:

PR #5364, opened 2016-09-10
Incense bugfixing

davidakachaos:
## Short Description:
Trying my first at catching incense Pokemon. Rebased on current dev.

- Catch Pokemon attracted by incense
- Check if an incense is active on a player
- Updated documentation for this change

Please review my changes and tell me if I broke something

### DIFF
--- a/pokemongo_bot/cell_workers/use_incense.py
+++ b/pokemongo_bot/cell_workers/use_incense.py
@@ -21,7 +21,7 @@ class UseIncense(BaseTask):
       }
     
     def _have_applied_incense(self):
-      for applied_item in inventory.applied_items:
+      for applied_item in inventory.applied_items().all():
         if applied_item.expire_ms > 0:
           mins = format_time(item.expire_ms * 1000)
           self.logger.info("Not applying incense, currently active: %s, %s minutes remaining", applied_item.item.name, mins)
--- a/pokemongo_bot/cell_workers/use_incense.py
+++ b/pokemongo_bot/cell_workers/use_incense.py
@@ -3,6 +3,7 @@ from pokemongo_bot.base_task import BaseTask
 from pokemongo_bot.worker_result import WorkerResult
 from pokemongo_bot.item_list import Item
 from pokemongo_bot import inventory
+from utils import format_time
 
 class UseIncense(BaseTask):
     SUPPORTED_TASK_API_VERSION = 1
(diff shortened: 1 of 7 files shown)


pokemongo_bot/cell_workers/use_incense.py
```diff
@@ -12,14 +13,23 @@ def initialize(self):
...
+
       self.types = {
         401: "Ordinary",
         402: "Spicy",
         403: "Cool",
         404: "Floral"
       }
-      
+
+    def _have_applied_incense(self):
```
sohje: Duplicate code from catch_pokemon.py ?

pokemongo_bot/cell_workers/use_incense.py
```diff
@@ -30,35 +40,36 @@ def _get_type(self):
...
+
+      self._update_inventory()
+      if self._has_count() and self.use_incense:
+        return True
 
     def work(self):
       if self._should_run():
         self.start_time = time.time()
-        type = self._get_type()        
+        type = self._get_type()
```
sohje: type is a built-in func. Really bad idead to assign something to it.

pokemongo_bot/inventory.py
```diff
@@ -319,6 +319,85 @@ def has_space_for_loot(cls):
...
+                ret = item
+                self.parse(item)
+
+        return ret
+
+    def __str__(self):
+        return self.name
+
+
+class AppliedItems(_BaseInventoryComponent):
```
sohje: There is already:
```python
class Items(_BaseInventoryComponent):
    TYPE = 'item'
    ID_FIELD = 'item_id'
    STATIC_DATA_FILE = os.path.join(_base_dir, 'data', 'items.json')
```
Why did you add another 2 classes? Just extend items class. Also there is class Item(object)

pokemongo_bot/inventory.py
```diff
@@ -1415,6 +1499,13 @@ def items():
     """
     return _inventory.items
 
+def applied_items():
```
sohje: There is cached a _inventory.items object.

davidakachaos:
Thanks for all the feedback @sohje 
I will have another go at it!

pokemongo_bot/inventory.py
davidakachaos: AppliedItem(s) is a seperate class in pgoapi, it are items that are in use by the player with a countdown. Like a LuckyEgg and a Incense. 
So it isn't the same as a incense in your inventory, it's a item that is being used.
my knowledge of Python is minimal (I learn as I go) so I just created two classes with Item and Items as example.

pokemongo_bot/inventory.py
davidakachaos: Yes, but here I'm not interested in the items in the bag. I want to get the items (and time left on those items) that are in use. Am I doing it wrong?

pokemongo_bot/cell_workers/use_incense.py
davidakachaos: Yes, how would I dry this up?

solderzzc:
👍

javajohnHub:
Cant wait to try this out

Issue #5367, opened 2016-09-10
Proposal for ConcurentWalking "Rickshaw" sharing between the tasks that have to walk


th3w4y:
**Below proposal is related to all tasks that are changing bot.position [via StepWalker]**

Currently we achieve already a great deal (almost everywhere) to standardise the bot.position changes to only use the StepWalker.step() [be it the PolylineWalker(StepWalker)]

**Assumptions:**
- 90% of users don't deviate much from the documented config.json.example
- We expect all the walking Tasks to befriend each other!
- Tasks have a .work() they calculate distances and do stuff here,  they create a step_walker = StepWalker() object and when they call step_walker.step() they expect True if bot has reached  the requested destination of False if walk is in progress...
- We don't want to spend time explaining the users what are valid walking task combinations and what not! or to explain them in what order the tasks should be configured....
- Things should just work!

**Identified Issues:**
- Different Tasks Concurrently request different Destinations [Classic example FollowSpiral conflicting with MoveToFort (_well actually conflicting with anything_)]
- Certain tasks want the bot.position to not change [loitering for example] and others want it to change.
- Tasks that use PolylineWalker are conflicting with task that don't use it

**Reason for the proposal against a refactoring alternative:**
- To much refactoring would be needed to solve above mentions points, in too many Tasks...,  Walkers... and so on... and so on...
- If i fuck up something the revert is to just undecorate the step() method in StepWalker

th3w4y:
```
import inspect

class ConcurrentWalking(object):
    """
    Proposal for decorator that stores all tasks destinations and returns them to the Task call only in certain cases
    Assumming bot.position (0,0)

    Will check:


    - When two tasks are requesting different destinations Task1(0,0), (5, 5)):   Task2((0,0), (6, 6))
        We identify that the cost of reaching dest (5,5) is small than reaching dest (6, 6)
        cost(Task1) < cost(Task2)
        Thus:
            a) When Task2.work() calls walker.step()  step() will NOT be executed but will return False immediately
            b) When Task1.work() calls walker.step() step() will be executed and method will return whateve is needed

    - When two task have same destination Will check if one of them is a PolylineWalker and if that is the case
        will execute the .step() of the identified PolylineWalker object also for the Task that requested StepWalker!!!

    - Other things to check

        - Possibly check the priority based on the order of the tasks in config.json !!!


    """
    _WALKER_CONTAINER = data # [] pr {} - to be defined!!!

    def _step(foo):
[...]

Issue #5333, opened 2016-09-09
Polyline Walker moves in zigzag


kureijir:
### Expected Behavior
Move path should look like this: (screenshot taken with commit ca94424f688bc8080db055d1b9443c3c8d8212fb)
http://imgur.com/rqfxWVu

### Actual Behavior
After updating to commit 11123a82cb18c172a262dc9e46e7b2c175666828 it looks like this:
http://imgur.com/xcYokPc
It slows down overall movement speed of the bot a bit and it looks annoying.

### Your FULL config.json (remove your username, password, gmapkey and any other private info)
http://pastebin.com/RP104P55

### Other Information
OS: 
Arch Linux ARM
Branch: 
master
Git Commit: 
11123a82cb18c172a262dc9e46e7b2c175666828
Python Version: 
2.7.12

crvfts:
try "enable_social": false, and remove the entire FollowSpiral task. Doing this solved this issue for me.

Anakin5:
Yes, it is your FollowSpiral that creates these pattern. Remove that tasks, you don't need it.

kureijir:
Thanks, just wonder why it's been working fine until I updated. And what's that task for anyway?

Anakin5:
It is for moving in zigzag and getting people to open issue tickets. It is pretty good at it ;)

Issue #5186, opened 2016-09-04
Polylinewalker is still drunk


orchid666:
I noticed a few days ago that , on the DEV branch, the polylinewalker stopped working smoothly.
Even after today's modification, it still has the "drunk" type of walking.
![poly](https://cloud.githubusercontent.com/assets/10350351/18232958/776518ec-72a0-11e6-94cb-5ce4ff9a06ee.jpeg)

Any idea what's going on?

maxprzemo:
In the file config.json remove lines
      {
        "type": "FollowSpiral",
        "config": {
          "enabled": true,
          "diameter": 4,
          "step_size": 70
        }
      }
I use the method "Follow Path" working as it should
![zrzut ekranu z 2016-09-04 21 10 53](https://cloud.githubusercontent.com/assets/4314537/18233298/907141c4-72e3-11e6-9753-2adc02aba406.png)

orchid666:
I removed followspiral and now polyline is working nicely. Aren't they supposed to work well together? It was the case last week.

th3w4y:
@orchid666 no they are not of course...
they give conflicting directions....

th3w4y:
One one task should use the PolylineWalker if you want consistent results..

Gobberwart:
Fixed: Told Polylinewalker to go home and sleep it off :)

orchid666:
Well kind of, it puked all over the couch.

mjmadsen:
@th3w4y maybe we should disable it in the config by default (instead of deleting it).

th3w4y:
@mjmadsen I do believe we are very close to that moment... 

There are still some issues when multiple tasks call step walker..
We have so many follow BaseTask and each call their StepWalwer as they please!

th3w4y:
@orchid666 is this still a problem for you ?

orchid666:
@th3w4y you mean polyline + spiral?

th3w4y:
@orchid666 i mean Polyline without spiral

(In my opinion Spiral should be deprecated!)

orchid666:
Polyline without spiral always worked for me.  Polyline with spiral used to work nicely then it stopped working so i wanted to bring that issue, but apparently it's something planned so i guess thats fine.

Gobberwart:
Currently using MoveToFort with polyline and it looks a lot less wobbly. Guess it sobered up a little.

th3w4y:
[...]

PR #5382, opened 2016-09-11
Bug fix on FollowPath, MoveToFort and PolylineGenerator

### LINKED ISSUE #5362: "FollowPath" Did not work.
### Expected Behavior
Working "FollowPath"

### Actual Behavior
A week ago everything worked (I have not used a bot by week) . No changes to the settings (I do not see the changes to be introduced in config.json).

### FULL config.json
http://hastebin.com/wuduwiqaci.js

[2016-09-10 09:49:07] [MainThread] [FollowPath] [INFO] Great job, lot of calories burned! Taking a break now for 0:27:15, will resume at 10:16:22.

The first point of "path_file" It is the end of the loop.
I cleaned folder /PokemonGo-Bot/data , I cleaned files cells-XXXXX.json, last-location-XXXXXX.json, XXX.db, mqtt_client_id

"MoveToFort" works correctly

### Other Information
OS: 
Debian 8.5
Branch: 
master
Git Commit: 
./setup.sh -r  today morning

Anakin5:
- Fix a crash with PolylineGenerator when speed is zero
- Fix a bug in FollowPath that think destination is reached each time we walk one step of a Polyline
- Fix FollowPath and MoveToFort tasks not returning RUNNING when they are actually the ones dictating the position.

Should fix issues users are having when they put FollowSpiral after a MoveToFort or a FollowPath. With this fix, the FollowSpiral will only be used if MoveToFort/FollowPath has nothing to do.

Fix #5362, #5333, #5186, #5367

### DIFF
--- a/pokemongo_bot/cell_workers/move_to_fort.py
+++ b/pokemongo_bot/cell_workers/move_to_fort.py
@@ -109,7 +109,7 @@ class MoveToFort(BaseTask):
                     formatted='Arrived at fort.'
                 )
 
-        return WorkerResult.SUCCESS
+        return WorkerResult.RUNNING
 
     def _get_nearest_fort_on_lure_way(self, forts):
 
--- a/pokemongo_bot/test/polyline_walker_test.py
+++ b/pokemongo_bot/test/polyline_walker_test.py
@@ -1,11 +1,11 @@
-import unittest, pickle, os
-import datetime, time
+import os
+import pickle
+import unittest
 
 from geographiclib.geodesic import Geodesic
 from mock import MagicMock, patch, mock
 
 import requests_mock
-from pokemongo_bot.walkers.polyline_generator import Polyline
 from pokemongo_bot.walkers.polyline_generator import PolylineObjectHandler
 from pokemongo_bot.walkers.polyline_walker import PolylineWalker
 
@@ -16,10 +16,8 @@ ex_speed = 2.5
 ex_total_distance = 194
 ex_resp_directions = 'example_directions.pickle'
 ex_resp_elevations = 'example_elevations.pickle'
-ex_enc_polyline = 'o_|~Gsl~r@??h@LVDf@LDcBFi@AUEUQg@EKCI?G?GBG@EBEJKNC??'
-ex_nr_samples = 78
-
-
+ex_enc_polyline = 'o_%7C~Gsl~r@??h@LVDf@LDcBFi@AUEUQg@EKCI?G?GBG@EBEJKNC??'
+ex_nr_samples = 64
 
 
 class TestPolylineWalker(unittest.TestCase):
@@ -55,7 +53,7 @@ class TestPolylineWalker(unittest.TestCase):
             m.get("https://maps.googleapis.com/maps/api/elevation/json?path=enc:{}&samples={}".format(
                 ex_enc_polyline, ex_nr_samples
             ), json=ex_elevations, status_code=200)
-            self.polyline = PolylineObjectHandler.cached_polyline(ex_orig, ex_dest, ex_speed)
+            self.polyline = PolylineObjectHandler.cached_polyline(ex_orig, ex_dest)
 
         self.bot.position = [ex_orig[0], ex_orig[1], self.polyline.get_alt(ex_orig)]
 
@@ -66,11 +64,10 @@ class TestPolylineWalker(unittest.TestCase):
     def test_polyline_fetched(self):
         self.assertEqual(self.polyline._points[0], ex_orig)
         self.assertEqual(self.polyline._points[-1], ex_dest)
-        total_seconds = self.polyline.get_total_distance() / self.polyline.speed
+        total_seconds = self.polyline.get_total_distance() / 3
         self.assertAlmostEqual(total_seconds, ex_nr_samples, places=0)
         self.assertEquals(self.polyline.get_total_distance(), ex_total_distance)
         self.assertEquals(self.polyline.get_last_pos(), self.polyline._last_pos)
-
 
     def test_one_small_speed(self):
         walk_max = self.bot.config.walk_max
@@ -103,7 +100,6 @@ class TestPolylineWalker(unittest.TestCase):
         self.bot.config.walk_max = walk_max
         self.bot.config.walk_min = walk_min
 
-
     def test_one_small_speed_big_precision(self):
         walk_max = self.bot.config.walk_max
         walk_min = self.bot.config.walk_min
@@ -134,7 +130,6 @@ class TestPolylineWalker(unittest.TestCase):
 
         self.bot.config.walk_max = walk_max
         self.bot.config.walk_min = walk_min
-
 
     def test_intermediary_speed(self):
         walk_max = self.bot.config.walk_max
@@ -203,8 +198,6 @@ class TestPolylineWalker(unittest.TestCase):
         walk_min = self.bot.config.walk_min
         speed = 300
         precision = 0.0
-        dlat = 47.1700271
-        dlng = 8.518072999999998
 
         self.bot.config.walk_max = speed
         self.bot.config.walk_min = speed
@@ -234,8 +227,6 @@ class TestPolylineWalker(unittest.TestCase):
         walk_min = self.bot.config.walk_min
         speed = 300
         precision = 2.5
-        dlat = 47.1700271
-        dlng = 8.518072999999998
 
         self.bot.config.walk_max = speed
         self.bot.config.walk_min = speed
--- a/pokemongo_bot/walkers/polyline_generator.py
+++ b/pokemongo_bot/walkers/polyline_generator.py
@@ -22,7 +22,7 @@ class PolylineObjectHandler:
     _run = False
 
     @staticmethod
-    def cached_polyline(origin, destination, speed, google_map_api_key=None):
+    def cached_polyline(origin, destination, google_map_api_key=None):
         '''
         Google API has limits, so we can't generate new Polyline at every tick...
         '''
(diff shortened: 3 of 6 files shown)


pokemongo_bot/cell_workers/move_to_fort.py
```diff
@@ -109,7 +109,7 @@ def work(self):
                     formatted='Arrived at fort.'
                 )
 
-        return WorkerResult.SUCCESS
```
solderzzc: Why you need bypass the following tasks ?

Anakin5: Because you do not want another moving tasks to dictate the position. You are the one deciding at that point.

solderzzc:
I see. current RUNNING status is not good. but still usable.

solderzzc:
👍

pokemongo_bot/walkers/polyline_generator.py
```diff
@@ -94,7 +94,7 @@ def __init__(self, origin, destination, speed, google_map_api_key=None):
         self._step_keys = sorted(self._step_dict.keys())
         self._last_step = 0
 
-        self._nr_samples = int(min(self.get_total_distance() / self.speed + 1, 512))
+        self._nr_samples = int(min(1 + self.get_total_distance() / max(self.speed, 1), 512))
```
th3w4y: I see so you had no elevations on speed 0 right?

pokemongo_bot/cell_workers/move_to_fort.py
solderzzc: @th3w4y  Agreed.

th3w4y:
@Anakin5 Thanks for finding the bug in the Polyline 

I would have swear we had (self.speed +1)... must have been sleepy when i wrote it :)
`int(min(self.get_total_distance() / (self.speed + 1), 512))`

pokemongo_bot/walkers/polyline_generator.py
th3w4y: 👍

th3w4y: Looked a bit in depth... min nr_sample needs to be 2 in order to have elevations returned. otherwise is an invalid response

th3w4y: Wait a second... if you hit the ZeroDivizion you are using the PolylineWalker incorectly!

th3w4y:
It is not valid to do this:

lets assume bot position is 0,0

`step_walker = PolylineWalker(0,0)`
This is invalid because:
- You consume one API call for nothing... there will be no directions or elevations returned distance is 0 Polyline._points = one point encoded polyline has only one point... there are no elevations returned

th3w4y:
correct use case is 
bot position 0,0
step_walker = PolylineWalker(0.1,0.1)
step_walker.step(speed=0)

th3w4y:
There could still be a problem when the first call to PolylineObjectHandler is created by a step with speed zero.

pokemongo_bot/walkers/polyline_generator.py
Anakin5: The polyline is not initialized in the constructor of the Walker, it is initialized at the first step.

Anakin5:
I am not sure what issue you are talking about. I do not want to initialize a Polyline with the current location.

th3w4y:
Yes that is why i commented after still a corner case...
>I am not sure what issue you are talking about. I do not want to initialize a Polyline with the current location.

pokemongo_bot/walkers/polyline_generator.py
th3w4y: I found the problem https://github.com/PokemonGoF/PokemonGo-Bot/pull/5382#issuecomment-246173317

pokemongo_bot/cell_workers/move_to_fort.py
Anakin5: There is no point to put your catch Pokemon task after the moving task. There is no point offering this kind of flexibility. People want the bot to work, they do not care about all the mighty features that they never asked for.
If you really want to remove the WorkerResult because you still want to allow people to put their moving tasks in the middle, why not just having a single variable that we use as a "lock" to decide who set the position ?

in MoveToFort
```
if self.bot.move_lock is not None and self.bot.move_lock != "MoveToFort":
    return

self.bot.move_lock == "MoveToFort"
<do my logic here>
```

in FollowPath
```
if self.bot.move_lock is not None and self.bot.move_lock != "FollowPath":
    return

self.bot.move_lock == "FollowPath"
<do my logic here>
```

That way you can put every move tasks where you want, and it only cost you 2 lines of code per walker.

th3w4y:
I have to think now what will happen when speed is float("inf")

pokemongo_bot/walkers/polyline_generator.py
th3w4y: we should not handle ZeroSpeed here otherwise we'll mask future incorrect utilisation cases

Anakin5:
So why not 1 then ?

th3w4y:
I see you understood.. 👍 
> So why not 1 then ?

will fetch to many elevations... much more calculations later on in .get_alt() 

makes sense?

th3w4y:
i could even standardise on a fixed speed no matter what...
.. it will not affect walking!

Shall i do that? say like pick 3.16 no matter what?

pokemongo_bot/walkers/polyline_generator.py
Anakin5: What do you mean ? In what case do you get an invalid Polyline ?

Anakin5:
I would pick 1 each time it is possible (that is for every Polyline of 512 meters of less). And I would pick 2 for polyline between 512 and 1024 meters, and 3 for polyline between 1024 and 1536 etc...
The speed does not matter, we just need some elevation points, they don't need to be exact at each centimeters.

th3w4y:
@Anakin5 meaning you would always pick min(distance, 512) right?

Anakin5:
I convinced myself

pokemongo_bot/walkers/polyline_generator.py
```diff
@@ -94,7 +93,7 @@ def __init__(self, origin, destination, speed, google_map_api_key=None):
         self._step_keys = sorted(self._step_dict.keys())
         self._last_step = 0
 
-        self._nr_samples = int(min(1 + self.get_total_distance() / max(self.speed, 1), 512))
+        self._nr_samples = 1 + int(math.ceil(self.get_total_distance() / 512))
```
th3w4y: Incorrect!

for distance = 510
1 + math.ceil(510/512) = 1

th3w4y:
Incorrect!

for distance = 510
1 + math.ceil(510/512) = 1
we need at lest 2!

we actually want 
min(self.get_total_distance(), 512)
if we don't want to involve speed!

Anakin5:
```
1 + int(math.ceil(510/512.0))
2
```

I will add the point to 512 eve if get_total_distance return a float, to be sure

th3w4y:
and the you will have only 2 elevation samples!!!!!!

Anakin5:
You always want 512 points ?

th3w4y:
That was why i was involving speed to get kinda 1 elevation sample at each step!

Anakin5:
ok for the min then

th3w4y:
maybe `min(max(int(self.get_total_distance()/3), 2), 512)` to ease a big on the calculations...  not 3 elevations per step...

Anakin5:
ok

pokemongo_bot/walkers/polyline_generator.py
```diff
@@ -94,7 +93,7 @@ def __init__(self, origin, destination, speed, google_map_api_key=None):
         self._step_keys = sorted(self._step_dict.keys())
         self._last_step = 0
 
-        self._nr_samples = int(min(self.get_total_distance() / self.speed + 1, 512))
+        self._nr_samples = int(max(min(self.get_total_distance() / 3, 512), 2))
```
th3w4y: needs to be an int!

th3w4y:
ons small problem... get_total_distance returns float... i need to send int to the API!

pokemongo_bot/cell_workers/move_to_fort.py
th3w4y: There is no point but people will still do it
>There is no point to put your catch Pokemon task after the moving task. 

The point is less opened issues to spend our time with...
but if you want to "educate" users on how to configure the bot and what is the "supported" task issues and after that  look at all issues rising from this.. we'll i guest you do have a lot of time to spare..
>There is no point offering this kind of flexibility

th3w4y: But this is already a side discussion.. to be taken at some other point!

th3w4y:
Regarding the Changes in the moving task .RUNNING .SUCCES
My opinion is that the changes done here are 💯% in line with how we do the things **NOW.**
(of course we could do it differently regarding the order of the Tasks in the config file but those comments are not for this PR)

th3w4y:
Test cases for polyline_generator are breaking
you still call ex_speed at 
https://github.com/Anakin5/PokemonGo-Bot/blob/b31f5226586629dddc90779efc0a53316dd014da/pokemongo_bot/test/polyline_generator_test.py#L34 

and ex_nr_samples should be changed to 64

pokemongo_bot/test/polyline_walker_test.py
```diff
@@ -16,12 +16,10 @@
 ex_total_distance = 194
 ex_resp_directions = 'example_directions.pickle'
 ex_resp_elevations = 'example_elevations.pickle'
-ex_enc_polyline = 'o_|~Gsl~r@??h@LVDf@LDcBFi@AUEUQg@EKCI?G?GBG@EBEJKNC??'
+ex_enc_polyline = 'o_%7C~Gsl~r@??h@LVDf@LDcBFi@AUEUQg@EKCI?G?GBG@EBEJKNC??'
```
th3w4y: you even took the time to url encode 👍

Anakin5:
Can I merge ?

solderzzc:
I think @th3w4y meant we can.

solderzzc:
👍

Issue #5217, opened 2016-09-05
Bot is constantly relogging when hits desired lvl


Shentang:
### Expected Behavior
Bot should stop and not restart upon hitting desired lvl

### Actual Behavior
Bot constantly restarts when hitting desired lvl

### Your FULL config.json (remove your username, password, gmapkey and any other private info)

I'll post the part that makes the bot stop at desired level, the rest doesn't matter anyways.

      {
        "type": "CollectLevelUpReward",
        "config": {
          "collect_reward": true,
          "level_limit": 20

### Output when issue occurred
" You have reached your target level! Exiting now."

### Steps to Reproduce
1. Set "level_limit" in config to something like 1 in task CollectLevelUpReward
2. Start the bot

### Other Information
**OS:** Debian
**Branch:** Master
__Python Version:__ 2.7

### Bot log

2016-09-05 18:40:46,645 [       cli] [INFO] PokemonGO Bot v1.0
2016-09-05 18:40:46,651 [       cli] [INFO] commit: ca94424f
2016-09-05 18:40:46,655 [       cli] [INFO] Configuration initialized
2016-09-05 18:40:46,655 [pokemongo_bot.health_record.bot_event] [INFO] Health check is enabled. For more information:
2016-09-05 18:40:46,655 [pokemongo_bot.health_record.bot_event] [INFO] https://github.com/PokemonGoF/PokemonGo-Bot/tree/dev#analytics
2016-09-05 18:40:46,661 [requests.packages.urllib3.connectionpool] [INFO] Starting new HTTP connection (1): www.google-analytics.com
(23300) wsgi starting up on http://194.28.50.34:3000
2016-09-05 18:40:46,854 [requests.packages.urllib3.connectionpool] [INFO] Starting new HTTP connection (1): 194.28.50.34
2016-09-05 18:40:46,860 [requests.packages.urllib3.connectionpool] [INFO] Starting new HTTP connection (1): 194.28.50.34
[2016-09-05 18:40:46] [SleepSchedule] [INFO] [next_sleep] Next sleep at 12:00:00, for a duration of 0:01:45
[2016-09-05 18:40:46] [PokemonGoBot] [INFO] [set_start_location] Setting start location.
[2016-09-05 18:40:47] [PokemonGoBot] [INFO] [location_found] Location found: London (51.5073509, -0.1277583, 0.0)
[...]

Issue #5376, opened 2016-09-10
Logging in with wrong password in auth.json doesn't stop the bot


Shentang:
Currently, the behaviour of the bot is, that if the password is wrong, it will try 5 times before locking you out from your account.

I don't really see sense in this, since if the password was wrong on the first try, how would it be correct on th 2nd, 3rd, 4th and 5th? Bot should just stop running instead of retrying login. I barely made it to use "crtl+c" to stop the bot while it was trying for the 5th time...

Gobberwart:
Fixed

PR #5386, opened 2016-09-11
exit the bot if unable to login

rawgni:
## Short Description:

raising the right Exception instead of True or False when trying to login.

- #5376 
-
-


solderzzc:
👍

Shentang:
@rawgni @solderzzc do you think you could implement this also to fix #5217 ? It's similiar in behaviour.

PR #5392, opened 2016-09-11
standard average speed for Polyline initialization when step(speed=0)

th3w4y:
Fixes:

- force a standard average speed for Polyline initialization when step(speed=0)

https://github.com/PokemonGoF/PokemonGo-Bot/pull/5382

### DIFF
--- a/pokemongo_bot/walkers/polyline_generator.py
+++ b/pokemongo_bot/walkers/polyline_generator.py
@@ -20,6 +20,8 @@ class PolylineObjectHandler:
     _cache = None
     _instability = 0
     _run = False
+    _standard_average_speed = 3.16
+    _standard_max_speed = 4.16
 
     @staticmethod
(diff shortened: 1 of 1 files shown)


pokemongo_bot/walkers/polyline_generator.py
```diff
@@ -49,6 +51,7 @@ def cached_polyline(origin, destination, speed, google_map_api_key=None):
                 PolylineObjectHandler._run = True
                 PolylineObjectHandler._instability = 20  # next N moves use same cache
 
+            speed = min(speed or PolylineObjectHandler._standard_average_speed, PolylineObjectHandler._standard_max_speed)
```
th3w4y: If someone wonders why the hardcoded values... it does not matter that much... the SpetWalker.step(speed=x) will override those values... 

this values are used only for calculating the nr. of elevation samples... not for walking!

th3w4y:
was handled in #5382

PR #5394, opened 2016-09-11
Min pokemon to trigger evolve

developerllazy:
Add the config to set the minimum no. of pokemon to trigger evolve so that it can maximum the experience earned when using lucky egg.


solderzzc:
👍

PR #5395, opened 2016-09-11
Stop moving to lures if no lures active anymore

davidakachaos:
## Short Description:
Small addition; if the lures expire before we get there, stop moving toward the cluster.


solderzzc:
👍

Anakin5:
Maybe we should not start at all. We know the distance, the speed and the expiration time after all.

davidakachaos:
@Anakin5 isn't this more human like? ;-) I agree that it would be more optimal not to move to the cluster if we can't make it in time, but in the game we can't see how long the lures will be active.

PR #5414, opened 2016-09-12
Master

javajohnHub:
Rename pokemon now works in telegram.
Added progress list so others can confirm or deny the events are
displaying in telegram.


solderzzc:
uh, this PR overrides some changes. It's better to redo on a refresh clone.

javajohnHub:
Will do and resubmit

PR #5421, opened 2016-09-13
Update to latest api

joelgreen:
## Short Description:
Updating to latest api at https://github.com/pogodevorg/pgoapi

Signature is now generated by the api instead of the bot, device data is still passed in from bot
Latest protos should support buddy system


pokecli.py
```diff
@@ -64,7 +64,7 @@
 try:
     import pkg_resources
     pgoapi_version = pkg_resources.get_distribution("pgoapi").version
-    if pgoapi_version < '1.1.8':
+    if pgoapi_version != '1.1.6':
```
DBa2016: Please help me understand... you are suggesting to change the API version requirements from 1.1.8 to 1.1.6?... Does not sound right to me.

solderzzc:
👍

pokecli.py
joelgreen: The bot was using my api fork at https://github.com/joelgreen/pgoapi which i incremented the version number, however since switching the changes i made have become obsolete so I'm suggesting we switch back to the main branch which is now about 30 commits ahead. However they never incremented the version number.

crvfts:
how do we get around "The plan is to install the git repository https://github.com/pogodevorg/pgoapi.git/" ? I deleted /python27/src/pgoapi but still get this message.

javajohnHub:
Bot wont start after merge

joelgreen:
If you're having trouble updating requirements try removing the -e in requirements.txt
`-e git+https://github.com/pogodevorg/pgoapi.git/@5e550f656b318227f021801c796b8a80ad917b93#egg=pgoapi`
to
`git+https://github.com/pogodevorg/pgoapi.git/@5e550f656b318227f021801c796b8a80ad917b93#egg=pgoapi`

DBa2016:
errr, can we revert this PR and you fix it so it can be run without manipulating? I understand this is the "dev" branch but no PR should be breaking it.

joelgreen:
@DBa2016 its not broken its just a bug with pip for some people

PR #5425, opened 2016-09-13
Improve performance of cluster computation in CampFort task

Anakin5:
Improve performance of cluster computation in CampFort task.
Compute clusters only one time and update them afterwards.. It could take several seconds to compute each time we are looking for a camp.

### DIFF
--- a/pokemongo_bot/cell_workers/camp_fort.py
+++ b/pokemongo_bot/cell_workers/camp_fort.py
@@ -25,10 +25,11 @@ class CampFort(BaseTask):
         super(CampFort, self).__init__(bot, config)
 
     def initialize(self):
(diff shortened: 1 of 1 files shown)


solderzzc:
👍

davidakachaos:
:+1: :smile:

Anakin5:
That is strange, I rebased this PR because somebody added event stuff. But now the event stuff disappear in the repo and I am credited with the changes...

javajohnHub:
I added the events to campfort task. Not sure what happened.

Anakin5:
@javajohnHub And I don't know why what you added has been merge because it breaks the task. One event is not declared.

PR #5432, opened 2016-09-13
Revert "Update to latest api"

DBa2016:
The referenced PR seems to produce many issues and needs manual edits (which will show up as diffs later) to be working at all. I suggest reverting it in order to allow further development while it is being tested more thoroughly.

Reverts PokemonGoF/PokemonGo-Bot#5421

### DIFF
--- a/pokecli.py
+++ b/pokecli.py
@@ -64,7 +64,7 @@ if sys.version_info >= (2, 7, 9):
 try:
     import pkg_resources
     pgoapi_version = pkg_resources.get_distribution("pgoapi").version
-    if pgoapi_version != '1.1.6':
+    if pgoapi_version < '1.1.8':
         print "Run following command to get latest update: `pip install -r requirements.txt --upgrade`"
         sys.exit(1)
 except pkg_resources.DistributionNotFound:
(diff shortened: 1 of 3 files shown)


solderzzc:
That commit is running on my machine.

halsafar:
The PR in question works fine.  Don't revert it just because you didn't read the instructions.

```
rm -rf src/
source bin/activate
pip install --upgrade -r requirements.txt
(let it switch repo when it asks if it asks).
```

Maybe run the pip command twice.  

All that was required for me anyway.

joelgreen:
This PR should fix the issues some people are having updating https://github.com/PokemonGoF/PokemonGo-Bot/pull/5434

DBa2016:
If this does not work "out of the box", it should be kept in a separate branch and tested before it is merged into the main "dev".

joelgreen:
@DBa2016 it does work out of the box for most. Some people were having trouble with pip and the PR i submitted is a workaround, along with a fix for an oversight in the run.sh file

RedViper9:
The new API seems to work for most things, but pokemon_optimizer doesn't seem to function anymore. #5439

solderzzc:
@RedViper9  Mine works well (optimizer). Is your optimizer config contain 'rules' ?

RedViper9:
@solderzzc Yes it does. The issue #5439 is mine. I've been struggling with the new enhancements to the pokemon optimizer since yesterday. Had my bot running on the older version of optimizer.py while I tried to figure it out, but this morning that version fails to work as well. Using the Evolve / Transfer tasks works just fine.

I'll try a fresh install tonight.

To update I do
```
git fetch --all
git reset --hard origin/dev
git pull origin dev
```
Which should overwrite any local changes but leave my config intact.

PR #5434, opened 2016-09-13
Fix issues people are having with updating to new api

joelgreen:
For most people theres no need for the `-e` for installing the api and it often causes issues installing with pip.

Also fix run.sh


solderzzc:
I guess this is no longer needed since many users force update the repo?

PR #5437, opened 2016-09-13
Catch max cp, catch certain moves

DBa2016:
## Short Description:

- introduced "catch_below_cp" parameter for the "catch" config section (numeric value)
- introduced "fast_attack" and "charged_attack" to specify desired moves for a pokemon (list of strings); omitting these parameters or supplying an empty lis ([]) has an effect of "any"


solderzzc:
I think you don't want to waste poke ball on high CP pokemon.

solderzzc:
👍

DBa2016:
This is actually not about saving pokeballs. :)

This is to auto-capture Pokemons for "Bubblestrat". You need a level 1 Diglett with Scratch or Mud Slap and a level 1 (or 1.5) Horsea with Bubble (other Bubble-capable Pokemons might do, too). You can capture them only when being level 1 or 2 (otherwise they will have higher levels). The idea is: ascend to level 5 or higher, place your CP20-Horsea in a gym as a defender; then you attack it with your Diglett (10CP). Scratch and Mud Slap are fast enough to kill the Horsea before its Bubble hits and you get 1k XP within seconds. In theory, it is even possible to make 1.5 million XP per day this way (without botting!).

solderzzc:
@DBa2016 Uha, I know Bubblestrat, this is really cool :) But I can hardly find a gym w/o other guys in range and I don't want to take risk do the Bubblestrat during midnight, the gun shot news scared me a lot.

solderzzc:
I'd like to have this instruction in commit:)
>This is to auto-capture Pokemons for "Bubblestrat". You need a level 1 Diglett with Scratch or Mud Slap and a level 1 (or 1.5) Horsea with Bubble (other Bubble-capable Pokemons might do, too). You can capture them only when being level 1 or 2 (otherwise they will have higher levels). The idea is: ascend to level 5 or higher, place your CP20-Horsea in a gym as a defender; then you attack it with your Diglett (10CP). Scratch and Mud Slap are fast enough to kill the Horsea before its Bubble hits and you get 1k XP within seconds. In theory, it is even possible to make 1.5 million XP per day this way (without botting!).

solderzzc:
👍

pokemongo_bot/cell_workers/pokemon_catch_worker.py
```diff
@@ -16,6 +16,9 @@
 from datetime import datetime, timedelta
 from utils import getSeconds
 
+from pprint import pprint
```
Anakin5: Is it used somewhere ?

pokemongo_bot/cell_workers/pokemon_catch_worker.py
```diff
@@ -284,15 +291,32 @@ def _pokemon_matches_config(self, config, pokemon, default_logic='and'):
         if pokemon.cp > catch_cp:
             catch_results['cp'] = True
 
+        catch_below_cp = pokemon_config.get('catch_below_cp', 12000)
```
Anakin5: 12000 is the default value ?

pokemongo_bot/cell_workers/pokemon_catch_worker.py
```diff
@@ -284,15 +291,32 @@ def _pokemon_matches_config(self, config, pokemon, default_logic='and'):
...
+        if pokemon.cp < catch_below_cp:
+            catch_results['cp'] = True
+
         catch_iv = pokemon_config.get('catch_above_iv', 0.8)
         if pokemon.iv > catch_iv:
             catch_results['iv'] = True
 
+
+        catch_results['fa'] = ( len(pokemon_config.get('fast_attack', [])) == 0 or unicode(pokemon.fast_attack) in map(lambda x: unicode(x), pokemon_config.get('fast_attack', [])))
```
Anakin5: You do not need all the Unicode things. If you want to manipulates string as Unicode, just add `from __future__ import unicode_literals` at the start of the file.

pokemongo_bot/cell_workers/pokemon_catch_worker.py
DBa2016: Was used during development and does not hurt as it is very useful for future development.

pokemongo_bot/cell_workers/pokemon_catch_worker.py
DBa2016: Yes, to make sure that no pokemons are excluded if this is not specified

pokemongo_bot/cell_workers/pokemon_catch_worker.py
DBa2016: They are needed. Comparisons were failing without the unicode cast.

DBa2016:
do not merge yet please, need one mor efix (logic error)

DBa2016:
okay, good for merge now

solderzzc:
👍

cianmce:
@javajohnHub Yes this code changed in `pokemongo_bot/cell_workers/pokemon_catch_worker.py` does result in all pokemon being considered VIP.
Reverting to an the previous file in: 02390dce72ddac2a60ba14a786906485ffa61f96 fixes this error

PR #5440, opened 2016-09-13
Telegram formatting changes

javajohnHub:


DBa2016:
1) Travis build fails
2) there is definitely a description needed. By just looking at your code nobody can tell whether it is doing what was intended or not.

javajohnHub:
Telegram formatting is more than enough information... the rest is trying to get travis passing. But thanks.

Issue #5378, opened 2016-09-10
MoveToMapPokemon  Teleport not working !!!!


cbogithub:
```
[2016-09-11 01:25:46] [PokemonCatchWorker] [INFO] You now have 47 Growlithe candy!
[2016-09-11 01:25:52] [UpdateLiveStats] [INFO] Uptime : 0:00:30 | Level 24 | Earned 100 Stardust | +160 XP | 19,152 XP/h | Visited 0 stops | Caught 1 pokemon
[2016-09-11 01:25:53] [MoveToMapPokemon] [INFO] Teleporting to Magikarp. (4.91km)
[2016-09-11 01:25:57] [MoveToMapPokemon] [INFO] Encountered Pokemon: Magikarp
[2016-09-11 01:26:00] [MoveToMapPokemon] [INFO] Teleporting back to previous location (-33.8904865206, 151.222649271)...
[2016-09-11 01:26:02] [MoveToMapPokemon] [INFO] Teleporting to Magikarp. (5.09km)
[2016-09-11 01:26:06] [MoveToMapPokemon] [INFO] Encountered Pokemon: Magikarp
[2016-09-11 01:26:08] [MoveToMapPokemon] [INFO] Teleporting back to previous location (-33.8904865206, 151.222649271)...
[2016-09-11 01:26:10] [MoveToFort] [INFO] Moving towards pokestop Kippax Lake - 0.14km
[2016-09-11 01:26:13] [UpdateLiveStats] [INFO] Uptime : 0:00:51 | Level 24 | Earned 100 Stardust | +160 XP | 11,207 XP/h | Visited 0 stops | Caught 1 pokemon
[2016-09-11 01:26:14] [MoveToMapPokemon] [INFO] Teleporting to Magikarp. (5.08km)
[2016-09-11 01:26:18] [MoveToMapPokemon] [INFO] Failure! Magikarp doesnt exist anymore. Skipping...
[2016-09-11 01:26:20] [MoveToMapPokemon] [INFO] Teleporting back to previous location (-33.8903356483, 151.222636764)...
[2016-09-11 01:26:22] [MoveToMapPokemon] [INFO] Teleporting to Magikarp. (5.10km)
[2016-09-11 01:26:27] [MoveToMapPokemon] [INFO] Encountered Pokemon: Magikarp
[2016-09-11 01:26:29] [MoveToMapPokemon] [INFO] Teleporting back to previous location (-33.8903356483, 151.222636764)...
[...]

Issue #3672, opened 2016-08-12
Public Sniping List Integration


TheSavior:
# Utilizing Sniping Lists
There are a bunch of services right now that provide lists of pokemon that are live, and snipeable. Examples include:

 - [PokeSnipers](http://pokesnipers.com/)
 - [Pokezz](http://pokezz.com/)
 - [PokeWatchers](https://pokewatchers.com/)
 - [Haxton](http://haxton.io/api/pokemon)
 - [Skiplagged](https://skiplagged.com/pokemon)
 - and many more

We want to be able to allow our bots to support sniping from these lists. Each site has a different way to read data from their api. 

## Requirements for a valid solution:
 - The user of the bot must be able to configure which list(s) they want to pull from
 - The user must be able to configure which pokemon they want to snipe
 - Different websites have different data formats and needs (only sending VIPs or perfect IVs for example)
 - Creating a new data source should be possible using plugins
 - We must be able to report found pokemon back to the configured servers.

### Example Configuration
I imagine the configuration for this task to be something like this:
```
{
  // ...
  "tasks": [
    {
      "type": "SnipePokemon",
      "config": {
        "providers": [
          "PokeSnipers",
          "Haxton"
        ]
      }
    },
  ],
  "catch": {
    "Growlith": { "catch_above_cp": 1000, "catch_above_iv": 0.95, "logic": "or" }
  },
  "vips" : {
    "Dragonite": {},
    "Snorlax": {}
  }
  // ...
}
```

### Providers
In this solution I will use the term `Provider` to mean a python class that provides the data from a single source.

There will be a `SnipingProviderBase` class that providers can implement. Here is a sample implementation:

```python
class SnipingProviderBase(object):
  def initialize(self):
    self.search_config = self.config.get('search_config', {})

  def encountered_pokemon(self, pokemon):
    """
    format the pokemon and send the data to the server
    """
    pass

  def search_for_pokemon(self):
    """
    connect to the provider, possibly using search_config with the provider api
    return a list of pokemon location objects
    """
    pass
```

And then for an idea of what the `SnipePokemon` task might look something like this:

```python
class SnipingTask(BaseTask):
  def initialize(self):
    self.raw_providers = self.config.get('providers', []);
    self.search_config = self.config.get('search_config', {})

    self.providers = [
      # Class references to instantiated providers
    ]

    # self.bot.event_manager.add_handler(...
    # register an event listener that listens for 
[...]

Issue #3666, opened 2016-08-12
Suggestion the GUI bot can support pokemon sinper 


zyxelgg:
Sounds Good
It feasible ?

k4n30:
#3355 #3278 #3226

Issue #3656, opened 2016-08-11
[Feature] Move map should accept multiple pokemon sources


shivamkalra:
One of te best feature of bot is that it can be guided using data from pokeman-map.

Since we already have the logic of guiding the bot based on priority and distance to the pokemon that needs to be caught.

I suggest guiding code should be separated out from data collection code. Such that I can collect pokemon data from various sources (twitter, pokemonmap, pokesniper and etc), normalize them and feed them to bot's guide module.

Basically my suggestion is to remove dependency on pokemap instead use some universal format which contains pokemon and their locations which guides the bot. However source can be either pokemon map or some twitter alerts or anything.

What do you think? I could work on it if you people like the idea?

DanielVolz:
It's a great idea! I actually use a hack to feed the movetopokemonmap worker with location data from a discord channel and it works really great as a sniper! But I would be happy to have a proper integration of this feature. :+1:

LDengXYZ:
Hi @DanielVolz how exactly do you do that?

stevelacey:
I got this far:

```
from datetime import datetime
import dateutil.parser

def get_pokemon_from_pokesniper(self):
    try:
        req = requests.get('http://pokesnipers.com/api/v1/pokemon.json')
    except requests.exceptions.ConnectionError:
        self._emit_failure('Could not get Pokemon data from PokemonGo-Map: '
                           '{}. Is it running?'.format(
                               self.config['address']))
        return []

    try:
        raw_data = req.json()
    except ValueError:
        self._emit_failure('Map data was not valid')
        return []

    pokemon_list = []
    now = int(time.time())

    for pokemon in raw_data['results']:
        pokemon['encounter_id'] = None
        pokemon['spawn_point_id'] = None
        pokemon['latitude'], _, pokemon['longitude'] = pokemon['coords'].partition(',')
        pokemon['latitude'], pokemon['longitude'] = float(pokemon['latitude']), float(pokemon['longitude'])
        pokemon['disappear_time'] = int((dateutil.parser.parse(pokemon['until'].replace('Z', '')) - datetime(1970, 1, 1, 0, 0, 0)).total_seconds())
        pokemon['name'] = pokemon['name']
        pokemon['is_vip'] = pokemon['name'] in self.bot.config.vips
```

Seemed to get as far as encountering the first thing off of the pokesnipers list and then idling – I don't know what to do about encounter or spawn ids – @DanielVolz presumably you have something like this?

[...]

PR #5443, opened 2016-09-15
Snipping V2

YvesHenri:
Snipping was first introduced, back in the days when there wasn't any snipping sources, to integrate the PokemonGo-Map project into the bot and this is brilliant. However, things have changed ever since. There are lots of sources nowadays that provides these informations, eg.: pokesnipers, pokespawns, pokecrew and the list goes on. Mostly of them offer APIs (urls) to get these results in a JSON formated fashion way and that's what this update has come for! Lots of efforts were put into it and everything were meticulously thought, including this text.

Below are some considerations that have been taken and why they have been added/removed:

- **Reaching a target by walk** (or running faster than a giant evil flying unicorn at the speed of light) that is 3k kilometers farther within a maximum of 15 minutes (if we're lucky), for example, is pretty much impossible and we all know that, so why walk when you can teleport? The closest that people would look forward rares would be within 2~8 kilometers, something near their houses, but still that would be hard, or even impossible, to reach it within 6 minutes due to traffic, for example. Not to mention that if you enabled to visit forts on the path, you would very likely run out of time. You get the idea.
- **Maintaining the PokemonGo-Map** is hard for newcomers on its own, due to the python requirements and other stuff. It has done a great job and was even the kick for the snipping websites era (including the crowd-sourceds, which was the solution to the recent API updates). Its pointless to keep it up when there are so many other options. Updating PokemonGo-Map location automatically while walking/teleporting around is just... pointless. Once again, it was made to manually analize certain areas to find some rare pokemons, not automatic.

These are the most important subjects but I could keep going forever, otherwise I wouldn't have spent all this time.

The new sniper idea is to make it as flexible as possible, by allowing the user to specify the attribute names of a JSON response, since they are almost always different, depending on the source. This also offers a much more "complex" way to prioritize your targets, instead of VIPs and threshold only. Surely there are lots of stuff to do to cover all these possibilities.

There is no point to keep the old sniper, since **theyre basically the same**, except that the old one allows you to reach the target by movements but **only works strictly with PokemonGo-Map**.

Related issues: #3672 #3656 #3666 #3355 #3278 #3226.


crvfts:
Please also keep original MoveToMapPokemon task.

YvesHenri:
Updated first post.

crvfts:
In my opinion removing MoveToMapPokemon is the equivalent of removing one of the most human-like functions of the bot. To me, as the bot has progressed it has always been in the direction to be more human-like.

I disagree with your statement in terms of walking-to-pokemon. For example, I play in an area that has a total radius of about 2.5km. Generally, the bot ends up in one of 3 'zones'. The reason MoveToMapPokemon works well for me is that I can set the walk range to a real-life scenario (human-like) and if a desired Pokemon comes within range (generally 350-700km) then the bot walks to it and encounters there.  I do this to avoid making a capture that has a spawn_id nowhere near where I'm playing. I understand using bots is "dangerous" overall, but the idea is to try maintain human-like play. I am able to easily achieve this using the recommend speeds that come in default config.

Maintaining PokemonGo-Map (itself) is quite easy. If the user already has PokemonGo-Bot then they already have the dependencies. All they have to do after that is get a working API key (that they may already have from utilizing PolyLineWalker), config pogom's config, and then run the map scanner. In my experience I have seen more users enter chat due to conflicts with map-chat than I have seen users having issues connecting with pogom. I believe both pogom and map-chat have their advantages and both do lead to confusing events.

If the new Sniper intention is to be as flexible as possible, why would something that already works be removed with an unforeseen future? Reduction here would make using the bot more restrictive.

In conclusion, there has been a lot of work put into MoveToMapPokemon, and even though I have not contributed any coding to the project, I have a great appreciation for the human-like characteristics of MoveToMapPokemon. Hopefully others think similarly.

YvesHenri:
I just noticed that @TheSavior was planning to do this a month ago and there it is. Lots of related issues were found as well, and I listed some of them on the first post (updated).

@crvfts Thank you for your feedback. The only bot feature that goes out of the human-like scope is the sniper. I have yet to see someone else who have been using walking sniper. To be honest, I dont even know why it was ever implemented. Someone has already explained to you how this exploit works but I'll remind you:

1. Teleport to a target very far away. This wont get you banned unless you catch it right there or spin some pokestops.
2. Send an encounter event with the target to the game server.
3. Teleport back to your previous position.
4. Catch the target. Here, even thought it's not confirmed, the game thinks it was encountered and caught at your position, near to you.

I strongly believe that you have never been lucky enough to catch something like a Snorlax or a Dragonite with your sniping settings, because its clearly very unlikely to happen exclusively to sniping feature, but luck. With teleport sniping you would of have much more Snorlaxes and Dragonites by now. Also, there's this new PokemonHunter task that might fit your walking requirements. Discussing bot/sniping safeness is like discussing religion and/or politics and I'm not going to do that, however I can tell you that I highly believe that **people aren't being banned for sniping, but botting**. I bet @joelgreen, one of the API contributors, would love to discuss that with you. :yum: 

Maintaining PokemonGo-Map can be easy for you, but wasn't for me when I started at all this python thing world. There were times that both the map and the bot used different requirements and you would need to use some third-party containers for this. Why wouldn't you want to use someone else's source (no hosting costs, no setup work, nothing) instead of all this? **Either way, you will still be able to use PokemonGo-Map with this new sniper**. I'm offering a much bigger flexibility than only sticking with PokemonGo-Map and it's not unforeseen, its how things grows.

crvfts:
Again, I'll reiterate- I UNDERSTAND HOW SNIPING WORKS. Again, my concern- that you catch from a spawn_id that is nowhere near it's location, ie. bot is in NY, encounters pokemon in Japan, catches a spawn_id originating in Japan in NY. I understand it's a glitch but it also something ridiculously simple that Niantic could implement (if they don't already). I'm arguing that the assumption that it is "not confirmed" is not erring on the side of caution. Yes, I understand using a bot whatsoever is the MAIN way to get banned. What I am arguing is a level between utility and fun. Your sniper has fantastic utility, I will not deny that, but for me personally it disengages fun. I am only one person and I completely understand that the direction the bot will take will not hinge on a single member that admittedly has not contributed in the coding of the bot. That's fair.

`"I have yet to see someone else who have been using walking sniper. "`
That's why I'm here, I'M one of the people that uses the walking sniper and LOVES it! I like watching the logger click by, take a look at my scanner, see desired Pokemon, then use live config to enable MoveToMapPokemon and boom- my bot WALKS to the Pokemon and catches it. I've caught Snorlax this way and many Dratini. For me it's abhorrently boring watching webUI create a series of straight lines. I understand botting is to utilize exploits but it also fun to play. 'Luck' is a factor of the game and the surprise of what I was 'lucky' to catch is more stimulating than knowing when I come home I'll have 5 new Snorlax, 5 new Dragonite, etc etc. While your approach has maximum utility it minimizes the fun factor- there is little excitement in predictability. 

`To be honest, I dont even know why it was ever implemented`
I'm sorry but with this statement it is my opinion that you have come short of appreciating the insight and development of others and at times it appears you refuse to accept that someone else appreciates this train of thought. To me, this type of statement dismisses and diminishes the work of others and what others have come to enjoy about the bot (human-like function). An honest question- what is more important in the development of the bot: utility or enjoyability? If a happy medium of both then again I suggest to keep MoveToMap. 

It DOES make sense to me that people are being banned for botting, not sniping. At the same time, if spawn_ids are run against the catch location they simply will never be within a range that makes sense and I find that to be a simple solution for Niantic to detect botting. This change would force reliance upon a 'glitch'. Currently, walking-to-snipe doesn't rely on a glitch in the same sense. Why rely on a glitch when it is not necessary?

MoveToMapPokemon is currently working perfectly. Perhaps removing it should be saved for the time when it requires such overhaul to remain properly implemented no longer makes sense. Ease of use should be a factor but from my experience in Slack people that have a hard time utilizing pogom with the bot have a horrendous time getting bot to work to begin with simply due to not being familiar with any coding or command line whatsoever. This bot can be quite difficult to use for sure but it's pretty awesome! I also prefer to use my own computer because i have a lot of money in it haha. (on a side-note I would HAPPILY contribute my scanner data to something like maps.pikabot.org.)

It's nice that I would still be able to use pogom BUT I would no longer be able to use it in the fashion I have come to love. Again, I think the removal of MoveToMapPokemon will not increase flexibility but rather limit it, the opposite of your intention.

cbogithub:
now in brach dev  MoveToMapPokemon  not worked 
https://github.com/PokemonGoF/PokemonGo-Bot/issues/5378

pokemongo_bot/tree_config_builder.py
```diff
@@ -31,13 +31,16 @@ def build(self):
...
             task_type = task.get('type', None)
-            if task_type is None:
-                raise ConfigException('No type found for given task {}'.format(task))
-            elif task_type == 'EvolveAll':
-                raise ConfigException('The EvolveAll task has been renamed to EvolvePokemon')
-
             task_config = task.get('config', {})
 
+            # Validate task name refactors
+            if type is None:
```
sohje: type is None?

pokemongo_bot/tree_config_builder.py
```diff
@@ -31,13 +31,16 @@ def build(self):
...
+                raise ConfigException('No type found for given task {}'.format(type))
+            elif type == 'EvolveAll':
```
sohje: task_type you mean?

pokemongo_bot/cell_workers/sniper.py
```diff
@@ -0,0 +1,370 @@
...
+import os
+import time
+import json
+import requests
+
+from random import uniform
+from pokemongo_bot import inventory
+from pokemongo_bot.inventory import Pokemon, Pokemons
+from pokemongo_bot.base_task import BaseTask
+from pokemongo_bot.base_dir import _base_dir
```
sohje: _base_dir isnt used?

pokemongo_bot/cell_workers/sniper.py
```diff
@@ -0,0 +1,370 @@
...
+from pokemongo_bot.constants import Constants
```
sohje: is Constatns used?

pokemongo_bot/cell_workers/sniper.py
```diff
@@ -0,0 +1,370 @@
...
+from pokemongo_bot.worker_result import WorkerResult
+from pokemongo_bot.cell_workers.utils import format_dist, format_time, fort_details
+from pokemongo_bot.walkers.walker_factory import walker_factory
```
sohje: 15, 16 lines import

pokemongo_bot/cell_workers/sniper.py
```diff
@@ -0,0 +1,370 @@
+from __future__ import unicode_literals
+
+import os
+import time
+import json
```
sohje: os and json imports

pokemongo_bot/cell_workers/sniper.py
```diff
@@ -0,0 +1,370 @@
...
+from pokemongo_bot.worker_result import WorkerResult
+from pokemongo_bot.cell_workers.utils import format_dist, format_time, fort_details
+from pokemongo_bot.walkers.walker_factory import walker_factory
+from pokemongo_bot.cell_workers.pokemon_catch_worker import PokemonCatchWorker
+
+POKEBALL_ID = 1
+GREATBALL_ID = 2
+ULTRABALL_ID = 3
+
+class OrderMode():
```
sohje: remove parentheses or add (object)

pokemongo_bot/cell_workers/sniper.py
```diff
@@ -0,0 +1,370 @@
+from __future__ import unicode_literals
+
+import os
+import time
+import json
+import requests
+
+from random import uniform
+from pokemongo_bot import inventory
+from pokemongo_bot.inventory import Pokemon, Pokemons
```
sohje: Pokemon import

pokemongo_bot/cell_workers/sniper.py
```diff
@@ -0,0 +1,370 @@
...
+    VALUES = [IV, VIP, MISSING, THRESHOLD]
+    DEFAULT = [MISSING, VIP, THRESHOLD]
+
+class SnipingMode():
+    URL = 'url'
+    SOCIAL = 'social'
+    VALUES = [URL, SOCIAL]
+    DEFAULT = SOCIAL
+
+class ResponseMapper():
```
YvesHenri: Why is it strange? From what I've heard by other python fans, it's a good practice. Python is all based on dictionaries, so keeping a constant key name should make things less confusing for maintainances.

pokemongo_bot/cell_workers/sniper.py
```diff
@@ -0,0 +1,370 @@
...
+        targets = []
+
+        # Retrieve the targets
+        if self.mode == SnipingMode.SOCIAL:
+            targets = self._get_pokemons_from_social()
+        elif self.mode == SnipingMode.URL:
+            targets = self._get_pokemons_from_url()
+
+        # Order the targets (descending)
+        for attr in self.order_by:
```
YvesHenri: It does sort by the attributes the user has specified, if thats what youre asking. So yes, it sorts correctly (tested).

pokemongo_bot/cell_workers/sniper.py
```diff
@@ -0,0 +1,370 @@
...
+            # If mappings contain errors, log it instead of forwarding the exception
+            try:
+                self._validate_mappings(pokemon_dictionary_list[0])
+            except Exception as exception:
+                self._error(exception)
+                return pokemons
+
+        # Build up the pokemon. Pops are used to destroy random attribute names and keep the known ones!
+        for pokemon_dictonary in pokemon_dictionary_list:
+            pokemon_dictonary['iv'] = pokemon_dictonary.pop(self.mappings.get(ResponseMapper.IV), 0)
```
YvesHenri: Not sure again what you mean. This line will assign a value to the 'iv' key. This value will be the corrresponding to the IV mapping name that you have specified. Even if for some reason this key does not exist, it will return 0 and pop fill fail (do nothing). The pops are used to keep a clean dictionary with single keys. By doing that, we will avoid having something like that: { 'id': 143, 'pokemon_id': 143 ... }.

pokemongo_bot/tree_config_builder.py
YvesHenri: Havent actually changed this, except reordering the lines. Feel free to change.

pokemongo_bot/tree_config_builder.py
YvesHenri: After posting about the "if type is None" line, I checked the changes and I mightve probably confused about task_type and type vars, although I dont get any errors, but I'll look at it.

pokemongo_bot/cell_workers/sniper.py
YvesHenri: Havent changed this. Just kept the old imports. I might take it off, if its not really used, though.

pokemongo_bot/cell_workers/sniper.py
YvesHenri: Not sure. Havent changed this. Just kept the old imports. I might take it off, if its not really used, though.

pokemongo_bot/cell_workers/sniper.py
YvesHenri: ?

pokemongo_bot/cell_workers/sniper.py
YvesHenri: os is probably not used, indeed (no files). I might take it off.

pokemongo_bot/cell_workers/sniper.py
YvesHenri: Yea, its already fixed. There's a new commit comming soon. See below.

pokemongo_bot/cell_workers/sniper.py
YvesHenri: Correct. I might take it off.

YvesHenri:
So, I've still been working after the lastest commit for a new feature, on which you'll be able to specify multiple sources (urls). Lot's of fixes have been done. This PR can still be merged and will even fix issue #5378

cbogithub:
waiting for merger :+1:

pokemongo_bot/cell_workers/sniper.py
sohje: Oh, i was commenting from the phone. Its about line 258.
```python
if not pokemon_dictonary['name']
```
Now i see that pokemon_dict**i**onary should be filled with None or some positive value

pokemongo_bot/tree_config_builder.py
sohje: [built-in](https://docs.python.org/2/library/functions.html#type)

pokemongo_bot/cell_workers/sniper.py
sohje: not used i think

YvesHenri:
I'm closing this PR because I'm opening another soon, which will be the Sniper V2.1, on which will keep the old sniper and add the "multiple sources" feature to the new one (2.1).

Dr8g0n:
@YvesHenri Please consider the longer sleep time after teleporting to encounter the Pokemon. Currently, it sleeps only 3 second which is too short. See my test below with 2 bots running in parallel, trying to catch the same Pokemon, one with longer wait time at the destination successfully caught it. I reported in the issue and pull request but it was closed prematurely. My apology if you happened to read my comment again but longer sleep is important so people don't miss the rare Pokemons. Let me know if you need a reliable source of confirmed rare Pokemon to test with.

Thank you for an awesome work on sniping.

I tested with 2 bots:
# this one sleeps 3 secs (by default) and missed the Pokemon
[2016-09-26 20:19:52] [MoveToFort] [INFO] Moving towards pokestop USPS Sign - 0.05km
[2016-09-26 20:19:57] [    Sniper] [INFO] Teleporting to meet Growlithe (41.7646115485; -72.9637643552)...
[2016-09-26 20:19:57] [    Sniper] [INFO] Sleeping 3 secs during teleporting...
[2016-09-26 20:20:00] [    Sniper] [INFO] Damn! Its not here. Reasons: too far, caught, expired or fake data. Skipping...
[2016-09-26 20:20:00] [    Sniper] [INFO] Teleporting back to the old position (36.1876341128; -122.986435657)...

#This one sleeps a little bit more and caught the same exact Pokemon
[2016-09-26 20:20:06] [    Sniper] [INFO] Teleporting to meet Growlithe (441.7646115485; -72.9637643552)...
[2016-09-26 20:20:06] [    Sniper] [INFO] Sleeping 3 secs during teleporting...
[2016-09-26 20:20:09] [    Sniper] [INFO] Waiting 8 secs for the Pokemon to appear...
[2016-09-26 20:20:17] [    Sniper] [INFO] Yay! There really is a wild Growlithe nearby!
[2016-09-26 20:20:17] [    Sniper] [INFO] Teleporting back to the old position (36.1250140604; -121.805482928)...
[2016-09-26 20:20:17] [    Sniper] [INFO] Sleeping 3 secs during teleporting...
[2016-09-26 20:20:20] [PokemonCatchWorker] [INFO] *A wild Growlithe appeared!* (CP: 1086) (NCP: 0.8) (Potential 1.0) (A/D/S 15/15/15)
[2016-09-26 20:20:23] [PokemonCatchWorker] [INFO] This is a VIP pokemon. Catch!!!
[2016-09-26 20:20:26] [PokemonCatchWorker] [INFO] Threw a Razz Berry! Catch rate with Pokeball is now: 25.46
[2016-09-26 20:20:31] [PokemonCatchWorker] [INFO] Great throw! Used Ultraball, with chance 46.60 (15 left)
[2016-09-26 20:20:32] [PokemonCatchWorker] [INFO] Captured Growlithe! [CP 1086] [NCP 0.8] [Potential 1.0] [15/15/15] (7/800) [+150 exp] [+100 stardust]
[2016-09-26 20:20:32] [PokemonCatchWorker] [INFO] You now have 603 Growlithe candy!

PR #5256, opened 2016-09-06
EvolvePokemon strip spaces from Nidoran M/F

Gobberwart:
## Short Description:

The bot stores Nidoran M/F as "NidoranM"/"NidoranF" (no spaces). Simple fix to remove spaces for matching.

## Fixes/Resolves/Closes (please use correct syntax):
- #5239

Issue #5239, opened 2016-09-06
Not evolving Nidoran M/F


jannickv:
All pokemon in my evolve_list are being evolved except for the Nidoran. Guess it has something to do with the space.

```
{
        "type": "EvolvePokemon",
        "config": {
          "enabled": true,

          "evolve_list": "Pidgey, Caterpie, Weedle, Zubat, Rattata, Sandshrew, Drowzee, Nidoran M, Tentacool, Spearow, Psyduck",
          "donot_evolve_list": "none",

          "first_evolve_by": "cp",
          "evolve_above_cp": 0,
          "evolve_above_iv": 0,
          "logic": "or",
          "min_evolve_speed": 22,
          "max_evolve_speed": 30,
          "use_lucky_egg": false
        }
      },

```

Anakin5:
No it is correct with one space. You are missing the female versionthough:
`Nidoran M` `Nidoran F`

jannickv:
I don't want to evolve the female version so thats ok. But i checkend, and even when i added it, it doesn't evolve them.

Gobberwart:
Remove the space, eg. "NidoranM" not "Nidoran M". Even though the pokemon is called "Nidoran M" the bot strips out the space before comparing.

Gobberwart:
Putting through a PR to fix this now.

jannickv:
@Gobberwart thanks for your work!

Anakin5:
Note that PokemonOptimizer accept both !!

PR #5447, opened 2016-09-15
use strip instead of replace for names with spaces

ChaosMarc:
once again because my pull request is now at [PokemonGo-Bot-Backup](https://github.com/PokemonGoF/PokemonGo-Bot-Backup/pull/21).

The configuration of evolve_list and donot_evolve_list doesn't work for "Nidoran F" and "Nidoran M" because when splitting the commaseparated string to an array their names are completely stripped of whitespaces with .replace(" ", "") leading to "NidoranF" and "NidoranM". To prevent this .strip() should be used to trim only leading and trailing whitespaces.

see [DeXtroTip's comment](https://github.com/PokemonGoF/PokemonGo-Bot-Backup/pull/21#issuecomment-247088572):
> Before, it was using strip() and then someone said it didn't work because "Nidoran F" and "Nidoran M" names in bot are stored as "NidoranF" and "NidoranM", that's why it was change to replace().
**EDIT** - I've checked and the bot actually stores the names as "Nidoran F" and "Nidoran M" so it should simply use strip() and not replace()

### DIFF
--- a/pokemongo_bot/cell_workers/evolve_pokemon.py
+++ b/pokemongo_bot/cell_workers/evolve_pokemon.py
@@ -42,10 +42,10 @@ class EvolvePokemon(BaseTask):
 
     def _validate_config(self):
         if isinstance(self.evolve_list, basestring):
-            self.evolve_list = [str(pokemon_name).lower().replace(" ","") for pokemon_name in self.evolve_list.split(',')]
+            self.evolve_list = [str(pokemon_name).lower().strip() for pokemon_name in self.evolve_list.split(',')]
             
         if isinstance(self.donot_evolve_list, basestring):
-            self.donot_evolve_list = [str(pokemon_name).lower().replace(" ","") for pokemon_name in self.donot_evolve_list.split(',')]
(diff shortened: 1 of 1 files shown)


solderzzc:
👍

solderzzc:
Why all checks have passed, but the status is still yellow(Required statuses must pass before merging)

solderzzc:
👍

Gobberwart:
No. Replace was used specifically for Nidoran M/F which was being represented as NidoranM NidoranF (no spaces). If you change this, something will break. Give me a minute to get the specifics. Please do not merge!

Gobberwart:
Here is the PR where strip was changed to replace #5256
And here's the issue it addressed: #5239

The bot was unable to match Nidoran M/Nidoran F. Changing to replace allowed it to match NidoranM/NidoranF. Changing back to strip may reintroduce this issue (although it may have been fixed elsewhere since PR #5256)

Gobberwart:
OK, checked and someone has corrected the way pokemon names are stored in inventory, so this can now be reverted to strip.

solderzzc:
👍

alexyaoyang:
👍

PR #5454, opened 2016-09-15
Improve performance of cluster computation in CampFort task

Anakin5:
Improve performance of cluster computation in CampFort task.
Compute clusters only one time and update them afterwards.. It could take several seconds to compute each time we are looking for a camp.


Anakin5:
Also if somebody could give me explanation on what are those "event" and why I need to declare every single line of log I want to print, that would be nice.

javajohnHub:
What do you mean? The output should not have changed. Replaced self.bot.logger with an event.
I don't know what you think I did, but that is it. Very basic.

solderzzc:
👍

Issue #5457, opened 2016-09-15
Latest commit treats all pokemon it see as vip


mmixx:
### Expected Behavior

All below 1200 CP or 0.9 IV will not be treated as VIP pokemon.

### Actual Behavior

All pokemons it see... the bot treats it as VIP.

### Your FULL config.json (remove your username, password, gmapkey and any other private info)

http://pastebin.com/rqhT89Hp

### Output when issue occurred

```
[2016-09-16 02:43:18] [PokemonCatchWorker] [INFO] *A wild Venonat appeared!* (CP: 545) (NCP: 0.53) (Potential 0.64) (A/D/S 12/5/12)
[2016-09-16 02:43:21] [PokemonCatchWorker] [INFO] This is a VIP pokemon. Catch!!!
```
```
[2016-09-16 02:46:45] [PokemonCatchWorker] [INFO] *A wild Bellsprout appeared!* (CP: 268) (NCP: 0.24) (Potential 0.73) (A/D/S 14/9/10)
[2016-09-16 02:46:48] [PokemonCatchWorker] [INFO] This is a VIP pokemon. Catch!!!
```
```
[2016-09-16 02:50:29] [PokemonCatchWorker] [INFO] *A wild Venonat appeared!* (CP: 260) (NCP: 0.25) (Potential 0.51) (A/D/S 7/3/13)
[2016-09-16 02:50:32] [PokemonCatchWorker] [INFO] This is a VIP pokemon. Catch!!!
```

### Steps to Reproduce

Update to the latest dev commit and run the bot.

### Other Information
OS: Ubuntu 16.04
 
Branch: dev
 
Git Commit: a45943e
 
Python Version: 2.7.12
 
Any other relevant files/configs (eg: path files)

cowst:
I confirm

einszA:
same for me after pull from dev e209c2f82b17d418f6842772e1af8257a6a0868f

javajohnHub:
Happening to me also. I'm looking for the cause but cannot pinpoint it.

clubza1:
same here

MerlionRock:
Yup... every single one = VIP....

madzul:
same here

rawgni:
i think it's because you did not specify charge_attack / fast attack. so it assumes that all type of attacks are VIP.

madzul:
@rawgni why should we? isn't fast_/charge_attack is parameter in catch config? not in the vips?

[...]

PR #5479, opened 2016-09-16
Hot fix for pokemon_catch_worker.py

alexyaoyang:
- #5457

@DBa2016 If you do something like this: `catch_ncp = pokemon_config.get('catch_above_ncp', pokemon.cp_percent)`.
`pokemon.cp_percent >= catch_ncp:` is a tautology for all our users who don't have `catch_above_ncp ` set. E.g. some users have `any: {"always_catch" : true}`

I've set some fall-back values for you here and changed the logic a little (`catch_cp <= pokemon.cp <= catch_below_cp` seems to be a contradiction, I'm not sure). 

Feel free to edit them once you've found a stable fix. Please be more careful next time. :)

### DIFF
--- a/pokemongo_bot/cell_workers/pokemon_catch_worker.py
+++ b/pokemongo_bot/cell_workers/pokemon_catch_worker.py
@@ -283,18 +283,18 @@ class PokemonCatchWorker(BaseTask):
         if pokemon_config.get('always_catch', False):
             return True
 
-        catch_ncp = pokemon_config.get('catch_above_ncp', pokemon.cp_percent)
-        if pokemon.cp_percent >= catch_ncp:
+        catch_ncp = pokemon_config.get('catch_above_ncp', 0.8)
+        if pokemon.cp_percent > catch_ncp:
             catch_results['ncp'] = True
 
-        catch_cp = pokemon_config.get('catch_above_cp', pokemon.cp)
-        catch_below_cp = pokemon_config.get('catch_below_cp', pokemon.cp)
-        if catch_cp <= pokemon.cp <= catch_below_cp:
+        catch_cp = pokemon_config.get('catch_above_cp', 1200)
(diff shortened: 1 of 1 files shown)


Anakin5:
Why do we have default values ? If the parameter is not in configuration then the user is not interested to limit catch with these values, so they should not be taken into consideration.

sohje:
@alexyaoyang 330L in pokemon_catch_worker.py, remove it - treats all pokemons as vip

alexyaoyang:
@Anakin5 I'm not sure what you mean. We are retrieving from config based on key, it is always good to have a fall back value. Sure you can add checks to see if user have that key in config or not but at then end of the day, still good to have a fall back value.

@sohje Hot fix already fixed that bug 👍

Anakin5:
@alexyaoyang What I mean is that it looks like that, by default, we are not going to catch Pokemon with ncp,cp or iv below the values your set as default.

Anakin5:
If I get it correctly, following configuration:
```
"catch": {
     "any": {"catch_above_cp": 10, "catch_above_iv": 0.1, "logic": "and" }
}
```

will not catch pokemon with `ncp` less than 0.8, which is weird since it is nowhere mentioned in my config.

alexyaoyang:
@Anakin5 This is a hot fix. In my opinion, it's more important to have dev branch working (currently a lot of users have all pokemon encountered are VIP), rather than a new feature which may not have as many users. 

That's why I reverted part of his changes which broke the pokemon catch worker and told @DBa2016 to fix it as he wants, when he finds a stable fix.

Gobberwart:
I've done more work on this in #5494 which accounts for the default values issue as well as fixes the "is_vip" logic. Would prefer to use that instead of this as it's a more complete fix.

alexyaoyang:
@Gobberwart Awesome, closing this then 👍

Issue #5455, opened 2016-09-15
Regular Session stale, followed by heartbeat crash


Anakin5:
### Expected Behavior
No session stale. And, if it ever happen, reconnect and continue to work.

### Actual Behavior
Regular session stale and position is not updated anymore after reconnexion.

### Your FULL config.json (remove your username, password, gmapkey and any other private info)
```
{
    "auth_service": "ptc",
    "username": "***",
    "password": "***",
    "location": "***",
    "location_cache": true,
    "gmapkey": "***",
    "encrypt_location": "",
    "//websocket": {
        "start_embedded_server": true,
        "server_url": "127.0.0.1:4000"
    },
    "heartbeat_threshold": 10,
    "enable_social": false,
    "live_config_update": {
        "enabled": false,
        "tasks_only": false
    },
    "tasks": [
        {
            "type": "CompleteTutorial",
            "config": {
                "enabled": true,
                "nickname": "***",
                "team": 1
            }
        },
        {
            "type": "CollectLevelUpReward",
            "config": {
                "collect_reward": true,
                "level_limit": -1
            }
        },
        {
            "type": "IncubateEggs",
            "config": {
                "infinite_longer_eggs_first": false,
                "breakable_longer_eggs_first": true,
                "min_interval": 120,
                "infinite": [
                    10,
                    2,
                    5
                ],
                "breakable": [
                    10
                ]
            }
        },
        {
            "type": "PokemonOptimizer",
            "config": {
                "enabled": true,
                "min_slots_left": 5,
                "transfer": true,
                "transfer_wait_min": 3,
                "transfer_wait_max": 5,
                "evolve": true,
                "evolve_time": 25,
                "evolve_for_xp": true,
                "evolve_only_with_lucky_egg": false,
                "evolve_count_for_lucky_egg": 80,
                "may_use_lucky_egg": true,
                "upgrade": true,
                "upgrade_level": 60,
                "buddy": "Dragonite",
                "groups": {
                    "gym": [
                        "Dragonite",
                        "Snorlax",
                        "Lapras",
                        "Arcanine"
                    ]
                },
                "rules": [
                    {
                        "mode": "by_family",
                        "top": 1,
                        "sort": [
                            "iv"
                        ],
                        "evolve": {
                            "iv": 0.9
                        }
                    },
                    {
                        "mode": "by_family",
                        "top": 1,
                        "sort": [
                            "ncp"
                        ],
                        "evolve": {
[...]

PR #5482, opened 2016-09-16
Fix api issues when session stale

Anakin5:
Fix bot not working after a session stale. api should not be persisted anywhere else than in the bot object.

#5455


solderzzc:
👍

PR #5442, opened 2016-09-15
Py3convert

JSchwerberg:
## Short Description:
Update bot to be cross compatible with Python 3.5+ and Python 2.7+.  

Was already being tested at
https://github.com/PokemonGoF/PokemonGo-Bot-Backup/pull/26

## Fixes/Resolves/Closes (please use correct syntax):
N/A

PR #5490, opened 2016-09-16
Python 3 Compatibility

JSchwerberg:
## Short Description:
Reopen of PR #5442 giving Py3 compatibility.  PR wouldn't update thanks to problems with multiple forks of the repo.

N/A


solderzzc:
👍  There're many files changes, we should better merge first before others to get rid of conflict.
Is that possible  add pylint check for python3 format?

JSchwerberg:
pylint run in a python 3 virtual environment will check using Py3's formatting.  Not sure about running them side by side

solderzzc:
From searching, it's easier to add version number in the .travis.yml, need check if works as expect:
https://docs.travis-ci.com/user/languages/python
https://github.com/PokemonGoF/PokemonGo-Bot/blob/dev/.travis.yml

Issue #5491, opened 2016-09-16
catch_pokemon: catch filter does not work


liejuntao001:
I set to catch above 400 cp or above 0.85iv
but seems bot is catching all

### Your FULL config.json (remove your username, password, gmapkey and any other private info)

    "catch": {
      "any": {"catch_above_cp": 400, "catch_above_iv": 0.85, "logic": "or"},
...
    },

### Output when issue occurred

[2016-09-16 15:10:45] [PokemonCatchWorker] [INFO] *A wild Pidgeotto appeared!* (CP: 397) (NCP: 0.33) (Potential 0.51) (A/D/S 1
/13/9)
[2016-09-16 15:10:59] [PokemonCatchWorker] [INFO] Captured Pidgeotto! [CP 397] [NCP 0.33] [Potential 0.51] [1/13/9] (367/700) 
[+110 exp] [+100 stardust]

Branch: 
master
Git Commit: 
a8ee31256d412413b107cce81b62059634e8c802

liejuntao001:
I modified rule as such

      "any": {"candy_threshold" : 1 ,"catch_above_cp": 400, "catch_above_iv": 0.85, "logic": "or"},

now the filter seems working, but it do log spam:

[2016-09-16 15:23:14] [PokemonCatchWorker] [INFO] Skip ignored Slowpoke! (CP 205) (Potential 0.36) (A/D/S 4/10/2)
[2016-09-16 15:23:14] [PokemonCatchWorker] [INFO] Amount of candies for Slowpoke is 28, greater than threshold 1
[2016-09-16 15:23:14] [MoveToFort] [INFO] Moving towards pokestop Captain of the Helm - 0.28km
[2016-09-16 15:23:15] [PokemonCatchWorker] [INFO] Amount of candies for Slowpoke is 28, greater than threshold 1
[2016-09-16 15:23:16] [PokemonCatchWorker] [INFO] Amount of candies for Slowpoke is 28, greater than threshold 1
[2016-09-16 15:23:16] [PokemonCatchWorker] [INFO] Amount of candies for Slowpoke is 28, greater than threshold 1
[2016-09-16 15:23:17] [PokemonCatchWorker] [INFO] Amount of candies for Slowpoke is 28, greater than threshold 1
[2016-09-16 15:23:18] [PokemonCatchWorker] [INFO] Amount of candies for Slowpoke is 28, greater than threshold 1
[...]

Issue #5476, opened 2016-09-16
a8ee312, All pokemon appear as VIP


Pjumpod:
In latest update, a8ee312.

it seem all pokemon appears will be vip.

dogeatdog:
Hey, Pjumpod. We meet again :)

Yes, second that. I have very important pidgeys and rattattas as well :)

[2016-09-16 10:47:29] [MoveToFort] [INFO] Moving towards pokestop Public Clock - 0.10km
[2016-09-16 10:47:35] [MoveToFort] [INFO] Moving towards pokestop Public Clock - 0.07km
[2016-09-16 10:47:36] [PokemonCatchWorker] [INFO] *A wild Rattata appeared!* (CP: 186) (NCP: 0.32) (Potential 0.84) (A/D/S 8/15/15)
[2016-09-16 10:47:39] [PokemonCatchWorker] [INFO] This is a VIP pokemon. Catch!!!
[2016-09-16 10:47:44] [PokemonCatchWorker] [INFO] Threw a Razz Berry! Catch rate with Pokeball is now: 64.82
[2016-09-16 10:47:53] [PokemonCatchWorker] [INFO] OK throw! Used Ultraball, with chance 101.63 (98 left)
[2016-09-16 10:47:54] [PokemonCatchWorker] [INFO] Captured Rattata! [CP 186] [NCP 0.32] [Potential 0.84] [8/15/15] (404/800) [+100 exp] [+100 stardust]
[2016-09-16 10:47:54] [PokemonCatchWorker] [INFO] You now have 7 Rattata candy!
[2016-09-16 10:47:59] [RecycleItems] [INFO] Forcing item recycle based on schedule
[2016-09-16 10:47:59] [RecycleItems] [INFO] Next forced item recycle at 10:50:12
[2016-09-16 10:48:15] [ItemRecycler] [INFO] Discarded 1x Max Potion.
[2016-09-16 10:48:20] [ItemRecycler] [INFO] Discarded 1x Revive.
[2016-09-16 10:48:20] [MoveToFort] [INFO] Moving towards pokestop Public Clock - 0.07km
[2016-09-16 10:48:21] [IncubateEggs] [INFO] Eggs incubating: [1.70/5.0 km] (Eggs left: 8, Incubating: 1)
[...]

PR #5494, opened 2016-09-17
Bot treats all pokemon as VIP fix

Gobberwart:
## Short Description:

Bot treats all pokemon as VIP. Additionally, default rules always return true if not configured. For example:

```
    "catch": {
      "any": {"catch_above_cp": 500, "catch_above_iv": 0.65, "logic": "or"}
    },
```

Given example pokemon with cp 30, iv 0.2 and ncp 0.15, the bot would still always catch it because no "catch_above_ncp" rule is defined. This is incorrect behaviour, should return false if no condition matches or condition doesn't exist.

- #5457 
- #5476 
- #5491

### DIFF
--- a/pokemongo_bot/cell_workers/pokemon_catch_worker.py
+++ b/pokemongo_bot/cell_workers/pokemon_catch_worker.py
@@ -51,6 +51,7 @@ class PokemonCatchWorker(BaseTask):
     def __init__(self, pokemon, bot, config):
         self.pokemon = pokemon
         super(PokemonCatchWorker, self).__init__(bot, config)
+        if self.config.get('debug', False): DEBUG_ON = True
 
     def initialize(self):
         self.position = self.bot.position
@@ -264,6 +265,8 @@ class PokemonCatchWorker(BaseTask):
             'ca': True
         }
 
+        catch_logic = pokemon_config.get('logic', default_logic)
+
         candies = inventory.candies().get(pokemon.pokemon_id).quantity
         threshold = pokemon_config.get('candy_threshold', -1)
         if (threshold > 0 and candies >= threshold):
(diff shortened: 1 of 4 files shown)


Gobberwart:
Replaces #5479

Anakin5:
`catch_results` are initialized to False. So if the don't have any rule on `cp`, `catch_results["cp"]` will be False, right ? 
So with logic at `and`, no Pokemon will get caught, right ?

pokemongo_bot/cell_workers/pokemon_catch_worker.py
```diff
@@ -44,6 +44,7 @@
     'andor': lambda x, y, z: x and y or z
 }
 
+DEBUG_ON = False
```
alexyaoyang: This is probably just a small issue, but should we introduce another debug flag here? We already have a debug flag in config now.

Gobberwart: this flag only applies to this rather than using global debug which generates a lot of garbage

Gobberwart:
@Anakin5 That's correct as it stands... The bot really shouldn't use a rule if it's not defined in config. Let me take a look.

pokemongo_bot/cell_workers/pokemon_catch_worker.py
alexyaoyang: Hmm ok noted

Gobberwart:
@Anakin5 Added some logic to handle different default values based on catch logic. It's mucky but should account for all configurations.

Having said that, the current LOGIC_TO_FUNCTION only works if there are exactly three values. Would be much better if we could only compare what's actually set in config and ignore whatever is not set. 

Leaving as-is for now so we can make the bot work properly again.

pokemongo_bot/cell_workers/pokemon_catch_worker.py
Gobberwart: It's not unusual. A lot of other code has the same thing. It's actually really handy when debugging a single module.

Aelfwin1996:
please merge these changes to master branch. I have updated every day since these commits but still have this issue on master branch

Issue #5480, opened 2016-09-16
"release_below_iv" feature doesn't work


Pjumpod:
I try this feature to automatic release monster and it doesn't work.
between I use "keep_best_iv: 1" feature for work around.

mmixx:
Mine didn't work too with the latest update...

```
[2016-09-16 23:37:08] [PokemonCatchWorker] [INFO] Pokemon went out of range!
[2016-09-16 23:37:09] [CatchPokemon] [INFO] Lured pokemon at fort Soul Waves (c49f13f1914644c7a0bdf6293362c3df.11)
[2016-09-16 23:37:09] [PokemonCatchWorker] [INFO] Your Pokemon inventory is full! Could not catch!
[2016-09-16 23:37:10] [PokemonCatchWorker] [INFO] Pokemon went out of range!
[2016-09-16 23:37:10] [PokemonCatchWorker] [INFO] Your Pokemon inventory is full! Could not catch!
[2016-09-16 23:37:10] [CatchPokemon] [INFO] Lured pokemon at fort Soul Waves (c49f13f1914644c7a0bdf6293362c3df.11)
[2016-09-16 23:37:10] [PokemonCatchWorker] [INFO] Pokemon went out of range!
[2016-09-16 23:37:11] [PokemonCatchWorker] [INFO] Your Pokemon inventory is full! Could not catch!
[2016-09-16 23:37:12] [PokemonCatchWorker] [INFO] Your Pokemon inventory is full! Could not catch!
[2016-09-16 23:37:13] [CatchPokemon] [INFO] Lured pokemon at fort Soul Waves (c49f13f1914644c7a0bdf6293362c3df.11)
[2016-09-16 23:37:13] [PokemonCatchWorker] [INFO] Your Pokemon inventory is full! Could not catch!
[2016-09-16 23:37:14] [PokemonCatchWorker] [INFO] Pokemon went out of range!
[2016-09-16 23:37:14] [PokemonCatchWorker] [INFO] Your Pokemon inventory is full! Could not catch!
[2016-09-16 23:37:15] [CatchPokemon] [INFO] Lured pokemon at fort Soul Waves (c49f13f1914644c7a0bdf6293362c3df.11)
[2016-09-16 23:37:15] [PokemonCatchWorker] [INFO] Your Pokemon inventory is full! Could not catch!
[...]

PR #5499, opened 2016-09-17
Fixed release trigger

Gobberwart:
## Short Description:

On TransferPokemon task, the bot randomizes min_free_slot by up to 20 (to simulate human-like behaviour?), so a value of 5 in config could actually be anything from 5 to -15. Bag limit never gets below zero, so bag has very large chance of filling up. Changed so will trigger if space <= 1 if random gives anything equal to or below 0.

- #5480


Gobberwart:
Very minor change, merging.

PR #5513, opened 2016-09-17
fix add pokemon

rawgni:
## Short Description:

We were comparing a literal to a json object which in our case is always True.
This causes self.pokemon to contains duplicate. Thus, creating multiple encounter api call to the server for the same pokemon.

-
-
-


solderzzc:
👍

PR #5529, opened 2016-09-18
Fix some cpm calculation issues

Anakin5:
Just found some issues with cp multiplier computation while doing something else.

It does not change anything functionally except the fact that you cannot pass a string as parameter of the 2 cp multiplier functions anymore.


solderzzc:
👍  Math is magic for me.

Issue #5527, opened 2016-09-18
Slow bot startup after telegram is enabled


gengstah:
### Expected Behavior

The bot used to start fast.

### Actual Behavior

After I enabled telegram, the bot startup is super slow.

### Your FULL config.json (remove your username, password, gmapkey and any other private info)

[config](http://pastebin.com/5XytpaT9)

### Output when issue occurred

Please note the time differences of each log line. Mostly ten seconds or more. Before enabling telegram, I had no problem with this. The bot just starts fine and fast.
[logs](http://pastebin.com/SKJ6CzcD)

### Steps to Reproduce

Just run the bot. I specifically run the bot with these features enabled:

- Social

- FollowPath

- MoveToMap

- PokemonOptimizer

- Telegram

### Other Information
OS: Kali Linux (Debian Jessie)
 
Branch: dev
 
Git Commit: 19639712f9a72ce056bedd5b0664d5e717e02db1
 
Python Version: Python 2.7.9
 
Any other relevant files/configs (eg: path files) 

[Path Config](http://pastebin.com/yejKz4wU)

gengstah:
This slowdown might be caused by slow internet connection. I take that the TelegramTask sends one message at a time, no? So for each log message, the bot sends an api request to telegram and if internet connection is sluggish, then the bot is also sluggish. Can someone confirm this? Maybe we can have an option to batch send logs to minimize bandwidth consumption.

javajohnHub:
I think there are more events that can be removed from the list. I'll check it out today.

javajohnHub:
This should fix it I think... I got rid of all login events in telegram so it should only start sending msgs after the bot starts. Avoiding any bog at startup.

javajohnHub:
Sorry @Gobberwart wants to display events that dont fire in telegram instead of accepting this fix.

Gobberwart:
Not quite. I said I don't think removing events will fix slow startup, but I agreed with removing imports. I never said to close the PR, just maybe put a bit more thought/testing into it.

Also.. what do you mean "don't fire in telegram"? Why don't they fire? And how do they display if they don't fire?

Gobberwart:
Also worth noting that the section where you're removing events doesn't even run until the bot goes to generate a list of available events (in response to /events), and takes microseconds to run. All you're doing is hiding events from the available list. Should have zero impact on performance.

javajohnHub:
[...]

PR #5532, opened 2016-09-18
Telegram-startup optimization

javajohnHub:
## Short Description:
Small change to telegram to remove more unneeded events that could be bogging the bot at startup
-possibly #5527
-
-

### DIFF
--- a/pokemongo_bot/event_handlers/__init__.py
+++ b/pokemongo_bot/event_handlers/__init__.py
@@ -4,3 +4,4 @@ from .socketio_handler import SocketIoHandler
 from .social_handler import SocialHandler
 from .telegram_handler import TelegramHandler
 from .discord_handler import DiscordHandler
+from .chat_handler import ChatHandler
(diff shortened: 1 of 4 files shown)


Gobberwart:
My brizzanch? LOL..izzle

I totally agree with removing unnecessary imports, but removing events... not so much. I'm really not sure they have any impact on startup performance, and people might want to know about them.

javajohnHub:
You can test pre removal vs post and see a drastic improvement, but the decision is yours. Thanks for the review.

javajohnHub:
And there is not really a point in knowing about events in telegram that do not produce output anyway.

Gobberwart:
Not sure I agree with removing events, as (a) they shouldn't have any impact on startup - or really anything since they only trigger if the event occurs, and telegram is running in a separate thread, and (b) some people might actually want to see those events.

Removing unneeded imports is a great idea, so long as you're absolutely sure they're not needed - and they shouldn't be since the functionality was moved to chat_handler.py

javajohnHub:
fair enough.

PR #5540, opened 2016-09-19
Give a breakdown of the remaining eggs instead of just showing many e…

kolinkorr839:
## Short Description:

In the logs, when IncubateEggs task is enabled, you can see something like this. The problem with this is that I don't know how many 10km/5km/2km eggs I have.

`[IncubateEggs] [INFO] Eggs incubating: [4.38/5.0 km] (Eggs left: 8, Incubating: 1)`

The change will allow to see a breakdown of how many of each particular egg is left, eg. 5km has 8 eggs.

`[IncubateEggs] [INFO] Eggs incubating: [4.38/5.0 km] (Eggs left: [(5.0, 8)], Incubating: 1)`

### DIFF
--- a/pokemongo_bot/cell_workers/incubate_eggs.py
+++ b/pokemongo_bot/cell_workers/incubate_eggs.py
@@ -1,3 +1,4 @@
+from collections import Counter
 from datetime import datetime, timedelta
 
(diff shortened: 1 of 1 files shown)


solderzzc:
👍

Issue #5523, opened 2016-09-18
Snipe using Pokemongo-Map not working


Gobberwart:
### Expected Behavior
If a snipeable pokemon exists in Pokemongo-Map data, teleport and snipe.

### Actual Behavior
Two issues:

1. Snipeable pokemon sometimes being reported as not-existing (even though still a long time to go before disappear)
2. If it does manage to find the pokemon, it does nothing after teleporting back.

First issue seems to be in this code:

```
                if nearby_pokemon_id == pokemon['pokemon_id']:
                    exists = True

                    # Also, if the IDs arent valid, update them!
                    if not pokemon['encounter_id'] or not pokemon['spawnpoint_id']:
                        pokemon['encounter_id'] = nearby_pokemon['encounter_id']
                        pokemon['spawn_point_id'] = nearby_pokemon['spawn_point_id']
                        pokemon['disappear_time'] = nearby_pokemon['last_modified_timestamp_ms'] if is_wild else nearby_pokemon['expiration_timestamp_ms']
                    break
```

"pokemon" is `{u'disappear_time': 1474172861, u'dist': 28654.516924508378, u'name': u'Rattata', u'pokemon_id': 19, u'spawnpoint_id': u'6ad63aea37d', u'longitude': 145.29003085906515, u'iv': 0, u'priority': 10, u'pokemon_rarity': u'Common', u'pokemon_types': [{u'color': u'#8a8a59', u'type': u'Normal'}], u'is_vip': False, u'pokemon_name': u'Rattata', u'latitude': -37.8142364646841, u'spawn_point_id': u'6ad63aea37d', u'encounter_id': u'MzgxNzg3NTgwMDIxNzU5NzEzMw=='}`

"nearby_pokemon" is `{'last_modified_timestamp_ms': 1474173240168L, 'longitude': 145.29339957978198, 'pokemon_data': {'pokemon_id': 19}, 'latitude': -37.81053711254291, 'spawn_point_id': u'6ad63abfc1b', 'encounter_id': 14315303638581835213L, 'time_till_hidden_ms': 469092}`
nearby_pokemon does not include 'expiration_timestamp_ms' if sniping so no match given

Second issue is unknown at this stage. I've spent a bit of time on this to identify issue, but have to attend to family stuff now, so if anyone wants to run with this, go ahead. Otherwise, I'll get back to it tomorrow.

### Your FULL config.json
[...]

PR #5542, opened 2016-09-19
MoveToMapPokemon sniping fix

Gobberwart:
## Short Description:

Two issues with MoveToMapPokemon:

- "Does not exist anymore" error

Fixed by changing "spawnpoint_id" to "spawn_point_id"

- If above doesn't occur, doesn't catch anything, just teleports back then moves on

Fixed by removing try block from pokemon_catch_worker. It is possible that a different exception MIGHT occur here now, but I haven't seen it, and it's better than doing nothing.

- #5523


YvesHenri:
How does removing a try/block fixes it? The "does not exist" have absolutely nothing to do with the "spawn_point_id" key.

PR #5536, opened 2016-09-18
Fixing issues in event fitlering (/events command in chat handler)

DBa2016:
## Short Description:
When sending "/events filter" to the bot, 2 issues occured:
1) the regular expression applied for filtering was ".*filter-*" instead of ".*filter.*"
2) removing apparently unused events was causing exceptions because these events were not in the list anymore after filtering

PR #5544, opened 2016-09-19
Fixing issues in event fitlering (/events command in chat handler)

DBa2016:
## Short Description:
- fixes issues in filtering of /events
- fixes issues when calling /info and no visited pokestops or caught pokemons counts are available
- minor reformatting of subscriptions shown by telegram handler

### DIFF
--- a/pokemongo_bot/event_handlers/chat_handler.py
+++ b/pokemongo_bot/event_handlers/chat_handler.py
@@ -26,9 +26,9 @@ class ChatHandler:
                     "*" + self.bot.config.username + "*",
                     "_Level:_ " + str(stats["level"]),
                     "_XP:_ " + str(stats["experience"]) + "/" + str(stats["next_level_xp"]),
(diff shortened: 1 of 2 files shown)


DBa2016:
Replaces #5536

DBa2016:
They were probably hidden because they are never really raised?...

I don't know, to be honest, why they were discarded at all.

javajohnHub:
I thought they were effecting startup and bogging...but i was wrong...@gobberwart fix is correct.

Gobberwart:
Correct, removing events has no effect at all if they're not actually being generated, and no effect on startup in any case. Suggest just not removing any events (although we could perhaps look at why these events even exist and remove the from the bot if they're not used)

DBa2016:
Okay, I removed the code discarding events in the list completely.

DBa2016:
Come on guys...

PR #5554, opened 2016-09-20
Updated Sniper docs, config.map.example and config.example files.

YvesHenri:
Minor Sniper fixes (including docs).


configs/config.json.example
```diff
@@ -367,16 +432,10 @@
...
-        "Moltres": {},
-        "Zapdos": {},
-        "Articuno": {},
 
         "// S-Tier pokemons (if pokemon can be evolved into tier, list the representative)": {},
-        "Mewtwo": {},
         "Dragonite": {},
         "Snorlax": {},
-        "// Mew evolves to Mewtwo": {},
-        "Mew": {},
```
Gobberwart: Why removing these? I know nobody's ever seen them, but what if one day... and the bot fails to snipe them :)

configs/config.json.example
```diff
@@ -1,7 +1,7 @@
 {
     "websocket_server": false,
     "heartbeat_threshold": 10,
-    "enable_social": true,
+    "enable_social": false,
```
solderzzc: You need keep this option as `true`.

cbogithub:
hi could you add   "enabled": true, to url  same below?

```
+				{  
                                         "enabled": true,
+					"url": "https://pokewatchers.com/grab/",
+					"mappings": {
+						"iv": { "param": "iv" },
+						"id": { "param": "pid" },
+						"name": { "param": "pokemon" },
+						"latitude": { "param": "cords" },
+						"longitude": { "param": "cords" },
+						"expiration": { "param": "timeend", "format": "milliseconds" }
+					}
```

Gobberwart:
good idea @cbogithub, but after the url not before it. That would let us keep semi-complicated configs in the file without having to delete them to disable.

Gobberwart:
Sorry about the conflicts @YvesHenri. I fixed that vip thing which seems to have upset this.

cbogithub:
and add timeout to connect for each url which want to snip because each url have different time to connect it ex https://pokewatchers.com/grab/ need 10sec - 15sec to respond a result . i think 7sec default not enough :


```

+				{  
                                         "enabled": true,
+					"url": "https://pokewatchers.com/grab/",
                                        "timeout": 30 // second,minute...
+					"mappings": {
+						"iv": { "param": "iv" },
+						"id": { "param": "pid" },
+						"name": { "param": "pokemon" },
+						"latitude": { "param": "cords" },
+						"longitude": { "param": "cords" },
+						"expiration": { "param": "timeend", "format": "milliseconds" }
+					}
```

Gobberwart:
@cbogithub In the interest of letting this PR get merged without too much complication, I recommend you open a new issue as a feature request with this stuff in it.

configs/config.json.map.example
```diff
@@ -455,22 +455,43 @@
...
-			"bullets": 3,
+			"bullets": 1,
 			"homing_shots": true,
 			"special_iv": 100,
-			"order": ["missing", "iv", "priority", "vip", "expiration_timestamp_ms"],
+			"order": ["missing", "vip", "priority"],
 			"sources": [
 				{
 					"url": "http://localhost:5000/raw_data",
 					"key": "pokemons",
```
DBa2016: Please add `"enabled": false`

I suggest disabling all sources by default, allowing user to enable exactly what he wants to

configs/config.json.map.example
```diff
@@ -455,22 +455,43 @@
...
-						"iv": { "param": "iv" },
 						"id": { "param": "pokemon_id" },
 						"name": { "param": "pokemon_name" },
 						"latitude": { "param": "latitude" },
 						"longitude": { "param": "longitude" },
 						"expiration": { "param": "disappear_time", "format": "milliseconds" }
 					}
+				},
+				{
+					"url": "https://pokewatchers.com/grab/",
```
DBa2016: add `"enabled": false,` here, same reason

configs/config.json.map.example
```diff
@@ -455,22 +455,43 @@
...
+						"iv": { "param": "iv" },
+						"id": { "param": "pid" },
+						"name": { "param": "pokemon" },
+						"latitude": { "param": "cords" },
+						"longitude": { "param": "cords" },
+						"expiration": { "param": "timeend", "format": "milliseconds" }
+					}
+				},
+				{
+					"url": "http://pokesnipers.com/api/v1/pokemon.json",
```
DBa2016: add `"enabled": false,` here, same reason

configs/config.json.example
YvesHenri: Those just appear in game events, which will me announced. Untill then, people are just sending fake reports to the broker, which triggers the sniper and waste a teleport.

YvesHenri:
At everyone requesting changes: this PR is about updating the docs/configs **ONLY**, since many people have been asking on slack about it. Create an issue and request features instead.

@cbogithub No. If relying solely on me, I will not do this. Like I said before, I want to keep the configs clean and simple. Setting timeout/fetch times will only trigger the possibility of more issues/nightmares. All the default times are good enough. I came to this conclusion after **A LOT** of tests. You are supposed to find those sources yourself and set them up, instead of relying on me. That's why this sniper has been made for.

YvesHenri:
Issues are already coming up: #5560.

configs/config.json.map.example
YvesHenri: Added.

YvesHenri:
Couple of updates:

1. Fixed the potential bug with homing shots under social mode
2. Fixed the bug where some pokemons were being classified as VIPs when they really werent (Gobb pushed a fix for this, but it is 'wrong' and was just a temporary one)
3. Increased the validation timeout from 10 to 7: this IS NOT going to be customizable!!!
4. Added the "enabled" flag for sources
5. Sniper's config validation wont be triggered if it is disabled

YvesHenri:
Conflicts resolved.

configs/config.json.example
Gobberwart: Yeah... not sure I agree with this approach. Those are definitely VIP pokemon and if an event happens while I'm not paying attention, I'd like sniper to catch them. Also, things other than Sniper use vip pokemon list.

DBa2016: Agree with @Gobberwart

docs/configuration_files.md
```diff
@@ -735,21 +735,25 @@ This task is an upgrade version of the MoveToMapPokemon task. It will fetch poke
 ### Options
 [[back to top](#table-of-contents)]
 
+* `sources` - Defines whether the **WHOLE** task is enabled or not. Please bear in mind that even if the task is enabled, all or any of its sources can be disabled. (default: false)
```
kureijir: Should this be `enabled` instead of `sources`?

Gobberwart: Good pickup

YvesHenri: "_`sources` - Defines whether the **WHOLE** task is enabled or not. Please bear in mind that even if the task is enabled, all or any of its sources can be disabled. (default: false)_" and "_`sources.enabled` - Defines whether this source is enabled or not. This has nothing to do with the task's `enabled`_"are two things completely different. Does this really sound confusing to you? Really?

YvesHenri:
Like I said before, those legendaries will not appear in game anytime soon. They're only catchable with masterballs and those dont even exist in the game yet (**acquirable**). It doesnt make any sense to have them on VIP while people can fuzz any of the sniping services out there (mqtt/broker, pokesnipers, slack/discord chats etc) like I've seen many times. Encounter/catch that use this fake data can easily lead users to a ban. I dont agree that these changes should hang this PR when there is clearly a potential bug regarding this task and social mode and many other fixes/enhancements done so far.

I will not make anymore changes, not even the ones requested by @Gobberwart. You can either accept this and make another PR customizing the config file the way you find correct or you can talk to @solderzzc about handling wrong data on the mqtt/broker side (might as well talk to all the other sniping source providers).

docs/configuration_files.md
kureijir: It does, with line 752, not 755.

Gobberwart: You've missed the point @YvesHenri . There is a line in configuration_files.md which reads 'sources' and it should read 'enabled'.

YvesHenri: Finally got it. Sorry

configs/config.json.example
Gobberwart: In the interest of getting this PR through, I'll let this one go. If anything happens as a result of it, we know where to find you ;)

Gobberwart: So, this is the only thing holding this PR up now? Bugger. I'll approve it if you change it back, otherwise... someone else will have to make the call.

configs/config.json.example
```diff
@@ -382,16 +450,10 @@
         "Any pokemon put here directly force to use Berry & Best Ball to capture, to secure the capture rate": {},
         "any": {"catch_above_cp": 1200, "catch_above_iv": 0.9, "logic": "or" },
         "Lapras": {},
-        "Moltres": {},
```
solderzzc: It doesn't hurt anything to place these inside an example file. So we should better keep it as it.

Gobberwart:
Approving/merging as-is since we can't reach agreement on vip list and that's the only thing holding this up. New PR to follow restoring vip list to how it currently is.

Issue #5556, opened 2016-09-20
Sniper is assuming false VIPs on social mode


YvesHenri:
It seems to be reporting false VIPs when under social mode. I'll investigate it tomorrow because I'm sick and my head hurts right now. You can close all the other issues about this.

Gobberwart:
You sure it's vip? could be any one of these conditions in here:

```
        if pokemon.get('pokemon_name', '') not in self.catch_list:
            # This is not in the catch list. Lets see if its a VIP one
            if not pokemon.get('vip'):
                # It is not a VIP either. Lets see if its IV is good (if any)
                if pokemon.get('iv', 0) < self.special_iv:
                    self._trace('{} is not listed to catch, nor a VIP and its IV sucks. Skipping...'.format(pokemon.get('pokemon_name')))
                    return False
```

Gobberwart:
I think this affects all modes, just that most urls don't contain non-vip pokemon :)

Gobberwart:
Never mind, it's VIP. Fixed in #5557, you can stop worrying about it now :)

PR #5557, opened 2016-09-20
Fix incorrect VIP in Sniper

Gobberwart:
## Short Description:

Fixes incorrect identification of a vip in Sniper

- #5556


Gobberwart:
Very minor change, merging.

Issue #5563, opened 2016-09-20
When optimizer tried to transfer it crashes bot


beltoft:
### Expected Behavior
optimizer runs

### Actual Behavior
bot crashes when bag limit is reached

### Your FULL config.json (remove your username, password, gmapkey and any other private info)
	{
			"type": "PokemonOptimizer",
			"config": {
				"enabled": true,
				"min_slots_left": 5,
        "action_wait_min": 2,
        "action_wait_max": 4,
        "transfer": true,
				"evolve": true,
        "evolve_to_final": false,
				"evolve_time": 20,
				"evolve_for_xp": true,
				"evolve_only_with_lucky_egg": false,
				"evolve_count_for_lucky_egg": 60,
				"may_use_lucky_egg": true,
				"upgrade": true,
				"upgrade_level": 64,
        "groups": {
					"gym": ["Dragonite", "Snorlax", "Lapras", "Arcanine"]
				},
				"rules": [
					{
						"mode": "by_family",
						"top": 1,
						"sort": ["iv"],
						"evolve": {"iv": 1.0},
						"upgrade": {"iv": 1.0}
					},
					{
						"mode": "by_family",
						"top": 1,
						"sort": ["iv"],
						"evolve": {"iv": 0.9}
					},
                    
          {
						"mode": "by_family",
						"top": 1,
						"sort": ["ncp"],
						"evolve": {"ncp": 0.9}
						
					}
          
				]
			}
		},

### Output when issue occurred

[2016-09-20 13:37:14] [PokemonOptimizer] [INFO] Pokemon Bag: 246 / 250
[2016-09-20 13:37:15] [       cli] [INFO]
[2016-09-20 13:37:15] [       cli] [INFO] Ran for 0:00:12
[2016-09-20 13:37:15] [       cli] [INFO] Total XP Earned: 0  Average: 0.00/h
[2016-09-20 13:37:15] [       cli] [INFO] Travelled 0.00km
[2016-09-20 13:37:15] [       cli] [INFO] Visited 0 stops
[2016-09-20 13:37:15] [       cli] [INFO] Encountered 0 pokemon, 0 caught, 0 released, 0 evolved, 0 never seen before ()
[2016-09-20 13:37:15] [       cli] [INFO] Threw 0 pokeballs
[2016-09-20 13:37:15] [       cli] [INFO] Earned 0 Stardust
[2016-09-20 13:37:15] [       cli] [INFO] Hatched eggs 0
[...]

Issue #5492, opened 2016-09-16
PokemonOptimizer it says Evolving the best, but apparently not


liejuntao001:
### Your FULL config.json (remove your username, password, gmapkey and any other private info)

       {
            "type": "PokemonOptimizer",
            "config": {
                "enabled": true,
                "min_slots_left": 5,
                "transfer": true,
                "transfer_wait_min": 4,
                "transfer_wait_max": 6,
                "evolve": true,
                "evolve_time": 25,
                "evolve_for_xp": false,
                "evolve_only_with_lucky_egg": false,
                "evolve_count_for_lucky_egg": 72,
                "may_use_lucky_egg": false,
                "upgrade": true,
                "upgrade_level": 60,
                "groups": {
                    "gym": ["Dragonite", "Snorlax", "Lapras", "Arcanine"],
                    "my_love": ["Pikachu"],
                    "vip": ["Exeggutor", "Gyarados", "Poliwrath", "Vaporeon", "Muk", "Venusaur", "Victreebel", "Dragonite", "Snorlax", "Lapras", "Arcanine", "Flareon", "Magmar", "Slowbro", "Machamp", "Blastoise", "Charizard", "Nidoqueen", "Vileplume"]
                },
                "rules": [
                    {
                        "mode": "by_pokemon",
                        "names": ["vip"]
                    },
                    {
                        "mode": "by_family",
                        "top": 1,
                        "sort": ["iv"],
                        "evolve": {"iv": 0.9}
                    },
                    {
                        "mode": "by_family",
                        "top": 1,
                        "sort": ["ncp"],
                        "evolve": {"ncp": 0.9}
                    },
                    {
                        "mode": "by_family",
                        "top": 1,
                        "sort": ["cp"]
                    },
                    {
                        "mode": "by_family",
                        "names": ["my_love"]
                    }
                ]
            }
        },

### Output when issue occurred

[2016-09-16 15:34:09] [PokemonGoBot] [INFO] #4 Charmander[2][30 candies]: (CP 507, IVCP 0.18, DPS 12.16, Moves: [Ember, Flamethrower]) | (CP 330, IVCP 0.41, DPS 15.12, Moves: [Scratch, Flamethrower])
[2016-09-16 15:34:09] [PokemonGoBot] [INFO] #14 Kakuna[1][57 candies]: (CP 259, IVCP 0.81, DPS 9.83, Moves: [Bug Bite, Struggle])

[...]

PR #5564, opened 2016-09-20
Fix crash with upgrade pokemon

Anakin5:
- Fix crash when assigning buddy
- Fix crash when upgrading pokemon
- Fix documentation and configuration example issue: evolve is true by default.

#5563 #5567 #5565 #5492

### DIFF
--- a/pokemongo_bot/cell_workers/pokemon_optimizer.py
+++ b/pokemongo_bot/cell_workers/pokemon_optimizer.py
@@ -56,11 +56,11 @@ class PokemonOptimizer(BaseTask):
         self.config_upgrade = self.config.get("upgrade", False)
         self.config_upgrade_level = self.config.get("upgrade_level", 30)
         self.config_groups = self.config.get("groups", {"gym": ["Dragonite", "Snorlax", "Lapras", "Arcanine"]})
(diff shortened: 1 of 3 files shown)


pokemongo_bot/cell_workers/pokemon_optimizer.py
```diff
@@ -508,8 +508,8 @@ def get_evolution_plan(self, family_id, family_list, keep, try_evolve, try_upgra
             full_upgrade_candy_cost = 0
             full_upgrade_stardust_cost = 0
 
-            for i in range(pokemon.level, upgrade_level, 0.5):
-                upgrade_cost = self.pokemon_upgrade_cost[2 * (i - 1)]
+            for i in range(int(pokemon.level * 2), int(upgrade_level * 2)):
+                upgrade_cost = self.pokemon_upgrade_cost[i - 2]
```
DBa2016: what kind of crash is it fixing? I don't see any difference between old and new code - it looks exactly the same (steps are twice as big, but "*2" is omitted when the couner variable is used, so...)

Can you show the logfile with the crash?

Anakin5: You cannot create a range with a decimal step

Anakin5:
need the pull approve stuff otherwise I cannot merge :s. Thank you

PR #5570, opened 2016-09-20
Task EvolvePokemon should give us a status updates so we know it is working

thejamespinto:
Hey guys,
I have done some local changes on this task so I know how far it is to start evolving my pokemon.

I really like the changes locally and I think they would be good for everybody.
Could anybody take a look and show me how to reduce the logging throughput.

Thanks ❤️

### DIFF
--- a/configs/config.json.example
+++ b/configs/config.json.example
@@ -153,7 +153,7 @@
         "type": "EvolvePokemon",
         "config": {
           "enabled": false,
-          "interval": 120,
+          "log_interval": 120,
 
             "// evolve only pidgey and drowzee": "",
             "// evolve_list": "pidgey, drowzee",
--- a/configs/config.json.example
+++ b/configs/config.json.example
@@ -153,6 +153,7 @@
         "type": "EvolvePokemon",
         "config": {
           "enabled": false,
+          "interval": 60,
 
             "// evolve only pidgey and drowzee": "",
             "// evolve_list": "pidgey, drowzee",
--- a/configs/config.json.example
+++ b/configs/config.json.example
@@ -153,7 +153,7 @@
         "type": "EvolvePokemon",
         "config": {
           "enabled": false,
-          "interval": 60,
+          "interval": 120,
 
             "// evolve only pidgey and drowzee": "",
             "// evolve_list": "pidgey, drowzee",
--- a/pokemongo_bot/cell_workers/evolve_pokemon.py
+++ b/pokemongo_bot/cell_workers/evolve_pokemon.py
@@ -18,7 +18,7 @@ class EvolvePokemon(BaseTask):
     def initialize(self):
         self.start_time = 0
         self.next_update = None
-        self.interval = self.config.get('interval', 60)
+        self.interval = self.config.get('interval', 120)
         self.evolve_list = self.config.get('evolve_list', [])
         self.donot_evolve_list = self.config.get('donot_evolve_list', [])
         self.min_evolve_speed = self.config.get('min_evolve_speed', 25)
--- a/pokemongo_bot/cell_workers/evolve_pokemon.py
+++ b/pokemongo_bot/cell_workers/evolve_pokemon.py
@@ -67,7 +67,7 @@ class EvolvePokemon(BaseTask):
                 candy = inventory.candies().get(pokemon.pokemon_id)
                 pokemon_to_be_evolved = pokemon_to_be_evolved + min(candy.quantity / (pokemon.evolution_cost - 1), filtered_dict[pokemon.pokemon_id])
 
-        self._log_interval_if_should(pokemon_to_be_evolved, self.min_pokemon_to_be_evolved)
+        self._log_update_if_should(pokemon_to_be_evolved, self.min_pokemon_to_be_evolved)
 
         has_minimum_to_evolve = pokemon_to_be_evolved >= self.min_pokemon_to_be_evolved
         if has_minimum_to_evolve:
@@ -78,7 +78,7 @@ class EvolvePokemon(BaseTask):
                 if pokemon.can_evolve_now():
                     self._execute_pokemon_evolve(pokemon, cache)
 
-    def _log_interval_if_should(self, has, needs):
+    def _log_update_if_should(self, has, needs):
         self._compute_next_log_update()
         if self._should_log_update:
             self.emit_event(
@@ -91,7 +91,7 @@ class EvolvePokemon(BaseTask):
         self.next_log_update = datetime.now() + timedelta(seconds=self.log_interval)
 
     def _should_log_update(self):
-        return self.next_log_update is None or datetime.now() >= self.next_log_update
+        return datetime.now() >= self.next_log_update
 
     def _should_run(self):
         if not self.evolve_list or self.evolve_list[0] == 'none':
--- a/pokemongo_bot/cell_workers/evolve_pokemon.py
+++ b/pokemongo_bot/cell_workers/evolve_pokemon.py
@@ -17,8 +17,8 @@ class EvolvePokemon(BaseTask):
 
     def initialize(self):
         self.start_time = 0
-        self.next_update = None
-        self.interval = self.config.get('interval', 120)
+        self.next_log_update = None
+        self.log_interval = self.config.get('log_interval', 120)
         self.evolve_list = self.config.get('evolve_list', [])
         self.donot_evolve_list = self.config.get('donot_evolve_list', [])
         self.min_evolve_speed = self.config.get('min_evolve_speed', 25)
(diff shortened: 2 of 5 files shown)


alexyaoyang:
@thejamespinto Thanks for the PR! I can see that it will be very helpful! 👍 

There's just one small question, how often does it show the status update? If it's too quick (once every second), we gotta limit it so it isn't too spammy.

thejamespinto:
@alexyaoyang this is my first PR and I don't know how to filter the log throughput, can you show me how it's done?

acurioustale:
Should have a configurable `"min_interval"`, imho.

alexyaoyang:
@thejamespinto As @acurioustale suggested or you could try using `"log_interval"` too :)

Gobberwart:
@thejamespinto using sleep(60) will halt bot execution completely for 60 seconds. We definitely don't want that.

pokemongo_bot/cell_workers/evolve_pokemon.py
```diff
@@ -66,12 +66,15 @@ def work(self):
                 pokemon_to_be_evolved = pokemon_to_be_evolved + min(candy.quantity / (pokemon.evolution_cost - 1), filtered_dict[pokemon.pokemon_id])
 
         if pokemon_to_be_evolved >= self.min_pokemon_to_be_evolved:
+            self.logger.info("Checking minimum pokemon to evolve: {}/{}. Hooray!".format(pokemon_to_be_evolved, self.min_pokemon_to_be_evolved))
```
Gobberwart: Do we REALLY need "Hooray!"? 

Also... it's not "checking". It has checked.

pokemongo_bot/cell_workers/evolve_pokemon.py
```diff
@@ -66,12 +66,15 @@ def work(self):
...
             if self.use_lucky_egg:
                 self._use_lucky_egg()
             cache = {}
             for pokemon in filtered_list:
                 if pokemon.can_evolve_now():
                     self._execute_pokemon_evolve(pokemon, cache)
+        else:
+            self.logger.info("Checking minimum pokemon to evolve: {}/{}. Not enough, gotta catch`em all (over and over again)".format(pokemon_to_be_evolved, self.min_pokemon_to_be_evolved))
```
Gobberwart: Again, do we need humorous status messages? Not a fan.

Suggest "Not enough pokemon for evolve" would do.

pokemongo_bot/cell_workers/evolve_pokemon.py
Gobberwart: On second thought, do we even need this? It will immediately start evolving so probably no need for another alert.

alexyaoyang:
@Gobberwart I agree with you but chill man, cut him some slack, it's his first time contributing to us. 👍 

@thejamespinto For examples of limiting throughput, take a look at `base_task.py` for `log_interval` or `update_live_stats.py` for `min_interval`. 

Give some revision to the messages (this bot does lean towards more simplistic logging) + have the throughput reduced and I'll approve your PR 👍 

PS: be careful when reducing the throughput so that it doesn't affect the frequency of the `pokemon_evolved` emit event too, we definitely want all the evolve messages to show :)

pokemongo_bot/cell_workers/evolve_pokemon.py
thejamespinto: I just think it's valid to state two versions of the same message so people know if something's wrong

thejamespinto:
@alexyaoyang yeah, thanks for holding my back against that awful bully 😢 

JK, we actually interact quite a lot on Slack and I had personally asked him to come and review this PR.

I have made some deeper changes and I'll add the changes in a few 👍 
I'm creating a new event named `pokemon_evolve_check`

thejamespinto:
@alexyaoyang I've re-patched it.

I wish my python skills were better so I could help out this project further with tests and such.

I hope these changes turn out valuable to the community.

Botta catch`em all!

pokemongo_bot/cell_workers/evolve_pokemon.py
```diff
@@ -65,19 +70,38 @@ def work(self):
...
             for pokemon in filtered_list:
                 if pokemon.can_evolve_now():
                     self._execute_pokemon_evolve(pokemon, cache)
 
+    def _print_check(self, pokemon_to_be_evolved):
+        has_minimum_to_evolve = pokemon_to_be_evolved >= self.min_pokemon_to_be_evolved
+        result_message = ("Gotta catch`em all!", "Gotta evolv`em all!")[has_minimum_to_evolve]
+        self.emit_event(
+            'pokemon_evolve_check',
+            formatted='Checking... Has {has}, needs {needs}. {message}',
```
alexyaoyang: I think it can be even more simplistic: `Evolvable: {has}/{need}` will be sufficient. 👍

pokemongo_bot/cell_workers/evolve_pokemon.py
```diff
@@ -65,19 +70,38 @@ def work(self):
...
+                'message': result_message
+            }
+        )
+
     def _should_run(self):
         if not self.evolve_list or self.evolve_list[0] == 'none':
             return False
-        return True
-    
+        return self.next_update is None or datetime.now() >= self.next_update
```
alexyaoyang: It seems like the duration that user set now affects how frequent evolve pokemon task is ran. I think only the logging should be reduced, but not the task itself. 

For example, if I set `min_pokemon_to_be_evolved` to 80, there's a chance that I might be evolving more than 80 pokemon due to evolve pokemon not running when I have exactly 80 pokemon.

While evolving more pokemon than the quota is not a big deal for me, I'm not sure if that applies to everyone else. Also if user accidentally sets `interval` to a huge number, that might be a bigger issue.

configs/config.json.example
```diff
@@ -153,6 +153,7 @@
         "type": "EvolvePokemon",
         "config": {
           "enabled": false,
+          "interval": 60,
```
alexyaoyang: Small issue, but I'm not sure if users will be confused by `interval` and `min_evolve_speed`. Perhaps a better name is `log_interval`?

thejamespinto: this number represents how often - in seconds - the task runs. I just realized I should also add that to the docs.

thejamespinto: @alexyaoyang I've added the documentation line. I have been trying this at home for 2 days now and I feel much more comfortable with 120 as the default value because we don't want this message to be spammy 😄

alexyaoyang:
@thejamespinto I think you missed a few of my comments :)

thejamespinto:
@alexyaoyang I did, sorry, I got caught up with the very last comment and I thought it was your only review :)

pokemongo_bot/cell_workers/evolve_pokemon.py
thejamespinto: I'm sorry, but I added documentation for this and the way I see this, people should read it so as not to put a huge number there. Also, how many extra pokemon can they catch in 120 seconds?

Having said that, I can still tweak this to a `log_interval` if you want ¯\_(ツ)_/¯

alexyaoyang: Hmmm yes I see what you mean and I agree with you, but I'd rather not change the behavior of the task since the purpose of this PR is for giving a status update, and not to reduce frequency of evolve task. 

Assuming that **not** all users read the documentation, we are essentially giving users the power/ability to screw up the evolve task.

IMO, if the frequency of evolve task should be reduced, it should be in another PR. What do you think?

Gobberwart: @alexyaoyang Hey man, chill. It's his first contribution :) :) :)

Just kidding, I agree with you. Keeping this PR to just the log message, not the task itself, would be better.

alexyaoyang: @Gobberwart Not sure why but I found that very funny. Now people looking at me kinda weird in Starbucks. Thanks!

alexyaoyang:
@thejamespinto Still waiting on your changes :)

thejamespinto:
@alexyaoyang I'm gonna release them, just kind of caught up with life and work. :)

thejamespinto:
@alexyaoyang there we go 😄

pokemongo_bot/cell_workers/evolve_pokemon.py
```diff
@@ -81,23 +78,25 @@ def work(self):
...
-            }
-        )
+    def _log_interval_if_should(self, has, needs):
+        self._compute_next_log_update()
+        if self._should_log_update:
+            self.emit_event(
+                'pokemon_evolve_check',
+                formatted='Evolvable: {has}/{need}',
+                data={'has': has, 'needs': needs}
+            )
```
alexyaoyang: One last question, will `self._should_log_update` always return false with `self._compute_next_log_update` above?

thejamespinto: no, it may return either `True` or `False`,
but I have just realized I made it impossible for `self.next_log_update` to be `None`. Nice catch. 👍

alexyaoyang:
👍 LGTM

Issue #5562, opened 2016-09-20
Not transferring pokemon


jannickv:
### Expected Behavior
Should be transferring pokemon. For example: i've got a 210 CP / 0.32 IV Ratatta in my inventory... that one should be transferred.

### Actual Behavior
[2016-09-20 11:05:17] Your Pokemon inventory is full! Could not catch!

### Your FULL config.json (remove your username, password, gmapkey and any other private info)
http://pastebin.com/FyT7MViA

### Other Information
OS: Mac OSx
Branch: DEV
Git Commit: d1e983f45f4687580c2a305591dd9178477be444
Python Version: Python 2.7.12

davidspeijer:
Same problem here...

Ryuspb:
I have same problem too ....

Cephis:
same problem too

arkachat:
ame problem here .. Mac os x  El Capitan 10.11.6

jannickv:
so we're all waiting on @sohje to check the pull request^^

Gobberwart:
Gave up waiting. Fixed in #5575

jannickv:
@Gobberwart MVP! Thanks a million!

PR #5510, opened 2016-09-17
Fix transfer, if some of release_below_iv/ivcp/cp missed and logic is…

sohje:
## Short Description:
Transfer wasn't working on config like:
```js
    "release": {
      "any": {"release_below_cp": 1700, "release_below_iv": 0.95, "logic": "and"}
    },
```
## Fixes/Resolves/Closes (please use correct syntax):
- possible #5507

Issue #5507, opened 2016-09-17
Transfer Pokemon not working


jury89:
Hi,

It looks like that automatically transfer Pokemons after every catch is not working correctly (or perhaps i'm missing something).
this is my config.json
```
    {
        "type": "TransferPokemon",
        "config": {
          "enabled": true,
          "min_free_slot": 9999,
          "// transfer_wait_min": 1,
          "// transfer_wait_max": 100
        }
      }
```
but after the bot succesfuly catch the Pokemon, it doesn't get transferred.

Any help?

0rgan1c:
Make sure you review this section: https://github.com/PokemonGoF/PokemonGo-Bot/blob/dev/docs/configuration_files.md#release-configuration

> Default configuration will not release any Pokémon.
`"release": {"any": {"release_below_cp": 0, "release_below_iv": 0, "logic": "or"}}`

If you never changed it, and it looks like that you will never release anything ever.

What I did is change the config.json to:
```
"release": {
      "any": {"keep_best_cp": 4},
      "//any": {"release_below_cp": 0, "release_below_iv": 0, "logic": "or"},
```
This will keep 4 of the same type of pokemon with the best CP, you can change this to however many you want. 

Alternatively, you can just comment (//) or uncomment any of of the examples in your config.json in the "RELEASE" section.

jury89:
I thought that the "release" section was to ignore some kind of Pokemon, not to transfer it after the Pokemon is caught. 
Btw your configuration works perfectly, thank for the hint.

PR #5575, opened 2016-09-20
TransferPokemon: Release logic fix

Gobberwart:
## Short Description:

Logic in transferpokemon task is incorrect. Works for "or" but not for "and". This fix is based on #5510 which appears to be inactive.

- #5507 
- #5480 
- #5562


Gobberwart:
Merged. Had enough discussion on #5510.

xiatianyu:
Yes, it's working again. Thanks for the fix!

Issue #5413, opened 2016-09-12
FollowPath json input


ukos-git:
I'm having issues with a path file. the file was acquired using http://www.gpsies.com as recommended in the help-file.
That site is returning float point values with no trailing zeroes like
`[{"location": "49.78837,9.93,191"]`

therefore the location is not recognized anymore by the regex pattern in `pokemongo_bot/__init__.py` line 1220ff
this fixed it for me:
`                "[-]?\d{1,3}[.]\d{2,7}", location_name`

don't know why the digits after the floating point have to be in that specific format. one digit after the "." should also be ok.

digging in the blame history, aabeace9 by @Calcyfer was the initial author of the regex.

I had to format the *.js file to match the correct syntax of "location": "x,y,z" but that's another story.

mmixx:
Is this location `[{"location": "49.78837,9.93,191"]` the one you put on your path config? It's missing a closing curly bracket...

ukos-git:
I extraced the line that caused the issue to boil down the problem. 
Forgot to post the closing braket. Just put the line in a path.json file. It will throw the error.

mmixx:
I'm currently using `FollowPath` and it's working for me... Try to remove the 3rd set of numbers... The `,191`...

ukos-git:
The altitude is an optionsl return value. 
As I said: the regex is not right. I will post a fix.

alexyaoyang:
@ukos-git Good catch. I think `[-]?\d{1,3}[.]?\d+` should do the trick. The front part before decimal should only have 1-3 digits (if I'm not mistaken latitude is [-80,80] and longitude is [-180,180]) while the # of significant digits can be [0,as precise as user wants]?

ukos-git:
if you write it like that, a number like 0, 1, 2 etc. would not be recognized because you made the "." optional but still want to have at least one digit after that "." with the "+"

alexyaoyang:
Yes my bad. `[-]?\d{1,3}([.]\d+)?` or yours should work then 👍

Shrihan-P:
Anyone, please help. I'm not able to install this on windows.  I have all the files ready, but having trouble putting them in place.? :(

Gobberwart:
@Shrihan-P This is not the correct place to post a general "I can't install the bot" question. Read the docs or visit pokemongo-bot.slack.com

alexyaoyang:
PR merged, closing issue

PR #5578, opened 2016-09-20
fix regex pattern for gps coordinates

ukos-git:
## Short Description:
fixes limited input for gps-coordinates in json files in FollowPath worker.
Too short Numbers like 1.23 were not recognised by the old logic.
Instead you would have to write 1.230.
The new pattern now also supports natural numbers like 1 and 2 as well as 1.1 and 2.2

## Fixes/Closes
https://github.com/PokemonGoF/PokemonGo-Bot/issues/5413


cowst:
did it result on crashes?
if yes, i experienced this issue  as well but thought was specific for sniping.

alexyaoyang:
👍

PR #5588, opened 2016-09-21
Fix PokemonOptimizer ValueError

cmezh:
Original output:
```
  File "pokecli.py", line 843, in <module>
    main()
  File "pokecli.py", line 202, in main
    bot.tick()
  File "/home/cmezh/pogo/pokemongo_bot/__init__.py", line 744, in tick
  File "/home/cmezh/pogo/pokemongo_bot/cell_workers/pokemon_optimizer.py", line 136, in work
    keep, try_evolve, try_upgrade, buddy = self.get_best_pokemon_for_rule(pokemon_list, rule)
ValueError: need more than 3 values to unpack
```

### DIFF
--- a/pokemongo_bot/cell_workers/pokemon_optimizer.py
+++ b/pokemongo_bot/cell_workers/pokemon_optimizer.py
@@ -313,7 +313,7 @@ class PokemonOptimizer(BaseTask):
         sorted_pokemon = self.sort_pokemon_list_to_keep(pokemon_list, rule)
 
         if len(sorted_pokemon) == 0:
-            return ([], [], [])
+            return ([], [], [], [])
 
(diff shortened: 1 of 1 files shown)


pokemongo_bot/cell_workers/pokemon_optimizer.py
```diff
@@ -313,7 +313,7 @@ def get_best_pokemon_for_rule(self, pokemon_list, rule):
         sorted_pokemon = self.sort_pokemon_list_to_keep(pokemon_list, rule)
 
         if len(sorted_pokemon) == 0:
-            return ([], [], [])
+            return ([], [], [], [])
```
cmezh: Since buddy was added this function must return tuple with length of 4, not 3 as it was before.

PR #5591, opened 2016-09-21
Avoid task init if task is disabled

### LINKED ISSUE #5501: RandomAlivePause RandomPause runs even disabled
I notice these run even disabled in config
RandomAlivePause
RandomPause

### Your FULL config.json (remove your username, password, gmapkey and any other private info)

     {
        "//NOTE: This task MUST be placed on the top of task list": {},
        "type": "RandomAlivePause",
        "config": {
          "enabled": false,
          "min_duration": "00:00:10",
          "max_duration": "00:10:00",
          "min_interval": "00:05:00",
          "max_interval": "01:30:00"
        }
      },
      {
        "type": "HandleSoftBan"
      },
      {
        "type": "RandomPause",
        "config": {
          "enabled": false,
          "min_duration": "00:00:10",
          "max_duration": "00:10:00",
          "min_interval": "00:10:00",
          "max_interval": "02:00:00"
        }
      },

[2016-09-16 21:09:07] [RandomAlivePause] [INFO] Next random alive pause at 21:24:19, for a duration of 0:06:03
[2016-09-16 21:09:07] [RandomPause] [INFO] Next random pause at 22:01:19, for a duration of 0:08:44

### Steps to Reproduce

start bot

### Other Information

Branch: 
master
Git Commit: 
a8ee31256d412413b107cce81b62059634e8c802

cmezh:
## Description
Disable task initialization if task is disabled in its config.

## Closes
#5501

### DIFF
--- a/pokemongo_bot/tree_config_builder.py
+++ b/pokemongo_bot/tree_config_builder.py
@@ -55,9 +55,15 @@ class TreeConfigBuilder(object):
                 continue
 
             if not task_config.get("enabled", True):
-                self.bot.logger.warning("Task %s is disabled" % task_type)
+                msg = "Task {task_name} is disabled"
+                self.bot.event_manager.emit(
+                    'task_disabled',
+                    sender=self,
+                    level='info',
+                    formatted=msg,
+                    data={'task_name': task_type}
+                )
                 continue
-
 
             if self._is_plugin_task(task_type):
                 worker = self.plugin_loader.get_class(task_type)
--- a/pokemongo_bot/tree_config_builder.py
+++ b/pokemongo_bot/tree_config_builder.py
@@ -54,6 +54,11 @@ class TreeConfigBuilder(object):
                                         'See config.json.*example for more information.')
                 continue
 
+            if not task_config.get("enabled", True):
+                self.bot.logger.warning("Task %s is disabled" % task_type)
+                continue
+
+
             if self._is_plugin_task(task_type):
                 worker = self.plugin_loader.get_class(task_type)
             else:
--- a/pokemongo_bot/__init__.py
+++ b/pokemongo_bot/__init__.py
@@ -214,6 +214,7 @@ class PokemonGoBot(object):
         self.event_manager.register_event('buddy_pokemon', parameters=('pokemon', 'iv', 'cp'))
         self.event_manager.register_event('buddy_reward', parameters=('pokemon', 'family', 'candy_earned', 'candy'))
         self.event_manager.register_event('buddy_walked', parameters=('pokemon', 'distance_walked', 'distance_needed'))
+        self.event_manager.register_event('task_disabled', parameters=('task_name',))
 
         #  ignore candy above threshold
         self.event_manager.register_event(
--- a/pokemongo_bot/base_task.py
+++ b/pokemongo_bot/base_task.py
@@ -16,9 +16,11 @@ class BaseTask(object):
     """
     self.bot = bot
     self.config = config
+    self.enabled = config.get('enabled', True)
+    if not self.enabled: return
+
     self._validate_work_exists()
     self.logger = logging.getLogger(type(self).__name__)
-    self.enabled = config.get('enabled', True)
     self.last_log_time = time.time()
     self.initialize()
 
--- a/pokemongo_bot/base_task.py
+++ b/pokemongo_bot/base_task.py
@@ -16,11 +16,9 @@ class BaseTask(object):
     """
     self.bot = bot
     self.config = config
-    self.enabled = config.get('enabled', True)
-    if not self.enabled: return
-
     self._validate_work_exists()
     self.logger = logging.getLogger(type(self).__name__)
+    self.enabled = config.get('enabled', True)
     self.last_log_time = time.time()
     self.initialize()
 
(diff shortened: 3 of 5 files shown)


cmezh:
I disagree with you. If task is disabled there is no reason to initialize it.
Take a look at tree_config_builder. If task is disabled it won't be appended to workers list.

YvesHenri:
This is a base task, or a base object. I shouldnt contain any such logic. This is my concern. 
Every concrete class has to handle this behaviour sepparately. Have you ever thought if I want some task to behave differently (on initialize) if its disabled, like print some error warnings, **_for example_**? This wouldnt be possible because the initialize call would never be hit due to that return you have added.

TheSavior:
This seems reasonable to me. Another option is to make this change in the tree_config_builder. Don't even call init if disabled is true.

alexyaoyang:
👍 This fix works fine, so I'll give it good to go. 

A slightly better place to have the check is in tree_config_builder. @cmezh If you'd like to make that change I'll approve it once I see it.

Gobberwart:
I agree with @YvesHenri. This is not the right place to put such logic.

In theory, sure... but in practice a task could do any number of things before it checks to see if its enabled or not. A blanket (if not enabled return) type statement that applies to everything that inherits BaseTask could have unintended results.

cmezh:
Agree with @TheSavior and @alexyaoyang. I will move this check to tree_config_builder and will check tasks init and initialize functions.
@Gobberwart On my opinion, there is no reason to let the task do anything if it is disabled. Maybe just throwing info message about it is enough. Anyway, I will check workers, maybe there is something I missed.

Gobberwart:
@cmezh Yeah, fair enough I guess. Would be nice to check first to see if it will break anything.

cmezh:
Moved check to tree_config_builder and added warning if task is disabled.
It shouldn't break anything, but testing is welcome.

cmezh:
Sorry, Travis is now broken, I have to fix test a bit later.

cmezh:
Fixed test.

alexyaoyang:
👍 LGTM

cmezh:
As I understand, it can't be merged without @YvesHenri approval?

YvesHenri:
It's now merged, @cmezh, even thought I still don't agree with this. You might have added a "warning" message and, lets say, its a white one. What if I want different colors for different task warnings? Well, you got it.

cmezh:
@YvesHenri bot's logger warning is white by default, and I think, we must change its color.
Setting different colors for different task warnings is useless in my opinion. Also the only way to do it is to add a bunch of additional events with different colors, it is not good practice.

YvesHenri:
Correction: it's not merged, my bad.

@cmezh I know the default color for logs. I'm saying that if I want my task to log in red/blue/yellow/pink/whatever if its disabled, it is not possible.

It's not good practice to handle such logic in a base class either. I myself did even handle this on the Sniper task for this very same reason.

cmezh:
@YvesHenri Don't you think it's better to use single color of task disable messages? In this case we can just add new event and emit it on disabled task loading (instead of using logger.warning in my case).

cmezh:
Looks like it won't be merged.

alexyaoyang:
Let's not get hung up by the color of the logs. That can be in another PR. I still think this PR is good. Re-opening.

YvesHenri:
@alexyaoyang It's not only about the log color. It's about handing such logic in a base class. All tasks inherit it and they may want to handle this situation in different ways. This is not right.

alexyaoyang:
@YvesHenri The logic has been shifted out of task base class since the first time you reviewed it.

It is safe to say that for all of our tasks, when user disable it, he/she don't want it to run. That's the entire purpose of disabling it (prior to having `enabled` field, we had to manually remove the respective tasks from the config). Having a log message (whatever color it is) to tell our user that task x is disable is as informative as it should be. The user probably doesn't want anymore logs from a task they disabled.

alexyaoyang:
@YvesHenri Ping

pokemongo_bot/tree_config_builder.py
```diff
@@ -54,6 +54,17 @@ def build(self):
...
+                msg = "Task {task_name} is disabled"
+                self.bot.event_manager.emit(
+                    'task_disabled',
+                    sender=self,
+                    level='info',
+                    formatted=msg,
+                    data={'task_name': task_type}
+                )
+                continue
+
```
Gobberwart: How often will this write to log? If a task is disabled, I don't want to know about it, apart from maybe at startup.

Gobberwart:
My only issue is whether there'll be output every tick. Otherwise, fine.

pokemongo_bot/tree_config_builder.py
alexyaoyang: From what I can find in the project, it is only called once in pokecli.py L98: `tree = TreeConfigBuilder(bot, config.raw_tasks).build()`.

cmezh: @alexyaoyang is right, it will log only at startup. I've added event instead of logger.warning to color its output (yellow).

YvesHenri:
I don't really 'approve' it. Just removed my review and you can decide whether to accept it or not (even though I've seen times where a PR was merged even with pending requested changes).
Anyways, @alexyaoyang, users do disable tasks when they don't want to use it and I completely agree with you. Although, there are times when we add new tasks and set them to disabled by default. There MIGHT be times that those are really required or whatsoever and you may want to behave/warn users differently. You can not treat this change as a generic behaviour. This is not generic.

DBa2016:
From the logic perspective, a "disabled" task is not there. It is not "doing something", it is completely absent - and should be treated as this (regardless whether in the Task base class or in the tree builder). 

Since the objection has bene withdrawn, I am merging the PR now.

PR #5594, opened 2016-09-21
added stardust per hour

maxer137:
i was playing with config and i saw that there was no stardust per hour so i added it

i don't know how to edit more files at once so i made 3


solderzzc:
👍

maxer137:
thanks this was my fist ever python script

Issue #5587, opened 2016-09-21
WARNING: release logic issue. 5 Snorlax and 2 Lapras  are released


seehuily:
dev: 
I just did a git pull (last pull is 3 days ago), 
OMG!!! some how the new release logic released my 5 Snorlax and 2 Lapras 

[2016-09-21 00:25:42] *Releasing Lapras* CP: 2153, IV: 0.53, IVCP: 0.66 | based on rule: CP < 0 AND IV < 0 IVCP < 0
[2016-09-21 00:25:43] Released Lapras (CP: 2153, IV: 0.53, IVCP: 0.66) You now have 2 Lapras candies
[2016-09-21 00:25:47] *Releasing Lapras* CP: 1687, IV: 0.73, IVCP: 0.75 | based on rule: CP < 0 AND IV < 0 IVCP < 0

[2016-09-21 00:26:01] *Releasing Snorlax* CP: 1590, IV: 0.84, IVCP: 0.75 | based on rule: CP < 0 AND IV < 0 IVCP < 0
[2016-09-21 00:26:02] Released Snorlax (CP: 1590, IV: 0.84, IVCP: 0.75) You now have 8 Snorlax candies
[2016-09-21 00:26:06] *Releasing Snorlax* CP: 1393, IV: 0.82, IVCP: 0.84 | based on rule: CP < 0 AND IV < 0 IVCP < 0
[2016-09-21 00:26:07] Released Snorlax (CP: 1393, IV: 0.82, IVCP: 0.84) You now have 9 Snorlax candies
[2016-09-21 00:26:11] *Releasing Snorlax* CP: 323, IV: 0.93, IVCP: 0.95 | based on rule: CP < 0 AND IV < 0 IVCP < 0
[2016-09-21 00:26:11] Released Snorlax (CP: 323, IV: 0.93, IVCP: 0.95) You now have 10 Snorlax candies
[2016-09-21 00:26:14] *Releasing Snorlax* CP: 225, IV: 0.67, IVCP: 0.78 | based on rule: CP < 0 AND IV < 0 IVCP < 0
[2016-09-21 00:26:15] Released Snorlax (CP: 225, IV: 0.67, IVCP: 0.78) You now have 11 Snorlax candies
[2016-09-21 00:26:18] *Releasing Snorlax* CP: 42, IV: 0.69, IVCP: 0.73 | based on rule: CP < 0 AND IV < 0 IVCP < 0
[...]

PR #5596, opened 2016-09-21
Hotfix: Fix release logic on no rule supplied

Gobberwart:
## Short Description:

If no cp/iv/ivcp release rules are provided for a given pokemon, the bot assumes "and" logic and releases it anyway. This is definitely not desirable. 

Changed so that if no cp/iv/ivcp rule is provided, the bot assumes false for everything.

- #5587


Gobberwart:
Merging this as a matter of urgency (lost pokemon)

Issue #5461, opened 2016-09-15
catch_pokemon: never_capture does not work


LazGit:
**Expected Behavior**
\*A wild Pidgey appeared!\* (CP: 235) (NCP: 0.35) (Potential 0.27) (A/D/S 6/1/5)
Great Curveball throw! Used Pokeball, with chance 38.66 (3 left)
Captured Pidgey! [CP 235] [NCP 0.35] [Potential 0.27] [6/1/5] (17/600) [+160 exp] [+100 stardust]

**Actual Behavior**
\*A wild Pidgey appeared!\* (CP: 235) (NCP: 0.35) (Potential 0.27) (A/D/S 6/1/5)
Skip ignored Pidgey! (CP: 235) (NCP: 0.35) (Potential 0.27) (A/D/S 6/1/5)

**Steps to Reproduce**
Update to the latest dev commit
And use 
"catch": { "Pidgey": { "never_capture": true } }

[docs/configuration_files.md](https://github.com/PokemonGoF/PokemonGo-Bot/blob/dev/docs/configuration_files.md#catch-configuration)
Additionally, you can specify always_capture and never_capture flags.
For example: "Pidgey": {"never_capture": true} will stop catching Pidgey entirely.

sech90:
You inverted the Expeted Behaviour and the Actual behaviour.
With the rule `"catch": { "Pidgey": { "never_capture": true } }` the bot is supposed to skip all Pidgeys. But it doesn't.

sech90:
replace "never_capture" in the config file with "never_catch".
 
from pokemon_catch_worker.py:
```python
if pokemon_config.get('never_catch', False):
    return False

if pokemon_config.get('always_catch', False):
    return True
```

Gobberwart:
never_capture is incorrect. should be never_catch as stated. PR #5598 will fix this.

PR #5598, opened 2016-09-22
Fix #5461wrong documentation

sech90:
#5461 is due to wrong syntax in the documentation, as it mentions "never_capture" instead of "never_catch" 

from pokemon_catch_worker.py:
```python
if pokemon_config.get('never_catch', False):
    return False

if pokemon_config.get('always_catch', False):
    return True
```


Gobberwart:
Can you please fix "always_capture" as well, eg. "you can specify **always_catch** and never_catch flags"

sech90:
Done :)

PR #5599, opened 2016-09-22
Restore example vip list

Gobberwart:
## Short Description:

VIP pokemon were removed from vip list in PR #5554, specifically "Articuno, Moltres, Zapdos, Mew and Mewtwo". While these pokemon have not yet been seen, there is no reason for them to be removed from vip list, with the exception that they may cause false positives in sniper data, which is a minor inconvenience at worst.

Restoring list to previous settings.


Gobberwart:
Merging this since @DBa2016 and @solderzzc already indicated their support in PR #5554.

Issue #5361, opened 2016-09-10
Log spam with candy threshold messages


lntn:
### Expected Behavior
Not sure

### Actual Behavior
Got spammed with messages about candy amount over threshold every second, without knowing what to do.
```
Amount of candies for Kakuna is 1994, greater than threshold 400
```
But then what? what am I supposed to do?

### Your FULL config.json (remove your username, password, gmapkey and any other private info)
http://pastebin.com/qWRy6i4e

### Output when issue occurred
http://pastebin.com/CYUCR1D8

### Other Information
OS: 
Mac
Branch: 
dev
Git Commit: 
f02ce439aa4e6f6c14c9a2aafb8fea848ac6a6ca

Anakin5:
It is in your config: `"any": {"candy_threshold" : 400, "catch_above_cp": 0, "catch_above_iv": 0, "logic": "or"}`

lntn:
@Anakin5 indeed

lntn:
@Anakin5 sorry but why closing this? the problem is with that log message appearing every second.

Anakin5:
Ok

Gobberwart:
Wow, this is definitely an issue. Why on earth are we displaying this event????

Gobberwart:
#5604 fixes this.

PR #5604, opened 2016-09-22
Remove event emit from catchworker

Gobberwart:
## Short Description:

Catchworker displays candy threshold alert every few seconds. Can't see why this needs to be displayed even once, so just removed it.

- #5361


solderzzc:
👍

PR #5605, opened 2016-09-22
Corrections and enhancement for Sniper task

GhosterBot:
Correction :
- false Name retrieve (at least for URL)
- incorrect sort when multiple order specified
- keep 'missing' Pokemon for sort option

Enhancement:
- New option in conf file to specify the minimum number of seconds
before the bot will try another Snipe task. Meantime the bot continue
the other tasks.
- Add the 'debug' option in conf file to display more informations on
Pokemons found and Snipe task details.
- Correct name 'priority' order
- Update of doc & config example

Attached print screen show details of Debug Snipping 
pgo_log
infos

### DIFF
--- a/configs/config.json.map.example
+++ b/configs/config.json.map.example
@@ -469,10 +469,12 @@
 		"type": "Sniper",
 		"config": {
 			"enabled": true,
+            "debug": false,
 			"mode": "social",
 			"bullets": 3,
 			"homing_shots": true,
 			"special_iv": 100,
+            "min_seconds_between_snipe": 360,
 			"order": ["missing", "iv", "priority", "vip", "expiration_timestamp_ms"],
 			"sources": [
 				{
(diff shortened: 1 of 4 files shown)


YvesHenri:
Hi @GhosterBot. Thank you for your PR. However, like you've already noticed, there were some fixes for this task lately, which invalidates all your work, unfortunatelly.
The name is now retrieved correctly (fixed), sorting is fine and sorting by missing pokemons does work.
The 'priority' ordering name has already been fixed as well and the debug flag wasnt specified because its supposed to be used by devs only. If a user really wants to see them for some reason, I'm sure they'll be able to find this 'secret' out by himself.
People are asking way too much questions about this tasks config the way it is. I can't imagine how its gonna be if we keep adding more params. That being said, I appreciate your intention but unfortunatelly I don't agree with that min_seconds_between_snipe. 
Since mostly everything on this PR has already been done, I'm closing this.

PR #5610, opened 2016-09-22
Fix order/display of IV and CP in Telegram /caught function

JaapMoolenaar:
## Short Description:
When using the /caught function in Telegram, the CP is always 0 (or 1 in case of a perfect iv) and IV holds the CP. This PR fixes that. 

It now also retrieves only the needed information from the database in a set order; something I suggest you/we do more often, seeing as you almost always select all columns. (Which might even cause these kinds of bugs in case the order of the columns changes in the table).

## Fixes/Resolves/Closes:
None


solderzzc:
👍

PR #5613, opened 2016-09-22
Sniper task: Keeping 'missing' Pokemon + sorting order correction

GhosterBot:
- A 'missing' only Pokemon (not on the catch_list, no VIP and with a bad IV) will now be keep on the list as a potential target if 'missing' is specified on the order option

- The sorting of the target list is now cumulative.
For example, if the order option is [missing, vip, priority], then the list will be ordered by 'missing' first, and then VIP's, and finally by Priority. Actually, the list is only ordered by the last order option specified on the config (only by priority on this example)

Screenshot example of a cumulative [missing, priority, iv, vip] order list

pgo_log

### DIFF
--- a/pokemongo_bot/cell_workers/sniper.py
+++ b/pokemongo_bot/cell_workers/sniper.py
@@ -6,6 +6,7 @@ import requests
 import calendar
 
 from random import uniform
+from operator import itemgetter, methodcaller
 from datetime import datetime
 from pokemongo_bot import inventory
 from pokemongo_bot.item_list import Item
--- a/pokemongo_bot/cell_workers/sniper.py
+++ b/pokemongo_bot/cell_workers/sniper.py
@@ -278,7 +278,7 @@ class Sniper(BaseTask):
                 if pokemon.get('vip', False):
                     self._trace('{} is not catchable and bad IV (if any), however its a VIP!'.format(pokemon.get('pokemon_name')))
                 else:
-                    if pokemon['missing']:
+                    if pokemon.get('missing', False):
                         self._trace('{} is not catchable, not VIP and bad IV (if any), however its a missing one.'.format(pokemon.get('pokemon_name')))
                     else:
                         self._trace('{} is not catchable, nor a VIP or a missing one and bad IV (if any). Skipping...'.format(pokemon.get('pokemon_name')))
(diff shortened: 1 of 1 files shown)


pokemongo_bot/cell_workers/sniper.py
```diff
@@ -277,8 +278,11 @@ def is_snipeable(self, pokemon):
                 if pokemon.get('vip', False):
                     self._trace('{} is not catchable and bad IV (if any), however its a VIP!'.format(pokemon.get('pokemon_name')))
                 else:
-                    self._trace('{} is not catachable, nor a VIP and bad IV (if any). Skipping...'.format(pokemon.get('pokemon_name')))
-                    return False
+                    if pokemon['missing']:
```
YvesHenri: Please change it from `if pokemon['missing']:` to `if pokemon.get('missing', False):`.

pokemongo_bot/cell_workers/sniper.py
```diff
@@ -362,8 +366,7 @@ def work(self):
 
             if targets:
                 # Order the targets (descending)
-                for attr in self.order:
-                    targets.sort(key=lambda pokemon: pokemon[attr], reverse=True)
+                targets = sorted(targets, key=itemgetter(*self.order), reverse=True)
```
YvesHenri: This is and `targets.sort(key=lambda pokemon: pokemon[attr], reverse=True)` are the same thing.

GhosterBot: Yes, but this method is called as many times as there is attr in self.order:
And everytime the targets list is sorted by pokemon[attr] only.
So if there is 4 order specified (missing, priority, iv, vip), the targets.sort will be called 4 times, and everytimes the sorting will only be on ONE attribute. First missing onyl, then priority only, the iv only, and finally ordering by vip only. So finally the list will only be sorted by vip (in this example).
I see it with my detailes .
Calling targets = sorted(targets, key=itemgetter(*self.order), reverse=True will do in one pass a cumulative ordering.

Please look at https://docs.python.org/2/library/operator.html :
 -> Changed in version 2.5: Added support for multiple item extraction.

pokemongo_bot/cell_workers/sniper.py
GhosterBot: Ho I see, sorry. Thanks to have handle that

pokemongo_bot/cell_workers/sniper.py
YvesHenri: I know. Thats what I meant. Although, the final/merged list will have around 30~50 items, and if you have a max of 5 ordering times, that should be an iteration of 150~250 times, which is worthless. I'll accept it anyways.

solderzzc:
👍

Issue #5602, opened 2016-09-22
No longer seeing pokemon caught in telegram


javajohnHub:
### Expected Behavior

A wild Weedle appeared! (CP: 254) (NCP: 0.57) (Potential 0.51) (A/D/S 12/10/1)

Great throw! Used Greatball, with chance 42.96 (118 left)

You caught a Weedle CP: 254 IV: 51

You now have 12 Weedle candy!

### Actual Behavior

pokemonmon-bot, [Sep 21, 2016, 10:47 PM]: 
A wild Weedle appeared! (CP: 254) (NCP: 0.57) (Potential 0.51) (A/D/S 12/10/1)

Great throw! Used Greatball, with chance 42.96 (118 left)

You now have 12 Weedle candy!

### Output when issue occurred

A wild Weedle appeared! (CP: 254) (NCP: 0.57) (Potential 0.51) (A/D/S 12/10/1)

Great throw! Used Greatball, with chance 42.96 (118 left)

You now have 12 Weedle candy!

### Steps to Reproduce

/sub all in telegram

### Other Information
OS: 
Mac
 
Branch: 
dev
 
Git Commit: 
 
Python Version: 
 
Any other relevant files/configs (eg: path files)

Gobberwart:
I'm getting notifications just fine, latest dev pull. What happens if you specifically subscribe to pokemon_caught instead of all?

javajohnHub:
No output. I will try another fresh install but this one was a fresh dev install

javajohnHub:
pokemonmon-bot, [Sep 21, 2016, 11:06 PM]: 
Subscriptions updated.

Great Curveball throw! Used Pokeball, with chance 27.08 (2 left)

You now have 39 Clefairy candy!

Gobberwart:
Running your branch? Or the plain dev branch? Oh and have you authenticated?

Gobberwart:
If it's a clean install, telegram won't know who you are yet until you send the bot a message and/or login.

Gobberwart:
![2016-09-22 03 36 08](https://cloud.githubusercontent.com/assets/20783738/18736358/a7f28ada-80c9-11e6-9e74-14033e7b0d93.png)

Latest dev, latest telegram

javajohnHub:
This is strange. Ill see if theres a telegram update.....and try dev again.

Gobberwart:
Also... I know this might seem silly, but check your alert_catch config :)

[...]

PR #5616, opened 2016-09-22
fix pokemon caught being displayed in telegram

javajohnHub:
## Short Description:
pokemon_caught event was not displaying in telegram.

-#5602
-
-


solderzzc:
👍

PR #5624, opened 2016-09-22
Add timeouts to sniper source config

Gobberwart:
## Short Description:

Sniper currently uses hard-coded values for timeout when trying to validate/retrieve data from sources. Some sources (eg. localhost:5000/raw_data) should have lower timeouts than others as they can be significantly slower.

This is now user-configurable per-source (default to 5 seconds) which allows users to configure higher timeouts for known slow sources, and lower timeouts for sources that should be fast.

- N/A

### DIFF
--- a/configs/config.json.example
+++ b/configs/config.json.example
@@ -224,6 +224,7 @@
 				{
 				    "enabled": false,
 					"url": "http://localhost:5000/raw_data",
+					"timeout": 3,
 					"key": "pokemons",
 					"mappings": {
 						"id": { "param": "pokemon_id" },
@@ -236,6 +237,7 @@
 				{
 				    "enabled": false,
 					"url": "https://pokewatchers.com/grab/",
+					"timeout": 10,
 					"mappings": {
 						"iv": { "param": "iv" },
 						"id": { "param": "pid" },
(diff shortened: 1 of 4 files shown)


YvesHenri:
Hi @Gobberwart. I've explained this **LOTS** of times. The only source that timeouts frequently is the pokewatchers. This is one source that **I** found, but it doesn't mean you **NEED** to use it. It is **ALSO** the only TROUBLESOME URL among the others. It doesnt only takes a few to load, but it also provides crap data. I almost removed it from the example file. I, actually, WILL remove it because people keep asking this over and over.
Also, please note that just because every source has a hard-coded value of 5 seconds for their timeout, it DOES NOT mean it'll take 5 seconds each ("_and lower timeouts for sources that should be fast_"), am I'm sure you know that. Making it user-configurable WILL NOT help in any way, since the issue is on their service.
And, again, I do not want to put any more params in the config file, unless they're **E-X-T-R-E-M-E-L-Y** relevant.

Gobberwart:
Sorry, I completely disagree with you, and this is not "Invalid". I am also well aware of how timeouts work. 

Setting a short timeout on a source that should be quick allows a failure to be identified quickly. And for a source that is often slow, it allows users to provide time for it to respond rather than timing out 50-90% of the time.

I cannot begin to tell you how many times I've had to modify the code of sniper.py to allow pokesnipers or pokegrab time to respond.

Gobberwart:
Anyone else want to weigh in on this one? Seems @YvesHenri and I don't agree, so we need an independent evaluation.

PR #5645, opened 2016-09-24
Dev Merge to Master

Gobberwart:
## Short Description:

I know we just did this, but the pgoapi fix is pretty critical for those using Google accounts.

Any objections?


Gobberwart:
Merging this. No reason to delay.

PR #5657, opened 2016-09-24
Add snipe_once to only capture pokemons which are not in the pokedex

kolinkorr839:
## Short Description:
If they only want to capture pokemons only once so they can fill up their pokedex, then this option can help.


acurioustale:
Wouldn't `"order": ["missing"],` do that already?

kolinkorr839:
Oh... let me check that.

YvesHenri:
`"order": ["missing"]` does this. Closing.

PR #5659, opened 2016-09-24
remove pprint since it is not being used

kolinkorr839:
## Short Description:
Just removing pprint since it is not being used at all.


Gobberwart:
:+1:

Issue #5652, opened 2016-09-24
Could not fetch data from PokemonGo-Map instance using Sniper


acurioustale:
### Expected Behavior
The bot fetches data from a local PokemonGo-Map instance and snipes Pokémon accordingly.

### Actual Behavior
The bot displays an error message and doesn't snipe Pokémon.

### Your FULL config.json (remove your username, password, gmapkey and any other private info)
[chemnitz.json](http://pastebin.com/b2G6sKej)

### Output when issue occurred
> [2016-09-24 10:40:52] [    Sniper] [INFO] Could not fetch data from 'http://localhost:5000/raw_data'. Details: 'NoneType' object has no attribute 'replace'. Skipping...

### Steps to Reproduce
1. Run PokemonGo-Map
2. Run PokemonGo-Bot

### Other Information
OS:  macOS 10.12 
Branch: dev
Git Commit: e262b95eb15f144c89c1f17ffe434cecb4b3e9e6
Python Version: Python 2.7.12
Any other relevant files/configs (eg: path files) [path1.gpx](http://pastebin.com/QMPPbbDw)

acurioustale:
I added the following to [sniper.py](https://github.com/PokemonGoF/PokemonGo-Bot/blob/dev/pokemongo_bot/cell_workers/sniper.py) to account for Nidoran F and Nidoran M as suggested [here](https://github.com/PokemonGoF/PokemonGo-Bot/commit/3b52047d0b3099ce89d4f571bcdc9300747dcc47):

```
name = name.replace("Nidoran\u2642","nidoran m")
name = name.replace("Nidoran\u2640","nidoran f")```

I also tried removing the name mapping, but the error still appears:

```"name": {
	"param": "pokemon_name"
},```

Same for a clean install of e262b95eb15f144c89c1f17ffe434cecb4b3e9e6.

Gobberwart:
Try adding `if name:` to the top of fixname, see if that helps. If so, we might need to put that in.

acurioustale:
That seems to have worked.

Thanks!

Gobberwart:
OK I'll put in a PR for it

Gobberwart:
#5661 should fix this for all

PR #5661, opened 2016-09-24
Update to sniper fixname method

Gobberwart:
## Short Description:

-Fixed issue where bot may crash if name == None supplied to fixname()
-Added Nidoran M/F

-#5652

### DIFF
--- a/README.md
+++ b/README.md
@@ -157,6 +157,7 @@ If you do not want any data to be gathered, you can turn off this feature by set
  * Breeze Ro
  * bruno-kenji
  * Gobberwart
+ * javajohnHub
 
 ## Disclaimer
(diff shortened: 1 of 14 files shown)


Anakin5:
If you need something more generic to fix names, there is https://github.com/PokemonGoF/PokemonGo-Bot/blob/master/pokemongo_bot/cell_workers/pokemon_optimizer.py#L314

Gobberwart:
Nice @Anakin5... I'll try that and see if it works better.

Gobberwart:
@Anakin5 Doesn't work so well...

Unknown Pokemon name [%s]. Assuming it is [%s] mr-mime Grimer

I'm going to stick with fixname, but add closest_name as a backup just in case we haven't explicitly defined something.

Result:

Unknown Pokemon name mr. mime. Assuming it is Mr. Mime

Anakin5:
Yes the match works better when using lowercase for all names. Not sure about Nidoran Unicode though.

Gobberwart:
I haven't seen it myself, but another user tested them and it works.

Gobberwart:
Ditching this as commits have gotten out-of-hand (git problem?). Will resubmit.

PR #5662, opened 2016-09-25
New task: CatchLimiter

Gobberwart:
## Short Description:

Adds a new task called "CatchLimiter" which enables users to automatically disable all catching tasks when balls reach a given threshold, and for a specified duration. 

While each task already has a min_balls parameter (excluding Sniper), this is generally very low and often does not permit sufficient recovery time. Eg. if min_balls = 5, and you spend all 5 balls trying to catch a pokemon, you can end up stuck in an "insufficient balls" loop. 

This new task basically switches off catch mode for a while so the bot can loot for a while to allow it to recover its ball collection.

I am very interested in feedback on this, and I realise it may not be useful for everyone (hence why it's disabled by default) but feedback I've had so far in slack has been positive. Users report that they often have to stop and disable catching tasks for a while when balls get low, so this should alleviate the problem.

### DIFF
--- a/configs/config.json.example
+++ b/configs/config.json.example
@@ -212,6 +212,14 @@
         }
       },
       {
+        "type": "CatchLimiter",
+        "config": {
+          "enabled": false,
+          "min_balls": 20,
+          "duration": 15
+        }
+      },
+      {
 		"type": "Sniper",
 		"config": {
 			"enabled": false,
--- a/configs/config.json.map.example
+++ b/configs/config.json.map.example
@@ -210,6 +210,14 @@
         }
       },
       {
+        "type": "CatchLimiter",
+        "config": {
+          "enabled": false,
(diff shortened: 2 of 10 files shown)


solderzzc:
👍  Nice to have this task.

crvfts:
Working excellently! Any chance this task could also include min_interval? I found myself hunting through logger to see when bot would start catching again.

Gobberwart:
min_balls doesn't affect what balls are used up, just disables catching when your total balls < threshold. What you actually use to catch, or what you find from pokestops while catching is disabled is a separate issue.

Re min_interval. I might just add catch_resume_at.strftime("%H:%M:%S") wherever the bot displays a message about catching disabled, eg.

`"Not catching magikarp. All catching tasks are currently disabled until 12:13:14"`

What do you think?

PR #5674, opened 2016-09-25
[Sniper] Fix cache, and name matching tweak

Gobberwart:
## Short Description:

Two issues fixed here. 

1. Existing cache is not persistent, and gets reset every time the task runs, which results in repeated failed attempts to catch non-existent/otherwise failed pokemon. Even if cache hits, still generates an unnecessary log message. Changed to persistent cache, and check before attempting to snipe.

2. Name matching can be flaky depending on sources. Added method to manually fix names we know about, or use difflib (as with pokemonoptimizer) to try and match any we haven't already caught. Note that difflib alone is not sufficient for this, eg. it thinks source name "mr-mime" is a Grimer. Close but no cigar.

### DIFF
--- a/pokemongo_bot/cell_workers/sniper.py
+++ b/pokemongo_bot/cell_workers/sniper.py
@@ -337,10 +337,8 @@ class Sniper(BaseTask):
             # Check if already in list of pokemon we've tried
             if self._is_cached(pokemon):
                 # Do nothing. Either we already got this, or it doesn't really exist
-                print "Already tried"
                 self._trace('{} already sniped! Skipping...'.format(pokemon['pokemon_name']))
             else:
-                print "Not already tried"
                 # Backup position before anything
                 last_position = self.bot.position[0:2]
 
--- a/pokemongo_bot/cell_workers/sniper.py
+++ b/pokemongo_bot/cell_workers/sniper.py
@@ -5,6 +5,7 @@ import json
 import requests
 import calendar
 import difflib
+import hashlib
 
 from random import uniform
 from operator import itemgetter, methodcaller
@@ -226,7 +227,7 @@ class Sniper(BaseTask):
     MIN_SECONDS_ALLOWED_FOR_CELL_CHECK = 10
     MIN_SECONDS_ALLOWED_FOR_REQUESTING_DATA = 5
     MIN_BALLS_FOR_CATCHING = 10
-    MAX_CACHE_LIST_SIZE = 200
+    MAX_CACHE_LIST_SIZE = 500
 
     def __init__(self, bot, config):
         super(Sniper, self).__init__(bot, config)
@@ -291,11 +292,6 @@ class Sniper(BaseTask):
             self._trace('{} is expired! Skipping...'.format(pokemon.get('pokemon_name')))
             return False
 
-        # Skip if already cached
-        if self._is_cached(pokemon):
-            self._trace('{} was already handled! Skipping...'.format(pokemon.get('pokemon_name', '')))
-            return False
-
         # Skip if not enought balls. Sniping wastes a lot of balls. Theres no point to let user decide this amount
         if all_balls_count < self.MIN_BALLS_FOR_CATCHING:
             self._trace('Not enought balls left! Skipping...')
@@ -334,9 +330,10 @@ class Sniper(BaseTask):
                 self.bot.sniper_unique_pokemon = []
             
             # Check if already in list of pokemon we've tried
-            if self._is_cached(pokemon):
+            uniqueid = self._build_unique_id(pokemon)
+            if self._is_cached(uniqueid):
                 # Do nothing. Either we already got this, or it doesn't really exist
-                self._trace('{} already sniped! Skipping...'.format(pokemon['pokemon_name']))
+                self._trace('{} was already handled! Skipping...'.format(pokemon['pokemon_name']))
             else:
                 # Backup position before anything
                 last_position = self.bot.position[0:2]
@@ -390,7 +387,7 @@ class Sniper(BaseTask):
                     self._teleport_back(last_position)
                     
                 # Save target and unlock heartbeat calls
-                self._cache(pokemon)
+                self._cache(uniqueid)
                 self.bot.hb_locked = False
 
         return success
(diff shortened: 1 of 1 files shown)


pokemongo_bot/cell_workers/sniper.py
```diff
@@ -476,20 +499,25 @@ def _hash(self, pokemon):
...
+        if not self._is_cached(uniqueid):
             # Free space if full and store it
-            if len(self.cache) >= self.MAX_CACHE_LIST_SIZE:
-                self.cache.pop(0)
-            self.cache.append(pokemon)
+            if len(self.bot.sniper_cache) >= self.MAX_CACHE_LIST_SIZE:
+                self.bot.sniper_cache.pop(0)
+            self.bot.sniper_cache.append(uniqueid)
+
+    def _build_unique_id(self, pokemon):
```
YvesHenri: This is totally unecessary. Use _hash instead.

pokemongo_bot/cell_workers/sniper.py
```diff
@@ -152,10 +154,25 @@ def validate(self):
             raise
             
     def _fixname(self,name):
-        name = name.replace("mr-mime","mr. mime")
-        name = name.replace("farfetchd","farfetch'd")
+        if name:
+            name = name.replace("mr-mime","mr. mime")
+            name = name.replace("farfetchd","farfetch'd")
+            name = name.replace("Nidoran\u2642","nidoran m")
+            name = name.replace("Nidoran\u2640","nidoran f")
```
YvesHenri: Seriously? Are you going to handle every name combination possible?

pokemongo_bot/cell_workers/sniper.py
```diff
@@ -152,10 +154,25 @@ def validate(self):
...
         return name
 
+    def _get_closest_name(self, name):
```
YvesHenri: What's this?

pokemongo_bot/cell_workers/sniper.py
```diff
@@ -210,13 +227,12 @@ class Sniper(BaseTask):
     MIN_SECONDS_ALLOWED_FOR_CELL_CHECK = 10
     MIN_SECONDS_ALLOWED_FOR_REQUESTING_DATA = 5
     MIN_BALLS_FOR_CATCHING = 10
-    MAX_CACHE_LIST_SIZE = 200
+    MAX_CACHE_LIST_SIZE = 500
```
YvesHenri: You  do realize people bot on raspberry devices, right? Increasing this this high might not be a good idea.

pokemongo_bot/cell_workers/sniper.py
```diff
@@ -232,6 +248,9 @@ def initialize(self):
         self.altitude = uniform(self.bot.config.alt_min, self.bot.config.alt_max)
         self.sources = [SniperSource(data) for data in self.config.get('sources', [])]
 
+        if not hasattr(self.bot,"sniper_cache"):
+            self.bot.sniper_cache = []
```
YvesHenri: What's this and what's it for?

pokemongo_bot/cell_workers/sniper.py
```diff
@@ -311,60 +325,70 @@ def snipe(self, pokemon):
...
-            if exists:
-                self._log('Yay! There really is a wild {} nearby!'.format(pokemon.get('pokemon_name')))
-                self._teleport_back_and_catch(last_position, pokemon)
+            # Have we already tried this pokemon?
+            if not hasattr(self.bot,'sniper_unique_pokemon'):
+                self.bot.sniper_unique_pokemon = []
+            
+            # Check if already in list of pokemon we've tried
+            uniqueid = self._build_unique_id(pokemon)
+            if self._is_cached(uniqueid):
```
YvesHenri: is_snipeable should check whether its been cached or not.

pokemongo_bot/cell_workers/sniper.py
Gobberwart: Read the description

pokemongo_bot/cell_workers/sniper.py
Gobberwart: How is it unnecessary? If there's an easy way to generate MD5 from string, then awesome. Please share your alternative code.

pokemongo_bot/cell_workers/sniper.py
Gobberwart: Again, read the description. Difflib is already used in pokemonoptimizer to TRY and get a real pokemon name from a supplied string. Eg. if source is "pigeot", it *should* return "pidgeot"... in many cases, it seems to work well.

pokemongo_bot/cell_workers/sniper.py
Gobberwart: You do realise that replacing dict with an md5 string makes it take FARRRRR less memory to store.

pokemongo_bot/cell_workers/sniper.py
Gobberwart: You had a "self.cache" which is nice but not persistent. Every time the task runs, it reset it to []. This creates a bot-global cache which will persist through task re-inits.

pokemongo_bot/cell_workers/sniper.py
Gobberwart: No need. It checks before it calls is_snipeable. Saves a wasted log message.

Gobberwart:
OK @YvesHenri I've answered all your questions. Talk to me in chat if something is still unclear.

pokemongo_bot/cell_workers/sniper.py
Gobberwart: is_snipeable currently lumps everything together as a general fail. I've removed one of those for clarity.

pokemongo_bot/cell_workers/sniper.py
Gobberwart: But OK, sure... I'll compromise at 300.

pokemongo_bot/cell_workers/sniper.py
Gobberwart: Unless you're talking about the len(self.bot.sniper_cache) thing... which is your code so you tell me why it's necessary.

Gobberwart:
Does anyone else have feedback for this?

Gobberwart:
No feedback received and YH's requested changes are irrelevant. Merging.

Issue #5669, opened 2016-09-25
Snipe wait time is too low to allow the Pokemon to show up during consecutive snipings


Dr8g0n:
### Expected Behavior
Longer wait time for Pokemon to appear before Teleport back to original location. The wait time is especially short ~1 sec after a recent snipe due to:

`# Wait a maximum of MIN_SECONDS_ALLOWED_FOR_CELL_CHECK seconds before requesting nearby cells
                if (seconds_since_last_check < self.MIN_SECONDS_ALLOWED_FOR_CELL_CHECK):
                    time.sleep(self.MIN_SECONDS_ALLOWED_FOR_CELL_CHECK - seconds_since_last_check)
`

Recommend adding time.sleep(randint(5,10))) outside of the MIN_SECONDS_ALLOWED_FOR_CELL_CHECK in pokemongo_bot/cell_workers/sniper.py

No longer miss any Pokemon after changing sleep time to allow more wait time especially during consecutive snipings.

### Actual Behavior
Wait time too short and often miss the Pokemon. Recommend adding time.sleep(randint(5,10))) outside of the MIN_SECONDS_ALLOWED_FOR_CELL_CHECK in pokemongo_bot/cell_workers/sniper.py

`# Wait a maximum of MIN_SECONDS_ALLOWED_FOR_CELL_CHECK seconds before requesting nearby cells
                if (seconds_since_last_check < self.MIN_SECONDS_ALLOWED_FOR_CELL_CHECK):
                    print "MIN_SECONDS_ALLOWED_FOR_CELL_CHECK:  " + str(self.MIN_SECONDS_ALLOWED_FOR_CELL_CHECK) + ", seconds_since_last_check: " + str(seconds_since_last_check)
                    print "Sleeping due to MIN_SECONDS_ALLOWED_FOR_CELL_CHECK"
                    print "Slept for: " +  str(self.MIN_SECONDS_ALLOWED_FOR_CELL_CHECK - seconds_since_last_check)
                    time.sleep(self.MIN_SECONDS_ALLOWED_FOR_CELL_CHECK - seconds_since_last_check)
                print "Slept a bit more... (randint(5,10))"
                time.sleep(randint(5,10)))
`

[2016-09-25 10:13:22] [    Sniper] [INFO] Teleporting to meet Geodude (12345, 6789)...
[2016-09-25 10:13:31] [    Sniper] [INFO] Yay! There really is a wild Geodude nearby!
[...]

PR #5677, opened 2016-09-26
Dev

Dr8g0n:
## Short Description:
Allow more time for Pokemons to appear during sniping from URLs.

Currently, if it falls into the if (seconds_since_last_check < self.MIN_SECONDS_ALLOWED_FOR_CELL_CHECK) statement, the wait time is around 1 second and that does not give enough time for the Pokemon to show up. It ended up missing some of the Pokemon.

#5669


DBa2016:
I am pretty sure this PR is a mistake, closing it.

PR #5678, opened 2016-09-26
Increase sniper's wait time, so as to avoid missing some pokemon

Dr8g0n:
## Short Description:
Currently, if it falls into the if (seconds_since_last_check < self.MIN_SECONDS_ALLOWED_FOR_CELL_CHECK) statement, the wait time is around 1 second and that does not give enough time for the Pokemon to show up. It ended up missing some of the Pokemon.

Added wait time between 5-10 second to allow the Pokemon to show up
Please refer to issue #5669

### DIFF
--- a/pokemongo_bot/cell_workers/sniper.py
+++ b/pokemongo_bot/cell_workers/sniper.py
@@ -330,6 +330,7 @@ class Sniper(BaseTask):
 
                 # Wait a maximum of MIN_SECONDS_ALLOWED_FOR_CELL_CHECK seconds before requesting nearby cells
                 if (seconds_since_last_check < self.MIN_SECONDS_ALLOWED_FOR_CELL_CHECK):
+		    time.sleep(self.MIN_SECONDS_ALLOWED_FOR_CELL_CHECK - seconds_since_last_check)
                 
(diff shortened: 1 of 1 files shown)


pokemongo_bot/cell_workers/sniper.py
```diff
@@ -329,7 +330,10 @@ def snipe(self, pokemon):
 
                 # Wait a maximum of MIN_SECONDS_ALLOWED_FOR_CELL_CHECK seconds before requesting nearby cells
                 if (seconds_since_last_check < self.MIN_SECONDS_ALLOWED_FOR_CELL_CHECK):
-                    time.sleep(self.MIN_SECONDS_ALLOWED_FOR_CELL_CHECK - seconds_since_last_check)
+
```
DBa2016: Check your intendation: the statement as it is will sleep unconditionally!

DBa2016:
Oh, and by the way... Please give the PR a descriptive name. Just putting your name into it without any real contents is nad habit.

pokemongo_bot/cell_workers/sniper.py
Dr8g0n: Accidentally remove original sleep time. Added it back in the next commit.

Issue #5569, opened 2016-09-20
Sniper: pending validation source list and sleeping itself when out of balls


YvesHenri:
As I have stated a couple times, I DO NOT want to put more settings in the config file for the Sniper task. The more settings, the higher is the amount of issues we'll have to maintain everyday.

Running out of balls is a known 'issue' if your catch list is big and catches pretty much everything. You're supposed to be catching very rare pokemons ONLY, otherwise expect this to happen frequently!
However, we can make it sleep itself whenever it happens to find out we're out of balls. This will help a lot. Setting a "min_balls" param wouldn't do it. No, it will not.

As we already know, the source has to reply something by the time they're being validated, otherwise we can't know whether the mappings have been specified correctly, so they have to be removed from the source list. We can, however, add them to a 'pending list', which will be validated in the task's work method, if any.

Please don't close this issue untill its been implemented (so people can stop talking about this - I hope).

Gobberwart:
And yes, I realise each function has a min_balls type parameter, but that is different. This would just stop all catching for a while.

YvesHenri:
It does look good, although I have no idea how you're going to make this interact with all the catchable tasks you mentioned above. Do you plan to apply fixes on them all or put it centralize it somewhere?

Gobberwart:
Yep, each task would need changes to make this work. Not sure if I plan to do it or not. Will see :)

kureijir:
This is good, but I'd still love to have Sniper's `min_balls` differs from the other tasks, either hardcoded or configurable, given that we usually snipe rare pokemon only, and they're harder to catch, so it's more prefered to have more balls when sniping. For example, I want to have at least 20 pokeballs with me if I were to teleport to a Snorlax or Dragonite, but trying to catch an Eevee appeared next to me with 5 balls left is fine.

YvesHenri:
This feature request will solve this problem @kureijir. Having a min balls set means trouble, since you'll be always running with this amount (no matter how many balls you got, you'll go down to that min quantity VERY quickly). If the min balls is 25, for example, and you have 100 balls, as soon as you go down to 25 (and, if sniper is enabled, this is very fast) and now, lets say you teleport to a dragonite, like you said; you'll be at around ~15 balls after this and it'll be skipping sniping untill you have 25 again, but then, again, you'll go down to 15~20 balls for the 42904th time. This will repeat forever. To avoid this, it'll need to sleep untill you have 100 balls again, lets say.

[...]

PR #5690, opened 2016-09-26
Sniper enhancements and code cleaning

YvesHenri:
1) Reverted some sugar-coat code:

Removed difflib usage, since it recognizes "mr. mime" as being a "grimmer", therefore using it can lead to unexpected behaviour.
Removed that non sense name fixing (sniper consumes sources, it doesnt fix them. if it's wrong, it's wrong. we cant predict all the possibilities of wrong names either). Pokemon names are used for debugging purposes only, anyways.
Removed that 'unique id building' since theres a similar function already: _hash(). MD5 or any encrypting algorithm is totally unecessary here, since there won't be different pokemons spawned at the same location.
Added the logic back to the is_snipeable() function (cached verification): it was removed because it would 'waste' a debug information (?).
2) Added pending validation for sources that doesn't get to reply something by the time they're being validated for the first time (fixes issue #5569).
3) Replaced catch and VIP lists for dictionaries: faster lookup and solves the naming handling (see above).
4) Other minor refactors

coat code:


javajohnHub:
Shouldn't fixing the errors sniper causes a higher priority than cleaning up code.

javajohnHub:
Please leave a description of your changes so we can properly review.

YvesHenri:
Read above, @javajohnHub.

Gobberwart:
OK, let me go through these issues one by one:

_Removed difflib usage, since it recognizes "mr. mime" as being a "grimmer", therefore using it can lead to unexpected behaviour._

That's why the "non sense name fixing" is there. It's not terribly elegant but it fixes known problematic names *if* difflib can't handle them. Difflib usually does, though, which is why it's used in pokemonoptimizer. I don't believe this should be removed as it's useful.

_Removed that non sense name fixing (sniper consumes sources, it doesnt fix them. if it's wrong, it's wrong. we cant predict all the possibilities of wrong names either). Pokemon names are used for debugging purposes only, anyways._

Not true at all. They are not just used for "debugging purposes".

```
        if pokemon.get('pokemon_name', '') in self.catch_list:
            self._trace('{} is catchable!'.format(pokemon.get('pokemon_name')))
```

Fixing names as best we can resolves issue where a named pokemon is in the catch_list, but the name doesn't quite match. This has happened to myself and others numerous times.

I agree that this method of fixing names is far from ideal. In the short-term, however, it is useful. Please leave it in unless it is actually causing harm.

_Removed that 'unique id building' since theres a similar function already: _hash(). MD5 or any encrypting algorithm is totally unecessary here, since there won't be different pokemons spawned at the same location._

You yourself have expressed concern at the size of the pokemon cache. My thinking here is that it is less resource-intensive to store a list of MD5 strings than a list of pokemon dicts, from a memory perspective anyway.

The MD5 hash is built from pokemon name, latitude, longitude and expiration so can almost be 100% guaranteed to be unique.

Yes, there is a *miniscule* bit of overhead required to calculate the MD5, but not enough to be of concern compared to storing a list of dicts.

Note that the existing hash function does nothing like build_unique_id. All _hash does is return a string containing a truncated version of lat/long. Not the same at all.

_Added the logic back to the is_snipeable() function (cached verification): it was removed because it would 'waste' a debug information (?)._

Not quite. It was moved because "is_snipeable" doesn't give good information to the user. It just comes back with "Damn! Its not here. Reasons: too far, caught, expired or fake data. Skipping...". Moving this to BEFORE is_snipeable prevents a) unnecessary log info and b) attempting to catch something we've already tried.

Honestly, if a pokemon is skipped because we've already tried it, no log info is needed at all. Nobody wants to see their logs filled with the same messages repeatedly, and that's what was happening prior to this modification.


OK so hopefully you can at least see where I'm coming from here. I'm not sure why you'd want to undo it as it improves things overall for all sniper users.  Happy to discuss further.

pokemongo_bot/cell_workers/sniper.py
```diff
@@ -58,7 +56,7 @@ def fetch(self):
             for result in results:
                 iv = result.get(self.mappings.iv.param)
                 id = result.get(self.mappings.id.param)
-                name = self._get_closest_name(self._fixname(result.get(self.mappings.name.param)))
+                name = result.get(self.mappings.name.param).replace("\u2642", " M").replace("\u2640", " F")
```
Gobberwart: Seriously? Come on man, that's really not a better solution is it?

pokemongo_bot/cell_workers/sniper.py
```diff
@@ -261,32 +240,24 @@ def initialize(self):
...
         pokeballs_count = self.inventory.get(Item.ITEM_POKE_BALL.value).count
         greatballs_count = self.inventory.get(Item.ITEM_GREAT_BALL.value).count
         ultraballs_count = self.inventory.get(Item.ITEM_ULTRA_BALL.value).count
         all_balls_count = pokeballs_count + greatballs_count + ultraballs_count
 
+        # Skip if already handled
+        if self._is_cached(pokemon):
+            self._trace('{} was already handled! Skipping...'.format(pokemon.get('pokemon_name')))
+            return False
+
```
Gobberwart: Yes, but this will result in potential repeated vague "Damn! ...." messages. Better left out of is_snipeable so it just cleanly skips.

pokemongo_bot/cell_workers/sniper.py
```diff
@@ -325,86 +296,65 @@ def snipe(self, pokemon):
...
 
     def work(self):
-        # Do nothing if this task was invalidated
-        if self.disabled:
-            self._error("Sniper was disabled for some reason. Scroll up to find out.")
-            
-        elif self.bot.catch_disabled:
-            if not hasattr(self.bot,"sniper_disabled_global_warning") or \
-                        (hasattr(self.bot,"sniper_disabled_global_warning") and not self.bot.sniper_disabled_global_warning):
-                self._log("All catching tasks are currently disabled until {}. Sniper will resume when catching tasks are re-enabled".format(self.bot.catch_resume_at.strftime("%H:%M:%S")))
```
Gobberwart: Do not remove this! It was discussed and approved by multiple people and has proven EXTREMELY useful. If you do this, sniper will be the only catch task that does not respect the user's configuration.

pokemongo_bot/cell_workers/sniper.py
```diff
@@ -499,25 +448,50 @@ def _hash(self, pokemon):
...
-
-    def _build_unique_id(self, pokemon):
-        # Build unique id for this pokemon from id, latitude, longitude and expiration
-        uniqueid = str(pokemon.get('pokemon_id','')) + str(pokemon.get('latitude','')) + str(pokemon.get('longitude','')) + str(pokemon.get('expiration',''))
-        md5str = hashlib.md5()
-        md5str.update(uniqueid)
-        uniqueid = str(md5str.hexdigest())
-        return uniqueid
+                del self.bot.sniper_cache[:25]
+            self.bot.sniper_cache.append(self._hash(pokemon))
```
Gobberwart: I fail to see how passing around dicts is better than passing a string.

YvesHenri:
**# Naming handling:**
`difflib` does look cool, but not really useful in my opinion. Having multiple functions with multiple naming possibilities is not really the way to solve this, nor it's elegant. The only problematic stuff are the nidoran unicode characters (this is the only thing we'll 'fix') and that explains the double replace() calls. We can't predict all the naming possibilities, nor we will EVER try to fix an issue that's related to their servers (like the timeout thing). For example, some sourced replied today something like "**...'longitude': "IV (85,30%) 14,94091230**". In other words, they replied both _IV_ and _longitude_ all together. We can't predict such things. If something goes wrong with the source's replies, warn user and skip it.

**# Pokemon names:**
They **ARE** used **ONLY** for debugging purposes.
```
if pokemon.get('pokemon_name', '') in self.catch_list:
            self._trace('{} is catchable!'.format(pokemon.get('pokemon_name')))
```
The above code has been replaced by this:
 ```
if self.catch_dictionary.has_key(pokemon.get('pokemon_id', 0)):
              self._trace('{} is catchable!'.format(pokemon.get('pokemon_name')))
```
This uses the hashset O(1) complexity for checkup, against the O(N) for the list checkup, which greatly improves the speed. Game objects should be threatened by their ID, just like the app game does.

**# Encryption (MD5):**
I do agree that storing a string requires much less memory than storing dictionaries. That PR did consider this. All I did is replace that hash result for the truncated location string, on which can be safely used as an unique identifier. You'll **VERY** rarely find multiple pokemons with the EXACT same position (4 decimal places guarantees it). You could append the expiration time, but it is **NOT REALLY** necessary. 

**# Sniping logic:**
is_snipeable is not supposed to provide any log to the users. It does provide logs (traces), but for devs to help debugging. The "Damn! Its not here" message/log has nothing to do with the sniping logic. It is related to the `self.bot.get_meta_cell()` call. When the above message pops up, it means that we couldnt find the target within that meta cell, probably because the source supplied us wrong location, wrong expire time etc. Yes, that message might have looked confusing to you, but should be clear to users (can't explain them how the get_meta_cell() works).

**# CatchLimiter task:**
I did NOT remove those checkups you have done. Instead, they were moved to somewhere that best suits it. Indeed, it might now spam that disabled catching log on every work call, but this should be fixed/handled within that task, not on every catching task (I even put a TODO).

File: pokemongo_bot/cell_workers/sniper.py (hunk 4 above)
Comment (YvesHenri) [inline pokemongo_bot/cell_workers/sniper.py]: `self._hash(pokemon)` returns a string, not a dictionary.

File: pokemongo_bot/cell_workers/sniper.py (hunk 3 above)
Comment (YvesHenri) [inline pokemongo_bot/cell_workers/sniper.py]: It was NOT removed. It was moved.

File: pokemongo_bot/cell_workers/sniper.py (hunk 2 above)
Comment (YvesHenri) [inline pokemongo_bot/cell_workers/sniper.py]: `is_snipeable`should not provide logs. The "Damn" messages have nothing to do with `is_snipeable`.

**Comment by Gobberwart** [comment]
**Naming handling**

Agree it's not particularly elegant, but we're dealing with sources that can be mangled. I'd still like to make *some* effort to fix names, either generically (with difflib) or manually using fixnames. That really should reduce the number of repeated questions for the same thing.

**Pokemon names**

Yes, but.... if a source does not supply a pokemon id, we have to get it from somewhere to be able to match in the catch_dictionary. So yes, we still need to have valid names wherever possible.

**Encryption (MD5)**

I think MD5 of name, lat, long, expiry is more unique than hash based purely on first 4 digits of lat/long. Why not keep it? Reverting seems like a backward step.

**Sniping logic**

Ok, sure, my mistake. 

**Catch Limiter**

Ah, didn't see it hiding down the bottom there in is_enabled. Cool.
Re spamming though, I already had code there to prevent it, which you HAVE removed, specifically:

```
        elif self.bot.catch_disabled:
            if not hasattr(self.bot,"sniper_disabled_global_warning") or \
                        (hasattr(self.bot,"sniper_disabled_global_warning") and not self.bot.sniper_disabled_global_warning):
                self._log("All catching tasks are currently disabled. Sniper will resume when catching tasks are re-enabled")
            self.bot.sniper_disabled_global_warning = True
            
        else:
            self.bot.sniper_disabled_global_warning = False
```

If you put that back, it will alert once the first time catching is disabled, then shut the hell up :)

pokemongo_bot/cell_workers/sniper.py
Gobberwart: Yep, I see that.

pokemongo_bot/cell_workers/sniper.py
Gobberwart: Moved, but with reduced functionality. Log spam is an issue again.

pokemongo_bot/cell_workers/sniper.py
Gobberwart: You are passing the pokemon dict to _is_cached. But that's definitely an improvement over what you had before.

YvesHenri:
**# Naming handling:**

We can NOT predict what will happen. Many of these sources get their data from slack/discord and parses it. Wrong parsings can lead to something I meantioned above.
I remember when you first added that 'fixname' function. It started off with 2 different namings and then sooner you added 2 more (and THEN added ANOTHER get_closest_name function...). I can't imagine what would happen in a month, other than Sniper's becoming a **GIANT HAIRBALL** with a lot of extra code (if that doesn't get patched by Niantic).
Good exception messages for laypersons would do the job.

**# Pokemon names:**

`catch_dictionary` is a dictionary where the key is the pokemon ID (not name) and the value is its priority. The ONLY time we need a name is when a source does not provide an ID and you'll need to use `Pokemons.id_for()`, but then again, if the name is wrong, we should throw an exception (see above).

**# Encryption (MD5):**

The thing is you do **N-O-T** need that extra overhead at all. `_hash()` is capable of providing an 'unique id' (it is NOT _based purely on the first 4 digits_ - `"{0:.4f}"` will truncate the first 4 decimal digits, KEEPING THE INTEGER VALUES, eg.: **123.4567**890 will result in 123.4567, as well as **0.1234**5678 will result in 0.1234). If ever 2 pokemons spawn at the same position (1.234; 1.234), they both will be encrypted to the very same value. Also, this is not really a 'revert'. I was using SHA1 (another encryption algorithm) before I commited the first version, untill I realized the extra encryption overhead is **POINTLESS**.

**# Catch Limiter:**

This is wrong. No catching tasks should be responsible of logging such warning, except the one that maintains such logic (CatchLimiter). I left the logs on Sniper as it is until someone fixes the CatchLimiter logs (by the way, the on/off warnings could be green and red instead, respectively, like a semaphore).
 `_log()` could be replaced by `_trace()` if you REALLY want. That's a lot better than maintaining and understanding (my eyes kinda pulled out of the box when I first saw it) this giant if statement and those on/off flags.

YvesHenri:
Fixed issue #5714.

Gobberwart:
I think you're way off base with your approach to name handling. As you said, we *need* either a valid id (preferable) or valid name to be able to match against catch lists. Some sources do not provide this, and we KNOW it, so to completely ignore the issue and *remove* any kind of fix, is a huge step backwards. Be it on your head if you remove this.

Hash... Can you at least include the expiration time in that? There are plenty of pokemon "nests" around where the same rare pokemon may appear on a fairly regular basis within a few metres of each other. Your _hash method assumes any pokemon with a similar location is the same one and just ignores it. Adding expiration time to _hash gives a little better validation.

CatchLimiter... uh-huh. Something else for us to disagree about :) You've apparently decided to change this from a _log message to a _trace (debug message) so it will never appear unless debug is on. Seems pointless and another backward step to me. 

Anyway, I think I'm done with this now. We disagree on some points so can't move forward until we get someone else to evaluate our comments and offer feedback and/or approve.

Gobberwart:
I guess nobody else wants this revert either.

YvesHenri:
If you took the time, you would notice that there aren't many people active. There are some other requests that's been hanging ever since I left.

Gobberwart:
This is still here I see. I realise that there hasn't been much activity here, but honestly I think you're the only one who wants this change, and your reasoning for much of it is still questionable. Should be closed IMHO, and each change you want to make placed in a separate PR.

PR #5696, opened 2016-09-26
Revert "Avoid task init if task is disabled" (#5591)

YvesHenri:
As expected, I just came across with the following situation:

_If the user wants to disable catching nearby pokemons **ONLY** and enables Sniper and/or MoveToMapPokemon, the PokemonCatchWorker will never be initialized._ But what's so bad about this? Well, I need to store the `daily_catch_limit` value (PokemonCatchWorker task's config), since it'll be used from both Sniper/MTMP, but if that task is never initialized, I will never have this value.

This is just **ONE MORE** case that proves why we shouldn't handle such logic in the TreeConfigBuilder/BaseTask, like I said, unless you guys can come up with something better (post below).

These are definitely not viable:

- Load up and parse the whole config file and retrieve only the PokemonCatchWorker portion within both Sniper/MTMP tasks if need be

### DIFF
--- a/pokemongo_bot/__init__.py
+++ b/pokemongo_bot/__init__.py
@@ -217,7 +217,6 @@ class PokemonGoBot(object):
         self.event_manager.register_event('buddy_pokemon', parameters=('pokemon', 'iv', 'cp'))
         self.event_manager.register_event('buddy_reward', parameters=('pokemon', 'family', 'candy_earned', 'candy'))
         self.event_manager.register_event('buddy_walked', parameters=('pokemon', 'distance_walked', 'distance_needed'))
-        self.event_manager.register_event('task_disabled', parameters=('task_name',))
 
         #  ignore candy above threshold
         self.event_manager.register_event(
--- a/pokemongo_bot/event_handlers/logging_handler.py
+++ b/pokemongo_bot/event_handlers/logging_handler.py
@@ -81,7 +81,6 @@ class LoggingHandler(EventHandler):
         'softban_log':                       'magenta',
         'spin_limit':                        'red',
         'spun_pokestop':                     'cyan',
-        'task_disabled':                     'yellow',
         'threw_berry_failed':                'red',
         'transfer_log':                      'magenta',
         'unknown_spin_result':               'red',
--- a/pokemongo_bot/tree_config_builder.py
+++ b/pokemongo_bot/tree_config_builder.py
@@ -54,17 +54,6 @@ class TreeConfigBuilder(object):
                                         'See config.json.*example for more information.')
                 continue
 
-            if not task_config.get("enabled", True):
-                msg = "Task {task_name} is disabled"
-                self.bot.event_manager.emit(
-                    'task_disabled',
-                    sender=self,
-                    level='info',
-                    formatted=msg,
-                    data={'task_name': task_type}
-                )
-                continue
-
             if self._is_plugin_task(task_type):
                 worker = self.plugin_loader.get_class(task_type)
             else:
--- a/tests/tree_config_builder_test.py
+++ b/tests/tree_config_builder_test.py
@@ -1,7 +1,6 @@
 import unittest
 import json
 import os
-from mock import MagicMock
 from pokemongo_bot import PokemonGoBot, ConfigException, MismatchTaskApiVersion, TreeConfigBuilder, PluginLoader, BaseTask
 from pokemongo_bot.cell_workers import HandleSoftBan, CatchPokemon
 from pokemongo_bot.test.resources.plugin_fixture import FakeTask, UnsupportedApiTask
@@ -11,9 +10,7 @@ def convert_from_json(str):
 
 class TreeConfigBuilderTest(unittest.TestCase):
     def setUp(self):
-        self.bot = MagicMock()
-        self.bot.event_manager = MagicMock()
-        self.bot.event_manager.emit = lambda *args, **kwargs: None
+        self.bot = {}
 
     def test_should_throw_on_no_type_key(self):
         obj = convert_from_json("""[{


Gobberwart:
I tend to agree with you on this one, however I noticed yesterday an issue with the way Sniper instantiates PokemonCatchWorker...

`catch_worker = PokemonCatchWorker(pokemon, self.bot, self.config)`

Passes Sniper's config to PokemonCatchWorker, which results in daily_catch_limit (and other values) being set to their defaults instead of reading from config. I added these print statements to show the issue. (note: daily_catch_limit in config is set to 2000)

```
Initializing catch worker
config: {u'snipe_high_prio_threshold': 500, u'//NOTE: Change the max_sniping_distance to adjust the max sniping range (m)': {}, u'enabled': True, u'update_map_min_time_sec': 120, u'max_extra_dist_fort': 10, u'prioritize_vips': True, u'snipe_sleep_sec': 2, u'snipe_high_prio_only': False, u'snipe_max_in_chain': 2, u'//NOTE: Change the max_walking_distance to adjust the max walking range when snipe is off (m)': {}, u'max_sniping_distance': 20000000, u'walker': u'StepWalker', u'update_map': True, u'address': u'http://localhost:5000/raw_data', u'skip_rounds': 5, u'snipe': True, u'max_walking_distance': 500, u'map_path': u'raw_data', u'min_ball': 15, u'update_map_min_distance_meters': 500, u'mode': u'priority', u'debug': False, u'catch': {u'Oddish': 100, u'Weezing': 500, u'Graveler': 500, u'Caterpie': 10, u'Paras': 100, u'Kadabra': 600, u'Growlithe': 200, u'Kangaskhan': 1000, u'Beedrill': 500, u'Hitmonlee': 400, u'Poliwrath': 800, u'Machamp': 800, u'Butterfree': 500, u'Raichu': 1000, u'Omanyte': 150, u'==========T1 Evolvers==========': 0, u'Tangela': 300, u'Slowpoke': 200, u'Chansey': 800, u'Diglett': 200, u'==========Legendaries==========': 0, u'Poliwhirl': 400, u"Farfetch'd": 1000, u'Parasect': 500, u'Tentacool': 200, u'Magnemite': 250, u'Ditto': 900, u'Aerodactyl': 1000, u'Koffing': 200, u'Exeggcutor': 500, u'Shellder': 200, u'Magmar': 900, u'Mankey': 150, u'Dratini': 1000, u'Nidoqueen': 10, u'Charmeleon': 600, u'Psyduck': 150, u'Slowbro': 500, u'Sandslash': 10, u'Arcanine': 700, u'Omastar': 500, u'Snorlax': 1000, u'Articuno': 1000, u'Blastoise': 1000, u'Golem': 800, u'Pinsir': 150, u'Cloyster': 500, u'Eevee': 200, u'Tauros': 1000, u'Fearow': 10, u'Bulbasaur': 400, u'Jigglypuff': 150, u'Abra': 300, u'Arbok': 10, u'Doduo': 10, u'Muk': 500, u'==========Semi Rare==========': 0, u'Marowak': 800, u'Wartortle': 600, u'Wigglytuff': 500, u'Raticate': 10, u'Magikarp': 150, u'Starmie': 800, u'Mew': 1000, u'Geodude': 100, u'Pidgeotto': 10, u'Rattata': 10, u'Primeape': 500, u'Squirtle': 400, u'Vulpix': 200, u'Bellsprout': 100, u'Jolteon': 800, u'Venusaur': 1000, u'Spearow': 10, u'Meowth': 250, u'Golduck': 500, u'==========Region Locked==========': 0, u'Scyther': 200, u'Ekans': 10, u'==========Trash==========': 0, u'Alakazam': 800, u'Kabutops': 500, u'==========Very Rare==========': 0, u'Seel': 300, u'Voltorb': 200, u'Dragonite': 1000, u'Gyarados': 800, u'Vaporeon': 800, u'Metapod': 10, u'Dragonair': 1000, u'Mewtwo': 1000, u'Lickitung': 500, u'Rhydon': 500, u'Haunter': 500, u'Ivysaur': 600, u'Vileplume': 600, u'Gastly': 200, u'Drowzee': 600, u'Goldeen': 150, u'Pidgey': 10, u'Hypno': 600, u'Machoke': 400, u'Porygon': 200, u'Poliwag': 200, u'==========Starters==========': 0, u'Sandshrew': 10, u'==========T2 Evolvers==========': 0, u'Venomoth': 500, u'Gloom': 200, u'Nidorino': 10, u'Nidorina': 10, u'Pidgeot': 300, u'Tentacruel': 500, u'Kingler': 500, u'Exeggcute': 250, u'Weepinbell': 400, u'Golbat': 10, u'Gengar': 1000, u'Dodrio': 10, u'Rapidash': 500, u'Dugtrio': 500, u'Krabby': 100, u'Seadra': 600, u'Persian': 500, u'Nidoking': 10, u'Zapdos': 1000, u'Zubat': 10, u'Charmander': 400, u'Electrode': 500, u'Moltres': 1000, u'Victreebel': 800, u'Flareon': 800, u'Kabuto': 150, u'==========Uncommon==========': 0, u'Electabuzz': 900, u'Venonat': 100, u'Charizard': 1000, u'Pikachu': 600, u'Machop': 150, u'Jynx': 200, u'Kakuna': 10, u'Horsea': 200, u'Seaking': 500, u'Dewgong': 500, u'Hitmonchan': 400, u'Clefable': 500, u'Nidoran F': 10, u'Ponyta': 200, u'Rhyhorn': 200, u'Nidoran M': 10, u'Ninetales': 600, u'Cubone': 300, u'Weedle': 10, u'Onix': 600, u'Clefairy': 150, u'Grimer': 200, u'Magneton': 500, u'Mr. Mime': 1000, u'Lapras': 900, u'Staryu': 200}}
daily_catch_limit: 800
```



So, I'm not sure the task disabling is actually the cause of this.

YvesHenri:
That's exactly what I'm trying to fix.
I came up with the following:

```
class PokemonCatchWorker(BaseTask):
    DAILY_CATCH_LIMIT = 0

    def __init__(self, pokemon, bot, config):
        self.pokemon = pokemon
        super(PokemonCatchWorker, self).__init__(bot, config)

        if PokemonCatchWorker.DAILY_CATCH_LIMIT == 0:
            PokemonCatchWorker.DAILY_CATCH_LIMIT = self.config.get('daily_catch_limit', 800)
```

When PokemonCatchWorker is initialized for the first time (and that should probably be before the `catch_worker = PokemonCatchWorker(pokemon, self.bot, self.config)`), it'll update the value of DAILY_CATCH_LIMIT, which can be then retrieved with PokemonCatchWorker.DAILY_CATCH_LIMIT.

I am not sure if the above approach is the best but definitely, again, we have to remove this global task initialization logic out of the TreeConfigBuilder/BaseTask.

alexyaoyang:
Well, **instead** of doing all those stuffs to PokemonCatchWorker, since multiple tasks want to share a common field, just move the `daily_catch_limit` to global config. It's a smaller **and** simpler change compared to your suggestion. On top of that, moving `daily_catch_limit` to global config makes total sense to user that `daily_catch_limit` applies to not only PokemonCatchWorker.

Unless the PR in mention has broken something completely, I don't think you should be reverting it. Moreover when the PR in mention actually does fix quite a few issues.

Gobberwart:
It's not just daily_catch_limit though, it's all this stuff:

```
        self.min_ultraball_to_keep = self.config.get('min_ultraball_to_keep', 10)
        self.berry_threshold = self.config.get('berry_threshold', 0.35)
        self.vip_berry_threshold = self.config.get('vip_berry_threshold', 0.9)
        self.treat_unseen_as_vip = self.config.get('treat_unseen_as_vip', DEFAULT_UNSEEN_AS_VIP)
        self.daily_catch_limit = self.config.get('daily_catch_limit', 800)
        self.vanish_settings = self.config.get('vanish_settings', {})
```
Can we get the bot to call PokemonCatchWorker with its own config instead of supplying a new, unrelated one?

Gobberwart:
Annoyingly, tasks do not seem to persist their config, and are instead completely reinitialized every time they're called. In an ideal world, the tasks would not have to reinit every time. Maybe there's a bigger bug that we haven't spotted yet that causes this.

alexyaoyang:
I agree, we got to solve this problem at its roots. Reverting the PR will just be playing ping pong without solving the actual problem.

Gobberwart:
Yeah, it seems we're calling tasks as:

`task = PokemonCatchWorker(bot, config)`

This generates a new instance, instead of being able to reference an existing task/worker instance.

Not sure where the existing instance is kept... self.bot.workers... something?

YvesHenri:
@alexyaoyang Reverting this PR will solve the trouble it has already caused. It should of have been dispatched on the other 2 tasks that were acting strangely. I warned many of you.
Making `daily_catch_limit` (and all the other params Gobberwart mentioned) globaly is not an option either.

Gobberwart:
OK I've got a fix for it. I have no doubt someone with more python experience can make it a bit more elegant, but it works. PR shortly.

Gobberwart:
I'm not entirely sure that issue is related to PR #5591. I think we already had it, because of the way PokemonCatchWorker is called by sniper and movetomappokemon.

I still approve reverting this change, as it has no real benefit other than to remove control from individual tasks.

Gobberwart:
PR #5708 addresses that specific issue.

alexyaoyang:
Well, if you guys think so, go ahead. 

However I still think that if multiple tasks are sharing common config, that config **shouldn't** be in one of the task. It's just pure confusing.

Moving them to global is a great and simple idea that fixes this problem we have, and it's not too hard to implement, just move them to global and handle them in pokecli.py.

Gobberwart:
For now, revert this to fix problems it has caused. We can always revisit later.

Gobberwart:
@alexyaoyang I agree, it IS confusing. CatchPokemon passes its config to PokemonCatchWorker, which is fine, but then Sniper and MTMP need to pass the *correct* config, and they don't. 

Should we have config section 'type':'PokemonCatchWorker' which contains these values instead of having them in CatchPokemon?

For now, PR #5708 has a fix for this.

PR #5697, opened 2016-09-26
Loiter new meaning

janpascal:
## Short Description:

Add a `loiter` option to the json path used by `FollowPath`.  Similar to `wander` but it doesn't allow lower priority task to run while loitering.

Also make the progress log from `FollowPAth` a bit more informative by logging the location as found in the json file instead of the raw coordinates.

### DIFF
--- a/docs/configuration_files.md
+++ b/docs/configuration_files.md
@@ -880,13 +880,27 @@ This task is an upgrade version of the MoveToMapPokemon task. It will fetch poke
 ### Description
 [[back to top](#table-of-contents)]
 
-Walk to the specified locations loaded from .gpx or .json file. It is highly recommended to use website such as [GPSies](http://www.gpsies.com) which allow you to export your created track in JSON file. Note that you'll have to first convert its JSON file into the format that the bot can understand. See [Example of pier39.json] below for the content. I had created a simple python script to do the conversion.
+Walk to the specified locations loaded from .gpx or .json file. It is highly 
+recommended to use website such as [GPSies](http://www.gpsies.com) which allow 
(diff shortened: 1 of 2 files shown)


janpascal:
@alexyaoyang This one also OK with you?

alexyaoyang:
I think this is OK, but do we really need it? The exact same effect can be achieved with wander by not having any task below FollowPath, am I right?

janpascal:
Not exactly, since `loiter`/`wander` are per-point options; one might want to `wander` at one point of the route, and `loiter` at another. E.g., `loiter` at (one or more) known spawn points, and `wander` visiting a number of pokestops inbetween.

Gobberwart:
So... correct me if I'm wrong, but this basically means go to point, then do absolutely nothing at all for a specified number of seconds. No catching, no spinning fort, no inventory maintenance, just standing still doing nothing. Is that right? If so, what's the purpose?

alexyaoyang:
It could be useful to make the bot behave more human-like, most people take breaks at/near pokestops rather than randomly in the middle of a street for example. Similar to CampFort but without the need for lured forts.

janpascal:
No, during loiter the bot would still spin forts, catch pokemon etc but not any lower priority tasks like, in the default config, MoveToFort. Keep in mind that you can configure this for every point in the path, so if one of the points in the route is a spawn point you can configure to stay exactly there (`loiter`) for a while to catch the pokemon spawning there. Then walk to a known area with some forts: add a  `wander` to that point in your `path.json`  to walk to the some of the forst and spin a  them. Then continue with the next point in the route.

solderzzc:
👍

PR #5706, opened 2016-09-26
[WIP] pokemon_optimizer: fix transfer and evolve for xp

mudmud2k:
## Short Description:

Trying to fix the optimizer logic

Right now it always tries to evolve for xp, so it's holding back many useless pokemon and ignores a part of the config ('evolve_for_xp', 'evolve_only_with_lucky_egg')

## Fixes/Resolves/Closes (please use correct syntax):
- https://github.com/PokemonGoF/PokemonGo-Bot/issues/5685

Issue #5702, opened 2016-09-26
PokemonOptimizer Buddy/Family Bug


RedViper9:
### Expected Behavior
Log shows correct family for pokemon.

### Actual Behavior
PokemonOptimizer seems to think that dragonair is in the mew family.

### Your FULL config.json 
I use the BuddyPokemon task and do not define a buddy in PokemonOptimizer.
http://pastebin.com/JNVpNQ6s

### Output when issue occurred
[2016-09-26 15:33:18] 1 Dratini candy earned. You now have 49 candy!
[2016-09-26 15:33:19] Buddy Dragonair rewards Mew candies [+0 candies] [0 candies]
[2016-09-26 15:33:24] Buddy Dragonair rewards Mew candies [+0 candies] [0 candies]

### Steps to Reproduce
Walk a dragonair.

### Other Information
OS: Mint 18
Branch: Dev
Git Commit: c53a57e9a2fc43844e62c6492883b5346042e5c3
Python Version: 2.7.12

Anakin5:
It is because of the other buddy task getting the candy first.

PR #5707, opened 2016-09-27
Fix for evolve_for_xp False + others

Anakin5:
- Fix for evolve_for_xp False that was not taken into account (Same intend as #5706)
- Do not display any message when zero candy earn from buddy (#5702)
- A bit stronger pokemon name fixing
- Some refactoring
- Update one config example
- Add more things to doc


solderzzc:
👍  This is pretty nice.

YvesHenri:
Can't review now. Someone please do it and merge, if agreed.

PR #5708, opened 2016-09-27
Load CatchPokemon config in PokemonCatchWorker

Gobberwart:
## Short Description:

If PokemonCatchWorker is instantiated from Sniper or MoveToMapPokemon, the config passed is not the correct config for PokemonCatchWorker, which results in defaults being set instead of using the config file values.

This PR loads the CatchPokemon config into PokemonCatchWorker if no config is provided.

This solution is probably not great python, but it works. If anyone can see a better way to grab this config from the already-existing list of workers, please advise.

Fixes issues like those seen in #5696 and #5684 

IMPORTANT NOTE: If the CatchPokemon task is disabled (PR #5591 which completely ignores CatchPokemon task config if the task is disabled), default values will be used instead. 

- #5696
- #5684

### DIFF
--- a/pokemongo_bot/cell_workers/move_to_map_pokemon.py
+++ b/pokemongo_bot/cell_workers/move_to_map_pokemon.py
@@ -255,7 +255,7 @@ class MoveToMapPokemon(BaseTask):
         # If target exists, catch it, otherwise ignore
         if exists:
             self._encountered(pokemon)
-            catch_worker = PokemonCatchWorker(pokemon, self.bot, self.config)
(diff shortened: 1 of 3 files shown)


alexyaoyang:
👍 LGTM

solderzzc:
Great, finally have it fixed.

khoaanh2212:
i've still receive error message from snipper : Snipping gives error: daily catch limit reached.
Can you help me? thanks.

khoaanh2212:
it fixed,i add :
"daily_catch_limit": 5000
into movetomappokemon and sniper and it works.
thanks.

Issue #5620, opened 2016-09-22
Feature request: FollowPath with different vehicles


janpascal:
### Short Description
It would be nice if, for every point in the path for the FollowPath task, you could indicate which vehicle should be simulated.

- walking (default)
- bicycle (speed around 20 km/h)
- car (max speed as indicated by the google maps api?)
- plane (900 km/h, no connection to the API for the flying time)

### Possible solution
In the FollowTask walker, use a different speed depending on the vehicle for the current stretch. When 'flying', pause for the flying time. 

### How it would help others
Helps to get to an interesting location more quickly without teleporting. Combine with loitering to, e.g. fly to California, walk around there from pokestop to pokestop, catching pokemon on the way, for a couple of hours, then drive to another city (no egg incubating during the trip), walk around there for another couple of hours, then fly back home.

alexyaoyang:
Sounds interesting, could be useful! How should it work with walk_min and walk_max?

janpascal:
I imagine those would only be used if mode==walking. The speed implied by the Google directions reply (it gives time and distance for each step)  can be used for driving and bicycling. For flying a constant can be used. 


On 27 September 2016 07:58:29 CEST, Alex Yao <notifications@github.com> wrote:
>Sounds interesting, could be useful! How should it work with walk_min
>and walk_max?
>
>-- 
>You are receiving this because you authored the thread.
>Reply to this email directly or view it on GitHub:
>https://github.com/PokemonGoF/PokemonGo-Bot/issues/5620#issuecomment-249773483

-- 
Sent from my Android device with K-9 Mail. Please excuse my brevity.

janpascal:
I'm working on this. The path.json would look like this:
```
[
        {"location": "Lindelaan 33, Leiderdorp", "mode": "walking"},
        {"location": "Eikenlaan 25, Leiderdorp", "mode": "walking"},
        {"location": "Hoofdstraat 2, Leiderdorp", "mode": "walking"},
        {"location": "Aankomstpassage, Schiphol", "mode": "driving"},
        {"location": "Hartmann Road, London", "mode": "flying"},
        {"location": "4 Cannon Street, Londeon", "mode": "driving"}
]
```

[...]

PR #5717, opened 2016-09-27
New walking modes (bicycle, driving and flying) for FollowPath

janpascal:
## Short Description:

For the `FollowPath` task with `walker` set to `PolylineWalker`, allow `mode` fields in the points in the `path.json`. Simulates `walking`, `bicycling`, `driving` or `flying`.

Fixes: #5620

### DIFF
--- a/docs/configuration_files.md
+++ b/docs/configuration_files.md
@@ -79,8 +79,10 @@ Document the configuration options of PokemonGo-Bot.
 | `max_steps`        | 5       | The steps around your initial location (DEFAULT 5 mean 25 cells around your location) that will be explored
 | `forts.avoid_circles`             | False     | Set whether the bot should avoid circles |
 | `forts.max_circle_size`             | 10     | How many forts to keep in ignore list |
-| `walk_max`             | 4.16    | Set the maximum walking speed in m/s (1 is about 3.6km/hr) 4.16m/s = 15km/h
-| `walk_min`             | 2.16    | Set the minimum walking speed in m/s (1 is about 3.6km/hr) 2.16m/s = 7.8km/h
+| `walk_max`             | 4.16    | Set the maximum walking speed in m/s (1 is about 3.6km/h) 4.16m/s = 15km/h
+| `walk_min`             | 2.16    | Set the minimum walking speed in m/s (1 is about 3.6km/h) 2.16m/s = 7.8km/h
+| `fly_max`              |  282    | Set the maximum flying speed in m/s (1 is about 3.6km/h) 282m/s = 1015km/h
+| `fly_min`              |  294    | Set the minimum flying speed in m/s (1 is about 3.6km/h) 294m/s = 1058km/h
 | `action_wait_min`   | 1       | Set the minimum time setting for anti-ban time randomizer
 | `action_wait_max`   | 4       | Set the maximum time setting for anti-ban time randomizer
 | `debug`            | false   | Let the default value here except if you are developer                                                                                                                                      |
@@ -887,6 +889,23 @@ indicated the number of seconds the bot should wander after reaching the point.
 During this time, the next Task in the configuration file is executed, e.g. a
 MoveToFort task. This allows the bot to walk around the waypoint looking for
 forts for a limited time.
+
+For each point in the json file, the optional `mode` field indicates the
+simulated travel model. This is only supported when `walker` is set to
+`PolylineWalker`. When `mode` is set to `driving` or `bicycling`, the
+Google Directions API is used to get the appropriate route.
+
+The supported values for `mode` are:
+* `walking` - walk as usual, using a speed between `walk_min` and `walk_max` 
+  in `config.json`
+* `bicycling` - simulate riding a bicycle. The speed from the Google Directions
+  API is used
+* `driving` - simulate driving a car. The speed from the Google Directions
+  API is used, which depends on the road (e.g. 100 km/h on a highway and 30 km/h 
+  in a suburban area).
+* `flying` - simulate taking an aeroplane. Flies are typical airline cruising
+  speed, configurable with `fly_min` and `fly_max`. During the simulated flight
+  no other actions are taken by the bot.
 
 ### Options
 [[back to top](#table-of-contents)]
--- a/docs/configuration_files.md
+++ b/docs/configuration_files.md
@@ -81,8 +81,8 @@ Document the configuration options of PokemonGo-Bot.
 | `forts.max_circle_size`             | 10     | How many forts to keep in ignore list |
 | `walk_max`             | 4.16    | Set the maximum walking speed in m/s (1 is about 3.6km/h) 4.16m/s = 15km/h
 | `walk_min`             | 2.16    | Set the minimum walking speed in m/s (1 is about 3.6km/h) 2.16m/s = 7.8km/h
-| `fly_max`              |  282    | Set the maximum flying speed in m/s (1 is about 3.6km/h) 282m/s = 1015km/h
-| `fly_min`              |  294    | Set the minimum flying speed in m/s (1 is about 3.6km/h) 294m/s = 1058km/h
+| `fly_max`              |  222    | Set the maximum flying speed in m/s (1 is about 3.6km/h) 222m/s = 800km/h
+| `fly_min`              |  202    | Set the minimum flying speed in m/s (1 is about 3.6km/h) 202m/s = 727km/h
 | `action_wait_min`   | 1       | Set the minimum time setting for anti-ban time randomizer
 | `action_wait_max`   | 4       | Set the maximum time setting for anti-ban time randomizer
 | `debug`            | false   | Let the default value here except if you are developer                                                                                                                                      |
--- a/pokemongo_bot/cell_workers/follow_path.py
+++ b/pokemongo_bot/cell_workers/follow_path.py
@@ -81,8 +81,7 @@ class FollowPath(BaseTask):
             point["lat"] = float(point_tuple[0])
             point["lng"] = float(point_tuple[1])
             point["alt"] = float(point_tuple[2])
-            if "mode" not in point:
-                point["mode"] = "walking"
+            point["mode"] = mode.get("mode", "walking")
         return points
 
     def load_gpx(self):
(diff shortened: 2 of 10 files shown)


pokemongo_bot/cell_workers/follow_path.py
```diff
@@ -77,10 +77,12 @@ def load_json(self):
                     'position': point_tuple
                 }
             )
-            # Keep point['location']
+            # Keep point['location'] and other keys
             point["lat"] = float(point_tuple[0])
             point["lng"] = float(point_tuple[1])
             point["alt"] = float(point_tuple[2])
+            if "mode" not in point:
+                point["mode"] = "walking"
```
YvesHenri: Replace this if block for `point["mode"] = mode.get("mode", "walking")`. If a mode has been set, it'll stay as it is, otherwise assign "walking" to it.

pokemongo_bot/cell_workers/follow_path.py
```diff
@@ -158,11 +160,15 @@ def work(self):
             alt = uniform(self.bot.config.alt_min, self.bot.config.alt_max)
 
         if self.bot.config.walk_max > 0:
+            mode = "walking"
+            if "mode" in point:
+                mode = point["mode"]
```
YvesHenri: Do this instead: `mode = point.get("mode", "walking")`

YvesHenri:
There might be more changes. Can't give a full review now.

Anakin5:
Please don't store any speed info in the Polyline. The points can be travelled at any speed.
Also, the code can move to Step Walker. There is nothing specific to Polyline here.

janpascal:
I would think there is, because the speed is determined from the travelling times returned from the Google Directions API.

Anakin5:
what do you mean by
> the speed is determined from the travelling times returned from the Google Directions API

janpascal:
The Polyline (as it was before) asks the Google Directions API for directions from the starting point to the end point. These are the same directions you see on the Google Maps web site if you ask for directions there. Every segment of the route that the API replies has a distance (in meters) and an expected travel time (in seconds). In my PR I use these to calculate the speed at which the bot simulates travel for each segments. The speed is different for each segment, because Google takes into account the type of road (freeway, freeway ramp, main road, inner city road) when it calculates the expected travel time.

Anakin5:
And then ? Why does that explains the fact that you need to store a speed with the Polyline. I can choose to follow the points at whatever speed I want once the polyline is there. I can walk, I can take a bike, I can go backward. Why do you need to store that speed since the polyline is no going to be used with that speed anyway ?

Anakin5:
Also, you dd not answer what makes your feature specific to Polyline. Why not StepWalker ?

th3w4y:
@janpascal I agree with Anakin5 (that's a new one :) ) but we don't need speed on polyline...

When i wrote the poly line I decided to not use the information provided by google regarding speed per segment for 2 reasons:

- the speed variance per segment is insignificant.. (+/-10%)
- we don't want to be very precise... in following the Google Directions
- google Directions are just a statistical estimate (I for one usually walk about 45% faster then Google Directions...)

docs/configuration_files.md
```diff
@@ -79,8 +79,10 @@ Document the configuration options of PokemonGo-Bot.
 | `max_steps`        | 5       | The steps around your initial location (DEFAULT 5 mean 25 cells around your location) that will be explored
 | `forts.avoid_circles`             | False     | Set whether the bot should avoid circles |
 | `forts.max_circle_size`             | 10     | How many forts to keep in ignore list |
-| `walk_max`             | 4.16    | Set the maximum walking speed in m/s (1 is about 3.6km/hr) 4.16m/s = 15km/h
-| `walk_min`             | 2.16    | Set the minimum walking speed in m/s (1 is about 3.6km/hr) 2.16m/s = 7.8km/h
+| `walk_max`             | 4.16    | Set the maximum walking speed in m/s (1 is about 3.6km/h) 4.16m/s = 15km/h
+| `walk_min`             | 2.16    | Set the minimum walking speed in m/s (1 is about 3.6km/h) 2.16m/s = 7.8km/h
+| `fly_max`              |  282    | Set the maximum flying speed in m/s (1 is about 3.6km/h) 282m/s = 1015km/h
```
th3w4y: passenger flight usually travel at around 800Km/h

pokemongo_bot/walkers/polyline_generator.py
```diff
@@ -104,30 +117,38 @@ def __init__(self, origin, destination, google_map_api_key=None):
...
+    def _get_directions_points_and_speeds(self):
         points = []
-        for point in self._directions_encoded_points:
-            points += polyline.decode(point)
-        return [x for n, x in enumerate(points) if x not in points[:n]]
+        speeds = []
+        for point in zip(self._directions_encoded_points, self._directions_speeds):
+            new_points = polyline.decode(point[0])
+            points += new_points
+            speeds += [ point[1] * uniform(0.9, 1.1) for p in new_points ]
```
th3w4y: if you are modifying them with +/- 10% you are kinda removing canceling the variance anyway....

pokemongo_bot/walkers/polyline_walker.py
```diff
@@ -1,17 +1,51 @@
...
+            if self.mode == "walking":
+                speed = uniform(self.bot.config.walk_min, self.bot.config.walk_max)
+            if self.mode == "flying":
+                self.handle_flight()
+                return True
+            else:
+                try:
+                    cache = PolylineObjectHandler._cache
+                    _,_,speed = cache._step_dict[cache._step_keys[cache._last_step]]
+                    speed = speed * uniform(0.95, 1.05)
```
th3w4y: again you are canceling out the variance

pokemongo_bot/walkers/polyline_generator.py
```diff
@@ -59,19 +64,23 @@ def cached_polyline(origin, destination, google_map_api_key=None):
...
                 '{},{}'.format(*self.origin),
                 '{},{}'.format(*self.destination))
         if google_map_api_key:
             self.DIRECTIONS_URL = '{}&key={}'.format(self.DIRECTIONS_URL, google_map_api_key)
         self._directions_response = requests.get(self.DIRECTIONS_URL).json()
+
         try:
             self._directions_encoded_points = [x['polyline']['points'] for x in
                                                self._directions_response['routes'][0]['legs'][0]['steps']]
+            self._directions_speeds = [x['distance']['value'] / float(x['duration']['value']) for x in
```
th3w4y: x['duration']['value']  can be Zero!!!

pokemongo_bot/event_handlers/social_handler.py
```diff
@@ -83,7 +83,11 @@ def initialize(self):
             return
 
     def run(self):
-        self._mqttc.connect("broker.pikabot.org", 1883, 20)
+        try:
+            self._mqttc.connect("broker.pikabot.org", 1883, 20)
+        except Exception as e:
```
th3w4y: what was the Exception that you are trying to catch?

pokemongo_bot/walkers/walker_factory.py
```diff
@@ -8,5 +8,8 @@ def walker_factory(name, bot, dest_lat, dest_lng, dest_alt=None, *args, **kwargs
     if 'StepWalker' == name:
         ret = StepWalker(bot, dest_lat, dest_lng, dest_alt)
     elif 'PolylineWalker' == name:
-        ret = PolylineWalker(bot, dest_lat, dest_lng)
+        mode = "walking"
+        if "mode" in kwargs:
+            mode = kwargs["mode"]
+        ret = PolylineWalker(bot, dest_lat, dest_lng, mode=mode)
```
th3w4y:
```
    elif 'PolylineWalker' == name:
        ret = PolylineWalker(bot, dest_lat, dest_lng, mode=kwargs.get('mode', walking'))
```

instead of:

```
    elif 'PolylineWalker' == name:
        mode = "walking"
        if "mode" in kwargs:
            mode = kwargs["mode"]
        ret = PolylineWalker(bot, dest_lat, dest_lng, mode=mode)
```

th3w4y:
@janpascal  on second review I see what you try to do with the speed in poly line....

The would apply only for Driving directions (there can be a reason to have speed I agree)
But... (do we even have the "I am a passenger" implemented????)

Even if we would store the speeds in the poly line object... (I am still not convinced by its value...) we would need to modify the StepWalker there is the place where the speed variance appears on each tick... and You are not reusing the speed stored in the polyline in StepWalker in the proposal of this PR anyway... (or am I missing something...?)

pokemongo_bot/walkers/polyline_walker.py
```diff
@@ -1,17 +1,51 @@
...
 
 
 class PolylineWalker(StepWalker):
+    def handle_flight(self):
+        speed = uniform(self.bot.config.fly_min, self.bot.config.fly_max)
+        origin_lat, origin_lng, origin_alt = self.bot.position
+        remaining = Geodesic.WGS84.Inverse(origin_lat, origin_lng, self.dest_lat, self.dest_lng)["s12"]
+        time_to_sleep = remaining / speed
+        sleep(time_to_sleep)
+        self.bot.api.set_position(self.dest_lat, self.dest_lng, self.dest_alt)
```
th3w4y: There is only one class changing bot.position 

that is StepWalker.... it would be nice if we could keep it that way!!! in the end StepWalker is the super class...

Flight will teleport to destination ...
after sleeping the required time....

you can do that by using super(PolylineWalker, self).step(float("inf"))

pokemongo_bot/event_handlers/social_handler.py
janpascal: I don't remember the exact type, but you get an exception if your firewall doesn't allow outgoing traffic on port 1883, crashing the application

pokemongo_bot/walkers/polyline_walker.py
janpascal: That's a good suggestion, thanks

pokemongo_bot/cell_workers/follow_path.py
janpascal: Thanks, I'll change that

janpascal:
@th3w4y You're right, the speed per segment is only useful for driving directions. Maybe I should implement the car travel in a way similar to plane travel, e.g., pause and then teleport? There are usually not many pokestops or pokemon around highways anyway.

pokemongo_bot/walkers/polyline_generator.py
janpascal: You're right, that was dumb. Let's see though if we agree that driving should be implemented like flying (sleep and teleport), in which case this code is going out anyway

docs/configuration_files.md
janpascal: You're right, I got those numbers from somewhere, but a fresh search shows your numbers are more realistic

BreezeRo:
Any updates on this?

PR #5749, opened 2016-10-01
Separate the event of Pokemon_caught to one. One for normal and another for VIP…

developerllazy:
Separate the event of Pokemon_caught to two. One for normal pokemon was caught and 1 for VIP pokemon was caught. Therefore, you can subscrib the vip_caught event in telegram.

### DIFF
--- a/pokemongo_bot/__init__.py
+++ b/pokemongo_bot/__init__.py
@@ -477,6 +477,20 @@ class PokemonGoBot(object):
             )
         )
         self.event_manager.register_event(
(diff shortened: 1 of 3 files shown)


Gobberwart:
Needs color defined in logging_handler.py, eg. ` 'vip_pokemon_caught':                    'blue',`

developerllazy:
@Gobberwart add the color in logging_handler

solderzzc:
👍

Issue #5723, opened 2016-09-28
Bot stuck in loop of transferring Dratini


eldonormalkowski:
The bot ran out off space for new pokemon, so it attempted to transfer junk. It worked so far unless it attempted to transfer this dratini:

[2016-09-28 07:37:53] [MainThread] [TransferPokemon] [INFO] *Releasing Dratini* CP: 61, IV: 0.33, IVCP: 0.34 | based on rule: CP < 700 OR IV < 0.0 IVCP < 0

I have no idea how to recreate this exact issue. More log looks like this:
[2016-09-28 07:36:46] [MainThread] [TransferPokemon] [INFO] *Releasing Dratini* CP: 61, IV: 0.33, IVCP: 0.34 | based on rule: CP < 700 OR IV < 0.0 IVCP < 0
[2016-09-28 07:36:47] [MainThread] [PokemonCatchWorker] [INFO] Your Pokemon inventory is full! Could not catch!
[2016-09-28 07:36:50] [MainThread] [TransferPokemon] [INFO] *Releasing Dratini* CP: 61, IV: 0.33, IVCP: 0.34 | based on rule: CP < 700 OR IV < 0.0 IVCP < 0
[2016-09-28 07:36:51] [MainThread] [PokemonCatchWorker] [INFO] Your Pokemon inventory is full! Could not catch!
[2016-09-28 07:36:51] [MainThread] [MoveToFort] [INFO] Moving towards pokestop Foyer D'étudiantes - 0.04km
[2016-09-28 07:36:54] [MainThread] [TransferPokemon] [INFO] *Releasing Dratini* CP: 61, IV: 0.33, IVCP: 0.34 | based on rule: CP < 700 OR IV < 0.0 IVCP < 0
[2016-09-28 07:36:55] [MainThread] [PokemonCatchWorker] [INFO] Your Pokemon inventory is full! Could not catch!
[2016-09-28 07:36:56] [MainThread] [  SpinFort] [INFO] Spun pokestop Foyer D'étudiantes. Experience awarded: 50. Items awarded: 1x Pokeball, 2x Greatball
[2016-09-28 07:37:00] [MainThread] [MoveToFort] [INFO] Moving towards pokestop Paris11, Boulangerie Ounissi - 0.06km
[...]

PR #5752, opened 2016-10-03
Do not try to transfer buddy

Gobberwart:
## Short Description:

Transfer attempt of buddy pokemon currently fails and results in endless transfer loop. Added check to both TransferPokemon and PokemonOptimizer to skip trying to transfer buddy pokemon.

- #5723

### DIFF
--- a/pokemongo_bot/cell_workers/transfer_pokemon.py
+++ b/pokemongo_bot/cell_workers/transfer_pokemon.py
@@ -19,6 +19,7 @@ class TransferPokemon(BaseTask):
         self.min_free_slot = self.config.get('min_free_slot', 5)
         self.transfer_wait_min = self.config.get('transfer_wait_min', 1)
         self.transfer_wait_max = self.config.get('transfer_wait_max', 4)
+        self.buddy = self.bot.player_data.get('buddy_pokemon', {})
 
     def work(self):
         if not self._should_work():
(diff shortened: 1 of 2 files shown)


pokemongo_bot/cell_workers/transfer_pokemon.py
```diff
@@ -31,7 +32,7 @@ def work(self):
 
         if self.bot.config.release.get('all'):
             group = [p for p in inventory.pokemons().all()
-                     if p.in_fort is False and p.is_favorite is False]
+                     if not p.in_fort and not p.is_favorite and not (p.unique_id == self.buddy['id'])]
```
DeXtroTip: If there is no buddy, `self.buddy['id']` will give an error

Gobberwart: Good point. Give me a few minutes.

Gobberwart:
OK that should take care of buddy not set situation. @DeXtroTip, thoughts?

DeXtroTip:
It could have been something like `p.unique_id == self.buddy.get('id', 0)`, i think. But yeah, it's okay now

Gobberwart:
I thought about that, but figured it might still generate an error if NoneType isn't iterable. Then I thought about over-complicating that one line and just decided to palm it off to another method.

Gobberwart:
If you're happy with this change @DeXtroTip could you please approve for merge?

Issue #5753, opened 2016-10-03
docker-compose crashes on sleep > 60sec


gise88:
### Expected Behavior
Working without errors.

### Actual Behavior & Output when issue occurred
When bot need to sleep over 60 seconds:
```ERROR: An HTTP request took too long to complete. Retry with —verbose to obtain debug information.
If you encounter this issue regularly because of slow network conditions, consider setting COMPOSE_HTTP_TIMEOUT to a higher value (current value: 60).```


### Steps to Reproduce
Run the bot and wait for a sleep


### Other Information
OS:
Arch Linux 4.7.5-1-ARCH #1 SMP PREEMPT Sat Sep 24 13:04:22 CEST 2016 x86_64 GNU/Linux

Branch:
dev

Git Commit:
https://github.com/PokemonGoF/PokemonGo-Bot/commit/db650fbc216f99a46ff31fdf2914ec3f077d07f5

Python Version:
Working with the original docker-compose

PR #5754, opened 2016-10-03
COMPOSE_HTTP_TIMEOUT fix

gise88:
## Short Description:
Adding .env file in order to set the docker-compose timeout to 10 hours.
This will avoid the following error when bot need to sleep:
```ERROR: An HTTP request took too long to complete. Retry with —verbose to obtain debug information.
If you encounter this issue regularly because of slow network conditions, consider setting COMPOSE_HTTP_TIMEOUT to a higher value (current value: 60).```


- https://github.com/PokemonGoF/PokemonGo-Bot/issues/5753
-
-


Gobberwart:
I'm not sure adding an env file is a great idea... What happens if people have made their own env files and suddenly start getting conflicts/errors?

I'm also not sure this should be an issue. If the bot is asleep, it shouldn't be making any HTTP requests at all, let alone ones that take more than 60 seconds to complete. 

Is there an underlying cause for this that should be fixed instead?

gise88:
Oh, ok! I understand.. What I did is an easy workaround to bypass a more big problem, but you are right!

Close.

PR #5766, opened 2016-10-05
Added absolute candy limit

ChaosMarc:
In addition to `candy_limit` I've added a new config parameter `candy_limit_absolute`. It should be used to stop collecting candy with your current buddy when a candy threshold for your buddy's pokemon family is reached (e.g. 50 for evolving).
It also should fix a potential crash when your current buddy gets evolved and no other pokemon of the same type is available.


solderzzc:
👍

PR #5771, opened 2016-10-06
Some translation issue of zh-tw

kevinlife:
1. fix Lapras translation issue
2. fix Nidoran F key issue, instead of Nidoran♀, it should be "Nidoran F".
3. fix Nidoran M key issue, instead of Nidoran♂, it should be "Nidoran M".


solderzzc:
👍

PR #5791, opened 2016-11-08
Changing the pgoapi tag on requirements.txt

mmixx:
## Short Description:
The tag set **1a112879f424b518f994b99f43859dd95cd43fdf** is not the right tag that **pgoapi** devs set on their branch... and when you try to update using using that tag, it will say:

tag-error

And `pip` will not try update your `pgoapi` copy...


solderzzc:
👍

PR #5704, opened 2016-09-26
pokecli.py: wait_time for every exception randomized

mudmud2k:
Feature Request (https://github.com/PokemonGoF/PokemonGo-Bot/issues/5579)

Short Description:
it's create a random waiting time based on config for all exception (connection lost, throttling, and so on), so no static time span will be used anymore
---

randint(min,max)
config.reconnecting -> min
config.reconnection \* 60 -> max

PR #5800, opened 2016-11-09
Fix ValueError crash

Gobberwart:
## Short Description:

FIxes bot crash introduced by PR #5704.

`ValueError, "non-integer arg 1 for randrange()"`

Error is caused by trying to pass FLOATs to the randrange function instead of INTs.

Does not happen when reconnecting_timeout in config is set to default 15, as 15 * 0.8 = 12, and 15 * 1.2 = 18, so both are INTs, however if reconnecting_timeout is changed from the default, this error occurs.

### DIFF
--- a/pokecli.py
+++ b/pokecli.py
@@ -186,7 +186,9 @@ def main():
         finished = False
 
         while not finished:
-            wait_time = randint((config.reconnecting_timeout * 0.8 * 60), (config.reconnecting_timeout *  1.2 * 60))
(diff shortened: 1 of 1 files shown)


mudmud2k:
Instead of random.randint it also could use random.uniform, so there is no need for helping variables or an int converting...

Sorry that I forgot to add the int() and missed the int/float-error, it should had been:

`wait_time = randint(int(config.reconnecting_timeout * 0.8 * 60), int(config.reconnecting_timeout * 1.2 * 60))`

PR #5803, opened 2016-11-10
Better search for encrypt/hash files

Gobberwart:
## Short Description:

- Fixes file name issue for macos users (libniantichash-osx-64.so not found)
- Searches multiple locations for required files... encrypt_location as specified in config, bot root dir, and ./src/pgoapi/pgoapi/lib.


Gobberwart:
Was meant to be in dev, sorry.

Issue #5798, opened 2016-11-09
Bot Won't Start


yelnatscoding:
// ♥ ./run.sh
2016-11-09 05:02:11,052 [       cli] [INFO] PokemonGO Bot v1.0
2016-11-09 05:02:11,062 [       cli] [INFO] commit: 5439a712
2016-11-09 05:02:11,066 [       cli] [INFO] Configuration initialized
2016-11-09 05:02:11,067 [pokemongo_bot.health_record.bot_event] [INFO] Health check is enabled. For more information:
2016-11-09 05:02:11,067 [pokemongo_bot.health_record.bot_event] [INFO] https://github.com/PokemonGoF/PokemonGo-Bot/tree/dev#analytics
2016-11-09 05:02:11,083 [requests.packages.urllib3.connectionpool] [INFO] Starting new HTTP connection (1): www.google-analytics.com
(16640) wsgi starting up on http://127.0.0.1:4000
[2016-11-09 05:02:11] [SleepSchedule] [INFO] Next sleep at 12:17:27, for a duration of 05:15:13
[2016-11-09 05:02:11] [PokemonGoBot] [INFO] Setting start location.
[2016-11-09 05:02:11] [PokemonGoBot] [INFO] [x] Coordinates found in passed in location, not geocoding.
[2016-11-09 05:02:11] [PokemonGoBot] [INFO] Location found: 40.772250, -73.971413 (40.77225, -73.971413, 8.0)
[2016-11-09 05:02:11] [PokemonGoBot] [INFO] Now at (40.77225, -73.971413, 8.0)
[2016-11-09 05:02:11] [PokemonGoBot] [INFO] Login procedure started.
_inventory was not initialized
_inventory was not initialized
[2016-11-09 05:02:11] [       cli] [INFO] 
[2016-11-09 05:02:11] [       cli] [INFO] Ran for 0:00:00
[2016-11-09 05:02:11] [       cli] [INFO] Total XP Earned: 0  Average: 0.00/h
[2016-11-09 05:02:11] [       cli] [INFO] Travelled 0.00km
[2016-11-09 05:02:11] [       cli] [INFO] Visited 0 stops
[2016-11-09 05:02:11] [       cli] [INFO] Encountered 0 pokemon, 0 caught, 0 released, 0 evolved, 0 never seen before ()
[2016-11-09 05:02:11] [       cli] [INFO] Threw 0 pokeballs
[...]

PR #5804, opened 2016-11-10
Fix hash/encrypt library searching

Gobberwart:
## Short Description:

- Fixes file name issue for macos users (libniantichash-osx-64.so not found)
- Searches multiple locations for required files... encrypt_location as specified in config, bot root dir, and ./src/pgoapi/pgoapi/lib.
- Update pgoapi src in requirements.txt to known working copy.


solderzzc:
👍

krummrey:
libniantichash-osx-64.so can't be found. 
The merge is broken on OS X!!! Even after a reset and/or clean install.
#5798

PR #5826, opened 2016-11-20
API Validation - Killswitch on API Change

eRobot1:
## Short Description:
Check if Niantic API is equal to current API. If not, warn and prompt user .
- Running old bot when Niantic force updates version on client

Validate if Official API has changed

### DIFF
--- a/pokemongo_bot/api_wrapper.py
+++ b/pokemongo_bot/api_wrapper.py
@@ -4,6 +4,8 @@ import logging
 import random, base64, struct
 import hashlib
 import os
+import urllib
+import sys
 from pgoapi.exceptions import (ServerSideRequestThrottlingException,
                                NotLoggedInException, ServerBusyOrOfflineException,
                                NoPlayerPositionSetException, EmptySubrequestChainException,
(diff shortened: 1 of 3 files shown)


houkop:
Valid api is 0.45.0, not 0.43.0

eRobot1:
@houkop correct. This code checks and compares

MZorzy:
? 0.45 is in official app
0.43 is on api for bot/map
so now it must not run, -> capi 0.43

tiagoAnon:
I believe the safe approach would be to check the current version on every login. The bot will have to login every time the session becomes stale. If Niantic changes the api version mid-botting, they will probably flag the account immediately.

Instead of prompting if the user wants to take a risk a flag should be added to the config.json with the default value being false.

reigelgallarde:
should be better if we can add this on the config file...
default to true on the python script. So we can change the config file if we really don't care about the version (example, I'm developing, the version doesn't matter, log me in...)
````
 ....
    "check_niantic_api": {
      "enabled": false
    },
 ....
````

also, I think the better place for this is on api_wrapper.py. We could add another def check_niantic_api. And use it inside def login... this way, it would always check every time it logs in. But yes, there's should be an option to turn this off.. then maybe add a warning when this is off...

Anyway, just a suggestion... I'm not a python programmer.. 😄

**Comment by showei** [comment]
@reigelgallarde I have a bash script wrapping run.sh to check api version and switch accounts every hour, not sure if it's a better place to hold it.

**Comment by eRobot1** [comment]
@tiagoAnon @reigelgallarde  Took your advice. Added logic to API Wrapper to handle to re-loggin on stale session. Added flag to config as well to override API check.

**Comment by eRobot1** [comment]
@ChaosMarc can you merge this?

File: pokemongo_bot/api_wrapper.py [hunk 1]
```diff
@@ -90,6 +92,17 @@ def create_request(self):
     def login(self, provider, username, password):
         # login needs base class "create_request"
         self.useVanillaRequest = True
+        if self.config.check_niantic_api is True:
+            capi = '0.45.0'
+            link = "https://pgorelease.nianticlabs.com/plfe/version"
+            f = urllib.urlopen(link)
+            myfile = f.read()
+            self.config.check_niantic_api
```
reigelgallarde: is line 100 an extra?

solderzzc:
👍

cierzniak:
Why kill switch was merged into master but no dev?

PR #5835, opened 2016-11-23
Custom admdebian

Admdebian:
I think that is very usefull to have coordinates in event messages.

When I got a message on telegram I know where the event appens.

Where should we put this change?


MZorzy:
@Admdebian is `, 8` altitude ? 
u added lat,long.. missed altitude. can be usefull to see if is not fixed (bot detection)

PR #5839, opened 2016-11-25
Improved logged informations for caught pokemons

brunifrancesco:
## Short Description:
Log better formatted informations in the caught pokemon log file. 
Now current datetime is added to result dictionary.
Each line is a valid JSON document and it also includes the coordinates where the pokemon has been caught.
Keys order is always remains constant for each written JSON document.


solderzzc:
👍

Issue #5845, opened 2016-12-03
incompatible version of websocket websocket-client from fresh install


johnscancella:
### Expected Behavior
After a fresh install it starts when running `./run.sh`

### Actual Behavior
I get the following output
```
An incompatible websocket library is conflicting with the one we need.
You can remove the incompatible library and install the correct one
by running the following commands:

yes | pip uninstall websocket websocket-client
pip install -U websocket-client
Fri Dec 2 19:49:39 EST 2016 Pokebot  Stopped.
Press any button or wait 20 seconds to continue.
```

### config.json 
```
{
    "websocket_server": false,
    "heartbeat_threshold": 10,
    "enable_social": false,
    "check_niantic_api": true,
    "live_config_update": {
      "enabled": false,
      "tasks_only": false
    },
    "tasks": [
      {
        "type": "TelegramTask",
        "config": {
          "enabled": false,
          "master": null,
          "password": null,
          "// old syntax, still supported: alert_catch": ["all"],
          "// new syntax:": {},
          "alert_catch": {
            "all": {"operator": "and", "cp": 1300, "iv": 0.95},
            "Snorlax": {"operator": "or", "cp": 900, "iv": 0.9}
          }
        }
      },
      {
        "type": "DiscordTask",
        "config": {
          "enabled": false,
          "master": null,
          "// old syntax, still supported: alert_catch": ["all"],
          "// new syntax:": {},
          "alert_catch": {
            "all": {"operator": "and", "cp": 1300, "iv": 0.95},
            "Snorlax": {"operator": "or", "cp": 900, "iv": 0.9}
          }
        }
      },
      {
        "//NOTE: This task MUST be placed on the top of task list": {},
        "type": "RandomAlivePause",
        "config": {
          "enabled": false,
          "min_duration": "00:00:10",
          "max_duration": "00:10:00",
          "min_interval": "00:05:00",
          "max_interval": "01:30:00"
        }
      },
      {
        "type": "HandleSoftBan"
      },
      {
        "type": "RandomPause",
        "config": {
          "enabled": false,
          "min_duration": "00:00:10",
          "max_duration": "00:10:00",
          "min_interval": "00:10:00",
          "max_interval": "02:00:00"
        }
      },
      {
        "type": "CompleteTutorial",
        "config": {
          "enabled": false,
          "// set a name": "",
          "nickname": "",
          "// 0 = No Team, 1 = Blue, 2 = Red, 3 = Yellow": "",
[...]

PR #5849, opened 2016-12-16
issue #5845

knightcraw:
## Short Description:
User receives the following error on new installs, while wbesocket-client is already at latest version.
'An incompatible websocket library is conflicting with the one we need.
You can remove the incompatible library and install the correct one
by running the following commands:

yes | pip uninstall websocket websocket-client
pip install -U websocket-client'

The error is from socketIO_client which require websocket-client to be 0.37.0


BreezeRo:
Approved

PR #5871, opened 2017-01-13
Disable hunting when no more balls

davidakachaos:
## Short Description:
When the bot isn't catching Pokemon because a lack of balls, don't hunt for Pokemon.

- Nothing, small enhancement


MerlionRock:
I thought there's already such function?

in config, add

`      {
        "type": "CatchLimiter",
        "config": {
          "enabled": true,
          "min_balls": 30,
          "duration": 15
        }
      },`

davidakachaos:
Yes this function works great :smile: But the Pokemon Hunter will still move toward a Pokemon of interest and starts searching for it, while we won't catch it (because the CatchLimiter is active)

This PR makes the Pokemon Hunter stop searching for or moving toward a Pokemon when we don't catch Pokemon. Thus the bot will keep moving toward Pokestops instead of needlessly searching for Pokemon.

I hope this makes this PR a bit clearer

Jcolomar:
Asap devs log they merge all changes

solderzzc:
👍

PR #5873, opened 2017-01-14
Don't wait at a lured fort if not catching Pokemon

davidakachaos:
## Short Description:
There is no need to hang around a lured fort if we aren't going to catch any Pokemon.
Also resets the boolean for the Pokemon Hunter message.


davidakachaos:
Work in Process :wink:

davidakachaos:
Work finished. Seems to run well. When on the road for balls, doesn't care about finding Pokemons :-)

solderzzc:
👍

PR #5876, opened 2017-01-15
Get rid of the evolved pokemons after XP evolve

davidakachaos:
## Short Description:
Let's cleanup the Pokemon bag after we evolved trash Pokemons for XP. We will find them again if we need them, we only need their candies anyway :smile:


solderzzc:
👍

PR #5877, opened 2017-01-16
Feature Enhancement: Snip request through telegram

MerlionRock:
Re-Created due to conflicts

Added new feature:
Snipping through telegram, enable it in config, Sniper Task "mode": "telegram"

Also fixed the problem of not able to find Pokemons without EncounterID and SpawnPointID.

Fixes#5863

### DIFF
--- a/pokemongo_bot/cell_workers/sniper.py
+++ b/pokemongo_bot/cell_workers/sniper.py
@@ -449,7 +449,6 @@ class Sniper(BaseTask):
         else:
             self.bot.sniper_disabled_global_warning = False
             targets = []
-            target = {}
 
             # Retrieve the targets
             if self.mode == SniperMode.SOCIAL:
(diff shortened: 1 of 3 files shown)


pokemongo_bot/cell_workers/sniper.py
```diff
@@ -406,12 +449,15 @@ def work(self):
         else:
             self.bot.sniper_disabled_global_warning = False
             targets = []
+            target = {}
```
solderzzc: Seems nowhere to use 'target' in the following code.

MerlionRock: Noted. Will delete it off. I changed this files too many times and lost track what I did.

solderzzc: Got it.

MerlionRock: Done :)

solderzzc:
👍

PR #5879, opened 2017-01-18
Fix step walker and move to fort issue by Anakin5

MerlionRock:
Originally fixed by @Anakin5 

Merged with latest commits, Monkey Patching problem still exist. But would want to see these issues fixed first and hopefully @Anakin5 will come back...

Original Comments:

- Fix step walker timing issue. This solve some problems where we were moving more than the configured speed from time to time.
- Fix move to fort task infinite loop with Polyline. Because MoveToFort was always taking the nearest fort, and because we do not always choose the straight line to move (Polyline), you might end up in a scenario where you infinitely switch between 2 forts.
- Improve a bit the computation speed of step_walker get_next_step
- pep8 fixes

### DIFF
--- a/pokemongo_bot/cell_workers/move_to_fort.py
+++ b/pokemongo_bot/cell_workers/move_to_fort.py
@@ -2,6 +2,7 @@
 from __future__ import unicode_literals
 from __future__ import absolute_import
 
+from geopy.distance import great_circle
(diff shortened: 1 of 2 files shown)


MerlionRock:
Found problem with these files. Closing this request

Issue #5883, opened 2017-01-20
Stardust count error on new account


davidakachaos:
### Expected Behavior
No crash, just set it to 0 and continue

### Actual Behavior
```
Traceback (most recent call last):
  File "pokecli.py", line 865, in <module>
    main()
  File "pokecli.py", line 206, in main
    bot.tick()
  File "/home/david/Broncode/personal/PokemonGo-Bot/pokemongo_bot/__init__.py", line 770, in tick
    if worker.work() == WorkerResult.RUNNING:
  File "/home/david/Broncode/personal/PokemonGo-Bot/pokemongo_bot/cell_workers/pokemon_optimizer.py", line 127, in work
    self.open_inventory()
  File "/home/david/Broncode/personal/PokemonGo-Bot/pokemongo_bot/cell_workers/pokemon_optimizer.py", line 294, in open_inventory
    self.ongoing_stardust_count = self.bot.stardust
  File "/home/david/Broncode/personal/PokemonGo-Bot/pokemongo_bot/__init__.py", line 75, in stardust
    return filter(lambda y: y['name'] == 'STARDUST', self._player['currencies'])[0]['amount']
KeyError: u'amount'
```
This happens on a fresh new account.

### Output when issue occurred
Se above

### Steps to Reproduce
Create a new account, run the bot from master.

### Other Information
OS:  Ubuntu
Branch: master 
Git Commit:  ce549716d20b5a512a65cfa9de93e808218cfb9e

PR #5884, opened 2017-01-20
Fixes stardust crash on new account. See #5883

davidakachaos:
This fixes an issue when you have a brand new account (no stardust yet)
and run the bot. This makes sure the key is there before we do something
with it.

###Fixes
See #5883


solderzzc:
👍

PR #5407, opened 2016-09-12
Fix step walker and move to fort issues

Anakin5:
- Fix step walker timing issue. This solve some problems where we were moving more than the configured speed from time to time.
- Fix move to fort task infinite loop with Polyline. Because MoveToFort was always taking the nearest fort, and because we do not always choose the straight line to move (Polyline), you might end up in a scenario where you infinitely switch between 2 forts.
- Improve a bit the computation speed of step_walker get_next_step
- pep8 fixes

Will fix #5384 but potentially other weird behavior as well

PR #5892, opened 2017-01-27
Fix step walker and move to fort issues

MerlionRock:
Changes to Move to Fort 

- Added a simple logic to prevent move to fort from taking shortest distance when a bounce between 2 forts are detected. This happened when using Polyline walker.

Original Changes by @Anakin5 which was not merged #5407 

- Fix step walker timing issue. This solve some problems where we were moving more than the configured speed from time to time.
- Improve a bit the computation speed of step_walker get_next_step
- pep8 fixes

### DIFF
--- a/configs/config.json.example
+++ b/configs/config.json.example
@@ -142,7 +142,8 @@
                     "stardust_earned",
                     "xp_earned",
                     "xp_per_hour",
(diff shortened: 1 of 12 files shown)


solderzzc:
👍

PR #5896, opened 2017-01-31
Fix an error in getting award km

davidakachaos:
## Short Description:
I noticed a crash when a buddy was already set, the start KM would be
zero, crashing this logic. Simple fix; check if present and set to 0
when not found. Seems to do the trick.

- A buddy crash I noticed :smile:


solderzzc:
👍

PR #5901, opened 2017-02-06
Captcha, POGOProtos and PogoAPI Check

MerlionRock:
Added a simple captcha check. Manual captcha solving is planned. WILL NOT add 2captcha service.

Also checking the version of POGOProtos to alert user of a possible PogoAPI changes. Can't think of a better solution as I can't find anyway to get versioning from PogoAPI.

Changes to PogoAPI check using float instead of matching against strings. In the event that PogoAPI is ahead of official Niantic API, the bot still can run.

### DIFF
--- a/pokemongo_bot/__init__.py
+++ b/pokemongo_bot/__init__.py
@@ -204,6 +204,8 @@ class PokemonGoBot(object):
         )
         self.event_manager.register_event('api_error')
         self.event_manager.register_event('config_error')
+        
+        self.event_manager.register_event('captcha')
 
         self.event_manager.register_event('login_started')
         self.event_manager.register_event('login_failed')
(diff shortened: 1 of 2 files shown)


solderzzc:
👍

snooter:
Lack of 2Captcha is really disappointing.

MerlionRock:
@snooter The original contributors here are not keen and they don't really approve 3rd party  services. Unless they change their stand, this will not change.

However, if someone else is interested, they can always code them in and submit as a PR and it is up to the reviewers to approve.

solderzzc:
@MerlionRock  It's welcome to have captcha service integration since it always depends on user to enable or not.

MerlionRock:
@solderzzc haha~ If you say so.

I'll find some time to look at it. I'm just treating this as a hobby. Not coding for at least 15 years, just don't want to loose a skill that I once good at. LOL

Issue #5922, opened 2017-02-19
New evolution - changes required


pogarek:
Together with 0.57 we have new Pokemons released. This includes movesets and new evolutions.
Now it is possible, that some pokemons can evolve to another (previously not available) Pokemons.
Some of these evolution, requires a new item(s) : stones.

With 0.57 these new items has been added to the game:
- 1101: Sun Stone
- 1102: Kings Rock
- 1103: Metal Coat
- 1104: Dragon Scale
- 1105: Upgrade

Currently bot knows about these new items, but it not aware, about evolution with them. So we have a challenges:

1. Decide (by config? ) which evolution to take, if candies/stones requirements are met : evolve to Pokemon which doesn't require stone or to this one, which requires stone/rock ?
2. Check if requirements are met. This can be done by hardcoding stones/rock in evolution logic, or by changing pokemon.json file (by moving NextEvolutionRequirements to Next Evolotuions/Pokemon )
3. Changing pokemon.json will have a bigger impact on code change

We are open to suggestions and PRs around that. :-)

MerlionRock:
Actually, there's a 3rd option easier option for is coders....

Can be used as a temporary measure until a preferred method has been chosen.

At login. If bot detected any of the evolution stones, just check if user has any of the Pokemons that can use it and throw a message: "you have (stone name), your (Pokemon name) can evolve to (new evolved name). Please login to game if you want to do so."

Just a suggestion before a more effective way is implemented

pogarek:
might be. In this week I'll prepare the new pokemon.json, with new requirements within next evolution. So we could start looking at code.

pogarek:
Adding json file with evolution requirements inside. Let's assume that structure of this JSON will be the one used going forward - when new evolutions will be done in bot
[pokemon.json.txt](https://github.com/PokemonGoF/PokemonGo-Bot/files/791660/pokemon.json.txt)

pogarek:
@davidakachaos : thanks for your PR about evolutions. I think that file as above would be more readable and more flexible for future changes in this area. Can you try to change evolution according to the structure ?

MerlionRock:
@pogarek Looking at your new changes, using Onix as an example, there is no information on  "Next evolution(s)", game master doesn't have it? since Onix will evolve to Steelix, where will that information be?

pogarek:
[...]

PR #5927, opened 2017-02-21
First try at the new evolution items for Optimizer

davidakachaos:
This will check if the needed evolution items are present for a Pokemon.
At first will add only here and there, more a proof of concept I think?

@MerlionRock , @pogarek is this a good way to start integrating the new evolution items?


solderzzc:
👍

pogarek:
looks nice. Let's continue discussion on #5922

PR #5929, opened 2017-02-21
First try at the new evo items (#5927)

solderzzc:


pogarek:
looks nice. Let's continue discussion on #5922

PR #5936, opened 2017-03-02
Version Checking & Minor Fixes

MerlionRock:
Version checking is now fully automatic between PGoAPI and Niantic's API, using hashing endpoint provided by bossland

The check will pass even if PGoAPI is ahead of Niantic's force API

Minor Fixes:
- Corrected error throw when facing captcha during login
- Removed version checking against POGOProtos


solderzzc:
👍

solderzzc:
The Github color changed. Nicer :)

Issue #5949, opened 2017-03-05
Acid missing as Koffling's fast attack in pokemon.json


xiatianyu:
### Description of the issue
As stated in the title, Acid is missing as one of Koffling's fast attacks in [pokemon.json](https://github.com/PokemonGoF/PokemonGo-Bot/blob/720c90a3962783e871896618e2353995cf6f94e9/data/pokemon.json). Although Acid has become a legacy move now for Koffling, it should NOT be removed from the JSON file. 

### Relevant section in pokemon.json
```
{
    "Number": "109",
    "Name": "Koffing",
    "Classification": "DataNotInGameMaster",
    "Type I": [
      "Poison"
    ],
    "Weaknesses": [
      "DataNotInGameMaster"
    ],
    "Fast Attack(s)": [
      "Tackle",
      "Infestation"
    ],
    "Weight": "1 kg",
    "Height": "0,6 m",
    "Candy": {
      "Name": "Koffing candies",
      "FamilyID": 109
    },
    "Next Evolution Requirements": {
      "Amount": 50,
      "Family": 109,
      "Name": "Koffing candies"
    },
    "Next evolution(s)": [
      {
        "Number": "110",
        "Name": "Weezing"
      }
    ],
    "Special Attack(s)": [
      "Sludge",
      "Sludge Bomb",
      "Dark Pulse"
    ],
    "BaseAttack": 119,
    "BaseDefense": 164,
    "BaseStamina": 80,
    "CaptureRate": 0.5,
    "FleeRate": 0.1,
    "BuddyDistanceNeeded": 3
  }
```

### Output when issue occurred
```
[2017-03-05 08:09:23] [MoveToFort] [INFO] Moving towards pokestop Coral Gazebo - 0.14km
[2017-03-05 08:09:25] [   Pokemon] [ERROR] Unexpected moveset [Acid, Dark Pulse] for #109 Koffing, please update info in pokemon.json and create issue/PR
[2017-03-05 08:09:25] [   Pokemon] [ERROR] Unexpected moveset [Acid, Dark Pulse] for #109 Koffing, please update info in pokemon.json and create issue/PR
[2017-03-05 08:09:28] [MoveToFort] [INFO] Moving towards pokestop Coral Gazebo - 0.12km
```

### Other Information
OS: 
Ubuntu 16.04.1 LTS 64-bit
Branch: 
dev
Git Commit: 
720c90a3962783e871896618e2353995cf6f94e9
[...]

PR #5950, opened 2017-03-05
Update pokemon.json

xiatianyu:
## Short Description:
Add missing fast attack (Acid) for Koffing
## Fixes/Resolves/Closes:
-#5949

### DIFF
--- a/data/pokemon.json
+++ b/data/pokemon.json
@@ -4897,6 +4897,7 @@
       "DataNotInGameMaster"
     ],
     "Fast Attack(s)": [
+      "Acid",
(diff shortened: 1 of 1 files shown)


MerlionRock:
👍
@pogarek You want to take a look at your tool how it miss out that attack? it's not in game master?

pogarek:
legacy attack are removed from game master. Currently my tool appends legacy moves from previous pokemon.json . In that file Acid was missing. I've added it now to the old one. 
Any pokemon.json can be used as legacy pokemon.json.

PR #5954, opened 2017-03-06
Eliminate error caused by typo

xiatianyu:
Currently when "pinapberries" is specified in the item list under `UpdateLiveInventory` in `config.json`, the following exception is raised:
```
File "./pokemongo_bot/cell_workers/update_live_inventory.py", line 200, in get_item
    raise ConfigException("item '{}' isn't available for displaying".format(item))
pokemongo_bot.tree_config_builder.ConfigException: item 'pinapberries' isn't available for displaying
```

It turns out "pinapberries" is misspelled as "pinabberries"; this update serves to eliminate the typo.


pogarek:
👍

pogarek:
merged #5954

PR #5957, opened 2017-03-07
Random eggs in infinite incubator

davidakachaos:
Since it is the unbreakable incubater, we might just want to put a random egg in there.
This can be viewed as more humanlike behaviour


pogarek:
👍

pogarek:
merged #5957

PR #5962, opened 2017-03-14
Sniper fixes

davidakachaos:
## Short Description:
Fixes for these issues I noticed:
- Fixes an issue with [a source](http://pokedex100.com/tie9/data/) hiding latitude and longitude of Pokemons when using it without a subscription.
- Also converts IV to float instead of an integer. Some sites use IVs like 96.7
- Adds a cooldown to sniping, to prevent Niantic from seeing you [jumping all over the world](https://www.youtube.com/watch?v=ZrPaEUjtYvQ) :smile:
- Fixes an logic issue, where sniper would snipe Pokemon below the special IV instead of that and above.

selfie-0
[_GitHub Selfies_](https://github.com/thieman/github-selfies/)

### DIFF
--- a/configs/config.json.example
+++ b/configs/config.json.example
@@ -291,6 +291,8 @@
                 "mode": "social",
                 "bullets": 1,
                 "homing_shots": true,
+                "cooldown_enabled": false,
+                "loiter_after_snipe": false,
                 "special_iv": 100,
                 "order": [
                     "missing",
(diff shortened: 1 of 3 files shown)


MerlionRock:
Wondering how does cool down prevent Niantic not seeing you jumping around the world.... Lol....

davidakachaos:
It was an idea I had; after sniping Niantic might "flag" an account to be more closely monitored. Nothing based on proof, just feeling lke something I would do when I was combatting bots...
And no, it wouldn't prevent it, but it does screw with a detecting strat to see if some accounts are sniping...
I would be happy to remove the cooldown if you think it has no use at all 😉

pogarek:
I think it would be good to expose cooldown to config.json , with default behavior like it was before the PR. :) Plus, of course, a proper modification to sample config and docs :)

davidakachaos:
@pogarek Done 😉

pogarek:
👍 
let's test it out on dev :) However I'll stick with my current config. Works for a last few months :-P

pogarek:
merged #5962

Issue #5942, opened 2017-03-03
TODO: New berries support


pogarek:
Currently bot seems to use only Razz Berries. 
It would be good to enhance the functionality to use new berries, with controlled behavior (by config.json). 

Help appreciated.

MeisterLLD:
I tried a very brutal idea.
In catch_pokemon_worker.py, replaced the 

> ITEM_RAZZBERRY = 701

by 

> ITEM_RAZZBERRY = 705

to try and have the bot us Pinap as if it were using Razz. Encountered a VIP mon and it tried to throw it but answered 

> Unknown response when throwing berry

I then opened the real game myself and indeed, only 3 candies were gotten instead of 6. So I guess the bot didn't really succeed at throwing the Pinap.
So the changes will probably be more subtle than that :)

MerlionRock:
I'm working on it right now.

MerlionRock:
Closing it. Merged :)

pogarek:
:-)

PR #5963, opened 2017-03-15
Changes to using berries

MerlionRock:
Changes from using old API call to new API call

- Added Pinap Berry Support #5942
- Pinap berry will be used instead of Razz Berry if use_pinap_on_vip is set to true (Under ctach_pokemon task)
- Razz berry will still be used when running out of Pinap Berries
- Amount of candies gained will be reported as it is no longer fixed at 3
- Updated documenation
- Updated config examples

### DIFF
--- a/configs/config.json.example
+++ b/configs/config.json.example
@@ -291,6 +291,8 @@
                 "mode": "social",
                 "bullets": 1,
                 "homing_shots": true,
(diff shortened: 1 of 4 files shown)


pogarek:
👍

pogarek:
merged #5963

Issue #5986, opened 2017-03-30
[Feature Request] pinap on low lvl mons


WayneUong:
Can we have `pinap_on_level_below` or `pinap_on_cp_below` options?

MerlionRock:
My hash key expired.

Once I have a vaild hash key, I'll work on it. That's not hard.

MerlionRock:
This option is for use_pinap_on_vip right?

Doesn't make sense if anyone want to use on any other pokemon that's not VIP....

davidakachaos:
I disagree Merlion, Maybe someone wants to use a pinap on a Pokemon when they can almost evolve that Pokemon in a missing or better combat Pokemon.

I vote for a ```pinap_on_candy_needed``` kind of config.

MerlionRock:
I've added 3 new configuration, they are: 

```
"pinap_on_level_below": 0,
"pinap_operator": "or",
"pinap_ignore_threshold": false,
```

set "pinap_on_level_below" to 0 if you don't want to use it at all
"pinap_operator" -> or | and to be used together with "use_pinap_on_vip"
"pinap_ignore_threshold" to ignore threshold if set to true

Any more request before I sumbit a PR?

MerlionRock:
PR added #5990

MerlionRock:
Merged therefore Close

PR #5990, opened 2017-04-01
Update Pinap Uses

MerlionRock:
Added the following:
"pinap_on_level_below" - Set at what level (and below) of the pokemon should Pinap berry br use on
"pinap_operator" - Set if Pinap berry going to be use together with "use_pinap_on_vip" or without (Operator "or" | "and")
"pinap_ignore_threshold": - Set if bot is going to ignore catch rate threshold when using pinap berry

#5986


Jcolomar:
Let's test

Issue #6004, opened 2017-04-10
How to stop transfering of your good pokemon ? Can somebody show me his code ?


Yasheshpandita:

Jcolomar:
Optimizer fix #6006

Issue #5999, opened 2017-04-05
Get rid of the evolved pokemons after XP evolve (#5876)


Jcolomar:
This has issues. The bot release all pokemons evolved in same round, trash and good ones:

[2017-04-05 21:07:19] [PokemonOptimizer] [INFO] Evolving 3 Pokemon (the best)
[2017-04-05 21:07:19] [PokemonOptimizer] [INFO] Evolved Geodude [IV 1.0] [CP 682] [2205 candies] [+500 xp]
[2017-04-05 21:07:45] [PokemonOptimizer] [INFO] Evolved Graveler [IV 1.0] [CP 1037] [2106 candies] [+500 xp]
[2017-04-05 21:08:11] [PokemonOptimizer] [INFO] Evolved Magnemite [IV 1.0] [CP 929] [9804 candies] [+500 xp]
[2017-04-05 21:08:38] [PokemonOptimizer] [INFO] Transferring 2 Pokemon
[2017-04-05 21:08:41] [PokemonOptimizer] [INFO] Exchanged Magneton [IV 1.0] [CP 1917] [9805 candies]
[2017-04-05 21:08:41] [PokemonOptimizer] [INFO] Exchanged Golem [IV 1.0] [CP 1666] [2107 candies]

MerlionRock:
@Jcolomar 
I just went through the codings.

Can't find anything wrong with it. It simply just run the checking 2 times. Post your Optimizer config again, I try to see what go wrong.

Jcolomar:
"rules": [
                    {
                        "mode": "overall",
                        "top": 1,
                        "sort": [
                            "-candy",
                            "iv"
                        ],
                        "evolve": false,
                        "buddy": true
                    },
                    {
                        "mode": "by_pokemon",
                        "names": [
                            "gym_attack"
                        ],
                        "top": 2,
                        "sort": [
                            "moveset.attack_perfection",
                            "dps_attack",
                            "npc"
                        ]
                    },
                    {
                        "mode": "by_pokemon",
                        "names": [
                            "gym_def"
                        ],
                        "top": 3,
                        "sort": [
                            "moveset.defense_perfection",
                            "dps_defense",
                            "npc"
                        ]
                    },
                    {
                        "mode": "by_family",
                        "top": 1,
                        "sort": [
                            "moveset.attack_perfection",
                            "iv",
                            "dps_attack",
                            "npc"
                        ],
                        "evolve": {
                            "iv": 1.0
[...]

PR #6006, opened 2017-04-10
Optimizer fix

davidakachaos:
## Short Description:
The getting rid of Pokemons after evolving trash seems to be bugged; it transfers Pokemon who just got evolved. This moves the 2 times loop up, so the loop will re-open the iventory (with the new pokemons) and it will re-evalutate the rules with the reloaded inventory.
I hope with this fix the transfer of good Pokemon will stop

- #5999 
- #6004 (no description given, I think this is related aswel? If not, please say so!)


Jcolomar:
Merged #6006

pogarek:
merging #6006 . Let's on dev.

pogarek:
@davidakachaos please do new PR to dev branch . We don't do PR to master.

PR #6007, opened 2017-04-10
[CatchLimiter] Resume at X balls

davidakachaos:
## Short Description:
CatchLimiter keeps us from running out of balls. But when botting in a PokeStop littered area (NYC) the number of balls can get high very fast.

This change lets you set a "resume catching when X balls reached" setting. When there are at least X (default 100) balls in the inventory, resume catching.


pogarek:
👍

pogarek:
merged #6007

PR #6008, opened 2017-04-10
Stop hunting Ivysaurs when we need a Bulbasaur

davidakachaos:
## Short Description:
Right now the Pokemon hunter will chase a Ivysaur when missing a Bulbasaur. This change will fix this behavior.


pogarek:
👍

pogarek:
merged #6008

PR #6012, opened 2017-04-11
Optimizer powerup fix

davidakachaos:
I noticed my Pokemons werent upgraded at all.
After look into this, I saw the stardust calculation subtracted the would-be used
dust always! But when we determin the Pokemon cant be upgraded, we didnt use the dust
This will "refund" this dust for the calculations and determin if the following mon
can be upgraded


pogarek:
👍

pogarek:
merged #6012

PR #6016, opened 2017-04-13
Fix Markdown in documentation

per1234:
The GitHub Markdown interpreter has recently been changed to strictly enforce the [GFM spec](https://github.github.com/gfm/) which has caused some Markdown to no longer display as originally intended.


solderzzc:
👍

Issue #6021, opened 2017-04-19
[Feature Request] Let's optimizer ignore Favorites


pogarek:
Currently optimizer upgrades and evolves favorites pokemons. It is only ignoring them for transfer.
It would be good to expand this behavior, so it could be added to groups , like touch_favorites: false.

Jcolomar:
Id like also set favorites by rule, and after compare them if they are same id or family id and keep bests one

pogarek:
or worst ones even (like for bubblestart)

Jcolomar:
yep

PR #6027, opened 2017-04-23
Added ability for Optimizer to handle favorites

davidakachaos:
This will enable the Pokemon Optimizer to mark Pokemon as favorite
accoring to rules set. Just like evolve and upgrade rules.
Also has the ability to unmark Pokemon as favorite if enabled - use with care!
Added options to ignore favorite Pokemon for evolve and/or upgrade rules

This was requested in #6021

### DIFF
--- a/configs/config.json.optimizer.example
+++ b/configs/config.json.optimizer.example
@@ -18,6 +18,9 @@
                 "evolve_only_with_lucky_egg": false,
                 "evolve_count_for_lucky_egg": 80,
                 "may_use_lucky_egg": true,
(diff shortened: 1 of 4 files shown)


pogarek:
Let me know when is done 😊😊😊
And thanks !!

davidakachaos:
@pogarek You're welcome 😉 
I think this is now ready for "battle testing" on dev. I haven't encountered crashes anymore, and I added action_delay to the favorite functions to emulate human behavior... 
So give it a go and see if this helps you out.

PR #6034, opened 2017-05-06
New Follow Path Option

MerlionRock:
Follow Path: Added option to disable location update infomation in console


Jcolomar:
Merged #6034

Jcolomar:
Merged #6034

MerlionRock:
I think this one has bug... I'll most likely use a new event instead of using existing one....

For now if anyone's bot crashes after disable follow path's location update, just enable it back.

PR #6049, opened 2017-05-26
Deploying of lures if available

MerlionRock:
## Short Description:
Deploying of lures is now possible!
Add "use_lure": true to "SpinFort" task to enable it!


Jcolomar:
Merged #6049

PR #6052, opened 2017-05-31
Pokemon Hunter - Mobster and Trash Hunter

davidakachaos:
In this version of the Pokemon Hunter, we can select if we want to hunt
down family members of VIPs (Like a mobster) and if we want to hunt for
certain "trash" Pokemon.

If a Pokemon is found to be hunted down, now the Hunter also checks
if the Pokemon is near a Pokestop. If so, it will start hunting there first.

Mobster
When setting "target_family_of_vip"to true, the Hunter will also mark
family of the VIPs you set as a valid target.
If you have "lock_on" enabled, and want to be sure you catch every last
family member of the VIPs, you can enable "treat_family_of_vip_as_vip".
This effectively means that the whole family of a VIP needs to be hunted
down as VIP Pokemons.

Trash Hunting
Trash hunting will trigger when a set amount of space is left in the
bag. If triggered, the Hunter will start hunting down Caterpies, Weedles
and Pidgeys. These only take 12 candy to evolve, so are useful for mass
evolving.

If a VIP Pokemon is detected nearby while hunting for trash, the Hunter
will change the target to the VIP (and locking onto it if enabled)

### DIFF
--- a/docs/configuration_files.md
+++ b/docs/configuration_files.md
@@ -1408,13 +1408,23 @@ Hunts down nearby Pokemon. Searches for Pokemon to complete the Pokedex, or if a
 [[back to top](#table-of-contents)]
 
 * `max_distance`: `Default: 2000`. Maxium of meters for the "nearby" part.
+* `enable_cooldown`: `Default: true`. After a hunt (succesful or not) have a cool down (stops hunting for a bit)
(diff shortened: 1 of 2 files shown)


davidakachaos:
Lol 😄 Thanks! I'm trying to keep this bot alive, so far no shadowban for me (stopped sniping immediately when they started softbanning snipers)

PR #6071, opened 2017-07-09
Inbox call added to heartbeat

davidakachaos:
This will return the notifications send by the server and display them if its a new notification.

It seems to only be related to gyms, but useful none the less.


MerlionRock:
👍

PR #6072, opened 2017-07-09
BadPokemon worker / Red Slashed Pokemon

davidakachaos:
If you have any Pokemon that Niantic has marked as bad (red slashes) this will notify you. If you set the option, it will also transfer those Pokemon.
These Pokemon are unusable for Gym and can't be powered up. These need to be transfered off your account.

### DIFF
--- a/docs/configuration_files.md
+++ b/docs/configuration_files.md
@@ -62,6 +62,7 @@
 - [CompleteTutorial](#completetutorial)
 - [BuddyPokemon](#buddypokemon)
 - [PokemonHunter](#pokemonhunter)
+- [BadPokemon](#badpokemon)
 
 # Configuration files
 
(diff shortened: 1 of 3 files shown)


MerlionRock:
I got the following when try to run the new changes:

  File "pokecli.py", line 885, in <module>
    main()
  File "pokecli.py", line 206, in main
    bot.tick()
  File "/Users/Thomas/PokemonGo-Bot/pokemongo_bot/__init__.py", line 830, in tick
  File "/Users/Thomas/PokemonGo-Bot/pokemongo_bot/cell_workers/bad_pokemon.py", line 25, in work
    bad_pokemons = [p for p in inventory.pokemons().all() if p.is_bad]
AttributeError: 'Pokemon' object has no attribute 'is_bad'

MerlionRock:
Hmm.... I think you need to add
`self.is_bad = data.get('is_bad', False)`
into inventory.py right?

davidakachaos:
Oops, you're right. Will fix this asap

Issue #6075, opened 2017-07-10
0.67.2


camnomis:
I know bossland/PokeFarmer has had all sorts of problems with capacity/authentication etc so haven't been focusing on the new API, I read a comment saying that even through the API has progressed form 0.67.1 the old endpoint can still be used, does anyone know if this is correct? 

No longer use the bot for catching, just sorting/mass evolving and wanted to know how "safe" it is

thelemonenergy:
Hi, it will for sure work, 
but it is not safe. Since the app was forced to Update niantic could see, that everybody who is still playing on 0.67.1 is not using the app.

You might propably get flagged!

davidakachaos:
Afaik the [pgoapi](https://github.com/pogodevorg/pgoapi) got bumped to 0.67.2 (not everything in the API yet) so I think if you use that latest version (setup.sh -u) the bot will identify itself as using the 0.67.2 API.
The hashing service (bossland) still identifies as 0.67.1 but the hashing hasn't changed between 0.67.1 and 0.67.2 (again, as far as I know)
That beinmg said, botting is never totally safe, Niantic could detect invalid calls from any app that is not the offical app. Always you at your own risk! For what its worth, I'm currently running 2 accounts without problems.

goedzo:
> Afaik the pgoapi got bumped to 0.67.2 (not everything in the API yet) so I think if you use that latest version (setup.sh -u)

Thanks for the info. I run the container version of Pika, so how can I update this command in my container? I now have build the new dev branche using >docker build --build-arg BUILD_BRANCH=dev -t pokemongo-bot .

But running this new branche gives me: 
>[2017-07-10 18:01:06] [PokemonGoBot] [INFO] Niantic Official API Version: 0.67.2
>[2017-07-10 18:01:06] [PokemonGoBot] [INFO] Latest Bossland Hashing API Version: 0.67.1

davidakachaos:
Yes, the Bossland Hashing API is still 0.67.1, but as far as I can see it works fine with 0.67.2
You can get the bot to work (don't do this with a main account, I do not assume risk for your account(s)) by setting in config.json
```json
"check_niantic_api": false
```
So the bot will run.

davidakachaos:
Also see #6070

MerlionRock:
Dear All, 

Pogo API now has a method that return latest hashing or API they are using now. So we no longer need to compare to Bossland's end point to know what API we are on.

[...]

PR #6077, opened 2017-07-11
check_niantic_api logic change

MerlionRock:
Now check api uses version number returned from PGoAPI rather than comparing to Bossland endpoint.

### DIFF
--- a/pokemongo_bot/__init__.py
+++ b/pokemongo_bot/__init__.py
@@ -1068,20 +1068,14 @@ class PokemonGoBot(object):
             formatted="Niantic Official API Version: {}".format(officalAPI)
         )
 
-        link = "https://pokehash.buddyauth.com/api/hash/versions"
-        f = urllib2.urlopen(link)
-        myfile = f.read()
-        f.close()
-        bossland_hash_endpoint = myfile.split(",")
-        total_entry = int(len(bossland_hash_endpoint))
-        last_bossland_entry = bossland_hash_endpoint[total_entry-1]
-        bossland_lastestAPI = last_bossland_entry.split(":")[0].replace('\"','')
-        hashingAPI_temp = 0
+        PGoAPI_version = PGoApi.get_api_version()
+        PGoAPI_version_str = str(PGoAPI_version)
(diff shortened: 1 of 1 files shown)


goedzo:
This commit does not work and causes the following error when building the container:
>AttributeError: class PGoApi has no attribute 'get_api_version'

Seems that the PGoApi is not updated in Dev branch yet.

MerlionRock:
@goedzo how do you usually do update? Is it possible to rebuild the container? Just update to latest Pgoapi. Else you are running older API and risk ban.

nimsnims:
Hi merlion. I use the command from the documeation:
>docker build --build-arg BUILD_BRANCH=dev -t pokemongo-bot .

This does not update the pgoapi, so that should be fixed too. I suspect that pgoapi update is stil in Dev branch, and container just gets release branch. But tbh i dont know how to update pgoapi in a container so please let me know how i can do that.

nimsnims:
Ps. Nimsnims & goedzo are both my github accounts.

MerlionRock:
@Nimsnims let me ask around in discord

MerlionRock:
@nimsnims refer to https://github.com/PokemonGoF/PokemonGo-Bot/issues/6075 and see last reply. See if that helps. If anymore questions, ask in that thread.

goedzo:
For Reference I will also post this here:

It seems that a lot of extra containers are also build in the process, so with above input I was able to find that

>    docker images -q

Gives all the containers. So I removed them all with the

>    docker rmi [image_id]

After this it seems to rebuild everything, including the pgo api.

Within the build process this seems to me the relevant line:

>    Collecting pgoapi from git+https://github.com/pogodevorg/pgoapi.git@develop#egg=pgoapi (from -r requirements.txt (line 4))
>    Cloning https://github.com/pogodevorg/pgoapi.git (to develop) to /tmp/pip-build-Rwwo7j/pgoapi

And now it seems to work correctly! So thank you for the help. I think the documentation on how to properly rebuild the container, should be added :)

This commit is actually perfect!

PR #6080, opened 2017-07-12
Updated Character Information

MerlionRock:
Updated character information at login. Removed items that we no idea when they will ever appear in game, update some name to be the same as in game, added new items.


Jcolomar:
Thx a lot

PR #6083, opened 2017-07-15
Deploy a pokemon in a gym when there is a free slot available

goedzo:
This is very basic functionality that deploys a pokemon in a free gym slot. By default this is disabled unless you add this, directly below "SpinFort" in the config file.

        {
            "type": "GymPokemon",
            "config": {
                "enabled": true,
                "feed_berries": true,
                "order_by": "cp",
                "min_interval":360,
                "min_recheck":30,
                "max_recheck":120
            }
        },

It will only use gyms that are in range and a max of 20 mons is hardcoded.


goedzo:
Will close this one, since I applied some bugfixing for special cases.

PR #6084, opened 2017-07-15
Basic gym support: Deploy a pokemon in a gym when there is a free slot available

goedzo:
This is very basic functionality that deploys a pokemon in a free gym slot. By default this is disabled unless you add this, directly below "SpinFort" in the config file.

        {
            "type": "GymPokemon",
            "config": {
                "enabled": true,
                "feed_berries": true,
                "order_by": "cp",
                "min_interval":360,
                "min_recheck":30,
                "max_recheck":120
            }
        },

It will only use gyms that are in range and a max of 20 mons is hardcoded.


MerlionRock:
I think you need to rebase your fork and do another PR for this.

Anyway, more or less we have come to a conclusion. I am okay to merge this and the forked API to Dev and eventually to Master. @solderzzc  However, I still do not agree with using berries. At most a first heal on deploy ONLY and if enabled. If we allow berries healing, it's unfair to real active users. Imagine someone use the bot to patrol a route, and only give up 8 hrs later, that's a bit too much. Real active users don't do that.

Let us know again if you really object to my thought process. While I agree we can't prevent people from misusing it, we at least, on our part, minimise the chances of people misusing it.

davidakachaos:
Let's rip out the feeding of Pokemon for now. It doesn't work yet anyway, and as such should be removed but also I agree that it can be to easy to abuse.... I mean, you setup a route with a gym or 4 and keep feeding all the Pokemon... I hadn't thought about that yet...
@goedzo can you rip that part out? Or do you need some help with that?

MerlionRock:
@goedzo Okay, follow @davidakachaos instructions. rip out the berries feeding

Issue #6063, opened 2017-06-28
Docker image does not build on new version of windows docker


goedzo:
Just upgraded docker to the newest stable version 17.0.6.0-ce Win 18 (12627).

When I now build a new docker container (dev or master branch doesnt matter) the following error occurs. Log included:
`
docker build --build-arg BUILD_BRANCH=dev -t pokemongo-bot .
Sending build context to Docker daemon  5.494MB
Step 1/13 : FROM alpine
 ---> 665ffb03bfae
Step 2/13 : ARG BUILD_REPO=PokemonGoF/PokemonGo-Bot
 ---> Using cache
 ---> 6caa94da9d81
Step 3/13 : ARG BUILD_BRANCH=master
 ---> Using cache
 ---> e2f4732759f0
Step 4/13 : LABEL build_repo $BUILD_REPO build_branch $BUILD_BRANCH
 ---> Using cache
 ---> a2285a315904
Step 5/13 : WORKDIR /usr/src/app
 ---> Using cache
 ---> 3b5425538398
Step 6/13 : VOLUME /usr/src/app/configs /usr/src/app/web
 ---> Using cache
 ---> d1d8a8f24be9
Step 7/13 : RUN apk -U --no-cache add python py-pip tzdata     && rm -rf /var/cache/apk/*     && find / -name '*.pyc' -o -name '*.pyo' | xargs -rn1 rm -f
 ---> Using cache
 ---> 53da740a6ac3
Step 8/13 : ADD http://pgoapi.com/pgoencrypt.tar.gz /tmp/pgoencrypt.tar.gz
Downloading  243.6kB/243.6kB
 ---> Using cache
 ---> 9471fde1345a
Step 9/13 : ADD https://raw.githubusercontent.com/$BUILD_REPO/$BUILD_BRANCH/requirements.txt .
Downloading     633B
 ---> Using cache
 ---> edc53490f26c
[...]

PR #6085, opened 2017-07-16
Docker fix for #6063

goedzo:
I have found the reason of this error. It seems that it was a problem with the CA certificates on Alpine Linux. It does not succesfully loads http://pgoapi.com/pgoencrypt.tar.gz because it cannot connect to the web domain, because of missing certificates. This pull request fixes the Docker file.


davidakachaos:
And this still works for older versions as well?

goedzo:
Yep! In older versions the certificates were already there but that increases the base footprint, so that is why they removed them in newer versions of Alpine. Also it was seen as a security issue to use pre-loaded certificates so that is why they always should be refreshed when creating a new container.

davidakachaos:
👍

PR #6088, opened 2017-07-17
Full API Support for 0.67.2 based on latest compile from POGOProtos

### LINKED ISSUE #6075: 0.67.2

goedzo:
Recompiled latest https://github.com/AeonLucid/POGOProtos API and added to my own maintained repository.

fixes #6075

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,7 +1,7 @@
 numpy==1.11.0
 networkx==1.11
 six==1.10
-git+https://github.com/pogodevorg/pgoapi.git@develop#egg=pgoapi
+git+https://github.com/goedzo/pgoapi.git@develop#egg=pgoapi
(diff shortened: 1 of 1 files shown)


MerlionRock:
@solderzzc I am okay to merge this so that we will be using a forked API instead of official in order to have full set of API support. Are you okay with it too?

solderzzc:
@MerlionRock I have no reason to refuse if the fork API has extra features and well maintained.

PR #6090, opened 2017-07-17
Fixed all bugs in latest HealPokemon Module

goedzo:
Event handlers were not registered and inventory was missing importand method.


MerlionRock:
👍

Issue #6065, opened 2017-06-29
[help!] Bot sticks in moving towards a same pokestop after about 15 minutes of running


fatalerrortan:
<h3>The problem is after the recent update！In the forward but never to the destination！</h3>

[2017-06-29 13:34:20] [MoveToFort] [INFO] Moving towards pokestop Eingangswächter Schillstraße - 0.13km
[2017-06-29 13:34:25] [MoveToFort] [INFO] Moving towards pokestop Eingangswächter Schillstraße - 0.13km
[2017-06-29 13:34:31] [MoveToFort] [INFO] Moving towards pokestop Eingangswächter Schillstraße - 0.12km
[2017-06-29 13:34:36] [MoveToFort] [INFO] Moving towards pokestop Eingangswächter Schillstraße - 0.13km
[2017-06-29 13:34:49] [MoveToFort] [INFO] Moving towards pokestop Eingangswächter Schillstraße - 0.12km
[2017-06-29 13:34:58] [MoveToFort] [INFO] Moving towards pokestop Eingangswächter Schillstraße - 0.13km
[2017-06-29 13:35:06] [MoveToFort] [INFO] Moving towards pokestop Eingangswächter Schillstraße - 0.12km
[2017-06-29 13:35:14] [MoveToFort] [INFO] Moving towards pokestop Eingangswächter Schillstraße - 0.12km

<h3>config.json</h3>
<pre>
{
    "websocket_server": false,
    "heartbeat_threshold": 10,
    "enable_social": false,
    "check_niantic_api": false,
    "solve_captcha": true,
    "live_config_update": {
        "enabled": false,
        "tasks_only": false
    },
    "tasks": [
        {
            "type": "TelegramTask",
            "config": {
                "enabled": false,
                "master": null,
                "password": null,
                "// old syntax, still supported: alert_catch": [
                    "all"
                ],
                "// new syntax:": {},
                "alert_catch": {
                    "all": {
                        "operator": "and",
                        "cp": 1300,
                        "iv": 0.95
                    },
                    "Snorlax": {
                        "operator": "or",
                        "cp": 900,
                        "iv": 0.9
                    }
                }
            }
        },
        {
            "type": "DiscordTask",
            "config": {
                "enabled": false,
                "master": null,
                "// old syntax, still supported: alert_catch": [
                    "all"
                ],
                "// new syntax:": {},
                "alert_catch": {
                    "all": {
                        "operator": "and",
[...]

PR #6092, opened 2017-07-18
Move to fort and gyms

davidakachaos:
## Short Description:
MoveToFort and SpinFort shoudl also include gyms now because we can spin them as we can Pokestops. This also adds some error handling for times the task gets stuck moving to a stop, or at least tries to detect and handle those cases

## Resolves:
- #6065 hopefully


MerlionRock:
👍

PR #6103, opened 2017-07-21
removed encrypt references from batch files

nbq:
## Short Description:

Cleaned up some text in the batch files and removed confusing old references to encrypt files

~Holoshed on Discord


davidakachaos:
👍

PR #6105, opened 2017-07-22
Cleanup encrypt references in linux, mac and docker setups

nbq:
## Short Description:

- Removes references and compilation of not needed encrypt files from the old API for posix based systems and docker.
- Removed the references of encrypt files from .gitignore.


MerlionRock:
👍

PR #6110, opened 2017-07-23
Notify of shortage once and dont heal max health Pokemon

davidakachaos:
## Short Description:
Only heal when we have the potions and/or revives. Notify of shortage only once!

## Fixes
- Log spamming when out of potions and/or revives
- Trying to heal a Pokemon that is already at full health


MerlionRock:
👍

PR #6118, opened 2017-07-23
Forcing PGO Api upgrade

goedzo:
Because of most people don't upgrade pogoapi when updating bot, this makes sure they actually do. This commit will check for latest version of pogo api.


BreezeRo:
Approved

MerlionRock:
@BreezeRo We need to merge this to master. Else those on master branch will be affected

PR #6122, opened 2017-07-24
Cleanup manual install document and fix windows updater batch file

nbq:
## Short Description:

- Fixes documentation to remove reference to old API
- Added reference image for adding python to path
- Patched windows updater batch file to always have up to date pgoapi


MerlionRock:
👍

Issue #6123, opened 2017-07-24
Dev branch: sudden break when arrived at Gym ...


venney:
Hi,

Not sure what happened.
My bot is "dev branch" that is cloned at 7/24 16:00 (TimeZone: +8)
It just breaks and restart to login ...

[2017-07-24 17:47:02] [GymPokemon] [INFO] Moving towards open Gym 春長鶴松 - 0.06km
[2017-07-24 17:47:05] [GymPokemon] [INFO] Moving towards open Gym 春長鶴松 - 0.06km
[2017-07-24 17:47:07] [GymPokemon] [INFO] Moving towards open Gym 春長鶴松 - 0.06km
[2017-07-24 17:47:10] [GymPokemon] [INFO] Moving towards open Gym 春長鶴松 - 0.05km
[2017-07-24 17:47:13] [GymPokemon] [INFO] Moving towards open Gym 春長鶴松 - 0.05km
[2017-07-24 17:47:15] [GymPokemon] [INFO] Moving towards open Gym 春長鶴松 - 0.05km
[2017-07-24 17:47:17] [GymPokemon] [INFO] Moving towards open Gym 春長鶴松 - 0.04km
[2017-07-24 17:47:21] [GymPokemon] [INFO] Moving towards open Gym 春長鶴松 - 0.04km
[2017-07-24 17:47:23] [GymPokemon] [INFO] Arrived at Gym 春長鶴松.
[2017-07-24 17:47:23] [       cli] [INFO] 
[2017-07-24 17:47:23] [       cli] [INFO] Ran for 0:01:25
[2017-07-24 17:47:23] [       cli] [INFO] Total XP Earned: 0  Average: 0.00/h
[2017-07-24 17:47:23] [       cli] [INFO] Travelled 0.07km
[2017-07-24 17:47:23] [       cli] [INFO] Visited 0 stops
[2017-07-24 17:47:23] [       cli] [INFO] Encountered 0 pokemon, 0 caught, 0 released, 0 evolved, 0 never seen before ()
[2017-07-24 17:47:23] [       cli] [INFO] Threw 0 pokeballs
[2017-07-24 17:47:23] [       cli] [INFO] Earned 0 Stardust
[2017-07-24 17:47:23] [       cli] [INFO] Hatched eggs 0
[2017-07-24 17:47:23] [       cli] [INFO] Next egg hatches in 0.45 km
[2017-07-24 17:47:23] [       cli] [INFO] 
[...]

PR #6125, opened 2017-07-24
Fixing #6123

goedzo:
When gym names have unicode names, bot crashes when using GymPokemon task. This fix makes the eventmanager more robust.

### DIFF
--- a/pokemongo_bot/event_manager.py
+++ b/pokemongo_bot/event_manager.py
@@ -52,8 +52,11 @@ class Event(object):
             self.sender = str(sender).encode('ascii', 'xmlcharrefreplace')
         
         self.level = str(level).encode('ascii', 'xmlcharrefreplace')
-        #self.formatted = str(formatted).encode('ascii', 'xmlcharrefreplace')
-        self.formatted = str(formatted).encode('utf-8', 'xmlcharrefreplace')
+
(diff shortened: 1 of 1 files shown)


MerlionRock:
👍

ChiefM3:
I'm afraid when arriving at gyms with unicode names, the bot still crashes

I changed this

>     self.formatted = str(formatted).encode('utf-8', 'ignore').decode('utf-8')
>     self.formatted = str(formatted).encode('ascii', 'xmlcharrefreplace')

into this

>         try:
>             self.formatted = str(formatted).encode('utf-8', 'ignore').decode('utf-8')
>         except UnicodeEncodeError:
>             self.formatted = str(unicode(formatted).encode('unicode_escape'))

After this change, no more crashing. Does this change make sense?

goedzo:
@ChiefM3  thanks for feedback. I understand issue now. Will fix the fix now :). Solution is almost similair like yours but a little different.

        #Fixing issue 6123 for gym names that are in utf format.
        try:
            self.formatted = str(formatted).encode('ascii', 'xmlcharrefreplace')
        except UnicodeEncodeError:
            self.formatted = str(unicode(formatted).encode('unicode_escape'))

Pull requesting now.

PR #6128, opened 2017-07-25
Updated Catch Limiter

MerlionRock:
Updated Catch Limiter to take "min_ultraball_to_keep" into calculation
Updated Gym Pokemon to be able to be disabled by Catch Limiter


davidakachaos:
👍

PR #6130, opened 2017-07-25
Setup.sh install now configures the map

nbq:
Added web tracker config to setup.sh
Added -w and --web to setup.sh for web config


nbq:
@MerlionRock Please close this one so I can submit it right, I picked master by mistake and then switched it which seems to have broken the compile bot.

MerlionRock:
As requested

PR #6136, opened 2017-07-25
Fix now correct for #6123

goedzo:
My first fix was not correct. @ChiefM3 pointed me to the right solution.


ChiefM3:
Glad I helped! Thank you for the correct fix!

PR #6142, opened 2017-07-26
Improved handling of Daily Catch Limit

davidakachaos:
This adds a new way to handle the daily catch limit. It will disable the
catch tasks just like we ran out of balls. It will disable all catch
tasks for an hour, after that it will check if the amount of catches is
lower than the limit and resume catching if so.
During the hour we still check if the amount of Pokemon caught is lower
than the daily limit minus 20, if so the bot will resume catching again,
ending the remaining wait time.
Notice if we enabled "exit_on_limit_reached", the bot will still exit if
the daily limit is reached!!


MerlionRock:
👍

PR #6143, opened 2017-07-26
Lowered the default catch limit to 500

davidakachaos:
In 72 hours the limit is 1500. Setting a high daily limit might cause problems. Added a warning about that.


MerlionRock:
👍

PR #6146, opened 2017-07-26
Catch Limiter Bug Fix

MerlionRock:
Added default values to daily_catch_limit and exit_on_limit_reached incase they are not found in CatchPokemon Task


davidakachaos:
👍

Issue #6099, opened 2017-07-19
Having difficulty walking to XXXX


camnomis:
### Expected Behavior

When walking to next Pokestop the bot should spin the stop

### Actual Behavior

Stop does not spin, message in log shows "Having difficulty walking to XXXX", bot continues to walk buddy and incubate eggs

### Your FULL config.json (remove your username, password, gmapkey and any other private info)
{
    "websocket_server": false,
    "heartbeat_threshold": 10,
    "enable_social": true,
    "check_niantic_api": true,
     "solve_captcha": false,
     "live_config_update": {
      "enabled": false,
      "tasks_only": false
    },
       "tasks": [
        {
            "type": "TelegramTask",
            "config": {
                "enabled": false,
                "master": null,
                "password": null,
                "// old syntax, still supported: alert_catch": [
                    "all"
                ],
                "// new syntax:": {},
                "alert_catch": {
                    "all": {
                        "operator": "and",
                        "cp": 1300,
                        "iv": 0.95
                    },
                    "Snorlax": {
                        "operator": "or",
                        "cp": 900,
                        "iv": 0.9
                    }
                }
            }
        },
        {
            "type": "DiscordTask",
            "config": {
                "enabled": false,
                "master": null,
                "// old syntax, still supported: alert_catch": [
                    "all"
                ],
                "// new syntax:": {},
                "alert_catch": {
                    "all": {
                        "operator": "and",
                        "cp": 1300,
                        "iv": 0.95
                    },
                    "Snorlax": {
                        "operator": "or",
                        "cp": 900,
                        "iv": 0.9
                    }
                }
            }
        },
      {
        "//NOTE: This task MUST be placed on the top of task list": {},
        "type": "RandomAlivePause",
        "config": {
          "enabled": false,
          "min_duration": "00:00:10",
          "max_duration": "00:10:00",
          "min_interval": "00:05:00",
          "max_interval": "01:30:00"
        }
      },
      {
        "type": "HandleSoftBan"
      },
      {
        "type": "RandomPause",
        "config": {
          "enabled": false,
          "min_duration": "00:00:10",
          "max_duration": "00:10:00",
          "min_interval": "00:10:00",
          "max_interval": "02:00:00"
        }
      },
      {
        "type": "CompleteTutorial",
        "config": {
          "enabled": false,
          "// set a name": "",
          "nickname": "",
          "// 0 = No Team, 1 = Blue, 2 = Red, 3 = Yellow": "",
[...]

Issue #6104, opened 2017-07-21
Random freezing of bot


goedzo:
All recent changes causes bot to freeze randomly. No single task is being executed anymore after a couple of hours. Did not try yet with debug on.

MerlionRock:
Can provide more information? Like what is the last task it did before it freeze? Which Commit? Dev or Master?

MerlionRock:
When I use webui, it shows that the bot is not moving, but when I see console, it is actually still moving with GymPokemon ... maybe webui cannot understand the gym moving function from main bot yet?

goedzo:
It's on Master branch. Here are the last details on console:

[2017-07-22 04:29:05] [PokemonCatchWorker] [INFO] Great Curveball throw! Used Pokeball, with chance 35.71 (19 left)
[2017-07-22 04:29:05] [PokemonCatchWorker] [INFO] Captured Paras! (CP: 574 IV: 0.6 5/9/13 NCP: 0.69 Shiny: False) Catch Limit: (461/800) +160 exp +100 stardust
[2017-07-22 04:29:05] [PokemonCatchWorker] [INFO] Candy gained: 3. You now have 190 Paras candy!
[2017-07-22 04:29:10] [GymPokemon] [INFO] We currently have 8 Pokemon in Gym(s):[2017-07-22 04:29:10] [GymPokemon] [INFO] Tyranitar (2688 CP)
[2017-07-22 04:29:10] [GymPokemon] [INFO] Flareon (2212 CP)
[2017-07-22 04:29:10] [GymPokemon] [INFO] Ursaring (2098 CP)
[2017-07-22 04:29:10] [GymPokemon] [INFO] Vaporeon (2239 CP)
[2017-07-22 04:29:10] [GymPokemon] [INFO] Pinsir (2028 CP)
[2017-07-22 04:29:10] [GymPokemon] [INFO] Starmie (2107 CP)
[2017-07-22 04:29:10] [GymPokemon] [INFO] Tyranitar (2082 CP)
[2017-07-22 04:29:10] [GymPokemon] [INFO] Jynx (1944 CP)
[2017-07-22 04:29:10] [GymPokemon] [INFO] Inspecting 1 gyms.
[2017-07-22 04:29:10] [GymPokemon] [INFO] Recent gyms: 18
[2017-07-22 04:29:10] [GymPokemon] [INFO] Active raid gyms: 0
[...]

PR #6150, opened 2017-07-27
Crvfts patch 1

crvfts:
Summary: 
Appears to fix bot getting stuck and not moving to fort and bot moving back and forth between forts but never getting to either.

Resolves:
#6099 #6104 #6065


MerlionRock:
👍

ChiefM3:
Wow thanks for the fix. I should definitely test this.

MerlionRock:
I don't think it's fixed....  It make it worst.... I am reverting it back on my own project and see

MerlionRock:
I think need to revert, this commit causes the bot behave incorrectly and no longer visit gyms.

Plecoholic:
it is not because of this change ...
if you look on the gym objects returned from cell parsing they all have atribute "closed".

Issue #6154, opened 2017-07-27
Cleaning up JSON files. 


VoIP:
### Short Description

Cleaner .json files. 

catch.json
```
"catch": {
"x":"y",
"a":"b"
}
```

evolve.json
```
"evolve": {
"x":"y",
"a":"b"
}
```
### Possible solution

By cleaning up the config, you not only make it easier to diagnose json errors, but also make it easy to find and write. Plus you would have easier understanded variables:
```
bot.config.evolve['rattata'] = evolve()
```
### How it would help others

By creating cleaner json files, and not bundling up all the data  into one huge file, it limits issues like 
JSON errors, and also makes it easier to modify what is needed instead of searching through a wall of text.

MerlionRock:
The json files for item and pokemon are auto generated with a tool that grab the gamemaster. 

I can point you to the tool to generate this two files if you need it

MerlionRock:
I misunderstood the original post meaning.

It's for the config.json. 

What are the views of the rest of dev?

davidakachaos:
On one end it will cleanup the config for a end user. But how often do you change the config?
We currently have code in place that will restart the bot/reload the task config when the config file is adjusted. I don't know if it can scale to multiple files, and if so what impact that will have on the bot?

VoIP:
As per our chat: The idea is to do:
`{
"pokemons": { 
  "Bulbasaur": {
    "catch": true,
    "evolve": false
  }
    "Ivysaur": {
      "catch": true,
      "evolve": false
  }
      "Venusaur": {
      "catch": true
  }
}`

davidakachaos:
So you want to extract things like 
```json
"catch": {
        "any": {
            "candy_threshold": 100000,
            "catch_above_cp": 0,
            "catch_above_iv": 0,
            "logic": "or"
        },
        "// Murkrow": {
          "// only_catch_better_cp": true,
          "// only_catch_better_iv": true
        },
        "Magikarp": {
            "always_catch": true
        },
        "// Always catch 'trash' mons:": {},
        "Caterpie": {
            "always_catch": true
        },
        "Weedle": {
            "always_catch": true
        },
        "Pidgey": {
            "always_catch": true
        },
        "Rattata": {
            "always_catch": true
        },
        "Nidoran F": {
            "always_catch": true
        },
        "Nidoran M": {
            "always_catch": true
        },
        "Zubat": {
            "always_catch": true
        },
        "Oddish": {
[...]

PR #6162, opened 2017-07-29
Added only_catch_better_iv and only_catch_better_cp 

davidakachaos:
Added ability for the bot to do a basic check if a Pokemon to be caught is a higher IV or CP than the one we have (highest one we have) and see if we need to try and catch it.

### DIFF
--- a/docs/configuration_files.md
+++ b/docs/configuration_files.md
@@ -423,6 +423,19 @@ For example:
 ```
 will stop catching Pidgey entirely.
 
+You can set a rule to only catch better IV or CP Pokemon by setting the only_catch_better_cp or only_catch_better_iv flags.
+
+
+For example:
+```
+"Pidgey": {
+    "only_catch_better_iv": true
+},
+"Gloom": {
+    "only_catch_better_cp": true
+}
+```
+
 ## Release Configuration
 [[back to top](#table-of-contents)]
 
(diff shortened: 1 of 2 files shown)


MerlionRock:
I don't understand the need of this... Just catch... Can earn some stardust... Later use optimizer to release... This sound redundant....

davidakachaos:
Agreed, but look at #6154 why someone wants to change catch parameters all the time

Jcolomar:
this is better than #6154

Jcolomar:
works only around IV and CP, but only see pokemon_id? or also family_id?

goedzo:
I am not liking the way this is coded:

if pokemon_config.get('only_catch_better_cp', False):

This will fail if the key is missing in json. So we need to avoid config fails whenever something new is added. Please default the values and allow the key not to be in config.

davidakachaos:
> This will fail if the key is missing in json.
It doesn't with me? I believe I provide a default value with the False here, so if no value found it will default to False?

davidakachaos:
@Jcolomar it works based on a Pokemon, not on a family of Pokemon, so no family_id

goedzo:
@davidakachaos sorry missed the false there :).

Jcolomar:
@davidakachaos so u need a rule for each pokemon...

davidakachaos:
Yes, I mean this rule isn't for every Pokémon? Like the candy rule, it
works like that, so it checks the Pokémon about to be caught versus the
rule(s) matched, so it would be able to match multiple Pokémon?

Op 1 aug. 2017 14:38 schreef "Jhero" <notifications@github.com>:

solderzzc:
@davidakachaos @Jcolomar You want to merge this pr?

Jcolomar:
Ya np. We kept opened but @davidakachaos is in vacation.

PR #6168, opened 2017-07-30
revert multibot shell looping

nbq:
## Short Description:

- Removed un-needed looping code in shell from multibot shell


MerlionRock:
👍

PR #6169, opened 2017-07-31
new settings, comments added

Xairooo:
add


MerlionRock:
I think the comment should use a invalid key instead. Like /threads

MerlionRock:
👍

PR #6181, opened 2017-08-02
Added optional hash end point

MerlionRock:
Added optional hash end point. If will use default api endpoint if key not found in auth.json


nbq:
👍

MerlionRock:
Closing, mistake. Is suppose to merge to dev instead

nbq:
Fixed to be towards dev instead of master and will commit as soon as it is done compiling

PR #6187, opened 2017-08-03
Locale by location

MerlionRock:
# About locale\_by\_location

By default the bot will tell game server that current timezone is `"America/Chicago"`. By setting `locale_by_location` to `true`, the bot will use google API to determine Country Code and Timezone base on bot's location and send this information to game server.

Please make sure you have both Google Maps Geocoding API and Google Maps Time Zone API enabled in https://console.developers.google.com/ if you want to set this option to true.

### DIFF
--- a/configs/auth.json.example
+++ b/configs/auth.json.example
@@ -10,6 +10,7 @@
         }
     ],
     "gmapkey": "GOOGLE_MAPS_API_KEY",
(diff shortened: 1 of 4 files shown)


MerlionRock:
Also make sure you pull in latest API that support it.
Your bot should be saying it's running on 0.69.1 API

PR #6191, opened 2017-08-04
Tele snipe change

MerlionRock:
No longer need to delete comma from coordinates. Simply copy and paste.
Changes the return time format to Min:Sec


nbq:
👍

PR #6201, opened 2017-08-11
Warning Patch & Login Enhancement

MerlionRock:
1. Remove looping in run.sh
2. Updated get_inbox
3. Added "bypass_warning" defaulted to false
4. PGoAPI warning and Niantic Warning confirmation before bot continues. Can be by pass with "bypass_warning" set to true


nbq:
👍

PR #6213, opened 2017-09-12
Killswitch changes

MerlionRock:
Change it to allow users to continue if they want to.
Also allows developers to add message without pausing the bot.
 
However, the ability to totally kill the bot is retained.


nbq:
👍

Issue #6235, opened 2017-10-22
Problem to run runMultiBot.sh


tokar86a:
### Actual Behavior

https://pastebin.com/8iQHawN4

It dont help that i run ./setup.sh -u It still say the same thing.

### Output when issue occurred

https://pastebin.com/8iQHawN4

### Other Information
OS: Ubuntu
 
Branch: dev
 
Git Commit: 
 
Python Version: 2.7.13
 
Any other relevant files/configs (eg: path files)

MerlionRock:
PR done for this issue #6237

PR #6237, opened 2017-10-24
Minor fixes

MerlionRock:
1. Fixed a bug that doesn't allow bot to bypass warning when "bypass_warning" has been set to true
2. Fix api versioning in runMultiBot.sh

This will solve #6235


nbq:
👍

PR #6240, opened 2017-10-25
Multi bot fixes

MerlionRock:
1. Prevent errors when threads are more than accounts
2. Fix multi bot not running as it should be on Mac
3. Fix proxies not properly set on certain platform (Please raise an issue if your platform is not supported) 
4. Fix bot not exiting when it should. (Example: When level reached)
5. Keyboard interrupts now able to stop the bot


nbq:
👍

PR #6249, opened 2017-11-12
Gym and shadowban fixes

MerlionRock:
Fixes:
1. Server busy message fixed when slotting pokemons
2. Bot crashed when shadow banned

Also includes codes of unfinished feature. End users will not be affected.


nbq:
👍

